PR #5, opened 2019-12-20
Setting loaded configurings through netifapi

michalrudolf:
set up lan interface through netifapi (should be thread-safe)
func docu: https://www.nongnu.org/lwip/2_1_x/group__netifapi__netif.html#ga31755ea6dbb213236bfce19bcbe8c973

### DIFF
--- a/src/gui/screen_lan_settings.c
+++ b/src/gui/screen_lan_settings.c
@@ -5,11 +5,13 @@
  *      Author: Migi
  */
 
+#define LAN_CONFIG_DEVELOPMENT
+
 #include "screen_lan_settings.h"
 #include "lwip/dhcp.h"
+#include "lwip/netifapi.h"
 #include "lwip.h"
 #include <stdlib.h>
-#include "screen_menu.h"
 #include <stdbool.h>
 #ifdef LAN_CONFIG_DEVELOPMENT
     #include "ini.h"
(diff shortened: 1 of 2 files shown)


src/gui/screen_lan_settings.c
```diff
@@ -82,7 +83,7 @@ static void _addrs_to_str(char *param_str, uint8_t flg) {
 
 static void _parse_MAC_addr(char *mac_addr_str) {
     uint8_t mac_addr[6] = { 0, 0, 0, 0, 0, 0 };
-    for (uint8_t i = 0; i < MAC_ADDR_SIZE; i++)
+    for (uint8_t i = 0; i < 6; i++)
```
RadekVana: i would not use "6" but sizeof(mac_addr)

src/gui/screen_lan_settings.c
```diff
@@ -18,7 +20,6 @@
 #include <string.h>
 
 #define MAC_ADDR_START 0x1FFF781A //MM:MM:MM:SS:SS:SS
-#define MAC_ADDR_SIZE 6
```
RadekVana: why did you delete this define and used number?

src/gui/screen_lan_settings.c
JustAnother1: It is not necessary to put in the size of the array if you initialize it. So this also works: uint8_t mac_addr[] = { 0, 0, 0, 0, 0, 0 };

PR #48, opened 2020-01-03
Set configurations from loaded .ini file

michalrudolf:
A3IDES-279
A3IDES-280

Set configurations from loaded .ini file through thread-safe netapi function.

(Previous version of this PL(https://github.com/prusa3d/Prusa-Firmware-Buddy/pull/5) needed to be closed due to repository publishing. My private fork was no longer connected)

### DIFF
--- a/src/gui/screen_lan_settings.c
+++ b/src/gui/screen_lan_settings.c
@@ -5,11 +5,13 @@
  *      Author: Migi
  */
 
+#define LAN_CONFIG_DEVELOPMENT
+
 #include "screen_lan_settings.h"
 #include "lwip/dhcp.h"
+#include "lwip/netifapi.h"
 #include "lwip.h"
 #include <stdlib.h>
-#include "screen_menu.h"
 #include <stdbool.h>
 #ifdef LAN_CONFIG_DEVELOPMENT
     #include "ini.h"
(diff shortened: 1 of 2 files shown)


src/gui/screen_lan_settings.c
```diff
@@ -5,11 +5,13 @@
  *      Author: Migi
  */
 
+#define LAN_CONFIG_DEVELOPMENT
```
joshymjose: you may remove this macro

src/gui/screen_lan_settings.c
```diff
@@ -125,18 +127,17 @@ static void screen_lan_settings_init(screen_t *screen) {
     //============== DECLARE VARIABLES ================
 
     plan_str = (char *)gui_malloc(150 * sizeof(char));
-    char mac_addr_str[18];
 
     //============= FILL VARIABLES ============
 
-    _parse_MAC_addr(mac_addr_str);
+    _parse_MAC_addr(plsd->mac_addr_str);
```
joshymjose: why "_function_name" and "function_name"? why some functions starts with under score?

src/gui/screen_lan_settings.h
```diff
@@ -22,6 +22,7 @@ typedef struct {
     menu_item_t *items;
 
     window_text_t text;
+    char mac_addr_str[18];
```
joshymjose: magic number "18"  use maco to define it

src/gui/screen_lan_settings.c
michalrudolf: These are just "private" functions, that are not included in header and are used only in this source file.

PR #63, opened 2020-01-08
Fix formatting.

mkbel:


src/common/filament_sensor.c
```diff
@@ -23,20 +23,18 @@
...
     M600_on_level = 1,
-    M600_never    = 2
-}
-send_M600_on_t;
+    M600_never = 2
+} send_M600_on_t;
 
-typedef struct{
-    uint8_t M600_sent    : 1;
+typedef struct {
```
DRracer: this is the case when automatic formatting goes against readability

src/common/filament_sensor.c
```diff
@@ -23,20 +23,18 @@
 static volatile fsensor_t state = FS_NOT_INICIALIZED;
 static volatile fsensor_t last_state = FS_NOT_INICIALIZED;
 
-
-typedef enum{
-    M600_on_edge  = 0,
+typedef enum {
+    M600_on_edge = 0,
```
DRracer: this is the case when automatic formatting goes against readability

dragomirecky:
> please revert the enum values and struct bit number alignment, it is against readability (use clang-format-off / on for these parts), otherwise is looks acceptable.

I think there should be a discussion about this. I fully agree that comments aligned to block are more readable. Still, it is not version-control friendly, and it generates unnecessary conflicts that someone has to manually resolve: By aligning the comments to block, the format of the whole enum is dependant on the longest case of the enum. When you change the longest case (by adding a new longer one, or by deleting it), you subsequently edit the whole enum. This might seem like a rare case, but I believe it happens more often than it looks and thus being a reason why WebKit and other code-style guides chose not to go with automatic aligning to block.

dragomirecky:
Btw, if we decide to go against WebKit and align comments to a block, we should not turn the formatter off for that sections of the code, but we should instead set the clang-format's `AlignTrailingComments` to `true`. But still, I believe there is a good reason why WebKit has it set to `false`.

src/common/filament_sensor.c
```diff
@@ -23,20 +23,18 @@
...
 
-
-typedef enum{
-    M600_on_edge  = 0,
-    M600_on_level = 1,
-    M600_never    = 2
-}
-send_M600_on_t;
+typedef enum {
+    M600_on_edge,
```
DRracer: I know the previous code was kind of redundant but it was also more readable (explicit values)...

mkbel:
Reverted redundant enum values.

PR #64, opened 2020-01-08
Text roll

michalrudolf:
A3IDES-446

Text rolling is a new text render function.
Implemented in:
  window_scroll_text_t - class window that can be created with window_create_ptr(WINDOW_CLS_SCROLL_TEXT, ...) (its still buggy in some cases (BIG font), fix is waiting for GUI c++ rework)
  window_menu - if focused menu item is too long, after 4s it starts rolling
  window_file_list - if focused file name is too long, after 4s it starts rolling

### DIFF
--- a/src/guiapi/include/display_helper.h
+++ b/src/guiapi/include/display_helper.h
@@ -36,7 +36,7 @@ extern void render_text_align(rect_ui16_t rc, const char *text, font_t *font, co
 
 extern void scroll_text_phasing(int16_t win_id, font_t * font, txtroll_t * roll);
 
-extern void render_scroll_text_align(uint8_t focus, rect_ui16_t rc, const char *text, font_t *font,
+extern void render_scroll_text_align(rect_ui16_t rc, const char *text, font_t *font,
     padding_ui8_t padding, uint8_t alignment, color_t clr0, color_t clr1, txtroll_t * roll);
 
 extern void render_icon_align(rect_ui16_t rc, uint16_t id_res, color_t clr0, uint16_t flags);
--- a/src/guiapi/include/display_helper.h
+++ b/src/guiapi/include/display_helper.h
@@ -9,11 +9,35 @@
 #define RENDER_FLG_WORDB 0x1000 // multiline text
 #define RENDER_FLG(a, r) (a | r << 8) // render flag macro (ALIGN and ROPFN)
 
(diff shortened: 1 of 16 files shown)


src/guiapi/include/display_helper.h
```diff
@@ -9,12 +9,36 @@
...
 
+typedef enum {
+    ROLL_SETUP = 0,
+    ROLL_GO,
+    ROLL_STOP,
+    ROLL_RESTART,
+} TXTROLL_PHASE_t;
+
+typedef struct _txtroll_t {
+    uint8_t phase;
```
DRracer: reorder vars so that they can be better aligned - less unused space

DRracer: and be very careful where this struct gets included into other structs - beware of #pragma pack

src/guiapi/src/display_helper.c
```diff
@@ -52,6 +58,94 @@ void render_text_align(rect_ui16_t rc, const char *text, font_t *font, color_t c
...
+            break;
+        case ROLL_STOP:
+            gui_timers_delete_by_window_id(win_id);
+            roll->phase = ROLL_RESTART;
+            gui_timer_create_oneshot(TEXT_ROLL_INITIAL_DELAY_MS, win_id);
+            window_invalidate(win_id);
+            break;
+        case ROLL_RESTART:
+            roll->setup = 0;
+            gui_timer_create_oneshot(TEXT_ROLL_INITIAL_DELAY_MS, win_id);
```
DRracer: here you create a One-shot-timer and in next call of this function (which will be in state ROLL_SETUP) another timer gets created? This looks suspicious

src/guiapi/src/display_helper.c
```diff
@@ -52,6 +58,94 @@ void render_text_align(rect_ui16_t rc, const char *text, font_t *font, color_t c
...
+                    roll->progress++;
+                }
+                roll->px_cd--;
+                window_invalidate(win_id);
+            } else {
+                roll->phase = ROLL_STOP;
+            }
+            break;
+        case ROLL_STOP:
+            gui_timers_delete_by_window_id(win_id);
```
DRracer: this call tries to delete timers and is called from a timer handler function. So far it doesn't cause any issues, but may be dangerous in the future (the caller ma keep the timer for something and it will get destroyed here)

PR #65, opened 2020-01-08
Knob type detection (new type has 100n capacitor between reset and gnd)

XPila:
A3IDES-546

### DIFF
--- a/src/guiapi/src/st7789v.c
+++ b/src/guiapi/src/st7789v.c
@@ -312,7 +312,16 @@ void st7789v_init_ctl_pins(void) {
 void st7789v_reset(void) {
     st7789v_clr_rst();
     st7789v_delay_ms(15);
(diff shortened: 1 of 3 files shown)


src/guiapi/src/st7789v.c
```diff
@@ -312,7 +312,16 @@ void st7789v_init_ctl_pins(void) {
 void st7789v_reset(void) {
     st7789v_clr_rst();
     st7789v_delay_ms(15);
+    gpio_init(st7789v_config.pinRST, GPIO_MODE_INPUT, GPIO_PULLUP, GPIO_SPEED_FREQ_LOW);
+    volatile uint16_t delay = 0;
+    int irq = __get_PRIMASK() & 1;
```
DRracer: do you realize this is not a completely atomic/thread safe code ;) ?

PR #103, opened 2020-01-14
Use crlf on windows

mkbel:
I think this is the correct solution of CRLF vs LF.
I think it is fool-proof. As .gitattributes configuration has precedence over global configuration.
But it is not safe against malicious committer, as he can remove/modify .gitattributes file before commit and don't stage this change.

Alternative solution would be to force check-in as-is check-out as-is in .gitattributes.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto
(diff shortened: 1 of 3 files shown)


mkbel:
https://github.com/pre-commit/pre-commit-hooks/issues/436

asottile
the hook reads the file contents from disk -- on disk that  file has crlf (git may smudge it later, but that's the truth that the  hook sees) you'll probably have better mileage not using autocrlf with git and this hook
--

dragomirecky:
I don't think I am the right person to review this (as I have no experience with handling (CR)LF on Windows). Maybe @DRracer, can you please have a look?

PR #139, opened 2020-01-21
Refactor version-string handling

dragomirecky:
A bigger refactor of version-version handling (based on A3IDES-511 and requirements of the build server).

### DIFF
--- a/src/gui/screen_version_info.cpp
+++ b/src/gui/screen_version_info.cpp
@@ -25,11 +25,13 @@ struct version_t {
     uint8_t patch;
 };
 
+#define VERSION_INFO_STR_MAXLEN 150
 char *version_info_str = nullptr;
 
 void screen_menu_version_info_init(screen_t *screen) {
     //=============SCREEN INIT===============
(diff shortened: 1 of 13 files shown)


src/gui/screen_version_info.cpp
```diff
@@ -25,11 +25,13 @@ struct version_t {
     uint8_t patch;
 };
 
+#define VERSION_INFO_STR_MAXLEN 150
 char *version_info_str = nullptr;
 
 void screen_menu_version_info_init(screen_t *screen) {
     //=============SCREEN INIT===============
     screen_menu_init(screen, "VERSION INFO", 1, 0, 0);
+    version_info_str = (char *)gui_malloc(VERSION_INFO_STR_MAXLEN * sizeof(char));
```
DRracer: does it really need to allocate dynamic mem?

DRracer: I know it has been here before...

dragomirecky: Yep, good point. I think we should rewrite it as soon as the text-wrapping issue is fixed. Until then, it would be unreasonably hard to rewrite...

PR #148, opened 2020-01-23
Static dhcp switch

joshymjose:


src/common/eeprom.c
```diff
@@ -191,7 +317,7 @@ int eeprom_var_sprintf(char *str, uint8_t id, variant8_t var) {
 
 void eeprom_print_vars(void) {
     uint8_t id;
-    char text[16];
+    char text[21];
```
joshymjose: magic number 21

src/gui/screen_lan_settings.c
```diff
@@ -5,7 +5,7 @@
  *      Author: Migi
  */
 
-//#define STATIC_SAVE_LOAD_CONFIG
+#define STATIC_SAVE_LOAD_CONFIG
```
joshymjose: this macro define may be unnecessary? this feature is not any more optional i guess?

src/common/eeprom.c
```diff
@@ -21,6 +21,32 @@ const uint8_t eeprom_map_v1[] = {
     VARIANT8_UI8, // EEVAR_RUN_XYZCALIB
     VARIANT8_UI8, // EEVAR_RUN_FIRSTLAY
     VARIANT8_UI8, // EEVAR_FSENSOR_ENABLED
+    VARIANT8_UI8, // EEVAR_LAN_FLAG
+    VARIANT8_UI32, // EEVAR_LAN_IP4_ADDR
+    VARIANT8_UI32, // EEVAR_LAN_IP4_MSK
+    VARIANT8_UI32, // EEVAR_LAN_IP4_GW
+    VARIANT8_UI32, // EEVAR_LAN_IP4_DNS1
+    VARIANT8_UI32, // EEVAR_LAN_IP4_DNS2
+    VARIANT8_UI8, // EEVAR_LAN_HOSTNAME_0
```
joshymjose: good to comment that host name is char array of 20 chars?

src/common/eeprom.c
```diff
@@ -21,6 +21,32 @@ const uint8_t eeprom_map_v1[] = {
     VARIANT8_UI8, // EEVAR_RUN_XYZCALIB
     VARIANT8_UI8, // EEVAR_RUN_FIRSTLAY
     VARIANT8_UI8, // EEVAR_FSENSOR_ENABLED
+    VARIANT8_UI8, // EEVAR_LAN_FLAG
+    VARIANT8_UI32, // EEVAR_LAN_IP4_ADDR
```
joshymjose: comment on how the value x.x.x.x is "encoded" here

src/lwip.c
```diff
@@ -119,6 +136,11 @@ void MX_LWIP_Init(void) {
 
     /* USER CODE BEGIN 3 */
     netif_set_link_callback(&eth0, netif_link_callback);
+    netif_set_status_callback(&eth0, netif_status_callback);
```
joshymjose: netif_set_status_callback documentation says "Set callback to be called when interface is brought up/down or address is changed while up". may be  netif_set_link_callback is not needed when you use this?
otherwise both callback may be called, just check

src/common/eeprom.c
```diff
@@ -166,11 +270,33 @@ void eeprom_dump(void) {
     }
 }
 
+void eeprom_get_hostname(char * dest){
+    char hostname_str[LAN_HOSTNAME_MAX_LEN + 1];
+    for(uint8_t i = 0; i < LAN_HOSTNAME_MAX_LEN; i++){
+        hostname_str[i] = (char)eeprom_get_var(EEVAR_LAN_HOSTNAME_0 + i).ui8;
+    }
+    hostname_str[LAN_HOSTNAME_MAX_LEN] = '\0';
```
joshymjose: is this an extra measure to end the char string in case "dest" is not null terminated?

src/common/eeprom.c
```diff
@@ -166,11 +270,33 @@ void eeprom_dump(void) {
...
+    strncpy(dest, hostname_str, LAN_HOSTNAME_MAX_LEN + 1);
+}
+void eeprom_set_hostname(char * src){
+    uint8_t end = 0;
```
joshymjose: i would prefer bool (#include <stdbool.h>) if it is used for true/false

src/common/eeprom.c
michalrudolf: Yes

src/lwip.c
michalrudolf: Both callbacks are needed. There need to be a callback that will automaticaly set interface up, when I plug the cabel in. I did not find any situations when they both are called. Even if there are some, when interface is already set down, netif_set_down(&eth0) function has no efect.

src/common/eeprom.c
```diff
@@ -166,11 +271,33 @@ void eeprom_dump(void) {
...
+    char hostname_str[LAN_HOSTNAME_MAX_LEN + 1];
+    for(uint8_t i = 0; i < LAN_HOSTNAME_MAX_LEN; i++){
+        hostname_str[i] = (char)eeprom_get_var(EEVAR_LAN_HOSTNAME_0 + i).ui8;
+    }
+    hostname_str[LAN_HOSTNAME_MAX_LEN] = '\0';
+    strncpy(dest, hostname_str, LAN_HOSTNAME_MAX_LEN + 1);
+}
+void eeprom_set_hostname(char * src){
+    bool end = false;
+    for(uint8_t i = 0; i < LAN_HOSTNAME_MAX_LEN; i++){
```
DRracer: ARM is a 32bit platform, you don't need to hesitate using size_t or int in for cycles

src/common/eeprom.h
```diff
@@ -15,6 +15,36 @@
...
+#define EEVAR_LAN_HOSTNAME_10 0x1A
+#define EEVAR_LAN_HOSTNAME_11 0x1B
+#define EEVAR_LAN_HOSTNAME_12 0x1C
+#define EEVAR_LAN_HOSTNAME_13 0x1D
+#define EEVAR_LAN_HOSTNAME_14 0x1E
+#define EEVAR_LAN_HOSTNAME_15 0x1F
+#define EEVAR_LAN_HOSTNAME_16 0x20
+#define EEVAR_LAN_HOSTNAME_17 0x21
+#define EEVAR_LAN_HOSTNAME_18 0x22
+#define EEVAR_LAN_HOSTNAME_19 0x23 //End of 20char string
```
DRracer: I assume this is a temporal "brutality" just to workaround the fact, that there is no interface for storing strings into EEPROM yet. Right?

src/gui/screen_lan_settings.c
```diff
@@ -73,11 +77,23 @@ static void _addrs_to_str(char *param_str, uint8_t flg) {
...
-        snprintf(param_str, MAX_INI_SIZE, "[lan_ip4]\naddress=%s\nmask=%s\ngateway=%s", ip4_addr_str, ip4_msk_str, ip4_gw_str);
-    else
-        snprintf(plan_str, 150, "IPv4 Address:\n    %s\nIPv4 Netmask:\n    %s\nIPv4 Gateway:\n    %s\nMAC Address:\n    %s",
+    if (flg){
+        char save_hostname[LAN_HOSTNAME_MAX_LEN + 1];
+        eeprom_get_hostname(save_hostname);
+        snprintf(param_str, MAX_INI_SIZE, "[lan_ip4]\ntype=%s\nhostname=%s\naddress=%s\nmask=%s\ngateway=%s",
+            config.lan_flag & LAN_EEFLG_TYPE ? LAN_type_opt[1] : LAN_type_opt[0], save_hostname, ip4_addr_str, ip4_msk_str, ip4_gw_str);
+    } else {
+        snprintf(plan_str, 150, "IPv4 Address:\n  %s      \nIPv4 Netmask:\n  %s      \nIPv4 Gateway:\n  %s      \nMAC Address:\n  %s",
```
DRracer: not sure about the spaces here, wouldn't a more precisely specified string placeholder work better? E.g. %-12.12s ?

src/common/eeprom.h
michalrudolf: Yes, this "brutality" was advised by Robert as a temporal solution.

PR #160, opened 2020-01-28
STACK OVERFLOW DETECTION A3IDES-606

RadekVana:


src/common/bsod.c
```diff
@@ -294,3 +295,18 @@ void _bsod(const char *fmt, const char *file_name, int line_number, ...) {
...
+#ifdef configCHECK_FOR_STACK_OVERFLOW
+
+static TaskHandle_t tsk_hndl = 0;
+static signed char *tsk_name = 0;
+
+void vApplicationStackOverflowHook( TaskHandle_t xTask, signed char *pcTaskName )
+{
+	tsk_hndl = xTask;
+	tsk_name = pcTaskName;
+	if (strlen((const char *)pcTaskName) > 20) _bsod("STACK OVERFLOW\nHANDLE %p\n%s", 0, 0, xTask, pcTaskName);
```
mkbel: Isn't there risk, that no null character will be encountered before strlen will start reading outside of RAM end generating hard fault?

PR #166, opened 2020-01-30
Enable MPU

mkbel:
Replace free RTOS with free RTOS-MPU port.

Work in progress, at least it compiles for BOOT variants, but it dosn't run.


teotwaki:
Is this still relevant? If not, could you close it, @mkbel?

PR #185, opened 2020-02-05
Code safety: sprintf->snprintf + static const char [] + proper USB VID/PID

DRracer:
- replace `sprintf `with `snprintf `where possible
- force using `static const char [] = ""` when assigning these strings into some pointers, which reference these strings beyond their scope of definition
- there are still a few spots to be fixed/improved, marked with `@@TODO`

Originally this branch was made for setting proper USB VID/PID and device strings
A3IDES-770

### DIFF
--- a/src/common/marlin_client.c
+++ b/src/common/marlin_client.c
@@ -201,12 +201,13 @@ int marlin_wait_motion(uint32_t timeout) {
 }
 
 void marlin_gcode(const char *gcode) {
-    char request[MARLIN_MAX_REQUEST];
     marlin_client_t *client = _client_ptr();
     if (client == 0)
         return;
-    strcpy(request, "!g ");
-    strcat(request, gcode);
+    char request[MARLIN_MAX_REQUEST] = "!g ";
+//    strcpy(request, "!g ");
+    strncat(request, gcode, MARLIN_MAX_REQUEST - 3);
+    request[MARLIN_MAX_REQUEST-1] = 0; // safety, strncat does not add a \0 if buffer limit is reached
     _send_request_to_server(client->id, request);
     _wait_ack_from_server(client->id);
 }
--- a/src/common/marlin_client.c
+++ b/src/common/marlin_client.c
@@ -205,9 +205,7 @@ void marlin_gcode(const char *gcode) {
     if (client == 0)
         return;
     char request[MARLIN_MAX_REQUEST] = "!g ";
-//    strcpy(request, "!g ");
-    strncat(request, gcode, MARLIN_MAX_REQUEST - 3);
-    request[MARLIN_MAX_REQUEST-1] = 0; // safety, strncat does not add a \0 if buffer limit is reached
+    strlcat(request + 3, gcode, MARLIN_MAX_REQUEST - 3);
     _send_request_to_server(client->id, request);
     _wait_ack_from_server(client->id);
 }
@@ -217,7 +215,6 @@ int marlin_gcode_printf(const char *format, ...) {
     marlin_client_t *client = _client_ptr();
     if (client == 0)
         return 0;
-    // strcpy(request, "!g ");
     char request[MARLIN_MAX_REQUEST] = "!g ";
     va_list ap;
     va_start(ap, format);
--- a/src/common/marlin_client.c
+++ b/src/common/marlin_client.c
@@ -213,14 +213,14 @@ void marlin_gcode(const char *gcode) {
 
 int marlin_gcode_printf(const char *format, ...) {
     int ret;
-    char request[MARLIN_MAX_REQUEST];
     marlin_client_t *client = _client_ptr();
     if (client == 0)
         return 0;
-    strcpy(request, "!g ");
+    // strcpy(request, "!g ");
+    char request[MARLIN_MAX_REQUEST] = "!g ";
     va_list ap;
     va_start(ap, format);
-    ret = vsprintf(request + 3, format, ap);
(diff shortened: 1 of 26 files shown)


src/common/menu_support.cpp
```diff
@@ -25,7 +25,7 @@ void menu_support() {
     int ver_min = (FW_VERSION - 100 * ver_maj) / 10;
     int ver_sub = FW_VERSION % 10;
     const char *stages[] = { "pre-alpha", "alpha", "beta", "RC", "final" };
-    sprintf(version, " %d.%d.%d %s", ver_maj, ver_min, ver_sub, (char *)stages[FW_STAGENR]);
+    snprintf(version, " %d.%d.%d %s", ver_maj, ver_min, ver_sub, (char *)stages[FW_STAGENR]);
```
dragomirecky: I don't see the "buffer size" parameter here 👀

src/gui/dialogs/window_dlg_statemachine.c
```diff
@@ -108,7 +108,11 @@ void _window_dlg_statemachine_draw_frame(window_dlg_statemachine_t *window) {
 void progress_draw(rect_ui16_t win_rect, font_t *font, color_t color_back,
     color_t color_text, padding_ui8_t padding, uint8_t progress) {
     rect_ui16_t rc_pro = win_rect; //must copy it
-    char text[16];
+    // why 16 bytes for "0%" - "100%"? Thats just 5 bytes instead of 16
+    // Save stack, guys!
+    // @@TODO duplicit code for printing percentage!
+    static const uint32_t text_size = 5;//16;
```
dragomirecky: I would move the comment from here to a commit message ;)

src/gui/dialogs/window_dlg_wait.c
```diff
@@ -101,7 +101,10 @@ void window_dlg_wait_draw(window_dlg_wait_t *window) {
             rect_ui16_t rc_pro = rc;
             rc_pro.x = 10;
             rc_pro.w -= 20;
-            char text[16];
+            // why 16 bytes for "0%" - "100%"? Thats just 5 bytes instead of 16
+            // Save stack, guys!
+            static const uint32_t text_size = 5;//16;
```
dragomirecky: /another nitpick/
Again, let's move it to a commit message :)

src/gui/screen_lan_settings.c
```diff
@@ -80,12 +80,12 @@ static void _addrs_to_str(char *param_str, uint8_t flg) {
...
 }
 
-static void _parse_MAC_addr(char *mac_addr_str) {
+static void _parse_MAC_addr(char *mac_addr_str, uint32_t mac_addr_str_size) {
     volatile uint8_t mac_addr[] = { 0, 0, 0, 0, 0, 0 };
     for (uint8_t i = 0; i < MAC_ADDR_SIZE; i++)
         mac_addr[i] = *(volatile uint8_t *)(MAC_ADDR_START + i);
 
-    sprintf(mac_addr_str, "%x:%x:%x:%x:%x:%x", mac_addr[0], mac_addr[1], mac_addr[2], mac_addr[3], mac_addr[4], mac_addr[5]);
+    snprintf(mac_addr_str, mac_addr_str_size, "%x:%x:%x:%x:%x:%x", mac_addr[0], mac_addr[1], mac_addr[2], mac_addr[3], mac_addr[4], mac_addr[5]);
```
dragomirecky: We might consider using assertions or other ways of making sure we „fail safely“ if the mac_add_str_size is not sufficient.

dragomirecky:
Two more questions...

1. What is the benefit of defining a special `text_size` variable 
Screenshot 2020-02-06 at 12 30 19
instead of using `sizeof(text)`?

2. >force using static const char [] = "" when assigning these strings into some pointers, which reference these strings beyond their scope of definition
    
    What is the advantage of this? I believe that string literals are guaranteed to have static storage.

src/common/marlin_client.c
```diff
@@ -358,8 +359,8 @@ variant8_t marlin_set_var(uint8_t var_id, variant8_t val) {
     if (client) {
         retval = marlin_vars_get_var(&(client->vars), var_id);
         marlin_vars_set_var(&(client->vars), var_id, val);
-        n = sprintf(request, "!var %s ", marlin_vars_get_name(var_id));
-        marlin_vars_value_to_str(&(client->vars), var_id, request + n);
+        n = snprintf(request, sizeof(request), "!var %s ", marlin_vars_get_name(var_id));
+        marlin_vars_value_to_str(&(client->vars), var_id, request + n, MARLIN_MAX_REQUEST - n);
```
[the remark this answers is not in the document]
DRracer: ... equivalent from my point of view

src/common/menu_support.cpp
DRracer: and the b@#$%^ compiler didn't say a word ... thanks for pointing this out

src/gui/dialogs/window_dlg_wait.c
DRracer: but then probably no one would see it while reading the code ;)

src/gui/screen_lan_settings.c
DRracer: _Static_assert - gcc specific C-style compile-time assert, probably non-portable

dragomirecky: yeah, I think we have to go with a runtime check

espr14:
Po merge všech BFW-770 PR by se mohl tento PR zcela smazat.

XPila:
Solved in other pulrequest

PR #186, opened 2020-02-05
Release 4.0.2+176 dump + dumpread

XPila:


src/common/dump.c
```diff
@@ -0,0 +1,99 @@
...
+				f_write(&fil, (uint8_t*)(DUMP_FLASH_ADDR + addr), DUMP_BUFF_SIZE, &bw);
+			f_close(&fil);
+			return 1;
+		}
+	}
+	return 0;
+}
+
+//
+void dump_hardfault_test_0(void)
```
DRracer: keep this only for DEBUG

src/common/dump.c
```diff
@@ -0,0 +1,99 @@
...
+{
+#define test_0_var (*((volatile unsigned long *)(0x4b000000)))
+	test_0_var = 0;
+}
+
+//integer div by zero test
+void dump_hardfault_test_1(void)
+{
```
DRracer: keep this only for DEBUG

utils/dumpread/src/mapfile.cpp
```diff
@@ -0,0 +1,230 @@
...
+};
+
+// class cmapfile (encapsulates mapfile_t)
+class cmapfile : public mapfile_t
+{
+public:
+	cmapfile();
+	~cmapfile();
+	bool load(const char* fn);
+	cmapfile_mem_entry* get_mem_entry(const char* name);
```
DRracer: it is better to use stl-like names for functions like this - in this case "find"

utils/dumpread/src/mapfile.cpp
```diff
@@ -0,0 +1,230 @@
...
+
+
+// "C" interface implementation
+
+extern "C"
+{
+
+void mapfile_free(mapfile_t* pm)
+{
+    if (pm) delete ((cmapfile*)pm);
```
DRracer: no need to test if(pm), delete does nothing in case it gets a nullptr

src/common/dump.h
```diff
@@ -0,0 +1,109 @@
...
+    "    mrs r2, CONTROL        \n" /* CONTROL  */ \
+    "    str r2, [r1, #0x50]    \n" \
+    "    mrs r2, MSP            \n" /* MSP  */ \
+    "    str r2, [r1, #0x54]    \n" \
+    "    mrs r2, PSP            \n" /* PSP  */ \
+    "    str r2, [r1, #0x58]    \n" \
+    "    str r3, [r1, #0x5c]    \n" /* lrexc  */\
+)
+
+//#define DUMP_SNAPSHOT_TO_XFLASH() { DUMP_REGS_GEN_SNAPSHOT(); dump_to_xflash(); }
```
DRracer: is it necessary to keep this commented line?

PR #227, opened 2020-02-17
Make Ethernet option

michalrudolf:
A3IDES-631
Make macro BUDDY_ENABLE_ETHERNET dependent on BUDDY_ENABLE_WUI


src/common/appmain.cpp
```diff
@@ -62,10 +62,10 @@ void app_idle(void) {
 void app_run(void) {
     DBG("app_run");
 
-#ifdef ETHERNET
+#ifdef BUDDY_ENABLE_ETHERNET
     if (diag_fastboot)
         osThreadResume(webServerTaskHandle);
```
joshymjose: i dont understand why it is like this? could you confirm with Robert what is this meant for?

PR #233, opened 2020-02-18
Support passing custom compiler options via CMake/build.py

dragomirecky:
1. adds `-D<VARIABLE>:<TYPE>=<VALUE>` command-line option to build.py
    - a way to pass custom CMake cache variables
    - essentially the same, as passing the same option to `cmake` (has the same syntax and semantics, except that the `<TYPE>` is required when used in build.py)
2. adds a new CUSTOM_COMPILE_OPTIONS option to cmake
    - for example `cmake ... -DCUSTOM_COMPILE_OPTIONS:STRING=-Werror`
    - and thanks to 2) `build.py ... -DCUSTOM_COMPILE_OPTIONS:STRING=-Werror`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -69,6 +69,10 @@ set(BUILD_NUMBER
     ""
     CACHE STRING "Build number of the firmware. Resolved automatically if not specified."
     )
+set(CUSTOM_COMPILE_OPTIONS
+    ""
+    CACHE STRING "Allows adding custom C/C++ flags"
+    )
 
 # Validate options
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -69,6 +69,10 @@ set(BUILD_NUMBER
     ""
     CACHE STRING "Build number of the firmware. Resolved automatically if not specified."
     )
+set(CUSTOM_COMPILE_OPTIONS
```
DRracer: future work - think about a more standard/widely used solution like CFLAGS/CXXFLAGS
this is ok for now

PR #244, opened 2020-02-24
M876 usbprtdialog A3IDES-594

RadekVana:


src/gui/screen_printing_serial.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+void screen_printing_serial_init(screen_t *screen);
+void screen_printing_serial_done(screen_t *screen);
+void screen_printing_serial_draw(screen_t *screen);
+int screen_printing_serial_event(screen_t *screen, window_t *window, uint8_t event, void *param);
+
+static const char txt0[] = "Remote printing\nin progress    ";
+static const char txt1[] = "Remote printing\nin progress .  ";
+static const char txt2[] = "Remote printing\nin progress .. ";
+static const char txt3[] = "Remote printing\nin progress ...";
```
DRracer: this is brutal ... why don't you append the dots to the text when being printed?

src/guiapi/src/screen.c
```diff
@@ -95,6 +95,21 @@ void screen_close(void) {
     }
 }
 
+int _currnet_in_list(screen_t *const *list, size_t sz) {
```
DRracer: typo?
currnet -> current?

src/common/marlin_client.c
```diff
@@ -49,6 +49,7 @@ typedef struct _marlin_client_t {
     uint32_t command;            // processed command (G28,G29,M701,M702,M600)
     marlin_host_prompt_t prompt; // current host prompt structure (type and buttons)
     uint8_t reheating;           // reheating in progress
+    dialog_cb_t dialog_cb;
```
DRracer: document the meaning and usage of this var

src/gui/screen_printing_serial.cpp
RadekVana: I did try to redraw just dots. But there is an bug in textfield.
Redrawing dots erased parts of characters around it.

src/guiapi/src/screen.c
RadekVana: Thanks

src/common/marlin_client.c
RadekVana: comment added

src/guiapi/src/screen.c
RadekVana: fixed

PR #248, opened 2020-02-24
[WIP] SPI low frequency while idle

XPila:
A3IDES-181


XPila:
will be solved different way

PR #254, opened 2020-02-27
New Eeprom

XPila:
A3IDES-351, A3IDES-647, A3IDES-365


src/common/eeprom.c
```diff
@@ -2,230 +2,290 @@
...
+    float PID_BED_D;
+    uint8_t LAN_FLAG;
+    uint32_t LAN_IP4_ADDR;
+	uint32_t LAN_IP4_MSK;
+	uint32_t LAN_IP4_GW;
+	uint32_t LAN_IP4_DNS1;
+	uint32_t LAN_IP4_DNS2;
+	char LAN_HOSTNAME[LAN_HOSTNAME_MAX_LEN + 1];
+    char TEST[10];
+    uint32_t CRC32;
```
DRracer: the CRC is the same case as a version number - it should probably be among the first structure items to ensure fixed location for future expansion of the EEPROM struct

src/common/variant8.cpp
```diff
@@ -0,0 +1,374 @@
...
+
+#include "variant8.h"
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+
+extern "C" {
+
+variant8_t variant8_init(uint8_t type, uint16_t count, void* pdata)
```
DRracer: the void *pdata will probably force the compiler to assume the parameter must have an address -> crippling optimization possibilities

src/common/variant8.cpp
```diff
@@ -0,0 +1,374 @@
...
+				*((uint16_t*)ptr) = size;
+				variant8_total_malloc_size += (size - old_size);
+			}
+			return ptr;
+		}
+		variant8_free(ptr);
+		return 0;
+	}
+	return variant8_malloc(size);
+}
```
DRracer: since there is a realloc, I'm definitely missing the "move" semantics - a function, that will hand over the pointer to another instance of the variant - can be then used from the c++ interface

src/common/eeprom.c
```diff
@@ -2,230 +2,295 @@
...
+    { VARIANT8_PCHAR, LAN_HOSTNAME_MAX_LEN + 1 }, // EEVAR_LAN_HOSTNAME
+    { VARIANT8_PCHAR, 10 }, // EEVAR_TEST
+    { VARIANT8_PCHAR, EEPROM__PADDING }, // EEVAR__PADDING32
+    { VARIANT8_UI32,  1  }, // EEVAR_CRC32
 };
 
+// eeprom variable names
 const char *eeprom_var_name[] = {
     "VERSION",
+    "FEATURES",
```
DRracer: Keeping all these descriptor arrays in sync will be very hard to maintain.
I don't see any compile-time checks to prevent accidental errors.
Probably making a:
struct EEPROMItemDesc {
  char var_name[14];
  uint8_t type;
  uint8_t bytes_used;
  variantXX default_value;
};
would force users add all necessary components when adding a new eeprom var

src/common/eeprom.h
```diff
@@ -49,14 +76,22 @@ extern void eeprom_set_var(uint8_t id, variant8_t var);
 // fill range 0x0000..0x0800 with 0xff
 extern void eeprom_clear(void);
 
-// fill dest parameter with hostname (max 20 chars)
-extern void eeprom_get_hostname(char *dest);
+//
+extern int eeprom_load_bin_from_usb(const char* fn);
```
DRracer: eeprom serialization is an unnecessary dependence here
this can be done in a separate h/c/cpp file to minimize the interface of the eeprom itself

src/common/eeprom.c
XPila: CRC is not same case. There is no reason to place CRC into "header" at start of the structure other than estetic. We can do version independent check with crc at end also with more simple code because we do not need skip it and calculation is one single call.

src/common/variant8.cpp
XPila: Function variant8_init is used only for pointer types, we need single function which takes pointer to data, for pointer types variant8_init will copy the data using memcpy (fastest way). I do not see better solution.

src/common/variant8.cpp
XPila: This is not necessary for C because variant8_t is not class and has not instance. Different case is future C++ encapsulation.
Realloc is not used, i implemeted it and planned to use in string conersion functions but then I found simpler solution.

src/common/eeprom.h
XPila: i will move serialization functions to separate source

src/common/variant8.h
```diff
@@ -13,54 +13,213 @@
...
+    cvariant8& operator = (int8_t val);
+    cvariant8& operator = (uint8_t val);
+    cvariant8& operator = (int16_t val);
+    cvariant8& operator = (uint16_t val);
+    cvariant8& operator = (int32_t val);
+    cvariant8& operator = (uint32_t val);
+    cvariant8& operator = (float val);
+    cvariant8& operator = (const char* val);
+public: // extractors
+    operator int8_t() { return (type == VARIANT8_I8)?i8:0; }
```
DRracer: I'd recommend using a member function like
inline uint8_t AsUint8t(uint8_t default_value)const { return (type == VARIANT8_U8)?u8:default_value; }
benefits:
- it makes clear, that some default value is being returned in case of an error
- it may be easier to use in cases, where multiple overloaded operators might be applicable in a particular context - cases, that may be solved by using static_cast<> which doesn't look too nice

src/common/variant8.h
```diff
@@ -13,54 +13,213 @@
...
 
 #pragma pack(pop)
 
+
 #ifdef __cplusplus
+
+class cvariant8 : public variant8_t {
+public: // construction/destruction
+    cvariant8();
+    cvariant8(const cvariant8& var8);
```
DRracer: a move constructor may be beneficial too, since we can do variant_realloc
cvariant8(cvariant8&& v2)...

src/gui/screen_lan_settings.c
```diff
@@ -79,7 +79,9 @@ static void _addrs_to_str(char *param_str, uint8_t flg) {
 
     if (flg) {
         char save_hostname[LAN_HOSTNAME_MAX_LEN + 1];
-        eeprom_get_hostname(save_hostname);
+        variant8_t hostname = eeprom_get_var(EEVAR_LAN_HOSTNAME);
+        strcpy(save_hostname, hostname.pch);
+        variant8_done(&hostname);
```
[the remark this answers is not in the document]
XPila: This is not very confusing, only pointer types needs variant8_done to be called, you can omit variant8_done only if you are sure that variant variable contain simple data type. But you are right, variant8 will show its own magic with C++.

src/common/eeprom.h
XPila: commit 3c634b7

src/common/variant8.cpp
```diff
@@ -0,0 +1,557 @@
+// variant8.c
```
DRracer: variant8.cpp ;)

src/common/eeprom.c
XPila: Names moved to eeprom_tab structure (commit 0ef1558), in tis time definition of eeprom data structure has two parts - first is eeprom_tab (used to access vars) and second is structure with default values (used in reset defaults and import previous version).

src/common/eeprom_loadsave.c
```diff
@@ -0,0 +1,101 @@
...
+int eeprom_save_xml_to_usb(const char *fn) {
+    FIL fil;
+    uint8_t id;
+    char text[128];
+    variant8_t var8;
+    UINT bw;
+    uint8_t var_count = eeprom_get_var_count();
+    const char *var_name;
+    if (f_open(&fil, fn, FA_WRITE | FA_CREATE_ALWAYS) == FR_OK) {
+        f_write(&fil, "<?xml version=\"1.0\" encoding=\"UTF-8\" standalone=\"no\"?>\n", 55, &bw);
```
DRracer: the attribute standalone has the default value of "no"

src/common/variant8.h
```diff
@@ -13,54 +13,213 @@
...
+    cvariant8 &operator=(uint8_t val);
+    cvariant8 &operator=(int16_t val);
+    cvariant8 &operator=(uint16_t val);
+    cvariant8 &operator=(int32_t val);
+    cvariant8 &operator=(uint32_t val);
+    cvariant8 &operator=(float val);
+    cvariant8 &operator=(const char *val);
+
+public: // extractors
+    operator int8_t() { return (type == VARIANT8_I8) ? i8 : 0; }
```
DRracer: where are the extractors with default values?

src/gui/screen_PID.cpp
```diff
@@ -370,7 +371,10 @@ int screen_PID_event(screen_t *screen, window_t *window, uint8_t event, void *pa
                         SPIN_DIGITS, SPIN_PRECISION),
                     get_single_PIDparamFromDisp(pd->idsDigits_Kd_E,
                         SPIN_DIGITS, SPIN_PRECISION));
-                marlin_gcode("M500"); //store to eeprom
+                eeprom_set_var(EEVAR_PID_NOZ_P, variant8_flt(Temperature::temp_hotend[0].pid.Kp));
```
DRracer: is the formatting correct?

src/common/variant8.cpp
XPila: fixed

XPila: commit bcd7031

src/common/variant8.h
XPila: commit bcd7031

src/common/eeprom_loadsave.c
XPila: xml serialization will be improved/finalized in some next pull request, save to xml is now implemented just for debugging and code will be completely reworked.

src/common/eeprom.c
```diff
@@ -2,228 +2,303 @@
...
+    uint16_t data_size;
+    void *data_ptr;
+    if (id < EEPROM_VARCOUNT) {
+        eeprom_lock();
+        if (var.type == eeprom_map[id].type) {
             size = eeprom_var_size(id);
-            st25dv64k_user_write_bytes(addr, &(var.ui32), size);
-            //			osDelay(5);
+            data_size = variant8_data_size(&var);
+            if (size == data_size) {
```
michalrudolf: In case of setting/getting a string: variant8_data_size is returning actual size of string (without '\0') and eeprom_var_size() is returning size of a string with a MAX length (shorter string is permitted). so this condition will never pass with strings.

Should be if(size >= data_size)

src/common/variant8.h
XPila: I will add extractors with default values later, we need discuss how it should behave.

src/common/eeprom.c
XPila: This is right. You can set string variable using the variant_init with exact size, but this is an uncomfortable way and cannot be used with c++ interface only. I will change the condition to accept shorter data (this will be an exception only for strings - VARIANT8_PCHAR type).

XPila: fixed in commit 9d92782

src/common/variant8.cpp
```diff
@@ -0,0 +1,660 @@
...
+#ifdef VARIANT8_DBG_MALLOC
+    if (ptr) {
+        uint16_t size = ((uint16_t *)ptr)[-1];
+        variant8_total_malloc_size -= size;
+        free(((uint16_t *)ptr) - 1);
+    }
+#else  //VARIANT8_DBG_MALLOC
+    free(ptr);
+#endif //VARIANT8_DBG_MALLOC
+}
```
dragomirecky: Can be simplified by using `malloc_usable_size`

src/common/variant8.h
```diff
@@ -13,54 +13,239 @@
...
+// returns 1 for unsigned integer types (UI8, UI16, UI32), otherwise returns 0
+inline static int variant8_is_unsigned(variant8_t *pvar8) { return (pvar8) ? (((pvar8->type == VARIANT8_I8) || (pvar8->type == VARIANT8_I16) || (pvar8->type == VARIANT8_I32)) ? 1 : 0) : 0; }
 
-extern variant8_t variant8_flt(float flt); // returns VARIANT8_FLT
+// returns 1 for integer types (I8, I16, I32, UI8, UI16, UI32), otherwise returns 0
+inline static int variant8_is_integer(variant8_t *pvar8) { return (pvar8) ? ((variant8_is_signed(pvar8) || variant8_is_unsigned(pvar8)) ? 1 : 0) : 0; }
 
-extern variant8_t variant8_user(uint32_t usr32); // returns VARIANT8_USER
+// returns 1 for numeric types (I8, I16, I32, UI8, UI16, UI32, float), otherwise returns 0
+inline static int variant8_is_number(variant8_t *pvar8) { return (pvar8) ? ((variant8_is_integer(pvar8) || (pvar8->type == VARIANT8_FLT)) ? 1 : 0) : 0; }
```
dragomirecky: Again, why not using bool as the return value?

src/common/eeprom.h
```diff
@@ -5,30 +5,64 @@
...
+#define EEVAR_LAN_IP4_DNS1    0x16 // X.X.X.X address encoded in uint32
+#define EEVAR_LAN_IP4_DNS2    0x17 // X.X.X.X address encoded in uint32
+#define EEVAR_LAN_HOSTNAME    0x18 // 20char string
+#endif
+
+#define EEVAR__PADDING        0x19 // 1..4 chars, to ensure (DATASIZE % 4 == 0)
+
+#define EEVAR_CRC32           0x1a // uint32_t crc32 for
+
+// clang-format on
```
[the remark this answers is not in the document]
XPila: This is because i prefer manual code formating and more readable code.  Current clang-format settings sometime break readability in tables and constants declarations, for example tables in eeprom.c. In similar cases is better prevent this with clang-format-off, because after commit is original formatting lost and result is unreadable... :(

src/common/eeprom.h
```diff
@@ -5,30 +5,64 @@
...
+#define EEVAR_LAN_IP4_MSK     0x14 // X.X.X.X address encoded in uint32
+#define EEVAR_LAN_IP4_GW      0x15 // X.X.X.X address encoded in uint32
+#define EEVAR_LAN_IP4_DNS1    0x16 // X.X.X.X address encoded in uint32
+#define EEVAR_LAN_IP4_DNS2    0x17 // X.X.X.X address encoded in uint32
+#define EEVAR_LAN_HOSTNAME    0x18 // 20char string
+#endif
+
+#define EEVAR__PADDING        0x19 // 1..4 chars, to ensure (DATASIZE % 4 == 0)
+
+#define EEVAR_CRC32           0x1a // uint32_t crc32 for
```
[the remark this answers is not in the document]
XPila: I cannot give a simple answer to this question, sometimes it is better to use enum, you can find many discusions about this e.g. at stackowerflow. Use define is simpler for C, older C compilers does not support enums... so by default i am using defines.
I want replace defines with enums for future because of more simple debugging - this is reason no.1 for me, but in this time i want keep defines and make changes step by step. I can promise this change very soon.

src/common/eeprom.c
```diff
@@ -242,83 +317,113 @@ void eeprom_dump(void) {
...
-    char hostname_str[LAN_HOSTNAME_MAX_LEN + 1];
-    for (uint8_t i = 0; i < LAN_HOSTNAME_MAX_LEN; i++) {
-        hostname_str[i] = (char)eeprom_get_var(EEVAR_LAN_HOSTNAME_START + i).ui8;
+void eeprom_print_vars(void) {
+    uint8_t id;
+    char text[128];
+    variant8_t var8;
+    for (id = 0; id < EEPROM_VARCOUNT; id++) {
+        var8 = eeprom_get_var(id);
+        eeprom_var_format(text, 128, id, var8);
```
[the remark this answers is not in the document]
XPila: commit 01b8ef5

src/common/variant8.h
```diff
@@ -13,54 +13,239 @@
...
+#define VARIANT8_PUI32 (VARIANT8_UI32 | VARIANT8_PTR)
+#define VARIANT8_PFLT  (VARIANT8_FLT | VARIANT8_PTR)
+#define VARIANT8_PCHAR (VARIANT8_CHAR | VARIANT8_PTR)
+
+//variant errors
+#define VARIANT8_ERR_MALLOC 1 // memory allocation error (during conversion to strings and allocating pointer types)
+#define VARIANT8_ERR_UNSTYP 2 // unsupported conversion (during conversion)
+#define VARIANT8_ERR_UNSCON 3 // unsupported conversion (during conversion)
+#define VARIANT8_ERR_INVFMT 4 // invalid format (during conversion from string)
+#define VARIANT8_ERR_OOFRNG 5 // out of range (during conversion from bigger to lower range number)
```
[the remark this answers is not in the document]
XPila: =||= ...will be changed to enum in some next pull

src/common/variant8.h
```diff
@@ -13,54 +13,239 @@
 #define VARIANT8_I32   0x05 // signed long - 4byte
 #define VARIANT8_UI32  0x06 // unsigned long - 4byte
 #define VARIANT8_FLT   0x07 // float - 4byte
+#define VARIANT8_CHAR  0x08 // char - 1byte
 #define VARIANT8_USER  0x3f // user - up to 7 bytes
-#define VARIANT8_PTR   0x40 // pointer - 4 bytes
-#define VARIANT8_ARRAY 0x80 // array
+#define VARIANT8_PTR   0x80 // pointer - 4 bytes,
```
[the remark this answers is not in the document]
XPila: =||= ...will be changed to enum in some next pull

src/common/eeprom_loadsave.c
```diff
@@ -0,0 +1,101 @@
...
+        f_write(&fil, "<eeprom>\n", 9, &bw);
+        for (id = 0; id < var_count; id++) {
+            var8 = eeprom_get_var(id);
+            *text = 0;
+            eeprom_var_format(text, 128, id, var8);
```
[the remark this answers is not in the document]
XPila: commit d55f11d

src/common/variant8.cpp
```diff
@@ -0,0 +1,660 @@
...
+                pvar8->ui16 = (uint16_t)pvar8->flt;
+            else
+                *pvar8 = variant8_error(VARIANT8_ERR_OOFRNG, 0, 0);
+            break;
+        case VARIANT8_UI32:
+            if ((pvar8->flt >= 0) && (pvar8->flt <= 4294967295))
+                pvar8->ui32 = (uint32_t)pvar8->flt;
+            else
+                *pvar8 = variant8_error(VARIANT8_ERR_OOFRNG, 0, 0);
+            break;
```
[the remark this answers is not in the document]
XPila: commit e946f0c

src/common/eeprom_loadsave.h
```diff
@@ -0,0 +1,28 @@
...
+extern int eeprom_load_bin_from_usb(const char *fn);
+
+// save binary image (0x0000..0x0800) to usb flash (fatfs file)
+extern int eeprom_save_bin_to_usb(const char *fn);
+
+// TODO:
+extern int eeprom_load_xml_from_usb(const char *fn);
+
+// save xml file with all variables to usb flash (fatfs file)
+extern int eeprom_save_xml_to_usb(const char *fn);
```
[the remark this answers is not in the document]
XPila: bool is not defined in C. Possible way is include <stdbool.h>, but I prefer usage of basic data types (int). There are many conventions for return values. when the logic is 1/0, for me is more transparent use value 1 for succes and value 0 for fail.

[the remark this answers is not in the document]
XPila: serialization is not finished, I will change the suffix later. Maybe better suffix will be "_ffile" because fatfs functions are prefixed with "f_" (e.g. "f_open" instead of "open").

src/common/variant8.cpp
XPila: i choosed this solution because malloc_usable_size is not supported in all compilers. I have some testing code under mingw... i will hack this different way.
commit 3074321

src/common/eeprom.h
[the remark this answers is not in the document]
dragomirecky: It doesn't break readability here.

src/common/eeprom_loadsave.h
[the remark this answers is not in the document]
dragomirecky: `bool` is part of C specification for over 20 years now (ISO/IEC 9899), so I believe it is very much part of the language.

I am aware there are many conventions for return values, and that is the reason why I suggested using bool. When using `int`, the reader can't be sure what convention the API is using. However, when using `bool`, there is no dispute about it.

src/common/eeprom.h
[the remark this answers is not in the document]
XPila: commit 39e1459

src/common/eeprom.c
```diff
@@ -2,228 +2,303 @@
...
+    uint8_t LAN_FLAG;
+    uint32_t LAN_IP4_ADDR;
+    uint32_t LAN_IP4_MSK;
+    uint32_t LAN_IP4_GW;
+    uint32_t LAN_IP4_DNS1;
+    uint32_t LAN_IP4_DNS2;
+    char LAN_HOSTNAME[LAN_HOSTNAME_MAX_LEN + 1];
+    char _PADDING[EEPROM__PADDING];
+    uint32_t CRC32;
+} eeprom_vars_t;
```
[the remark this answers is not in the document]
XPila: EEPROM_FEATURES is reflected only in header, why it should be reflected here and how?
Adding variables to eeprom is not every day job. EEprom structure and table will be checked in eeprom unit test, because it cannot be simply checked in compile time. Only padding vs datasize can be checked with static_assert - this will be done with C++ eeprom encapsulation. Some runtime check can be done inside init, i will add this later.

src/common/eeprom.c
```diff
@@ -2,228 +2,303 @@
...
+    "",              // EEVAR__PADDING
+    0xffffffff,      // EEVAR_CRC32
 };
 
-uint16_t eeprom_crc_value = 0;
-uint8_t eeprom_crc_index = 0;
+// clang-format on
+
+// semaphore handle (lock/unlock)
+osSemaphoreId eeprom_sema = 0;
```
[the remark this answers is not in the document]
XPila: commit 2f2b4ba

src/common/eeprom.c
```diff
@@ -2,228 +2,303 @@
...
+    char _PADDING[EEPROM__PADDING];
+    uint32_t CRC32;
+} eeprom_vars_t;
+
+#pragma pack(pop)
+
+// clang-format off
+
+// eeprom map
+const eeprom_entry_t eeprom_map[] = {
```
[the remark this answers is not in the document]
XPila: commit 2f2b4ba

src/common/eeprom.c
```diff
@@ -2,228 +2,303 @@
...
-    "EEVAR_LAN_HOSTNAME_CHAR13",
-    "EEVAR_LAN_HOSTNAME_CHAR14",
-    "EEVAR_LAN_HOSTNAME_CHAR15",
-    "EEVAR_LAN_HOSTNAME_CHAR16",
-    "EEVAR_LAN_HOSTNAME_CHAR17",
-    "EEVAR_LAN_HOSTNAME_CHAR18",
-    "EEVAR_LAN_HOSTNAME_CHAR19",
-    "EEVAR_LAN_HOSTNAME_CHAR20",
+// eeprom variable defaults
+const eeprom_vars_t eeprom_var_defaults = {
```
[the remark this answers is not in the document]
XPila: commit 2f2b4ba

src/common/eeprom.c
```diff
@@ -2,228 +2,303 @@
...
-variant8_t eeprom_var_default(uint8_t id);
 void eeprom_dump(void);
+int eeprom_var_format(char *str, unsigned int size, uint8_t id, variant8_t var);
 void eeprom_print_vars(void);
-void eeprom_clear(void);
+int eeprom_convert_from_v2(void);
+int eeprom_convert_from(uint16_t version, uint16_t features);
+
+int eeprom_check_crc32(void);
+void eeprom_update_crc32(uint16_t addr, uint16_t size);
```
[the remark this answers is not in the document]
XPila: commit 2f2b4ba

src/common/eeprom.c
```diff
@@ -2,228 +2,303 @@
...
+                st25dv64k_user_write_bytes(addr, data_ptr, data_size);
+                eeprom_update_crc32(addr, size);
+            } else {
+                // TODO: error
+            }
+        } else {
+            // TODO: error
         }
+        eeprom_unlock();
     }
```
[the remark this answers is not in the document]
XPila: error checking will be improved later

src/common/eeprom.c
```diff
@@ -2,228 +2,303 @@
...
+        if (eeprom_convert_from(version, features) == 0)
+            defaults = 1;
     }
-    //eeprom_print_vars();
+    if (defaults)
+        eeprom_defaults();
+    eeprom_print_vars();
     //eeprom_dump();
-    return ret;
+    return defaults;
```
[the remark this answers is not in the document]
XPila: commit 255f656

src/common/eeprom_loadsave.h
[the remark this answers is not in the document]
XPila: bool is not supported in C, only with header stdbool.h ("bool", "true" and "false" are macros). I have different opinion to return values, i am not using bool datatype in C code and i will not change this. In future these functions should return system error codes.

src/common/eeprom_loadsave.c
```diff
@@ -0,0 +1,101 @@
...
+
+int eeprom_load_bin_from_usb(const char *fn) {
+    FIL fil;
+    uint8_t buff[128];
+    uint16_t size = 0x800;
+    uint16_t addr = 0x000;
+    UINT bs = sizeof(buff);
+    UINT br;
+    if (f_open(&fil, fn, FA_READ) == FR_OK) {
+        //        eeprom_lock();
```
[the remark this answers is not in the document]
XPila: commit c8f0d01

src/common/eeprom.c
```diff
@@ -242,83 +317,113 @@ void eeprom_dump(void) {
...
 
-void eeprom_print_vars(void) {
-    uint8_t id;
-    char text[16];
-    for (id = 0; id < EE_VAR_CNT; id++) {
-        eeprom_var_sprintf(text, id, eeprom_get_var(id));
-        _dbg("%s=%s", eeprom_var_name[id], text);
-    }
+// version independent crc32 check
+int eeprom_check_crc32(void) {
```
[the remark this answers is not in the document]
XPila: i want not use bool in C code (description bellow)

src/common/eeprom.c
```diff
@@ -242,83 +317,113 @@ void eeprom_dump(void) {
...
-    case EEVAR_UNUSED_3:
-        return sprintf(str, "%.1f", (double)var.flt);
-    case EEVAR_RUN_SELFTEST:
-    case EEVAR_RUN_XYZCALIB:
-    case EEVAR_RUN_FIRSTLAY:
-    case EEVAR_FSENSOR_ENABLED:
-        return sprintf(str, "%u", (unsigned int)var.ui8);
-    }
+// conversion function for new version format (features, firmware version/build)
+int eeprom_convert_from(uint16_t version, uint16_t features) {
```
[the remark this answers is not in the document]
XPila: i want not use bool in C code (description bellow)

src/common/eeprom.c
```diff
@@ -242,83 +317,113 @@ void eeprom_dump(void) {
...
+
+#define ADDR_V2_FILAMENT_TYPE  0x0400
+#define ADDR_V2_FILAMENT_COLOR EEPROM_ADDRESS + 3
+#define ADDR_V2_RUN_SELFTEST   EEPROM_ADDRESS + 19
+#define ADDR_V2_ZOFFSET        0x010e
+#define ADDR_V2_PID_NOZ_P      0x019d
+#define ADDR_V2_PID_BED_P      0x01af
+
+// conversion function for old version 2 format (marlin eeprom)
+int eeprom_convert_from_v2(void) {
```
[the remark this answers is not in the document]
XPila: i want not use bool in C code (description bellow)

PR #293, opened 2020-03-13
Use board independent hwio.h instead of board dependent hwio_a3ides.h

mkbel:


src/common/hwio.h
```diff
@@ -16,15 +16,30 @@
...
 
+//a3ides analog inputs
+enum {
+    ADC_HW_IDENTIFY,    //!< PA3 - chan 3
+    ADC_TEMP_BED,       //!< PA4 - chan 4
+    ADC_TEMP_2,         //!< PA5 - chan 5
+    ADC_TEMP_HEATBREAK, //!< PA6 - chan 6
+    ADC_TEMP_0,         //!< PC0 - chan 10
+    ONE_BEHIND_LAST_ADC,
+} typedef Adc;
```
dragomirecky: /very subjective/
For some reason, I hate the look of `Adc`. Can it be `ADC`? :D

mkbel: It can't be ADC, it would look like macro. But ADC_t would be possible.

dragomirecky: Yes, good point!

PR #298, opened 2020-03-14
Allow to define printer where marlin configures sensor-less homing an…

mkbel:
…d thresholds itself. It is disabled for MINI.


dragomirecky:
I am not sure I understand this PR. What does `MARLIN_ONLY_DRIVER` represent?

mkbel:
If MARLIN_ONLY_DRIVER is enabled configuration of Trinamic motor driver (or any other motor driver) is completely left on Marlin. In our case it means that sensorless homing can be enabled in Marlin and its configuration left on Marlin.

mkbel:
Macro has been renamed to make its meaning more clear.

PR #332, opened 2020-03-30
Remove screen externs

RadekVana:
BFW-726
externed pointers to screens replaced by functions returning pointers to screens
removed LCDSIM


src/guiapi/include/screen_pointers.h
```diff
@@ -0,0 +1,49 @@
...
+extern screen_t *pscreen_mesh_bed_lv;
+extern screen_t *pscreen_wizard;
+extern screen_t *const pscreen_print_preview;
+extern screen_t *const pscreen_lan_settings;
+extern screen_t *pscreen_marlin;
+extern screen_t *const pscreen_menu_fw_update;
+
+#ifdef __cplusplus
+}
+#endif //__cplusplus
```
dragomirecky: Why don't we have a header file for each screen? Even though that is a lot of files, I believe it would be a nicer and cleaner solution.

In any case – if you want to keep this one big header file with all the screens, I would be up for renaming it to `screens.h` ☺️

src/gui/screen_version_info.cpp
```diff
@@ -117,4 +118,6 @@ screen_t screen_version_info = {
     0,                          //pdata
 };
 
-const screen_t *pscreen_version_info = &screen_version_info;
+extern "C" {
+screen_t *pscreen_version_info = &screen_version_info;
+}
```
dragomirecky: You can write `extern "C" screen_t ...`, no need to have it in a block.

dragomirecky: (this applies to more files in the PR)

src/guiapi/include/screen_pointers.h
RadekVana: 1 header is compromise me and Robert agreed on
ok will rename it

RadekVana: renamed

src/gui/screen_version_info.cpp
RadekVana: fixed

src/gui/screen_menu_tune.cpp
```diff
@@ -23,6 +17,7 @@ enum {
     MI_FLOWFACT,
     MI_BABYSTEP,
     MI_FILAMENT,
+    MI_INFO,
```
dragomirecky, line 20: Why is the `INFO` screen back? Isn't it a mistake?

src/gui/test/screen_test_graph.c
```diff
@@ -86,7 +86,7 @@ int screen_test_graph_event(screen_test_term_t *screen, window_t *window, uint8_
     } else if (event == WINDOW_EVENT_CLICK) {
         switch ((int)param) {
         case 1:
-            //screen_open(pscreen_menu_service->id);
+            //screen_open(get_scr_menu_service()->id);
```
dragomirecky, line 89: Why do we have this commented code here?

src/gui/test_window.c
```diff
@@ -52,11 +52,9 @@ void ui_loop(void)
...
-extern const screen_t *pscreen_test;
-
 void test_window(void) {
-    //	pscreen_test->init(pscreen_test);
-    //	pscreen_test->init(pscreen_test);
+    //	get_scr_test()->init(get_scr_test());
+    //	get_scr_test()->init(get_scr_test());
     /*
 	font_t* font = resource_font(IDR_FNT_11x18_4);
 	int16_t id0 = window_frame_create(-1, rect_ui16(0,0,0,0), COLOR_GRAY);
```
dragomirecky, line 60: Can't we just remove the `test_window(void)`?

src/guiapi/include/screen_unloop.h
```diff
@@ -0,0 +1,19 @@
...
+#include "stdint.h"
+#include "screen.h" //screen.t
+
+#ifdef __cplusplus
+extern "C" {
+#endif //__cplusplus
+
+typedef enum {
+    sculp_timeout,
+    sculp_M876
```
dragomirecky, line 12: It took me some time to understand that `sculp` is short for `screen unlooping` (and in this case, I had the full context). In case I would find this at a random place in the codebase, I am quite sure I would have no idea what it means.

Can we rename it? `SCREEN_UNLOOPING_TIMEOUT`, `SCREEN_UNLOOPING_M876`...


Hmm, but what is the semantics of „screen unlooping“? From what I see, it is about „closing as many screens as possible“ – but there are _no loops involved_. 

What about just `close_screens(CLOSE_ON_M876)`?

src/guiapi/src/screen_unloop.cpp
```diff
@@ -0,0 +1,64 @@
...
+static screen_t *const m876_blacklist[] = {
+    get_scr_printing_serial(),
+    get_scr_home()
+#ifdef PIDCALIBRATION
+        ,
+    get_scr_PID()
+#endif //PIDCALIBRATION
+};
+
+int _current_in_list(screen_t *const *list, size_t sz) {
```
dragomirecky, line 31: `static`

src/guiapi/src/screen_unloop.cpp
```diff
@@ -0,0 +1,64 @@
+// screen_unloop.cpp
+//unlooping screens via C++ with C interface
+
+#include "screen_unloop.h"
+#include "gui.h"
+#include "screens.h"
+
+//I could not use functions to initialize those arrays in C
```
dragomirecky, line 8: Yes, that is the side effect of wrapping the screen pointers in `get_scr_` functions. All statically initialized data using screen pointers are now going to have to be initialized at runtime :(

src/gui/screen_menu_tune.cpp
RadekVana: why was it deleted?
i thought was deleted by a mistake

src/gui/test/screen_test_graph.c
RadekVana: there is more commented code in this file,
not my code to remove

src/gui/screen_menu_tune.cpp
RadekVana: removed

src/guiapi/include/screen_unloop.h
RadekVana: screen_close() also opens previous screen
i think unlooping is fine

src/guiapi/src/screen_unloop.cpp
RadekVana: original `extern screen_t*` behaved exactly same

RadekVana: > All statically initialized data using screen pointers are now going to have to be initialized at runtime :(
I want to rewrite all screens to Cpp, so I will be able to use constexpr

src/gui/test_window.c
RadekVana: removed

src/guiapi/include/screen_unloop.h
RadekVana: sculp_timeout, sculp_M876
renamed

PR #354, opened 2020-04-07
Remove unnecessary MOTHERBOARD macro usage.

mkbel:


src/common/config.h
```diff
@@ -3,57 +3,8 @@
...
-    #elif (BOARD == A3IDES2209_REV02)
-        #include "config_a3ides2209_02.h"
-    #elif (BOARD == A3IDES2209_REV03)
-        #include "config_a3ides2209_02.h"
-    #endif
-
-#else
-    #error "BOARD not defined"
-#endif
+#include "config_a3ides2209_02.h"
```
dragomirecky, line 7: What about using this opportunity to get rid of `a3ides` (and replacing it with `buddy`)? 🙏🙂

PR #355, opened 2020-04-08
Improved File Dialog (unlimited number of files + less RAM usage)

DRracer:
Implemented as a moving window over directory entries.
For now the engine is a template, but it may be changed later to a simple class with some preset window size.

BFW-712

### DIFF
--- a/src/gui/lazyfilelist-c-api.h
+++ b/src/gui/lazyfilelist-c-api.h
@@ -1,0 +1,20 @@
+// temporary C API for the lazy dir view
+#pragma once
+#include <stdint.h>
+#include <stdbool.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+bool LDV_ChangeDir(void *LDV, bool sortByName, const char *path);
+uint32_t LDV_TotalFilesCount(void *LDV);
+uint32_t LDV_VisibleFilesCount(void *LDV);
+bool LDV_MoveUp(void *LDV);
+bool LDV_MoveDown(void *LDV);
+const char *LDV_FileAt(void *LDV, int index, bool *isFile);
+void *LDV_Create(void);
+
+#ifdef __cplusplus
+}
+#endif
--- a/src/gui/lazyfilelist.h
+++ b/src/gui/lazyfilelist.h
@@ -1,0 +1,329 @@
+#pragma once
+#include <memory>
+#include <algorithm>
+#include "fatfs.h"
+#include <string.h>
+#include <limits.h>
+
+/// Lazy Dir View
+/// Implements a fixed size view over a directory's content.
+///
+/// The entries are sorted according to one of the selected policies: BY_NAME, BY_CRMOD_DATETIME
+/// Sorting is always done directories-first. Directories may be also sorted by name or by their CR/MOD datetime
+///
+/// The main benefits of this implementation over the previous one are:
+/// - 8KB less RAM consumption
+/// - unlimited number of dir entries supported
+/// - start from a given filename to be displayed first - support for restoring file browser's exact content when returning from One Click Print
+template <size_t WINDOW_SIZE>
+class LazyDirView {
+public:
+    /// Types of directory entries
+    enum class EntryType : uint8_t {
+        FILE,
+        DIR
+    };
+
+    /// Entries sort policies
+    enum class SortPolicy : uint8_t {
+        BY_NAME,
+        BY_CRMOD_DATETIME ///< Sort by combined Creation and Modification time stamp.
+                          ///< Expects support in FATfs - to return the most recent time stamp from both of them
+    };
+
+    LazyDirView()
+        : filesInWindow(0)
+        , totalFiles(0)
+        , windowStartingFrom(0) {
+        std::for_each(files.begin(), files.end(), [](Entry &e) { e.Clear(); });
+    }
+
+    /// @param windowIndex index within the window (i.e. [0 - WINDOW_SIZE-1])
+    /// @return pointer to a filename (or nullptr, if there is nothing at that index) and the type of the entry (FILE / DIR)
+    std::pair<const char *, EntryType> FileNameAt(size_t windowIndex) {
+        return std::make_pair(files[windowIndex].name, files[windowIndex].isFile ? EntryType::FILE : EntryType::DIR);
+    }
+
+    /// @return total number of files/entries in a directory
+    size_t TotalFilesCount() const { return totalFiles; }
+
+    /// @return number of visible files/entries in a directory supported by this instance
+    size_t VisibleFilesCount() const { return WINDOW_SIZE; }
+
+    /// Initial population of the window + sorting is a normal insert sort algorithm with one run through the directory.
+    /// Much of the comparison is done in RAM while avoiding slower USB/FATFS interface as much possible.
+    ///
+    /// Future implementation consideration for sorting policy:
+    /// - compile time - making ChangeDirectory a template: +faster -more code size
+    /// - run time - sort policy as a parameter: -slower +less code size
+    ///
+    /// @param p directory path (will be remembered inside this class)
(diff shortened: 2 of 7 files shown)


src/gui/lazyfilelist.h
```diff
@@ -0,0 +1,329 @@
...
+            }
+        }
+        ++windowStartingFrom;
+        return true;
+    }
+
+private:
+    struct Entry {
+        bool isFile;
+        char name[96];
```
dragomirecky, line 177: I think the maximum length is slightly higher. 

What about using the `_MAX_LFN` macro from `FatFS`?

src/gui/lazyfilelist.h
```diff
@@ -0,0 +1,329 @@
...
+        uint16_t date, time;
+        void Clear() {
+            isFile = false;
+            name[0] = 0;
+            date = time = 0;
+        }
+        void CopyFrom(const FILINFO &fno) {
+            strncpy(name, fno.fname, 96);
```
dragomirecky, line 185: 1. `str*l*cpy` 😉
2. `sizeof(name)`

There are more occurrences of this in this PR.

src/gui/lazyfilelist.h
```diff
@@ -0,0 +1,329 @@
...
+            isFile = (fno.fattrib & AM_DIR) == 0;
+            date = fno.fdate;
+            time = fno.ftime;
+        }
+        void SetDirUp() {
+            name[0] = name[1] = '.';
+            name[2] = 0;
+            isFile = false;
+            date = time = USHRT_MAX;
```
dragomirecky, line 194: /pedantic/ `UINT16_MAX` 😁

src/gui/lazyfilelist.h
```diff
@@ -0,0 +1,329 @@
...
+            date = time = 0;
+        }
+    };
+
+    std::array<Entry, WINDOW_SIZE> files; ///< roughly WINDOW_SIZE * 100B, can be placed in CCRAM if needed. For MINI it is only 9 entries -> only 900B
+    size_t filesInWindow;                 ///< number of files populated in the window - less than WINDOW_SIZE when there are less files in the dir
+    size_t totalFiles;                    ///< total number of entries in the directory
+    int windowStartingFrom;               ///< from which entry index the window starts (e.g. from the 3rd file in dir).
+                                          ///< intentionally int, because -1 means ".."
+    char path[256];                       ///< current directory path - @@TODO this may not be enough - needs checking
```
dragomirecky, line 208: We already have `F_MAXPATHNAMELENGTH`... (currently defined in `window_file_list.h`). Maybe would be nice to make sure all those things are using the same value originating from one place.

src/gui/screen_filebrowser.cpp
```diff
@@ -101,57 +101,55 @@ static int screen_filebrowser_event(screen_t *screen, window_t *window,
         return 0;
     }
 
-    const FILINFO *file_item = &(filelist->file_items[(intptr_t)param]);
+    // const FILINFO *file_item = &(filelist->file_items[(intptr_t)param]);
+
```
dragomirecky, line 105: Can we remove this?

src/gui/lazyfilelist-c-api.h
```diff
@@ -0,0 +1,20 @@
...
+extern "C" {
+#endif
+
+bool LDV_ChangeDir(void *LDV, bool sortByName, const char *path);
+uint32_t LDV_TotalFilesCount(void *LDV);
+uint32_t LDV_VisibleFilesCount(void *LDV);
+bool LDV_MoveUp(void *LDV);
+bool LDV_MoveDown(void *LDV);
+const char *LDV_FileAt(void *LDV, int index, bool *isFile);
+void *LDV_Create(void);
```
dragomirecky, line 16: Also, what about adding `typedef void* LDV_ptr_t`, so random people don't run away when seeing all the `void *` around? 😁

src/gui/screen_filebrowser.cpp
```diff
@@ -101,57 +101,55 @@ static int screen_filebrowser_event(screen_t *screen, window_t *window,
...
-            if ((strlen(filelist->altpath) + strlen(file_item->altname) + 1) >= MAXPATHNAMELENGTH) {
-                LOG_ERROR("path too long");
-                return 0;
-            }
-
-            size_t len = strlen(filelist->altpath);
-            if (filelist->altpath[len - 1] != '/') {
-                filelist->altpath[len++] = '/';
+    size_t altPathLen = strlen(filelist->altpath);
+    if ((altPathLen + strlen(currentFName) + 1) >= MAXPATHNAMELENGTH) {
```
dragomirecky, line 117: What is the difference between `MAXPATHNAMELENGTH` and `F_ MAXPATHNAMELENGTH` used in the header?

src/gui/window_file_list.c
```diff
@@ -142,31 +66,27 @@ void window_file_list_draw(window_file_list_t *window) {
...
     int i;
     for (i = 0; i < visible_count && i < window->count; i++) {
-        int idx = i + window->top_index;
-        uint16_t id_icon = IDR_NULL;
-        if (idx == 0) {
-            id_icon = (!strcmp(window->altpath, "/")) ? IDR_PNG_filescreen_icon_home : IDR_PNG_filescreen_icon_up_folder;
-        } else if (window->file_items[idx].fattrib & AM_DIR) {
-            id_icon = IDR_PNG_filescreen_icon_folder;
+        bool isFile = true;
+        const char *item = LDV_FileAt(window->ldv, i, &isFile);
```
dragomirecky, line 70: `LDV_FileAt` returns `nullptr` in case there is no file at the position. 
I believe it does not „usually happen“ here, but I also think it will happen in the case of `firstDirEntry` being set to a file „at the of the directory“. Am I right?

src/gui/lazyfilelist-c-api.h
DRracer: The C-API is just a temporary solution to plant the new implementation into the existing code, it will be removed soon, so I'd stick with the void * for now.

src/gui/lazyfilelist.h
DRracer: thanks, this wasn't polished yet

src/gui/lazyfilelist.h
DRracer: I don't want to pollute lazyfilelist.h with window_file_list.h or extracting the F_MAXPATHNAMELENGTH into a separate header for now, this will get sorted out when window_file_list gets upgraded to C++.

src/gui/screen_filebrowser.cpp
DRracer: Honestly, I have no idea what were the original intentions, this has been there already.
I'll make it more consistent...

src/gui/file_list_defs.h
```diff
@@ -0,0 +1,23 @@
...
+    #define MAX_DIR_DEPTH 10 // Maximum folder depth
+#endif
+
+#ifndef MAXDIRNAMELENGTH
+    #define F_MAXDIRNAMELENGTH _MAX_LFN + 1
+#endif
+
+#ifndef MAXPATHNAMELENGTH
+    #define F_MAXPATHNAMELENGTH (1 + (F_MAXDIRNAMELENGTH + 1) * (MAX_DIR_DEPTH) + 1 + _MAX_LFN)
+#endif
```
dragomirecky, line 23: Isn't this a mistake? (the fact that the macros in `#ifdef` don't have the `F_` prefix and the ones being defined do have it)

dragomirecky: (I just noticed you extracted the code from `src/gui/window_file_list.h`, but still...)

DRracer: This was reused from the original code, in fact it's been confusing already - there were and still are 2 types of macros: with "F_" and without. They look like being interchangeable, but their real values differ in some cases. This must be improved when rewriting to c++.

PR #387, opened 2020-04-24
Marlin SD card rework

XPila:
BFW-719


src/common/marlin_client.c
```diff
@@ -140,6 +130,7 @@ void marlin_client_loop(void) {
             if (client->flags & MARLIN_CFLG_LOWHIGH) {
                 *(((uint32_t *)(&msg)) + 1) = ose.value.v; //store high dword
                 _process_client_message(client, msg);      //call handler
+                variant8_done(&msg);
```
DRracer, line 133: are you 100% sure the pointer held in `msg` is to be dealocated here calling `variant8_done`?

src/common/marlin_client.c
```diff
@@ -732,17 +666,17 @@ uint32_t _wait_ack_from_server(uint8_t client_id) {
 
 // process message on client side (set flags, update vars etc.)
 void _process_client_message(marlin_client_t *client, variant8_t msg) {
-    char var_str[16];
+    char var_str[128];
```
DRracer, line 669: Magical constant? What is the reason for making it bigger? Did you check stack usage?

src/common/marlin_vars.c
```diff
@@ -243,6 +270,18 @@ void marlin_vars_value_to_str(marlin_vars_t *vars, uint8_t var_id, char *str) {
...
+            sprintf(str, "%u", (unsigned int)(vars->media_inserted));
+            break;
+        case MARLIN_VAR_PRNSTATE:
+            sprintf(str, "%u", (unsigned int)(vars->print_state));
+            break;
+        case MARLIN_VAR_FILENAME:
+            sprintf(str, "%s", vars->media_file_name);
+            break;
+        case MARLIN_VAR_FILEPATH:
+            sprintf(str, "%s", vars->media_file_path);
```
DRracer, line 283: snprintf at least when formatting texts

src/common/marlin_client.c
XPila: yes, this is the right place - immediately after message processed.

src/common/marlin_client.c
XPila: this buffer is just for trace I will make define and limit output with snprintf.

src/wui/wui.c
```diff
@@ -47,9 +48,7 @@ void update_web_vars(void) {
     web_vars.print_dur = wui.wui_marlin_vars->print_duration;
     web_vars.sd_precent_done = wui.wui_marlin_vars->sd_percent_done;
     web_vars.sd_printing = wui.wui_marlin_vars->sd_printing;
-    if (marlin_event(MARLIN_EVT_GFileChange)) {
-        marlin_get_printing_gcode_name(web_vars.gcode_name);
-    }
+    web_vars.gcode_name = wui.wui_marlin_vars->media_file_name;
```
joshymjose, line 51: how the other clients get informed about the file change?

XPila: filename is now variable, client can check change flag for this variable:
if (marlin_change_clr(MARLIN_VAR_FILENAME))
filename variable is changed when printing started

src/common/marlin_client.c
XPila: commit aa4f145

src/common/marlin_vars.c
XPila: aa4f145

src/common/media.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+void media_print_resume(void) {
+    if (media_print_state == media_print_state_PAUSED) {
+        if (f_open(&media_print_fil, media_print_filepath, FA_READ) == FR_OK) {
+            if (f_lseek(&media_print_fil, media_current_position) == FR_OK)
+                media_print_state = media_print_state_PRINTING;
+            else
+                f_close(&media_print_fil);
+        }
+    }
```
dragomirecky, line 106: Again, what if any of these functions fail?

src/common/media.cpp
```diff
@@ -0,0 +1,172 @@
...
+}
+
+void media_loop(void) {
+    _usbhost_reenum();
+    if (media_print_state == media_print_state_PRINTING) {
+        char buffer[MAX_CMD_SIZE + 1];
+        char *pch;
+        if (!f_eof(&media_print_fil))              //check eof
+            while (queue.length < (BUFSIZE - 1)) { //keep one free slot for serial commands
+                if (f_gets(buffer, MAX_CMD_SIZE, &media_print_fil)) {
```
dragomirecky, line 128: What if `f_gets` returns two (or more) lines in the buffer? It looks to me this throws them away, but maybe I am missing something.

src/gui/guimain.c
```diff
@@ -67,6 +66,9 @@ static void _gui_loop_cb() {
     marlin_client_loop();
 }
 
+char gui_media_filename[FILE_NAME_MAX_LEN + 1];
+char gui_media_filepath[FILE_PATH_MAX_LEN + 1];
+
```
dragomirecky, line 71: `static`

src/gui/screen_home.cpp
```diff
@@ -243,8 +246,9 @@ static bool find_latest_gcode(char *fpath, int fpath_len, char *fname, int fname
 
         if ((fname[0] == 0 || is_newer) && !skip) {
             const char *short_name = current_finfo.altname[0] ? current_finfo.altname : current_finfo.fname;
-            snprintf(fpath, fpath_len, "/%s", short_name);
-            snprintf(fname, fname_len, "%s", current_finfo.fname);
+            fpath[0] = '/';
+            strncpy(fpath + 1, short_name, fpath_len - 1);
+            strncpy(fname, current_finfo.fname, fname_len);
```
dragomirecky, line 251: `strlcpy`

src/gui/screen_printing.cpp
```diff
@@ -484,9 +462,9 @@ int screen_printing_event(screen_t *screen, window_t *window, uint8_t event, voi
             if (gui_msgbox("Are you sure to stop this printing?",
                     MSGBOX_BTN_YESNO | MSGBOX_ICO_WARNING | MSGBOX_DEF_BUTTON1)
                 == MSGBOX_RES_YES) {
-                abort_print(screen);
-                screen_close();
-                return 1;
+                marlin_print_abort();
+                //                screen_close();
+                //                return 1;
```
dragomirecky, line 467: Why is it commented out?

src/common/marlin_client.c
XPila: Resource is allocated in function marlin_vars_get_var, variant8_done does nothing in case of nonpointer type or empty data. Each client will receive own copy.

include/marlin/Configuration_A3ides_2209_MINI_adv.h
```diff
@@ -1348,7 +1348,7 @@
     #define FILAMENT_CHANGE_ALERT_BEEPS 10 // Number of alert beeps to play when a response is needed.
     #define PAUSE_PARK_NO_STEPPER_TIMEOUT // Enable for XYZ steppers to stay powered on during filament change.
 
-    #define PARK_HEAD_ON_PAUSE // Park the nozzle during pause and filament change.
+//    #define PARK_HEAD_ON_PAUSE // Park the nozzle during pause and filament change.
```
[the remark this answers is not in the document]
XPila: Screen_printing will be refactored soon (4.1.0), I will keep commented code for reference.
SD card and pause and printing state machine is completely reworked in this pull, maybe the title is confusing. Radek Vana will continue to use new pause in filament load/unlad, PAUSE_FEATURE and NOZZLE_PARK_FEATURE will be completely removed just like HOST_ACTIONS. This is not done in this pull, because we need keep filament load/unload functionality in 4.1.0 beta until new implementation will be ready.

src/gui/screen_printing.cpp
XPila: This is because new implementation is waiting for the stop state, stop is now asynchronous (parking head...). Also i changed the behavior and after stop print you can now "reprint". Maybe it will be better than previous behavior, maybe we will return to previous behavior. We can discuse.

src/common/media.cpp
XPila: I think this is clear - media_print_state stay unchanged and user can repeat resume action.

src/common/media.cpp
XPila: f_gets returns single lines, please read http://elm-chan.org/fsw/ff/doc/gets.html. But other case is not solved here - lines longer than buffer size are returned by parts and this generates parser error. I will fix this in some next pull.

XPila: lines longer than buffer is issue BFW-719/4

src/common/appmain.cpp
```diff
@@ -197,30 +198,6 @@ void app_tim14_tick(void) {
...
-        if ((timer) && (hUsbHostHS.gState == HOST_ENUMERATION) && (hUsbHostHS.EnumState == ENUM_IDLE)) {
-            // longer than 500ms
-            if ((tick - timer) > 500) {
-                _dbg("USB host reenumerating"); // trace
-                USBH_ReEnumerate(&hUsbHostHS);  // re-enumerate UsbHost
-            }
-        } else // otherwise update timer
-            timer = tick;
-    }
-}
```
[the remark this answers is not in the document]
XPila: Usb reenum is now handled in module "media", usb reenum is dirty hack and will be removed (i hope) after usb driver update. Functionality is unchanged.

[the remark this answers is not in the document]
dragomirecky: I just realized you only moved this to another file. My bad!

include/marlin/Configuration_A3ides_2209_MINI_adv.h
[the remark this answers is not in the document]
dragomirecky: Ok!

src/common/media.h
```diff
@@ -0,0 +1,52 @@
...
+
+extern media_print_state_t media_print_get_state(void);
+
+extern uint32_t media_print_get_size(void);
+
+extern uint32_t media_print_get_position(void);
+
+extern void media_loop(void);
+
+extern void media_set_inserted(uint8_t inserted);
```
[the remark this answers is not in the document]
dragomirecky: Do you disagree with me, that `bool` should be used for boolean values?

Or do you just think, that this can be refactored even further – into two separate functions?

src/common/media.cpp
dragomirecky: Yep, that's what I was missing. Seems fine then :)

src/common/media.h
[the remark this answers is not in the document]
RadekVana: I try to avoid `bool` parameters if possible.
I especially do not like `void set_something(bool set);`
I have nothing against `bool `return value or variables.
In this case I think it should be refactored even further.

I would have nothing against
`void media_set_state(some_enum state);`
since it can be extended with more states like disabled.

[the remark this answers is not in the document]
RadekVana: But 

> void media_set_inserted();
> void media_clear_inserted();

seems most clear to me.

src/common/marlin_client.c
```diff
@@ -16,22 +16,16 @@
...
-						MARLIN_VAR_MSK(MARLIN_VAR_POS_X) | \
-						MARLIN_VAR_MSK(MARLIN_VAR_POS_Y) | \
-						MARLIN_VAR_MSK(MARLIN_VAR_POS_Z) | \
-						MARLIN_VAR_MSK(MARLIN_VAR_POS_E) )*/
+//trace variable change notifications (client side), to disable trace undef DBG_VAR_MSK
+#define DBG_VAR     DBG
+#define DBG_VAR_MSK (MARLIN_VAR_MSK_ALL & ~MARLIN_VAR_MSK_TEMP_ALL)
+
+//maximum string length for DBG_VAR
+#define DBG_VAR_STR_MAX_LEN 128
```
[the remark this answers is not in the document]
XPila: fixed commit 97212a6

src/common/marlin_vars.c
```diff
@@ -177,74 +188,102 @@ void marlin_vars_set_var(marlin_vars_t *vars, uint8_t var_id, variant8_t var) {
...
+            break;
+        case MARLIN_VAR_FILENAME:
+            snprintf(str, size, "%s", vars->media_file_name);
+            break;
+        case MARLIN_VAR_FILEPATH:
+            snprintf(str, size, "%s", vars->media_file_path);
             break;
         default:
-            sprintf(str, "???");
+            snprintf(str, size, "???");
```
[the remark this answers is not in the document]
XPila: commit 472123a
in real this cannot happened because filename and filepath is read only, but I fixed it. We plan replace marlin_server input character queue with same message queue used for clients, so all this code will be deleted, but we cannot manage this in 4.1.0 beta.

src/common/marlin_client.c
```diff
@@ -582,6 +543,16 @@ void marlin_quick_stop(void) {
     _wait_ack_from_server(client->id);
 }
 
+void marlin_print_start(const char *filename) {
+    char request[MARLIN_MAX_REQUEST];
+    marlin_client_t *client = _client_ptr();
+    if (client == 0)
+        return;
+    sprintf(request, "!pstart %s", filename);
```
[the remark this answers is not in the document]
XPila: commit c313329

src/gui/screen_home.cpp
XPila: commit 7abfd1b

src/common/appmain.cpp
```diff
@@ -28,12 +28,14 @@
...
 #include "safe_state.h"
 #include "crc32.h"
+#include "ff.h"
 
 #include <Arduino.h>
 #include "trinamic.h"
 #include "../Marlin/src/module/configuration_store.h"
 #include "../Marlin/src/module/temperature.h"
 #include "../Marlin/src/module/probe.h"
+#include "../Marlin/src/gcode/queue.h"
```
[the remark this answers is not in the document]
XPila: commit  97730a9
removed also temperature and probe

src/common/marlin_vars.c
```diff
@@ -30,7 +30,10 @@ const char *__var_name[] = {
     "SD_PRINT",
     "SD_PDONE",
     "DURATION",
-    "FSENSOR",
+    "MEDIAINS",
+    "PRN_STAT",
+    "FILENAME",
+    "FILEPATH",
 };
```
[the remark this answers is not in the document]
XPila: commit 8f5b4e5

src/common/marlin_server.cpp
```diff
@@ -37,31 +36,25 @@
...
 #pragma pack(push)
 #pragma pack(1)
 
 typedef struct _marlin_server_t {
-    char gcode_name[GCODE_NAME_MAX_LEN + 1];         // printing gcode name
-    uint16_t flags;                                  // server flags (MARLIN_SFLG)
-    uint64_t notify_events[MARLIN_MAX_CLIENTS];      // event notification mask
-    uint64_t notify_changes[MARLIN_MAX_CLIENTS];     // variable change notification mask
-    marlin_vars_t vars;                              // cached variables
-    char request[MARLIN_MAX_REQUEST];                //
```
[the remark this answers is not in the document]
XPila: maybe we should adjust clang-format, I thing comments are ALWAYS better formatted by hand.

src/common/marlin_server.cpp
```diff
@@ -322,7 +314,6 @@ int marlin_server_idle(void) {
     else if ((marlin_server.flags & MARLIN_SFLG_BUSY) == 0) {
         //_dbg("SVR: BUSY");
         marlin_server.flags |= MARLIN_SFLG_BUSY;
-        _send_notify_event(MARLIN_EVT_Busy, 0, 0);
         if (parser.command_letter == 'G')
             switch (parser.codenum) {
             case 28:
```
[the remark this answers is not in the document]
XPila: i accepted that your solution is better, but we will remove this part soon, notification should be send directly from implementation of each command.

src/common/media.h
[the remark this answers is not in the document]
dragomirecky: I understand your point and I even agree, that splitting it is the nicer alternative 👍

I was just surprised you disagreed with me, while the only thing I said is „Use bool for boolean values.“. Do you not agree on using `bool` to represent boolean values?

My only point was, that I too often see boolean values being stored in `uint8_t` instead of `bool`.

src/common/marlin_server.cpp
```diff
@@ -157,27 +145,29 @@ void _server_update_pqueue(void);
...
+void _server_update_and_notify(int client_id, uint64_t update);
 
 //-----------------------------------------------------------------------------
 // server side functions
 
 void marlin_server_init(void) {
     int i;
     memset(&marlin_server, 0, sizeof(marlin_server_t));
-    osMessageQDef(serverQueue, 64, uint8_t);
+    osMessageQDef(serverQueue, 128, uint8_t);
```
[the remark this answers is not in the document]
XPila: commit 78bf4d1
also client queue size added to config

src/common/marlin_server.cpp
```diff
@@ -914,54 +994,75 @@ int _process_server_request(char *request) {
...
             }
         }
     }
     //	_dbg("_server_set_var %d %s %s", var_id, name_val_str, val_str);
     return 1;
 }
 
+// update variables defined by 'update' mask and send notification to client that requested updating
+// other clients will receive notification in next cycle
+void _server_update_and_notify(int client_id, uint64_t update) {
```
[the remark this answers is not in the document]
XPila: commit 145a2c6
all private client/server functions changed to "static"

src/common/media.cpp
```diff
@@ -0,0 +1,172 @@
...
+            }
+        } else // otherwise update timer
+            timer = tick;
+    }
+}
+
+uint8_t media_inserted = 0;
+media_error_t media_error = media_error_OK;
+media_print_state_t media_print_state = media_print_state_NONE;
+char media_print_filename[128] = { 0 };
```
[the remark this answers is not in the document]
XPila: commit 2aea0d3

src/gui/screen_home.cpp
```diff
@@ -172,17 +172,20 @@ int screen_home_event(screen_t *screen, window_t *window, uint8_t event, void *p
     if (p_window_header_event_clr(&(pw->header), MARLIN_EVT_MediaInserted) &&
 
         (HAL_GetTick() > 5000)) {
```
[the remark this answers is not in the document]
XPila: commit 2aea0d3

src/common/media.h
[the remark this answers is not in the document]
XPila: I am not happy with bool, finally I have chosen solution with enum.

src/common/marlin_server.cpp
```diff
@@ -76,12 +69,12 @@ typedef struct _marlin_server_t {
     marlin_mesh_t mesh;                              // meshbed leveling
     uint64_t mesh_point_notsent[MARLIN_MAX_CLIENTS]; // mesh point mask (points that are not sent)
     uint64_t update_vars;                            // variable update mask
+    marlin_print_state_t print_state;
+    float resume_pos[4];
```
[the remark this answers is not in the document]
XPila: doc added

src/common/marlin_vars.h
```diff
@@ -29,7 +29,11 @@
 #define MARLIN_VAR_SD_PRINT 0x15 // R:  uint8, card.flag.sdprinting
 #define MARLIN_VAR_SD_PDONE 0x16 // R:  uint8, card.percentDone()
 #define MARLIN_VAR_DURATION 0x17 // R:  uint32, print_job_timer.duration()
-#define MARLIN_VAR_MAX      MARLIN_VAR_DURATION
+#define MARLIN_VAR_MEDIAINS 0x18 // R:  uint8, media_is_inserted()
+#define MARLIN_VAR_PRNSTATE 0x19 // R:  marlin_print_state_t, marlin_server.print_state
+#define MARLIN_VAR_FILENAME 0x1a // R:  char*,
+#define MARLIN_VAR_FILEPATH 0x1b // R:  char*,
+#define MARLIN_VAR_MAX      MARLIN_VAR_FILEPATH
```
RadekVana, line 37: `static_assert(MARLIN_VAR_MAX < 64, "Too many variables");`

src/common/marlin_vars.h
```diff
@@ -50,13 +54,13 @@
...
-    MARLIN_VAR_MSK(MARLIN_VAR_MOTION) | MARLIN_VAR_MSK(MARLIN_VAR_GQUEUE) | MARLIN_VAR_MSK(MARLIN_VAR_PQUEUE) | MARLIN_VAR_MSK_IPOS_XYZE | MARLIN_VAR_MSK_POS_XYZE | MARLIN_VAR_MSK_TEMP_ALL | MARLIN_VAR_MSK(MARLIN_VAR_Z_OFFSET) | MARLIN_VAR_MSK(MARLIN_VAR_FANSPEED) | MARLIN_VAR_MSK(MARLIN_VAR_PRNSPEED) | MARLIN_VAR_MSK(MARLIN_VAR_FLOWFACT) | MARLIN_VAR_MSK(MARLIN_VAR_WAITHEAT) | MARLIN_VAR_MSK(MARLIN_VAR_WAITUSER) | MARLIN_VAR_MSK(MARLIN_VAR_SD_PRINT) | MARLIN_VAR_MSK(MARLIN_VAR_SD_PDONE) | MARLIN_VAR_MSK(MARLIN_VAR_DURATION))
+    MARLIN_VAR_MSK(MARLIN_VAR_MOTION) | MARLIN_VAR_MSK(MARLIN_VAR_GQUEUE) | MARLIN_VAR_MSK(MARLIN_VAR_PQUEUE) | MARLIN_VAR_MSK_IPOS_XYZE | MARLIN_VAR_MSK_POS_XYZE | MARLIN_VAR_MSK_TEMP_ALL | MARLIN_VAR_MSK(MARLIN_VAR_Z_OFFSET) | MARLIN_VAR_MSK(MARLIN_VAR_FANSPEED) | MARLIN_VAR_MSK(MARLIN_VAR_PRNSPEED) | MARLIN_VAR_MSK(MARLIN_VAR_FLOWFACT) | MARLIN_VAR_MSK(MARLIN_VAR_WAITHEAT) | MARLIN_VAR_MSK(MARLIN_VAR_WAITUSER) | MARLIN_VAR_MSK(MARLIN_VAR_SD_PRINT) | MARLIN_VAR_MSK(MARLIN_VAR_SD_PDONE) | MARLIN_VAR_MSK(MARLIN_VAR_DURATION) | MARLIN_VAR_MSK(MARLIN_VAR_PRNSTATE) | MARLIN_VAR_MSK(MARLIN_VAR_FILENAME) | MARLIN_VAR_MSK(MARLIN_VAR_FILEPATH))
```
RadekVana, line 63:
```
#include <limits.h>
#if MARLIN_VAR_MAX  == 63  
#define MARLIN_VAR_MSK_ALL ULLONG_MAX
#else
#define MARLIN_VAR_MSK_ALL ((((uint64_t)1)<<(MARLIN_VAR_MAX+1))-1)
#endif
```

RadekVana: no need for that `#if`
`#define MARLIN_VAR_MSK_ALL ((((uint64_t)1)<<(MARLIN_VAR_MAX+1))-(uint64_t)1)`
shold do since 1 << 64 == 0 and 0-1 ==   MARLIN_VAR_MAX

src/common/marlin_client.c
DRracer: ok, in this case the code path seems to be deterministic

src/common/marlin_vars.c
```diff
@@ -98,6 +101,14 @@ variant8_t marlin_vars_get_var(marlin_vars_t *vars, uint8_t var_id) {
             return variant8_ui8(vars->sd_percent_done);
         case MARLIN_VAR_DURATION:
             return variant8_ui32(vars->print_duration);
+        case MARLIN_VAR_MEDIAINS:
+            return variant8_ui8(vars->media_inserted);
+        case MARLIN_VAR_PRNSTATE:
+            return variant8_ui8(vars->print_state);
+        case MARLIN_VAR_FILENAME:
+            return variant8_pchar(vars->media_file_name, 0, 1);
```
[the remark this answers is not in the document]
DRracer: That's the dangerous part of the API - init==true means, the variant shall make a copy of the input string.
THE most deadly part of it is, that the variant doesn't remember, if it own the memory - thus the programmer must decide, whether to call variant_done() on destruction or not.
Everyone, who had to do something with strings and variants so far, failed to do it well.

src/common/marlin_vars.h
DRracer: there is even a non-standard static assert for plain C in gcc

src/wui/wui_vars.h
```diff
@@ -7,7 +7,7 @@
 #define Y_AXIS_POS 1
 #define Z_AXIS_POS 2
 typedef struct {
-    char gcode_name[GCODE_NAME_MAX_LEN + 1];
+    char *gcode_name;
```
DRracer, line 10: @joshymjose @XPila 
in case this var was previously used only as a temporary buffer to store the filename being printed, sholdn't it now be a const char * to prevent accidental write into it?

src/usb_host.c
```diff
@@ -132,19 +131,17 @@ static void USBH_UserProcess(USBH_HandleTypeDef *phost, uint8_t id) {
...
         FRESULT result = f_mount(&USBHFatFS, (TCHAR const *)USBHPath, 0);
         if (result == FR_OK)
-            media_set_inserted(true);
-        else {
-            // TODO: inform the user
-            _dbg("disk not mounted; error: %d", result);
-        }
+            media_set_inserted();
+        else
+            media_set_error(media_error_MOUNT);
```
DRracer, line 144: How do we visualize this for the user? Do we have an icon for this state, so that our tech-support guys can immediately tell the user, that his USB flash wasn't recognized?

src/gui/wizard/xyzcalib.c
```diff
@@ -80,9 +79,8 @@ int xyzcalib_xy_measure(int16_t id_body, xyzcalib_screen_t *p_screen, xyzcalib_d
             "points");
         window_set_icon_id(p_screen->icon.win.id, IDR_PNG_wizard_icon_measure);
         marlin_gcode("G29");
-        while (!(marlin_busy() || marlin_motion()))
-            osDelay(10);
-    } else if (!(marlin_busy() || marlin_motion()))
+        marlin_event_clr(MARLIN_EVT_CommandEnd);
+    } else if (marlin_event_clr(MARLIN_EVT_CommandEnd))
```
DRracer, line 83: error-prone and hard-to-read - I'm missing the { } for this if-block.

src/gui/window_header.c
```diff
@@ -58,7 +56,7 @@ void window_header_init(window_header_t *window) {
     window->icons[HEADER_ICON_WIFI] = HEADER_ISTATE_OFF;
     window->label = NULL;
 
-    if (media_is_inserted()) {
+    if (marlin_vars()->media_inserted) {
```
DRracer, line 59: what shall be displayed in case of a mount-error?

src/gui/screen_printing.cpp
```diff
@@ -779,11 +743,49 @@ void set_stop_icon_and_label(screen_t *screen) {
 }
 
 void change_print_state(screen_t *screen, printing_state_t st) {
-    _dbg("printstate %d entered", (int)st);
+    //_dbg("printstate %d entered", (int)st);
```
DRracer, line 746: is this comment block still relevant?

src/gui/screen_printing.cpp
```diff
@@ -404,6 +376,10 @@ int screen_printing_event(screen_t *screen, window_t *window, uint8_t event, voi
         close_popup_message(screen);
     }
 
+    /*    if (Lock::IsLocked())
```
DRracer, line 379: @RadekVana do you have prepared some cleaning-up of the commented code here in the GUI?

src/gui/screen_menu_calibration.c
```diff
@@ -78,13 +86,25 @@ int screen_menu_calibration_event(screen_t *screen, window_t *window, uint8_t ev
...
+            while (!marlin_event_clr(MARLIN_EVT_CommandBegin))
+                marlin_client_loop();
+            gui_dlg_wait(gui_marlin_G28_or_G29_in_progress);
             break;
         case MI_MESH_BED:
-            marlin_gcode("G28");
+            if (!marlin_all_axes_homed()) {
+                marlin_event_clr(MARLIN_EVT_CommandBegin);
+                marlin_gcode("G28");
+                while (!marlin_event_clr(MARLIN_EVT_CommandBegin))
```
DRracer, line 99: does marlin_event_clr() switch RTOS tasks, so that the CPU can do something else while waiting here (burning CPU cycles for nothing)?

src/gui/screen_menu_calibration.c
```diff
@@ -78,13 +86,25 @@ int screen_menu_calibration_event(screen_t *screen, window_t *window, uint8_t ev
...
+                    marlin_client_loop();
```
DRracer, line 100: I don't like these loops with explicit calls to marlin_client_loop - are they really necessary?

src/wui/wui_vars.h
joshymjose: this will break the code. original code keeps copy of the file change in WUI thread.

joshymjose: at the line 83 in wui_api.c the web_vars.gcode_name used as char string. now it is pointer to char string. hopefully it does not break?

src/common/marlin_client.c
```diff
@@ -732,17 +669,17 @@ uint32_t _wait_ack_from_server(uint8_t client_id) {
...
-    char var_str[16];
+    char var_str[DBG_VAR_STR_MAX_LEN + 1];
     uint8_t id = msg.usr8 & MARLIN_USR8_MSK_ID;
     if (msg.usr8 & MARLIN_USR8_VAR_FLG) // variable change received
     {
         marlin_vars_set_var(&(client->vars), id, msg);
         client->changes |= ((uint64_t)1 << id);
-        marlin_vars_value_to_str(&(client->vars), id, var_str);
 #ifdef DBG_VAR_MSK
+        marlin_vars_value_to_str(&(client->vars), id, var_str, sizeof(var_str) - 1);
```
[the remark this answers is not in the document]
XPila: commit 67db45e (+-1)
buffer is used just for trace, I want improve this trace in future using cvariant8 (without buffer).

src/common/media.h
```diff
@@ -0,0 +1,52 @@
...
+    media_print_state_PAUSED = 2,
+} media_print_state_t;
+
+#ifdef __cplusplus
+extern "C" {
+#endif //__cplusplus
+
+extern char media_print_filename[128];
+
+extern char media_print_filepath[128];
```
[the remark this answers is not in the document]
XPila: fixed

src/common/media.cpp
```diff
@@ -0,0 +1,172 @@
...
+char media_print_filepath[128] = { 0 };
+FIL media_print_fil;
+uint32_t media_current_position;
+uint32_t media_current_line;
+uint32_t media_queue_position[BUFSIZE];
+uint32_t media_queue_line[BUFSIZE];
```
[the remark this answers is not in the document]
XPila: commit d58e83b

src/common/marlin_server.cpp
```diff
@@ -37,31 +36,25 @@
...
 
-#define DBG_HOST DBG //trace host_actions
-//#define DBG_HOST(...)   //disable trace
-
-// host_actions definitions
-#define HOST_PROMPT_LEN_MAX 32 // max 32 chars prompt text
-#define HOST_BUTTON_LEN_MAX 16 // max 16 chars button text
-#define HOST_BUTTON_CNT_MAX 4  // max 4 buttons
+//#define DBG_FSM DBG //trace fsm
+#define DBG_FSM(...) //disable trace
```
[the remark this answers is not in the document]
XPila: commit 3caeea4

src/common/marlin_server.cpp
```diff
@@ -428,35 +410,149 @@ void marlin_server_quick_stop(void) {
...
-        planner.set_position_mm(x, y, z, 0);
-        xyz_pos_t park_point = NOZZLE_PARK_POINT;
-        nozzle.park(2, park_point);
+        planner.synchronize();
+        marlin_server.resume_pos[0] = current_position.x;
+        marlin_server.resume_pos[1] = current_position.y;
+        marlin_server.resume_pos[2] = current_position.z;
+        marlin_server.resume_pos[3] = current_position.e;
+        current_position.e -= 2 / planner.e_factor[active_extruder];
+        line_to_current_position(2000);
```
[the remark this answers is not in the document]
XPila: commit b7bd061

src/common/marlin_server.cpp
```diff
@@ -836,6 +918,7 @@ int _process_server_request(char *request) {
     uint32_t msk32[2];
     float offs;
     int ival;
+    int ival2;
```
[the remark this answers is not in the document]
XPila: commit 5e0dc53

src/common/marlin_server.cpp
```diff
@@ -904,7 +981,10 @@ int _process_server_request(char *request) {
         processed = 1;
     } else if (sscanf(request, "!change_msk %08lx %08lx", msk32 + 0, msk32 + 1)) {
         marlin_server.notify_changes[client_id] = msk32[0] + (((uint64_t)msk32[1]) << 32);
+        marlin_server.client_changes[client_id] = msk32[0] + (((uint64_t)msk32[1]) << 32);
         processed = 1;
+    } else {
+        //TODO: BSOD
```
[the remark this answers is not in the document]
XPila: commit 5e0dc53

src/common/marlin_vars.c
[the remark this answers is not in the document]
XPila: added doc - commit 6c8e4fa
Function variant8_done is problem only in C, C++ encapsulation is safe from this point of view.

src/common/media.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+uint8_t media_is_inserted(void) {
+    return media_inserted;
+}
+
+void media_print_start(const char *filepath) {
+    FILINFO filinfo;
+    if (media_print_state == media_print_state_NONE) {
+        strncpy(media_print_filepath, filepath, sizeof(media_print_filepath) - 1);
```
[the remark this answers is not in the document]
XPila: strlcpy already used

src/common/media.cpp
```diff
@@ -0,0 +1,172 @@
...
+        if (f_stat(media_print_filepath, &filinfo) == FR_OK) {
+            strncpy(media_print_filepath, filinfo.fname, sizeof(media_print_filepath) - 1);
```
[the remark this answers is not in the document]
XPila: strlcpy already used

src/common/media.cpp
```diff
@@ -0,0 +1,172 @@
...
+            if (f_open(&media_print_fil, media_print_filepath, FA_READ) == FR_OK) {
+                media_current_position = 0;
+                media_current_line = 0;
+                media_print_state = media_print_state_PRINTING;
+            }
+        }
+    }
+}
```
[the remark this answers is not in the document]
XPila: my comment is in next conversation, in future we can add more error states

src/common/marlin_vars.h
XPila: commit 7daf3c4

src/common/marlin_vars.h
XPila: Marlin var definitions will be replaced with enum just like events. I think mask_all will be not necessary.

src/wui/wui_vars.h
XPila: it is pointer and by default is set to null - in this case matlin api will not handle notifications with filename. Clients that need receive filename notifications must allocate this pointer. This is done for wui in this pulrequest (struct web_client_t, member media_file_name).

src/usb_host.c
XPila: We can improve this - now there is 3 state logic - inserted/removed/error. Maybe some next pull :)

src/gui/wizard/xyzcalib.c
XPila: {} is here, maybe the diff is confusing

src/gui/window_header.c
XPila: now it is not solved, this is part of header/footer rework.

src/gui/screen_printing.cpp
XPila: I commented this line out because it filled dbg output too much. This source will be refactored soon, we can keep commented/dead code for now.

src/gui/screen_printing.cpp
XPila: similar to previous conversation - i thing it is better to keep dead/unused code for reference and remove later.

src/common/marlin_vars.c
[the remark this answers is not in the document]
dragomirecky: Ok. I don't think this is related to this PR specifically, it is more of an ongoing issue to solve...

src/gui/screen_printing.cpp
RadekVana: Lock part should not be needed, since after Roberts change there are no more guiloops.
I do not care about that commented code.
To answer @DRracer, no I do not.

src/common/marlin_client.c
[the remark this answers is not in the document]
dragomirecky: /nonblocking/ The buffer still has a size of `+1`, which is unnecessary (`snprintf` always terminates the string).

src/gui/screen_menu_calibration.c
XPila: This is just synchronisation (waiting for command start). In real the loop is called 1-2 times, because command started immediately. In future this should be replaced with new FSM from Radek.

src/common/media.cpp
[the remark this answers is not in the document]
dragomirecky: The other comment does not really apply here. 

For example, if the `f_open` fails, you leave „the media“ in an undefined state (e.g. `media_print_size` is already set).

src/gui/screen_printing.cpp
dragomirecky: I am very much against having commented code in master – we have GIT to deal with those situations. But let's leave it for now; we can discuss this again after the git training next week :)

src/common/media.cpp
```diff
@@ -0,0 +1,182 @@
...
+media_state_t media_get_state(void) {
+    return media_state;
+}
+
+void media_print_start(const char *filepath) {
+    FILINFO filinfo;
+    if (media_print_state == media_print_state_NONE) {
+        strncpy(media_print_filepath, filepath, sizeof(media_print_filepath) - 1);
+        if (f_stat(media_print_filepath, &filinfo) == FR_OK) {
+            strncpy(media_print_filepath, filinfo.fname, sizeof(media_print_filepath) - 1);
```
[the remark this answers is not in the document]
XPila: this is already fixed

PR #390, opened 2020-04-27
New handy libray for C

espr14:
Will be used in following PRs.


src/common/cmath_ext.h
```diff
@@ -0,0 +1,46 @@
...
+// Library of flexible functions not defined in standard C99
+//
+
+#include <stdlib.h>
+
+#ifndef CMATH_EXT
+
+    #define MAX(a, b) \
+        ({ __typeof__ (a) _a = (a); \
+    __typeof__ (b) _b = (b); \
```
DRracer, line 11: What do you need those temporaries _a and _b for?
The ternary operator must have one return type - this won't solve the case if the macro is used as MAX(1, 2.0)

espr14: This solves MAX( n--, i++). I haven't found type-safe ternary macro but this might work. In case of different types it will report an error and return 0.

 #define MAX(a, b) \
({ __typeof__ (a) _a = (a); \
__typeof__ (b) _b = (b); \
__builtin_types_compatible_p(__typeof__ (a), __typeof__ (b)) ? \
 _a >= _b ? _a : _b; \
: report_error(); }) \

PR #392, opened 2020-04-27
Fix line length

espr14:


src/common/MindaRedscreen.c
```diff
@@ -101,7 +101,7 @@ void mbl_error(uint16_t moves, uint16_t points) {
     display->clear(COLOR_RED_ALERT);
 
     display->draw_text(rect_ui16(PADDING, PADDING, X_MAX, 22), "MBL ERROR", gui_defaults.font, COLOR_RED_ALERT, COLOR_WHITE);
-    display->draw_line(point_ui16(PADDING, 30), point_ui16(display->w - PADDING, 30), COLOR_WHITE);
+    display->draw_line(point_ui16(PADDING, 30), point_ui16(display->w - 1 - PADDING, 30), COLOR_WHITE);
```
DRracer, line 104: Is this to be merged after the draw_line's last point fix or before?

espr14: It can be merged anytime but both have to merged to give intentional result.

PR #397, opened 2020-04-27
Refactoring DialogStateful.cpp

espr14:


src/gui/dialogs/DialogStateful.cpp
```diff
@@ -44,11 +44,14 @@ IDialogStateful::~IDialogStateful() {
 }
 
 void IDialogStateful::draw_frame() {
-    rect_ui16_t rc = rect;
-    display->draw_line(point_ui16(rc.x, rc.y), point_ui16(239, rc.y), COLOR_GRAY);
-    display->draw_line(point_ui16(rc.x, rc.y), point_ui16(rc.x, 320 - 67), COLOR_GRAY);
-    display->draw_line(point_ui16(239, rc.y), point_ui16(239, 320 - 67), COLOR_GRAY);
-    display->draw_line(point_ui16(rc.x, 320 - 67), point_ui16(239, 320 - 67), COLOR_GRAY);
+    const rect_ui16_t rc = rect;
+    const auto w = display->w;
```
RadekVana, line 48: I would not use `auto`
`w - 1` will use it as int anyway

consider `const uint16_t  right = display->w - uint16_t (1);` instead `w` since there is 4x `w - 1`, 
same apply for `h`

DRracer: @RadekVana @espr14 I agree, please name a proper type if possible. It is much more readable then (even though there are situations, when one doesn't even want to know the name of a type ;) )

PR #406, opened 2020-04-29
Refactoring window_temp_graph

espr14:
Lot more work can be done here (variable width / height etc.) but let's keep it for another time.


.vscode/settings.json
```diff
@@ -18,5 +18,8 @@
     ],
     "cmake.buildDirectory": "${workspaceFolder}/build-vscode",
     "cmake.cmakePath": "${workspaceFolder}/.dependencies/cmake-3.15.5/bin/cmake",
-    "cmake.generator": "Ninja"
+    "cmake.generator": "Ninja",
```
DRracer, line 21: don't push changes to settings.json

PR #411, opened 2020-04-29
FIX GUI FREEZE

RadekVana:
always send MARLIN_EVT_Acknowledge
added static to "private" functions
BFW-745


RadekVana:
fixed in different PR

PR #432, opened 2020-05-11
SFN path format into a filebrowser

Chleba:
Moved getting SFN path into a filebrowser, because that's where is stored into a marlin_vars request when start printing. Media.cpp is getting path variable from there - so previous solution (BFW-468) was not effective for deeper nested files because of length.

BFW-795


src/common/media.cpp
```diff
@@ -51,51 +51,10 @@ media_state_t media_get_state(void) {
     return media_state;
 }
 
-void media_get_sfn_path(char *sfn, const char *filepath) {
```
DRracer, line 54: Why was this moved into GUI file browser? This is an operation of the file system, it has nothing to do with UI, it must be kept here.

PR #446, opened 2020-05-13
BFW-689

espr14:


espr14:
Let me know if I should squash it. Otherwise, it will be squashed during merging PR.

src/common/marlin_client.c
```diff
@@ -144,182 +148,196 @@ void marlin_client_loop(void) {
...
 int marlin_wait_motion(uint32_t timeout) {
-    marlin_vars_t *vars = marlin_update_vars(MARLIN_VAR_MSK(MARLIN_VAR_MOTION));
-    uint32_t tick = HAL_GetTick();
+    const marlin_vars_t *vars = marlin_update_vars(MARLIN_VAR_MSK(MARLIN_VAR_MOTION));
+    const uint32_t tick = HAL_GetTick();
     while ((vars->motion == 0) && ((HAL_GetTick() - tick) < timeout))
         marlin_update_vars(MARLIN_VAR_MSK(MARLIN_VAR_MOTION));
-    return (vars->motion) ? 1 : 0;
+    return SIGN(vars->motion);
 }
```
XPila, line 259: I thing use SIGN is confusing, motion is bitmask, original code is more readable.

espr14: What about `return (bool) vars->motion`?

DRracer: that's not the same - please keep the original code.

PR #468, opened 2020-05-18
BFW-770 Refactor to snprintf

espr14:
snprintf is safer than sprintf

### DIFF
--- a/src/common/support_utils.cpp
+++ b/src/common/support_utils.cpp
@@ -15,65 +15,65 @@ char *eofstr(char *str) {
     return (str + strlen(str));
 }
 
-void block2hex(char *str, uint8_t *pdata, size_t length) {
+void block2hex(char *str, uint32_t str_size, uint8_t *pdata, size_t length) {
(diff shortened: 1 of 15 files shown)


src/common/support_utils.cpp
```diff
@@ -15,65 +15,69 @@ char *eofstr(char *str) {
     return (str + strlen(str));
 }
 
-void block2hex(char *str, uint8_t *pdata, size_t length) {
+void block2hex(char *str, uint32_t str_size, uint8_t *pdata, size_t length) {
     for (; length > 0; length--)
-        sprintf(eofstr(str), "%02X", *(pdata++));
+        snprintf(eofstr(str), str_size - strlen(str), "%02X", *(pdata++));
```
DRracer, line 20: Since snprintf actually returns the number of characters, we may avoid computing strlen at all. Moreover when we do it twice now for obvious reasons. I doubt the compiler can optimize the strlen calls into a single one.

PR #474, opened 2020-05-18
BFW-770 Refactor 2 vsnprintf

espr14:


src/common/dbg.c
```diff
@@ -46,7 +46,7 @@ void _dbg_swo(const char *fmt, ...) {
     char line[DBG_MAXLINE];
     va_list va;
     va_start(va, fmt);
-    int len = vsprintf(line, fmt, va);
+    unsigned int len = vsnprintf(line, DBG_MAXLINE, fmt, va);
```
DRracer, line 49: why unsigned when vsnprintf returns int?

src/common/uartslave.c
```diff
@@ -127,10 +127,11 @@ void uartslave_cycle(uartslave_t *pslave) {
 }
 
 int uartslave_printf(uartslave_t *pslave, const char *fmt, ...) {
-    char text[32];
+    unsigned int text_len = 32;
```
DRracer, line 130: static const is probably better

PR #483, opened 2020-05-20
BFW 238 QR codes

espr14:


CMakeLists.txt
```diff
@@ -454,7 +454,8 @@ target_sources(
           src/common/marlin_server.cpp
           src/common/sim_motion.cpp
           src/common/st25dv64k.c
-          $<$<STREQUAL:${INIT_TRINAMIC_FROM_MARLIN_ONLY},NO>:src/common/trinamic.cpp>
+          src/common/trinamic.cpp
+          src/common/trinamic.h
           src/common/w25x.c
```
XPila, line 459: why is header file here?

espr14: trinamic.h is in the maseter for 5 moths already but not in CMakeLists. Should it be deleted?

PR #485, opened 2020-05-21
TMC stallguard sampling in measurement thread

XPila:
BFW-844

### DIFF
--- a/lib/Arduino_Core_A3ides/cores/arduino/HardwareSerial.cpp
+++ b/lib/Arduino_Core_A3ides/cores/arduino/HardwareSerial.cpp
@@ -15,6 +15,8 @@ extern "C" {
 
 extern UART_HandleTypeDef huart2;
 extern DMA_HandleTypeDef hdma_usart2_rx;
+extern osThreadId uart2_thread;
+extern int32_t uart2_signal;
 
 uint8_t rbuff[10];
 unsigned int rbufc = 0;
@@ -63,6 +65,7 @@ size_t HardwareSerial::write(uint8_t c) {
     if ((cnt > 2) && (buf[2] & 0x80))
         len = 8;
     if (cnt >= len) {
+        uart2_thread = osThreadGetId();
         int retry = 3;
         while (retry--) {
             if (len == 4)
@@ -74,7 +77,7 @@ size_t HardwareSerial::write(uint8_t c) {
             {
                 DBG("tx %02x %02x %02x %02x", buf[0], buf[1], buf[2], buf[3]);
                 osEvent ose;
-                if ((ose = osSignalWait(4, 100)).status == osEventSignal) {
+                if ((ose = osSignalWait(uart2_signal, 100)).status == osEventSignal) {
                     DBG("signal, received %d", 12);
                     DBG("rx %02x %02x %02x %02x %02x %02x %02x %02x %02x %02x %02x %02x", buf2[0], buf2[1], buf2[2], buf2[3], buf2[4], buf2[5], buf2[6], buf2[7], buf2[8], buf2[9], buf2[10], buf2[11]);
                     memcpy(rbuff, buf2 + 4, 8);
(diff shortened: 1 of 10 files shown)


lib/Arduino_Core_A3ides/cores/arduino/HardwareSerial.cpp
```diff
@@ -15,6 +15,8 @@ extern "C" {
 
 extern UART_HandleTypeDef huart2;
 extern DMA_HandleTypeDef hdma_usart2_rx;
+extern osThreadId uart2_thread;
+extern int32_t uart2_signal;
```
DRracer, line 19: Do we really need to screw up code again with externs in cpp files? Why isn't there a header file that would describe the necessary data types and items.

src/common/thread_measurement.c
```diff
@@ -3,14 +3,31 @@
 #include "cmsis_os.h" //osDelay
 #include "filament_sensor.h"
 #include "marlin_client.h"
+#include "trinamic.h"
 
 void StartMeasurementTask(void const *argument) {
     marlin_client_init();
+
+    // wait for startup
+    while (!marlin_event_clr(MARLIN_EVT_StartProcessing)) {
```
DRracer, line 12: This is a common function, which may be used by other tasks too. I recommend extracting the implementation into e.g.
`marlin_client_wait_for_start_processing()` or similar.

src/common/trinamic.cpp
```diff
@@ -5,27 +5,24 @@
...
-extern TMC2209Stepper stepperZ;
-extern TMC2209Stepper stepperE0;
-
 extern "C" {
 
-TMC2209Stepper *pStepX = 0;
-TMC2209Stepper *pStepY = 0;
-TMC2209Stepper *pStepZ = 0;
-TMC2209Stepper *pStepE = 0;
+TMC2209Stepper *pStep[4] = { 0, 0, 0, 0 };
```
DRracer, line 17: use nullptr in C++!

src/common/trinamic.cpp
```diff
@@ -35,51 +32,52 @@ void tmc_delay(uint16_t time) // delay for switching tmc step pin level
...
+    }
+    return mask;
 }
 
 void tmc_set_sgthrs(uint16_t SGT) {
-    pStepX->SGTHRS(SGT);
-    pStepY->SGTHRS(SGT);
-    pStepZ->SGTHRS(SGT);
-    pStepE->SGTHRS(SGT);
+    pStep[0]->SGTHRS(SGT);
```
DRracer, line 71: can we have symbolic constant names for the indices in pStep[]?
like `enum class Axes : uint_least8_t { X = 0, Y, Z, E }` ?

lib/Arduino_Core_A3ides/cores/arduino/HardwareSerial.cpp
XPila: same externs already contained in this source, this is because HardwareSerial for TMC is hack, we need use some handles from kernel. Arduino HardwareSerial should be removed in future and we should control TMC directly via HAL functions.

src/common/thread_measurement.c
XPila: marlin_event_clr is comon function for read and clear event state in single call. We do not need create own function for each event. This is special case of usage because we are waiting in loop (we need block until server ready, but also we need process client loop). I can allocate signal or mutex and synchronize startup using RTOS, but this is more simple and also more readable (principle is not hidden). My opinion is that using osDelay for synchronisation is proper way if we do not need react immediately (for example in DMA synchronization we are using signals).

DRracer: All what you wrote is acceptable, but it doesn't answer the first part of the comment - create a function `marlin_client_wait_for_start_processing()` which **does hide the implementation on purpose** and which also can be improved in the future by using OS synchronization primitives at one place.
I'd really like to avoid having each task implementing its own startup waiting (which is what you propose), when we want all tasks to wait the same way.

XPila: commit faa7a74

src/common/trinamic.cpp
XPila: commit f8ee683

src/common/trinamic.cpp
XPila: commit 45ddd2c

PR #487, opened 2020-05-21
Load rework: do not ask is in gear

RadekVana:
BFW-847

### DIFF
--- a/src/marlin_stubs/pause/pause.cpp
+++ b/src/marlin_stubs/pause/pause.cpp
@@ -52,9 +52,6 @@
 #include "filament.h"
 #include "RAII.hpp"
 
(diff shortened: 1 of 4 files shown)


src/marlin_stubs/pause/pause.cpp
```diff
@@ -52,6 +52,9 @@
 #include "filament.h"
 #include "RAII.hpp"
 
+#define MINIMAL_PURGE                   1
```
DRracer, line 55: please avoid using #define in C++ code, use enum or static const something etc.

RadekVana: fixed

PR #490, opened 2020-05-23
Serial Print - pause & disconnect

Chleba:
Added pause & disconnect button into a Serial printing screen.
Now FW can communicate with host (octoprint, printerface,..) via action commands.
Now handling only pause & disconnect.
When disconnect is happend by user interaction, M86 (safety timer) enabled (1800ms).

BFW-655, BFW-824, BFW-826


src/gui/screen_printing_serial.cpp
```diff
@@ -13,6 +13,35 @@
 #pragma pack(push)
 #pragma pack(1)
 
+typedef enum {
```
DRracer, line 16: please use enum class in C++ ... avoid old C-style constructs

DRracer: and why do you need `#pragma pack(1, push)` just for the enum? The enum doesn't care about alignment ...

src/gui/screen_printing_serial.cpp
```diff
@@ -75,45 +105,40 @@ void screen_printing_serial_init(screen_t *screen) {
...
-    //button tune
-    const rect_ui16_t rect_bt_tune = rect_ui16(8, 185, 64, 64);
-    id = window_create_ptr(
-        WINDOW_CLS_ICON, root,
-        rect_bt_tune,
-        &(pw->tune_button));
-    window_set_color_back(id, COLOR_GRAY);
-    window_enable(id);
-    window_set_icon_id(id, IDR_PNG_menu_icon_settings);
+    for (uint8_t col = 0; col < 3; col++) {
```
DRracer, line 113: uint8_t may not be efficient on 32bit platforms

src/gui/screen_printing_serial.cpp
```diff
@@ -131,28 +156,25 @@ int screen_printing_serial_event(screen_t *screen, window_t *window, uint8_t eve
...
 
-    int now = HAL_GetTick();
-
-    if ((now - pw->last_tick) > 500) {
-        pw->last_tick = now;
-        static int i = 0;
-        ++i;
-        i %= (sizeof(txts) / sizeof(txts[0]));
-        window_set_text(pw->w_message.win.id, txts[i]);
+    switch (((int)param) - 1) {
```
DRracer, line 163: OMG, win32 bullsh*t sending unspecified 4 bytes to handler functions. I really hate this reinterpret_cast of the `param` parameter.

src/marlin_stubs/pause/pause.cpp
```diff
@@ -67,13 +67,13 @@
...
     ENABLED(ADVANCED_PAUSE_CONTINUOUS_PURGE) || \
     BOTH(FILAMENT_UNLOAD_ALL_EXTRUDERS, MIXING_EXTRUDER) || \
-    ENABLED(HOST_ACTION_COMMANDS) || \
-    ENABLED(HOST_PROMPT_SUPPORT) || \
     ENABLED(SDSUPPORT)
 #error unsupported
 #endif
 // clang-format on
 
+PauseMenuResponse pause_menu_response;
```
DRracer, line 75: what is this and how is it related to the octoprint screen?

src/gui/screen_printing_serial.cpp
```diff
@@ -22,25 +51,19 @@ typedef struct
 
     window_text_t w_message; //Messages from onStatusChanged()
     window_icon_t octo_icon;
-    window_icon_t tune_button;
-    window_text_t tune_label;
+
+    window_icon_t w_buttons[3];
```
DRracer, line 55: magical constants?

src/marlin_stubs/pause/pause.cpp
Chleba: After added Marlin's Host features in need to be included, because it's needed in host_actions as a response handler. I can't make a build without it.

src/gui/screen_printing_serial.cpp
Chleba: added C++ enum definition

src/gui/screen_printing_serial.cpp
Chleba: used unsigned int insted of uint8_t

src/gui/screen_printing_serial.cpp
Chleba: param - void* will be obsolete, I hope, in refactored version of screens / windows - now used reinterpret_cast<int>

src/gui/screen_printing_serial.cpp
Chleba: I don't fully understand this comment - Added definition of buttons insted of fixed integers

src/gui/screen_printing_serial.cpp
```diff
@@ -10,8 +10,38 @@
 #include "stm32f4xx_hal.h" //HAL_GetTick
 #include "screens.h"
 
+#define BUTTON_TUNE       0
+#define BUTTON_PAUSE      1
+#define BUTTON_DISCONNECT 2
+
 #pragma pack(push)
-#pragma pack(1)
+
```
DRracer, line 18: leaving this #pragma pack(push) alone here is one of the fastest ways to hell - please remove it. You are missing the #pragma pack(pop) for this one!

src/gui/screen_printing_serial.cpp
```diff
@@ -10,8 +10,38 @@
...
+const char *serial_printing_labels[iid_count] = {
+    "Tune",
+    "Pause",
+    // "Stop",
+    // "Resume",
+    "Disconnect"
+};
+
+#pragma pack(push, 1)
```
DRracer, line 44: and another unterminated #pragma pack - very dangerous leaving it unterminated with #pragma pack(pop)

src/gui/screen_printing_serial.cpp
DRracer: can we have something like
`static const uint32_t number_of_buttons = 3` and use it at appropriate places?

src/gui/screen_printing_serial.cpp
```diff
@@ -75,50 +104,46 @@ void screen_printing_serial_init(screen_t *screen) {
...
-    //button tune
-    const rect_ui16_t rect_bt_tune = rect_ui16(8, 185, 64, 64);
-    id = window_create_ptr(
-        WINDOW_CLS_ICON, root,
-        rect_bt_tune,
-        &(pw->tune_button));
-    window_set_color_back(id, COLOR_GRAY);
-    window_enable(id);
-    window_set_icon_id(id, IDR_PNG_menu_icon_settings);
+    for (unsigned int col = 0; col < 3; col++) {
```
DRracer, line 112: here the `3` should be also replaced by a symbolic constant `number_of_buttons`, am I right?

src/gui/screen_printing_serial.cpp
DRracer: also, it is a good practice to check at compile time that the size of w_buttons and w_labels is enough for all the items you want to put in. I assume you have something like:
`enum { BUTTON_TUNE=0, BUTTON_PAUSE=1, BUTTON_DISCONNECT=2 };`
One of the tricks is to add another item and document, that it must be the last one with highest number:
`... BUTTONS_COUNT=3`
and use the BUTTONS_COUNT for the array's sizes
or even better:
`static_assert( BUTTON_DISCONNECT < number_of_buttons );` later on

src/gui/screen_printing_serial.cpp
```diff
@@ -75,50 +104,46 @@ void screen_printing_serial_init(screen_t *screen) {
...
-        240 - (240 + rect_bt_tune.x + rect_bt_tune.w - pt_txt.x) / 2, //cannot use 240-rect_txt.x .. stupid C
-        95);
-    id = window_create_ptr(WINDOW_CLS_TEXT, root, rect_txt, &(pw->w_message));
-    window_set_alignment(id, ALIGN_LEFT_TOP);
-    //window_set_padding(id, padding_ui8(0, 2, 0, 2));
-    window_set_text(id, txts[0]);
+    // -- CONTROLS
+    window_icon_t *sp_button;
+    // -- tune button
+    sp_button = &pw->w_buttons[BUTTON_TUNE];
```
DRracer, line 133: static_assert( BUTTON_TUNE < number_of_buttons );

src/gui/screen_printing_serial.cpp
```diff
@@ -75,50 +104,46 @@ void screen_printing_serial_init(screen_t *screen) {
...
+    set_icon_and_label(iid_tune, sp_button->win.id, pw->w_labels[BUTTON_TUNE].win.id);
+    // -- pause
+    sp_button = &pw->w_buttons[BUTTON_PAUSE];
```
DRracer, line 136: static_assert( BUTTON_PAUSE < number_of_buttons );

src/gui/screen_printing_serial.cpp
```diff
@@ -75,50 +104,46 @@ void screen_printing_serial_init(screen_t *screen) {
...
+    set_icon_and_label(iid_pause, sp_button->win.id, pw->w_labels[BUTTON_PAUSE].win.id);
+    // -- disconnect
+    sp_button = &pw->w_buttons[BUTTON_DISCONNECT];
```
DRracer, line 139: static_assert( BUTTON_DISCONNECT < number_of_buttons );
because this will fail at compile time warning the user of the array being too short

src/marlin_stubs/pause/pause.cpp
DRracer: ok

PR #494, opened 2020-05-25
BFW-853 Don't show unreal values in footer

espr14:


src/gui/status_footer.cpp
```diff
@@ -44,7 +44,7 @@ void status_footer_init(status_footer_t *footer, int16_t parent) {
         &(footer->wt_nozzle));
     footer->wt_nozzle.font = resource_font(IDR_FNT_SPECIAL);
     window_set_alignment(id, ALIGN_CENTER);
-    window_set_text(id, "0/0\177C");
+    window_set_text(id, "  / \177C");
```
DRracer, line 47: woudn't `"?/?\177C"` look better?

src/gui/status_footer.cpp
```diff
@@ -60,7 +60,7 @@ void status_footer_init(status_footer_t *footer, int16_t parent) {
         &(footer->wt_heatbed));
     footer->wt_heatbed.font = resource_font(IDR_FNT_SPECIAL);
     window_set_alignment(id, ALIGN_CENTER);
-    window_set_text(id, "0/0\177C");
+    window_set_text(id, "  / \177C");
```
DRracer, line 63: can we have one definition of this text and not two?

src/gui/status_footer.cpp
```diff
@@ -92,7 +92,7 @@ void status_footer_init(status_footer_t *footer, int16_t parent) {
         &(footer->wt_z_axis));
     footer->wt_z_axis.font = resource_font(IDR_FNT_SPECIAL);
     window_set_alignment(id, ALIGN_CENTER);
-    window_set_text(id, "999.95");
+    window_set_text(id, "0.00");
```
DRracer, line 95: is 0.00 a valid value or an estimate for the Z axis' position? Wouldn't a `"0.00?"` reflect the situation better?

PR #497, opened 2020-05-25
Integrate unit tests into the project

dragomirecky:


DRracer:
@dragomirecky aren't we c++14 (since Marlin is c++14 and gcc 7.3 handles it too)?
... just a question related to std::make_reverse_iterator...

DRracer:
anyway looking good, having separate dirs for various sets of tests with necessary simulated interfaces is probably the right way to go

tests/unit/gui/lazyfilelist/file_list_defs.h
```diff
@@ -0,0 +1,26 @@
...
+
+#ifndef MAXDIRNAMELENGTH
+    #define F_MAXDIRNAMELENGTH _MAX_LFN + 1
+#endif
+
+#ifndef MAXPATHNAMELENGTH
+    #define F_MAXPATHNAMELENGTH (1 + (F_MAXDIRNAMELENGTH + 1) * (MAX_DIR_DEPTH) + 1 + _MAX_LFN)
+#endif
+
+#define strlcpy strncpy
```
DRracer, line 26: if we can solve the non-standard strlcpy with clang, this can be removed then

utils/holly/TestDockerfile
```diff
@@ -0,0 +1,8 @@
+FROM gcc:10
```
DRracer, line 1: gcc10?

src/gui/lazyfilelist.h
```diff
@@ -135,7 +138,7 @@ class LazyDirView {
                     if (strcmp(files[0].lfn, dir.fno.fname) != 0) {
                         // i.e. we didn't get the same entry as the zeroth entry (which may occur when populating the window with non-null firstDirEntry)
                         // Make place in the window by standard item rotation downwards (to the right)
-                        std::rotate(files.rbegin(), files.rbegin() + 1, std::make_reverse_iterator(i)); // solves also the case, when there are less files in the window
+                        std::rotate(files.rbegin(), files.rbegin() + 1, std::reverse_iterator<typename decltype(files)::iterator>(i)); // solves also the case, when there are less files in the window
```
DRracer, line 141: I'd really like to move to C++14 ... if not, I'll make my own make_reverse_iterator, this statement looks horrible

dragomirecky:
> @dragomirecky aren't we c++14 (since Marlin is c++14 and gcc 7.3 handles it too)?
> ... just a question related to std::make_reverse_iterator...

Yes, the firmware uses c++11. I believe that Marlin is the only thing compiled as c++14.

I am down for upgrading to c++14, but that should be another PR.

tests/unit/gui/lazyfilelist/file_list_defs.h
dragomirecky: No clang for now. I decided to go back to GCC 🙈

utils/holly/TestDockerfile
dragomirecky: Yep ☝️

src/gui/lazyfilelist.h
dragomirecky: It does look horrible, I agree 😅. Fixed.

PR #501, opened 2020-05-26
Octoscreen fixes

RadekVana:
BFW-875 BFW-878


src/common/marlin_vars.h
```diff
@@ -40,6 +40,12 @@
 // variable masks
 #define MARLIN_VAR_MSK(v_id) ((uint64_t)1 << (v_id))
 
+#if (MARLIN_VAR_MAX == 64)
```
DRracer, line 43: what if the number of marlin vars exceeds 64? At least some #error should be here to prevent compilation

RadekVana: there is
static_assert(MARLIN_VAR_MAX < 64, "MarlinAPI: Too many variables");
in marlin_server.cpp

PR #505, opened 2020-05-27
MarlinAPI - fixed notification in marlin_server_set_temp_to_display

XPila:
BFW-857


src/common/marlin_server.cpp
```diff
@@ -605,6 +606,8 @@ int marlin_all_axes_known(void) {
 
 void marlin_server_set_temp_to_display(float value) {
     marlin_server.vars.display_nozzle = value;
+    for (int id = 0; id < MARLIN_MAX_CLIENTS; id++)
```
DRracer, line 609: Why don't we have a function which notifies the clients with appropriate mask? This is being done at several places and the code is duplicated.

PR #506, opened 2020-05-27
BFW-800 timestamp in printing screen

michalrudolf:
Change displayed time format from Remaining time to print end timestamp. Description in JIRA task.

additional changes:
In screen_menu_settings.cpp > When timezone is changed and sntp is already initialized, it sets up time immidiately.

### DIFF
--- a/src/common/wui_api.c
+++ b/src/common/wui_api.c
@@ -314,23 +314,23 @@ uint32_t stringify_timestamp(time_str_t *dest, timestamp_t *timestamp, uint8_t f
         }
 
         // DATE PARSING
-        if (flag & TIME_STR_DAYS) {
-            snprintf(dest->date, MAX_DATE_STR_SIZE, "%02d.", timestamp->date.d);
+        if (flag & TIME_STR_YEARS) {
+            snprintf(dest->date, MAX_DATE_STR_SIZE, "%d", timestamp->date.y);
         }
         if (flag & TIME_STR_MONTHS) {
             uint8_t length = strlen(dest->date);
             if (length > 0) {
-                snprintf(dest->date + length, MAX_DATE_STR_SIZE - length, "%02d.", timestamp->date.m);
+                snprintf(dest->date + length, MAX_DATE_STR_SIZE - length, "-%02d", timestamp->date.m);
             } else {
-                snprintf(dest->date, MAX_DATE_STR_SIZE, "%02d.", timestamp->date.m);
+                snprintf(dest->date, MAX_DATE_STR_SIZE, "%02d", timestamp->date.m);
             }
         }
-        if (flag & TIME_STR_YEARS) {
+        if (flag & TIME_STR_DAYS) {
             uint8_t length = strlen(dest->date);
             if (length > 0) {
-                snprintf(dest->date + length, MAX_DATE_STR_SIZE - length, "%d", timestamp->date.y);
+                snprintf(dest->date + length, MAX_DATE_STR_SIZE - length, "-%02d", timestamp->date.d);
             } else {
-                snprintf(dest->date, MAX_DATE_STR_SIZE, "%d", timestamp->date.y);
+                snprintf(dest->date, MAX_DATE_STR_SIZE, "%02d", timestamp->date.d);
             }
         }
         return 1;
--- a/src/common/wui_api.c
+++ b/src/common/wui_api.c
@@ -286,10 +286,53 @@ void sntp_set_system_time(uint32_t sec) {
     sntp_time_init = true;
 }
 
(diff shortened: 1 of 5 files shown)


src/common/wui_api.c
```diff
@@ -286,10 +286,53 @@ void sntp_set_system_time(uint32_t sec) {
...
+            uint8_t length = strlen(dest->time);
+            if (length > 0) {
+                snprintf(dest->time + length, MAX_TIME_STR_SIZE - length, ":%02d", timestamp->time.s);
+            } else {
+                snprintf(dest->time, MAX_TIME_STR_SIZE, "%02d", timestamp->time.s);
+            }
+        }
+
+        // DATE PARSING
+        if (flag & TIME_STR_DAYS) {
```
DRracer, line 317: do you realize this is not a valid English date format?

src/gui/screen_printing.cpp
```diff
@@ -429,6 +445,42 @@ static void update_remaining_time(screen_t *screen, time_t rawtime) {
...
 }
+
+static void update_end_timestamp(screen_t *screen, timestamp_t *now) {
+
+    auto &array = pw->text_etime;
+    char qmark[2] = "";
+
+    if (marlin_vars()->time_to_end == TIME_TO_END_INVALID) {
+        pw->w_etime_value.color_text = COLOR_VALUE_INVALID;
+        strlcpy(qmark, "?", 2);
```
DRracer, line 456: why do you copy the "?" string into qmark?
qmake could have been a `static const char qmark[] = "?"` and you either print qmark to get the "?" on the screen or qmark+1 to print an empty string (this part could be optimized further).

src/gui/screen_printing.cpp
```diff
@@ -104,7 +106,8 @@ typedef struct
     uint8_t last_sd_percent_done;
 
     std::array<char, 9> text_time;
-    std::array<char, 9> text_etime;
+    std::array<char, MAX_END_TIMESTAMP_SIZE> text_etime;
```
DRracer, line 109: using std array for static storage makes sense (iterators, algorithms etc.), but and abusing it with strlcpy looks weird

src/common/wui_api.c
RadekVana: i would use standard time library
http://www.cplusplus.com/reference/ctime/strftime/
it supports national mutations
http://zetcode.com/articles/cdatetime/

src/gui/screen_printing.cpp
michalrudolf: commit `187f789`

src/gui/screen_printing.cpp
michalrudolf: commit `187f789`

src/common/wui_api.c
michalrudolf: commit `187f789` ISO format, (We can reimplement to strftime anytime in the future, for now I use my own timestamp structure and my own print function)

src/gui/screen_printing.cpp
```diff
@@ -23,7 +24,8 @@
 #define BUTTON_PAUSE 1
 #define BUTTON_STOP  2
 
-#define POPUP_MSG_DUR_MS 5000
+#define POPUP_MSG_DUR_MS       5000
+#define MAX_END_TIMESTAMP_SIZE (MAX_DATE_STR_SIZE + MAX_TIME_STR_SIZE + 5) // "dd.mm.yyyy at hh:mm:ss" + safty measures for 3digit where 2 digits should be
```
RadekVana, line 28: avoid macros
use static const size_t for size constants

michalrudolf: Why? I think code consistency is more important

src/gui/screen_printing.cpp
```diff
@@ -429,6 +445,42 @@ static void update_remaining_time(screen_t *screen, time_t rawtime) {
...
+    if (marlin_vars()->time_to_end == TIME_TO_END_INVALID) {
+        pw->w_etime_value.color_text = COLOR_VALUE_INVALID;
+        strlcpy(qmark, "?", 2);
+    } else {
+        pw->w_etime_value.color_text = COLOR_VALUE_VALID;
+    }
+
+    timestamp_t print_end, tommorow;
+    print_end.epoch_secs = now->epoch_secs + (marlin_vars()->time_to_end / 1000);
+    tommorow.epoch_secs = now->epoch_secs += 86400; // now + one full day
```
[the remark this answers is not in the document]
michalrudolf: commit `c14a82d`

src/gui/screen_printing.cpp
DRracer: if you can, avoid macros in C++ code (code safety)

PR #507, opened 2020-05-27
Purge caption

RadekVana:
BFW-829


src/marlin_stubs/pause/M701_M702.cpp
```diff
@@ -40,17 +40,18 @@
...
 #include "pause_stubbed.hpp"
 #include "filament.h"
+#include <functional>
 
 #define DO_NOT_RESTORE_Z_AXIS
 #define Z_AXIS_LOAD_POS   40
 #define Z_AXIS_UNLOAD_POS 20
 
-typedef void (*load_unload_fnc)(const int8_t target_extruder);
+using Func = std::function<void()>;
```
DRracer, line 49: Are you sure std function doesn't allocate memory? Not sure about current state, but there were times when std::function allocated memory to store the pointer. Please check the template definition thoroughly

src/marlin_stubs/pause/pause.cpp
```diff
@@ -97,13 +94,28 @@ Pause &Pause::GetInstance() {
     return s;
 }
 
-//xyze_pos_t Pause::resume_position;
+void Pause::SetUnloadLenght(float len) {
+    unload_length = -ABS(len); // it is negative value
```
DRracer, line 98: use std functions in c++ (std::abs), no need to use the c-style 3rd party workarounds

src/marlin_stubs/pause/M701_M702.cpp
RadekVana: it is fine, it should pass pointer to anonymous functor (lambda)

src/marlin_stubs/pause/pause.cpp
RadekVana: thx
fixed

PR #516, opened 2020-05-29
[REVIEW AFTER BFW-859 MERGE] GUI: Unify header-body-footer sizes BFW-889 

espr14:


src/gui/dialogs/DialogStateful.hpp
```diff
@@ -80,12 +75,12 @@ class IDialogStateful : public IDialog {
     bool Change(uint8_t phs, uint8_t progress_tot, uint8_t progress); // = 0; todo should be pure virtual
     virtual ~IDialogStateful();
 
-    static constexpr rect_ui16_t get_radio_button_size() {
```
RadekVana, line 83: constexpr  has meaning here, removing it will increase code size
you should refactor gui_defaults first to satisfy constexpr

src/gui/screen_printing.cpp
```diff
@@ -222,18 +221,21 @@ void screen_printing_init(screen_t *screen) {
     window_hide(id);
     pw->message_flag = 0;
 
+    // buttons
+    const uint16_t icon_y = gui_defaults.footer_sz.y - gui_defaults.padding.bottom - 22 - 64;
```
RadekVana, line 225: 64 and 22 should be constants too

src/gui/dialogs/DialogStateful.hpp
espr14: I made BFW-954 to split the issue. It's not crucial to do it immediately.

src/gui/screen_printing.cpp
espr14: They should but I don't know what they represent. Or it needs another refactor through the code.

PR #517, opened 2020-05-30
ABS, HIPS, PP Filament

ovariludovit:
Added ABS Filament, added to MBL Auto Home if stepper is disabled


Chleba:
Hi,
Congratulation [unknow user], for Your first official PR for our new Buddy board.
We are very happy to see that :).
Thanks to this PR, we have found some small issues with our CI system (Holly) and we have to make some changes, so that next PR's from our community will be much smoother experience.
For now, we are having a talk about adding all of this materials into Prusa MINI and how to pick right solution for enabling and adding custom materials.
Stay put for review and new information.

Chleba:
Hi again,
We checked all material's temperatures as you've added and update them. You can have a look here - #552. These new materials preheat options with will be available in next firmware version.
Thanks again for your PR. We really appreciate it.

PR #520, opened 2020-06-01
Footer: Fix init & round BFW-871, BFW-899

espr14:


src/gui/status_footer.h
```diff
@@ -31,13 +25,15 @@ typedef enum heat_state_e {
 
 typedef struct
 {
-    float nozzle;                /// temperature of nozzle shown on display
-    float nozzle_target;         /// target temperature of nozzle shown on display
+    float nozzle;                /// actual temperature of nozzle
```
DRracer, line 28: actual? Seriously?

PR #522, opened 2020-06-02
Clear window before draw BFW-890 

espr14:


src/gui/dialogs/DialogStateful.hpp
```diff
@@ -83,12 +78,12 @@ class IDialogStateful : public IDialog {
     bool Change(uint8_t phase, uint8_t progress_tot, uint8_t progress); // = 0; todo should be pure virtual
     virtual ~IDialogStateful();
 
-    static constexpr rect_ui16_t get_radio_button_size() {
-        rect_ui16_t rc_btn = { 0, 32, 240, 320 - 96 }; //msg box size
-        rc_btn.y += (rc_btn.h - 40);                   // 30pixels for button (+ 10 space for grey frame)
-        rc_btn.h = 30;
-        rc_btn.x += 6;
-        rc_btn.w -= 12;
+    static rect_ui16_t get_radio_button_size() {                                // cannot be const(expr)
```
RadekVana, line 81: It should be constexpr
problem is gui_defaults is written in C and will be rewritten to C++

RadekVana: I recommend modifying gui_defaults to have `#ifdef __cplusplus` version

espr14: I would leave it for different PR focused on C -> C++ transition.

espr14: BFW-954 was made.

espr14: However, it's duplicit work to review this before 889 is merged.

src/gui/dialogs/DialogStateful.cpp
```diff
@@ -2,6 +2,11 @@
 #include "DialogRadioButton.hpp"
 #include "gui.h"
 
+constexpr uint8_t PROGRESS_BAR_X_PAD = 10;
```
RadekVana, line 5: should be
`constexpr static const` or `constexpr static` (equal in this case)

espr14: constexpr const == constexpr
I will add static.

src/gui/dialogs/DialogStateful.hpp
```diff
@@ -80,12 +75,12 @@ class IDialogStateful : public IDialog {
     bool Change(uint8_t phs, uint8_t progress_tot, uint8_t progress); // = 0; todo should be pure virtual
     virtual ~IDialogStateful();
 
-    static constexpr rect_ui16_t get_radio_button_size() {
-        rect_ui16_t rc_btn = { 0, 32, 240, 320 - 96 }; //msg box size
-        rc_btn.y += (rc_btn.h - 40);                   // 30pixels for button (+ 10 space for grey frame)
-        rc_btn.h = 30;
-        rc_btn.x += 6;
-        rc_btn.w -= 12;
+    static rect_ui16_t get_radio_button_size() {                                // cannot be const(expr)
```
RadekVana, line 78: I'm still not happy with removing constexpr (bigger code, less efficient)
If you insist, at least put definition of the method into cpp file.
And comment it with todo or fixme to be changed to constexpr after gui_defaults refactoring

PR #523, opened 2020-06-02
WUI - Print end timestamp and webpage fixes

michalrudolf:
BFW-798
Last thing is missing is Bruno's repaired new version of local web page,
then full NTP task will be completed


src/common/marlin_server.cpp
```diff
@@ -421,7 +421,7 @@ void marlin_server_print_start(const char *filename) {
     if ((marlin_server.print_state == mpsIdle) || (marlin_server.print_state == mpsFinished) || (marlin_server.print_state == mpsAborted)) {
         media_print_start(filename);
         for (int id = 0; id < MARLIN_MAX_CLIENTS; id++)
-            marlin_server.client_changes[id] |= MARLIN_VAR_MSK(MARLIN_VAR_FILEPATH);
+            marlin_server.client_changes[id] |= MARLIN_VAR_MSK(MARLIN_VAR_FILENAME);
```
DRracer, line 424: Shouldn't you notify both? LFN filename and SFN filepath?

lib/WUI/wui.c
```diff
@@ -32,7 +32,8 @@ osMutexDef(wui_thread_mutex);   // Mutex object for exchanging WUI thread TCP th
 osMutexId(wui_thread_mutex_id); // Mutex ID
 
 static marlin_vars_t *wui_marlin_vars;
-wui_vars_t wui_vars; // global vriable for data relevant to WUI
+wui_vars_t wui_vars;                              // global vriable for data relevant to WUI
+static char wui_media_LFN[FILE_NAME_MAX_LEN + 1]; // static buffer for gcode file name
```
DRracer, line 36: Do you really need another buffer for the filename? marlinvars already have one...

src/common/marlin_server.cpp
michalrudolf: commit `f7b231a`

PR #527, opened 2020-06-03
Print button disabled after USB disk inserted with no gcode in root

Chleba:
Updated `screen_home_disable_print_button` with 'disable' flag for update button state if screen is not redrawed after `find_latest_gcode` will not found file to print.

BFW-912


src/gui/screen_home.cpp
```diff
@@ -62,7 +62,7 @@ struct screen_home_data_t {
 #define pw ((screen_home_data_t *)screen->pdata)
 
 static bool find_latest_gcode(char *fpath, int fpath_len, char *fname, int fname_len);
-void screen_home_disable_print_button(screen_t *screen);
+void screen_home_disable_print_button(screen_t *screen, int disable);
```
DRracer, line 65: prefer bool in c++

src/gui/screen_home.cpp
```diff
@@ -250,13 +251,14 @@ static bool find_latest_gcode(char *fpath, int fpath_len, char *fname, int fname
     return result == FR_OK && fname[0] != 0 ? true : false;
 }
 
-void screen_home_disable_print_button(screen_t *screen) {
-    pw->w_buttons[0].win.f_disabled = 1;
-    pw->w_buttons[0].win.f_enabled = 0; // cant't be focused
-    window_set_text(pw->w_labels[0].win.id, labels[6]);
+void screen_home_disable_print_button(screen_t *screen, int disable) {
+    pw->w_buttons[0].win.f_disabled = disable;
```
DRracer, line 255: these f_enabled and f_disabled are very suspicious (at least naming), subject to future refactoring

src/gui/screen_home.cpp
Chleba: I know it is preferable use boolean in C++, but here I used int, because I can set a flags (f_enabled, f_disabled) right away and not after i recast it or using ternary operator to set 1 || 0 in those flags.

src/gui/screen_home.cpp
Chleba: Agreed

PR #539, opened 2020-06-06
Lan settings

RadekVana:
BFW-947 BFW-948 BFW-949 BFW-950


michalrudolf:
I still don't see displaying of the MAC address, it was part of the old implementation and if you didn't receive any request to erase it, it should be there.

PR #549, opened 2020-06-09
Preheat: Fix in preheat and filament menu

espr14:


src/gui/screen_menu_filament.cpp
```diff
@@ -14,7 +14,12 @@
 
 /// Sets temperature of nozzle not to ooze before print (MBL)
 void setPreheatTemp() {
-    marlin_gcode_printf("M104 S%d D%d", (int)PREHEAT_TEMP, (int)filaments[get_filament()].nozzle);
+    const marlin_vars_t *vars = marlin_vars();
+    if (!vars)
```
RadekVana, line 18: no need to check ` if (!vars)`, it cannot be nullptr

src/gui/screen_menu_preheat.cpp
```diff
@@ -14,9 +14,14 @@ class MI_Filament : public WI_LABEL_t {
     virtual void click(Iwindow_menu_t & /*window_menu*/) override {
         const filament_t filament = filaments[T];
         marlin_gcode("M86 S1800"); // enable safety timer
-        marlin_gcode_printf("M104 S%d", (int)filament.nozzle);
+        /// don't use preheat temp for cooldown
+        if (PREHEAT_TEMP >= filament.nozzle) {
```
RadekVana, line 18: use `if (PREHEAT_TEMP > filament.nozzle)` parameters S and D makes no sense when they are equal

src/gui/screen_menu_filament.cpp
RadekVana: also missing else

src/common/client_response.hpp
```diff
@@ -48,7 +48,7 @@ constexpr T GetEnumFromPhaseIndex(size_t index) {
 //and YES phase can have 0 responses
 //every enum must have "_first" and "_last"
 //"_first" ==  "previous_enum::_last" + 1
-//EVERY response shall have unique ID (so every button in GIU is unique)
+//EVERY response shall have unique ID (so every button in GUI is unique)
```
DRracer, line 51: when you are fixing the comment wording, please make sure it is proper English too ;) 
... shall have **a** unique ID

src/gui/screen_menu_filament.cpp
```diff
@@ -14,7 +14,12 @@
...
+        marlin_gcode_printf("M104 S0");
+
+    /// don't read from EEPROM since it's not in sync
```
DRracer, line 21: EEPROM?

espr14: Previous implementation that does not work.

src/gui/screen_menu_preheat.cpp
espr14: In case of equality "M104 S%d" is called.

espr14: S makes sense in both cases.

PR #554, opened 2020-06-10
str_utils: add multiline BFW-970

espr14:


espr14:
str_utils are redone nearly from scratch so using of diff tool does not make much sense.

src/common/str_utils.h
```diff
@@ -1,43 +1,33 @@
...
+#define NBSP "\xA0"                 /// Non Breaking Space
+constexpr char CHAR_NL = '\n';      /// New Line
+constexpr char NL[2] = { CHAR_NL }; /// New Line
+constexpr char EOS = '\0';          /// end of string
 
-#define QT_HSPACE "\x1F" // ~ <UnitSeparator>
-#define QT_LF     "\n"   // ~ <LineFeed>
-#define QT_NL     QT_LF
-#define QT_HYPHEN "\x1A" // ~ <Substitute>
+#ifdef __cplusplus
```
RadekVana, line 13: it is included in cpp files only
remove `#ifdef __cplusplus` ?

RadekVana: and please rename headers to .hpp

src/common/str_utils.h
```diff
@@ -1,43 +1,33 @@
...
+extern "C" int str2multiline(char *str, size_t max_size, const size_t line_width);
 
-#define EOS '\x00'
+#else
 
-enum class delimiter_t : uint8_t { NONE,
-    SPACE,
-    HYPHEN,
-    CUSTOM };
+typedef struct
```
RadekVana, line 22: please do not use `typedef struct` in C++ code

src/common/str_utils.cpp
```diff
@@ -1,168 +1,142 @@
...
+        return -1;
+    if (n == 0)
+        return 0;
 
-void set_hyphen_distance(int dist) {
-    hyphen_distance = dist;
-}
+    const size_t size = strlen(str);
+    if (size + n >= max_size) /// too much to add
+        return -2;
```
RadekVana, line 35: define some constants for -1 and -2 return values

PR #556, opened 2020-06-10
SNTP client cycle

michalrudolf:
make sntp client cycle

### DIFF
--- a/lib/WUI/sntp/sntp_client.c
+++ b/lib/WUI/sntp/sntp_client.c
@@ -15,3 +15,18 @@ void sntp_client_init(void) {
     }
     sntp_init();
 }
+
+void sntp_client_stop(void) {
+    sntp_stop();
+}
+
+void sntp_client_cycle(void) {
+    static bool sntp_init_done = false;
+    if (!sntp_init_done && internet_connected) {
+        sntp_client_init();
+        sntp_init_done = true;
+    } else if (sntp_init_done && !internet_connected) {
+        sntp_client_stop();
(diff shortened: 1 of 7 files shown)


lib/WUI/sntp/sntp_client.c
```diff
@@ -2,8 +2,8 @@
 #include "sntp_client.h"
 #include "wui_api.h"
 
-static ip_addr_t ntp_server; // testing ntp server located in Prague
-
+static ip_addr_t ntp_server;         // testing ntp server located in Prague
+static bool sntp_processing = false; // describes if sntp is currently running or not
```
DRracer, line 6: probably `sntp_running` will be a better name

lib/WUI/wui_api.h
```diff
@@ -55,6 +56,9 @@ typedef struct {
     char printer_state[PRI_STATE_STR_LEN]; // state of the printer, have to be set in wui
 } printer_info_t;
 
+extern ETH_STATUS_t eth_status; // discribes state of internet connection
```
DRracer, line 59: avoid exposing variables directly - use encapsulating functions instead (like c++ methods)

lib/WUI/sntp/sntp_client.c
michalrudolf: commit `e278f27`

lib/WUI/wui_api.h
michalrudolf: commit `e278f27`

PR #559, opened 2020-06-10
Sound refactoring

Chleba:
A little refactoring was needed, when getting feedback from testing department for better adding new sounds and defining it's values with less code needed.

BFW-977


src/common/sound.cpp
```diff
@@ -14,6 +14,34 @@ Sound::Sound() {
     frequency = 100.f; // frequency of sound signal (0-1000)
     volume = 0.50;     // volume of sound signal (0-1)
 
+    // -- values of sound signals - frequencies, volumes, durations
+    // -- durations of beep
+    durations[eSOUND_TYPE_ButtonEcho] = 100.f;
```
DRracer, line 19: I'd recommend doing some static initialization. The compiler might understand your code and optimize it away completely, but I wouldn't rely on it.
`static const float durations[] = {100.F, 500.F ....};`
or even better - make a struct and initialize all components together (will be more readable)
`struct DFV { float duration, freq, volume; }; DFV dfv[] = { { 100.F, 900.F, volume }, { .... } };` if volume is a `static const float volume = 0.5F`

src/common/sound.cpp
```diff
@@ -14,6 +14,34 @@ Sound::Sound() {
...
+    durations[eSOUND_TYPE_StandardPrompt] = 500.f;
```
DRracer, line 20: oh, durations are uin32_t, why do you assign floating point numbers?

src/common/sound.cpp
```diff
@@ -14,6 +14,34 @@ Sound::Sound() {
...
+    durations[eSOUND_TYPE_StandardAlert] = 200.f;
+    durations[eSOUND_TYPE_CriticalAlert] = 500.f;
+    durations[eSOUND_TYPE_EncoderMove] = 50.f;
+    durations[eSOUND_TYPE_BlindAlert] = 100.f;
+    durations[eSOUND_TYPE_Start] = 100.f;
+
+    // -- frequencies of beep
+    frequencies[eSOUND_TYPE_ButtonEcho] = 900.f;
+    frequencies[eSOUND_TYPE_StandardPrompt] = 600.f;
```
DRracer, line 29: frequencies are double? Why? Double is not HW accelerated on our CPU? Assigning a float into a double?

src/common/sound.h
```diff
@@ -30,6 +30,21 @@ class Sound {
     double frequency;
     double volume;
 
+    uint32_t durations[eSOUND_TYPE_count];
```
DRracer, line 33: use reasonable data types and initialize them properly

src/common/sound.cpp
```diff
@@ -51,125 +79,42 @@ void Sound::stop() {
...
-            this->soundEncoderMove(1, 50.f);
-        }
-        if (eSoundType == eSOUND_TYPE_BlindAlert) {
-            this->soundBlindAlert(1, 100.f);
-        }
-        if (eSoundType == eSOUND_TYPE_CriticalAlert) {
-            this->soundCriticalAlert(-1, 500.f);
-        }
+        t_size = sizeof(assistTypes) / sizeof(assistTypes[0]);
+        this->_playSound(eSoundType, assistTypes, assistRepeats, t_size);
```
DRracer, line 108: why do you use explicit `this->` ?

src/common/sound.cpp
DRracer: Also, when you write comments (which is good), please use the doxygen style and prefer documenting the public interface methods in the header file (their purpose and meaning of parameters).
Implementation-related comments are ok to put near the implementation (i.e. into the cpp file)

src/common/sound.cpp
Chleba: my mistake - i copied previous frequency values and compiler did not give a damn about that.

src/common/sound.cpp
Chleba: Another mistake of not checking my types. Fixed.

src/common/sound.cpp
Chleba: bad habit from previous languages - fixed

src/common/sound.cpp
Chleba: Doxygen commentary studied and implemented and static defining of signal's values changed.

src/common/sound.h
Chleba: done

src/guiapi/include/MItem_tools.hpp
```diff
@@ -216,14 +216,17 @@ class MI_SOUND_MODE : public WI_SWITCH_t<4> {
...
-class MI_SOUND_TYPE : public WI_SWITCH_t<5> {
-    constexpr static const char *const label = N_("Sound Type");
-
-    constexpr static const char *str_ButtonEcho = N_("ButtonEcho");
-    constexpr static const char *str_StandardPrompt = N_("StandardPrompt");
-    constexpr static const char *str_StandardAlert = N_("StandardAlert");
-    constexpr static const char *str_EncoderMove = N_("EncoderMove");
-    constexpr static const char *str_BlindAlert = N_("BlindAlert");
+class MI_SOUND_TYPE : public WI_SWITCH_t<8> {
+    constexpr static const char *const label = "Sound Type";
```
DRracer, line 220: translations!

PR #578, opened 2020-06-15
do not change button DHCP->static when failed

RadekVana:
BFW-990

### DIFF
--- a/src/gui/screen_menu_lan_settings.cpp
+++ b/src/gui/screen_menu_lan_settings.cpp
@@ -48,8 +48,8 @@ public:
     static bool IsOn();
     static void Init();
     static bool IsUpdated();
-    static void SetStatic();
-    static void SetDHCP();
+    static bool SetStatic();
+    static bool SetDHCP();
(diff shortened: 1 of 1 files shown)


src/gui/screen_menu_lan_settings.cpp
```diff
@@ -223,7 +225,9 @@ class MI_LAN_IP_t : public WI_SWITCH_t<2> {
     MI_LAN_IP_t()
         : WI_SWITCH_t<2>(Eth::IsStatic() ? 1 : 0, label, 0, true, false, str_DHCP, str_static) {}
     virtual void OnChange(size_t old_index) override {
-        old_index == 0 ? Eth::SetStatic() : Eth::SetDHCP();
+        bool success = old_index == 0 ? Eth::SetStatic() : Eth::SetDHCP();
```
DRracer, line 228: This condition could be reduced - SetDHCP only returns true (I understand you made SetDHCP and SetStatic with the same interface just to make it look better)
Can SetDHCP actually return false in the future?

PR #595, opened 2020-06-18
Cpp

RadekVana:
BFW-998
rename gui C files to Cpp and fix errors / warnings
remove extern "C"
replace "cmath_ext.h" with <algorithm>
replace ifndef - define with pragma once


src/gui/window_file_list.cpp
```diff
@@ -104,7 +104,7 @@ void window_file_list_draw(window_file_list_t *window) {
 
     int visible_slots = rc_win.h / item_height;
     int ldv_visible_files = LDV_VisibleFilesCount(window->ldv);
-    int maxi = MIN(MIN(visible_slots, ldv_visible_files), window->count);
+    int maxi = std::max(std::min(visible_slots, ldv_visible_files), window->count);
```
espr14, line 107: MIN(MIN vs max(min

RadekVana: fixed

PR #602, opened 2020-06-20
Remove wrappers

RadekVana:


src/gui/window_file_list.h
```diff
@@ -29,7 +36,7 @@ typedef struct _window_file_list_t {
     txtroll_t roll;
     int count;                        // total number of files/entries in a dir
     int index;                        // selected index - cursor position within the visible items
-    void *ldv;                        // I'm a C-pig and I need a pointer to my LazyDirView class instance ... subject to change when this gets rewritten to C++
+    LDV9 *ldv;                        // I'm a C-pig and I need a pointer to my LazyDirView class instance ... subject to change when this gets rewritten to C++
```
DRracer, line 39: and how about removing the comment when the code finally transformed into C++ ;) ?

RadekVana: screen can be created by memcpy, it is just renamed to cpp and compiles as C++, but behaves more like C

PR #614, opened 2020-06-23
Serial print fixes

Chleba:
- fixed tune menu - moving Z axis every time when user stepped into a tune menu
- fixed M876 P0 gcode making a BSOD
- fixed disconnect default confirm message box to have "NO" selected
- fixed safety timer (now cooldown) when disconnected from host (M876 P0 is called OR printer's DISCONNECT button is pressed)
- fixed marlin queue to make last gcode - parking and cooldown - now waiting for Marlin's gcode queue is empty then do the sequence

### DIFF
--- a/src/marlin_stubs/M876.cpp
+++ b/src/marlin_stubs/M876.cpp
@@ -23,6 +23,7 @@
 
 #if ENABLED(HOST_PROMPT_SUPPORT) && DISABLED(EMERGENCY_PARSER)
 
+		#include "../../lib/Marlin/Marlin/src/feature/safety_timer.h"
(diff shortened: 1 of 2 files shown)


src/marlin_stubs/M876.cpp
```diff
@@ -23,6 +23,7 @@
 
 #if ENABLED(HOST_PROMPT_SUPPORT) && DISABLED(EMERGENCY_PARSER)
 
+		#include "../../lib/Marlin/Marlin/src/feature/safety_timer.h"
```
ojrik, line 26: This include already exists two lines below.

Chleba: Fixed - thank you @ojrik

PR #616, opened 2020-06-23
Refactor - remove sound C wrap

Chleba:
Replaced all sound C wrap when refactoring into C++

BFW-1025, BFW-1027


src/common/sound.cpp
```diff
@@ -1,8 +1,9 @@
-#include "sound_C_wrapper.h"
-#include "sound.h"
+#include "sound.hpp"
 #include "hwio.h"
 #include "eeprom.h"
 
+uint8_t SOUND_INIT = 0;
```
RadekVana, line 5: make it static to be visible only in this file
use enum class instead

RadekVana: or bool if it is only true/false

Chleba: changed to - static bool

PR #618, opened 2020-06-24
appendix status info in the shared config

joshymjose:


src/common/shared_config.h
```diff
@@ -11,8 +11,14 @@
 #define FW_UPDATE_ENABLE  0xAA
 #define FW_UPDATE_DISABLE 0x00
 
-// pin PA13 state
-#define APPENDIX_FLAG_MASK 0x01
+// appendix state
+typedef enum {
+    APNDX_EXISTS = 0, /**< Appendix not broken, signature authentication needed */
+    APNDX_BROKEN      /**< Appendix broken, no signature authentication needed */
+} BT_APNDX_STATUS;
```
dragomirecky, line 18: /suggestion/ Can we either have all the cases + the enums name prefixed with `BT_`, or none of them?

Also, I think it would be better if the name of this enum would correspond to the name of the „appendix location macro“.

src/common/shared_config.h
```diff
@@ -11,8 +11,23 @@
...
+typedef enum {
+    appednix_status = 0, /**< appendix status flag */
+    reserved_1,
+    reserved_2,
+    reserved_3,
+    reserved_4,
+    reserved_5,
+    reserved_6,
+    reserved_7
+} model_specific_flags;
```
dragomirecky, line 30: 1. Where are the flags located in memory?
2. There is a typo in `appednix_status`
3. Why is the name of `APNDX_STATUS` capitalized and the name of `model_specific_flags` not?
4. The enum's cases `reserved_1`, `reserved_2`, etc are all global symbols. They should be prefixed with the enum's name.

joshymjose: refer to usage "volatile data_exchange_t ram_data_exchange __attribute__((section(".boot_fw_data_exchange")));" in src/common/sys.cpp, line 16 for location in memory

src/common/shared_config.h
```diff
@@ -11,14 +11,28 @@
 #define FW_UPDATE_ENABLE  0xAA
 #define FW_UPDATE_DISABLE 0x00
 
-// pin PA13 state
-#define APPENDIX_FLAG_MASK 0x01
+typedef enum {
+    BT_APPENDIX_EXISTS = 0, /**< Appendix not broken, signature authentication needed */
+	BT_APPENDIX_BROKEN	/**< Appendix broken, no signature authentication needed */
```
dragomirecky, line 16: bad indentation

src/common/shared_config.h
```diff
@@ -11,14 +11,28 @@
...
+// model_specific_flags bit locations from LSB
+typedef enum {
+	BT_APPENDIX_STATUS_LOC = 0,
+	BT_BOOT_FLAG_1,
+	BT_BOOT_FLAG_2,
+	BT_BOOT_FLAG_3,
+	BT_BOOT_FLAG_4,
+	BT_BOOT_FLAG_5,
+	BT_BOOT_FLAG_6,
+	BT_BOOT_FLAG_7
```
dragomirecky, line 28: tabs instead of spaces

src/common/shared_config.h
```diff
@@ -11,14 +11,28 @@
...
 
 #pragma pack(push)
 #pragma pack(1)
 typedef struct {
     uint8_t fw_update_flag;
-    uint8_t model_specific_flags; // ~ "reserved1" originally
+    uint8_t boot_flags;
     uint8_t reserved2;
     uint8_t reserved3;
     uint8_t reserved4;
```
dragomirecky, line 38: /nitpick/
Why not 

```C
typedef enum {
    FW_UPDATE_ENABLE = 0xAA,
    FW_UPDATE_DISABLE = 0xFF,
} fw_update_flag;

typedef struct  {
    fw_update_flag       fw_update_flag;
    bool                 appendix_broken : 1;
    uint8_t              __reserved[14];
} data_exchange_t;
```

Less code, clearer and safer code, the same memory layout.

joshymjose: yes this is better implementation, thanks for pointing out

dragomirecky:
There is a compilation error now 🙀

src/common/shared_config.h
```diff
@@ -7,32 +7,23 @@
...
-    uint8_t reserved10;
-    uint8_t reserved11;
-    uint8_t reserved12;
-    uint8_t reserved13;
-    uint8_t reserved14;
-    uint8_t reserved15;
+
+typedef enum {
+    FW_UPDATE_ENABLE = 0xAA,
+    FW_UPDATE_DISABLE = 0xFF,
```
DRracer, line 20: Is 0xFF the default after clearing the flash memory area? We don't want the FW update disabled by default ... actually we are getting reports that fw update got disabled in FW 4.1. Can you explain it?

src/common/shared_config.h
```diff
@@ -7,32 +7,23 @@
...
+} fw_update_flag;
+
+typedef struct  {
+    fw_update_flag       fw_update_flag;
+    bool                 appendix_broken : 1;
```
DRracer, line 25: Using bool this way is very unusual. I'm not saying the gcc won't accept this, but using the good old uint8_t makes things more clear. Bool may actually be represented as 4B or even 8B variable depending on target platform.

src/common/shared_config.h
joshymjose: The vales are taken from EEPROM in case of "update always" option, and in case of update on restart the values are from RAM, you meant stm flash area?

src/common/shared_config.h
joshymjose: good point!

joshymjose:
outdated

PR #620, opened 2020-06-24
Add unit test for extremely long file names

DRracer:
Expected to fail for now, such files are not accepted into the file browser, thus the user cannot print them. To be solved in the future.

BFW-1041

### DIFF
--- a/tests/unit/gui/lazyfilelist/fatfs.cpp
+++ b/tests/unit/gui/lazyfilelist/fatfs.cpp
@@ -6,8 +6,7 @@
 #include <iostream>
 #include <assert.h>
 
-// filesystem se bude simulovat
-// budou ruzne velky adresare s ruznejma casama
+// Simulated filesystem with FATfs interface
 
 #if 0
 static std::vector<FileEntry> testFiles0 = {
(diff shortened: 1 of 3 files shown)


tests/unit/gui/lazyfilelist/fatfs.cpp
```diff
@@ -51,14 +52,25 @@ std::vector<FileEntry> testFiles0;
 
 extern "C" {
 
+/// fill FILINFO with LFN and simulate a SFN based on file's index
+void MakeLFNSFN(FILINFO *fno, const char *lfn, size_t fileindex) {
+    assert(fileindex < 100); // to make the SFN numeric suffix within 2 digits for now
+    strncpy(fno->fname, lfn, sizeof(fno->fname));
+    if (strlen(lfn) >= 13) {
```
marekmosna-prusa3d, line 59: those 13 represent FF_SFN_BUF?

marekmosna-prusa3d (changes requested):

tests/unit/gui/lazyfilelist/fatfs.cpp
DRracer: yes, improved in the next commit

marekmosna-prusa3d (approved):

PR #624, opened 2020-06-24
Refactor hwio

mkbel:
### Known issues

- Application doesn't start when printer is powered on by power switch (doesn't apply for reset button) AND USB flash drive with firmware.bbf is present AND bootloader offers to update firmware. Whole screen is white and there is no beep after bootloarder checks the application. (Reproduced on pre-production MINI with broken apendix.)


src/common/putslave.cpp
```diff
@@ -369,29 +367,6 @@ int putslave_do_cmd_a_start(uartslave_t *pslave) {
...
-        //SCK - PB10
-        gpio_init(TPB10, GPIO_MODE_OUTPUT_PP, GPIO_NOPULL, GPIO_SPEED_FREQ_HIGH);
-        //MISO - PC2
-        gpio_init(TPC2, GPIO_MODE_INPUT, GPIO_PULLUP, GPIO_SPEED_FREQ_VERY_HIGH);
-        //MOSI - PC3
-        gpio_init(TPC3, GPIO_MODE_OUTPUT_PP, GPIO_NOPULL, GPIO_SPEED_FREQ_HIGH);
-    }
-    return UARTSLAVE_OK;
-}
-
```
mkbel, line 394: Putslave "stop" can not be removed, as it is used for retesting already tested and locked boards.

mkbel:
It is probably not a good idea to change Marlin and TMCStepper libraries in order to remove spi.h, better would be to ship empty spi.h

src/common/Pin.hpp
```diff
@@ -0,0 +1,207 @@
...
+    down = GPIO_PULLDOWN,
+};
+
+class InputPin : private ConfigurableIndestructible, public Pin {
+public:
+    InputPin(IoPort ioPort, IoPin ioPin, IMode iMode, Pull pull)
+        : Pin(ioPort, ioPin)
+        , m_mode(iMode)
+        , m_pull(pull) {}
+    GPIO_PinState read() {
```
dragomirecky, line 143: /suggestion/ Consider using our own data type for „pin state“, instead of exposing STM's `GPIO_PinState`.

src/common/Pin.hpp
```diff
@@ -0,0 +1,207 @@
...
+#define MARLIN_PIN_15 15
+
+#define MARLIN_PORT_A 0
+#define MARLIN_PORT_B 1
+#define MARLIN_PORT_C 2
+#define MARLIN_PORT_D 3
+#define MARLIN_PORT_E 4
+#define MARLIN_PORT_F 5
+
+#define MARLIN_PORT_PIN(port, pin) ((16 * (port)) + pin)
```
dragomirecky, line 39: /question/ Those macros seem to be an alternative interface for our HAL, that are not needed in the declarations below. What about having those Marlin-wrapper-related declarations in a separate file?

src/common/Pin.hpp
```diff
@@ -0,0 +1,207 @@
...
+
+/**
+ * @brief Container for all instances of its class
+ *
+ * Use only for objects of static storage duration. It is not possible meaningfully destroy object
+ * of this class. (Do not allocate on stack or heap.)
+ * Implicitly creates linked list of all created objects of its class. To conserve RAM, pointers to
+ * previous items in list are constant.
+ */
+class ConfigurableIndestructible {
```
dragomirecky, line 51: /suggestion/ What about having all the declarations/definitions in this file within a `buddy::hw` namespace? It would help us create boundaries between different modules and we wouldn't be polluting global namespace that much.

src/common/hwio_a3ides_2209_02.cpp
```diff
@@ -640,18 +632,18 @@ void digitalWrite(uint32_t ulPin, uint32_t ulVal) {
...
+        case MARLIN_PORT_PIN_Z_ENA:
+        case MARLIN_PORT_PIN_X_ENA:
+        case MARLIN_PORT_PIN_Z_STEP:
+        case MARLIN_PORT_PIN_E0_DIR:
+        case MARLIN_PORT_PIN_E0_STEP:
+        case MARLIN_PORT_PIN_E0_ENA:
+        case MARLIN_PORT_PIN_Y_DIR:
+        case MARLIN_PORT_PIN_Y_STEP:
+        case MARLIN_PORT_PIN_Y_ENA:
+        case MARLIN_PORT_PIN_Z_DIR:
```
dragomirecky, line 646: Marku, ty víš.

src/common/hwio_pindef_MINI.h
```diff
@@ -1,46 +1,124 @@
...
+extern InputPin zMin;
+
+#define MARLIN_PORT_X_DIAG     MARLIN_PORT_E
+#define MARLIN_PIN_X_DIAG      MARLIN_PIN_2
+#define MARLIN_PORT_PIN_X_DIAG MARLIN_PORT_PIN(MARLIN_PORT_X_DIAG, MARLIN_PIN_X_DIAG)
+extern InputPin xDiag;
+
+#define MARLIN_PORT_Y_DIAG     MARLIN_PORT_E
+#define MARLIN_PIN_Y_DIAG      MARLIN_PIN_1
+#define MARLIN_PORT_PIN_Y_DIAG MARLIN_PORT_PIN(MARLIN_PORT_Y_DIAG, MARLIN_PIN_Y_DIAG)
```
dragomirecky, line 25: /for discussion/ Having the `MARLIN_PORT_PIN_<NAME>` with every pin definition seems a wee bit repetitive. What about something like this:

```C
#define MARLIN_PORT_<NAME> <VALUE>
#define MARLIN_PIN_NUMBER_<NAME> <VALUE>
```
and then having a global `MARLIN_PIN(name)` macro, which would expand to `MARLIN_PORT_PIN(MARLIN_PORT_<NAME>, MARLIN_PIN_NR_<NAME>)`, changing the usage to `MARLIN_PIN(Y_DIAG)` in the case above.

src/main_MINI.cpp
```diff
@@ -158,6 +156,26 @@ volatile uint32_t Tacho_FAN1;
...
+OutputPin xStep(static_cast<IoPort>(MARLIN_PORT_X_STEP), static_cast<IoPin>(MARLIN_PIN_X_STEP), InitState::reset, OMode::pushPull, OSpeed::low);
+OutputPin yStep(static_cast<IoPort>(MARLIN_PORT_Y_STEP), static_cast<IoPin>(MARLIN_PIN_Y_STEP), InitState::reset, OMode::pushPull, OSpeed::low);
+OutputPin zStep(static_cast<IoPort>(MARLIN_PORT_Z_STEP), static_cast<IoPin>(MARLIN_PIN_Z_STEP), InitState::reset, OMode::pushPull, OSpeed::low);
+OutputPin e0Step(static_cast<IoPort>(MARLIN_PORT_E0_STEP), static_cast<IoPin>(MARLIN_PIN_E0_STEP), InitState::reset, OMode::pushPull, OSpeed::low);
+OutputPin xDir(static_cast<IoPort>(MARLIN_PORT_X_DIR), static_cast<IoPin>(MARLIN_PIN_X_DIR), InitState::reset, OMode::pushPull, OSpeed::low);
+OutputPin yDir(static_cast<IoPort>(MARLIN_PORT_Y_DIR), static_cast<IoPin>(MARLIN_PIN_Y_DIR), InitState::reset, OMode::pushPull, OSpeed::low);
+OutputPin zDir(static_cast<IoPort>(MARLIN_PORT_Z_DIR), static_cast<IoPin>(MARLIN_PIN_Z_DIR), InitState::reset, OMode::pushPull, OSpeed::low);
+OutputPin e0Dir(static_cast<IoPort>(MARLIN_PORT_E0_DIR), static_cast<IoPin>(MARLIN_PIN_E0_DIR), InitState::reset, OMode::pushPull, OSpeed::low);
+
```
dragomirecky, line 178: What about moving those definitions outside this file and having them within the `buddy::hw` namespace? It would allow us to merge configurations of different printers, plus it would be a next step in reducing the clutter in all the `main*` files.

lib/Middlewares/Third_Party/FreeRTOS/Source/CMSIS_RTOS/cmsis_os.h
```diff
@@ -26,7 +26,7 @@
  *
  *----------------------------------------------------------------------------
  *
- * Portions Copyright � 2016 STMicroelectronics International N.V. All rights reserved.
+ * Portions Copyright � 2016 STMicroelectronics International N.V. All rights reserved.
```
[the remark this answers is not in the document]
mkbel: fixed

src/common/hwio_pindef_MINI.h
mkbel: done

src/common/hwio_a3ides_2209_02.cpp
mkbel: done

src/common/Pin.hpp
dragomirecky: Still not addressed?

src/common/Pin.hpp
dragomirecky: Still not addressed?

mkbel:
> Looks good ok_hand
> 
> A few questions:
> 
>     1. How are we going to support dummy pins (pins with software implementation)?

To be done when needed.
>     2. Have you considered adding an option to pass a lambda function to the InputPin/OutputPin constructor for an advanced configuration of the pin?
It is not needed now, as there is nearly zero overhead for new pin type, you can make special pin type and implement its configure function.

lib/Drivers/CMSIS/Include/arm_math.h
```diff
@@ -39,7 +39,7 @@
  * -------------------------------------------------------------------- */
 
 /**
-   \mainpage CMSIS DSP Software Library
+   \file CMSIS DSP Software Library
```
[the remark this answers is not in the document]
mkbel: Discussed, will be left there.

src/common/Pin.hpp
mkbel: fixed

src/main_MINI.cpp
mkbel: Moved to hwio_pindef_MINI.h

src/common/Pin.hpp
mkbel: Extracted into separate file.

src/common/Pin.hpp
mkbel: done

DRracer (approved):

PR #627, opened 2020-06-25
M109: Shorten wait time BFW-1036

espr14:


espr14:
So you propose the same behavior but different implementation?

PR #643, opened 2020-06-30
Fix -BSOD/deadlock when connecting/disconnecting octoprint (M876)

XPila:
BFW-1059

### DIFF
--- a/src/gui/screen_home.cpp
+++ b/src/gui/screen_home.cpp
@@ -114,11 +114,12 @@ void screen_home_init(screen_t *screen) {
         }
     }
 
-    if (pw->is_starting)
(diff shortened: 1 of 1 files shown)


src/gui/screen_home.cpp
```diff
@@ -114,7 +114,13 @@ void screen_home_init(screen_t *screen) {
         }
     }
 
-    if (!marlin_update_vars(MARLIN_VAR_MSK(MARLIN_VAR_MEDIAINS))->media_inserted)
+    if (pw->is_starting)
```
DRracer, line 117: please wrap the if() block into { } ... especially usefull with the multi-line comment

PR #657, opened 2020-07-08
Refactor dbg.c/h BFW-850

espr14:


src/common/dbg.c
```diff
@@ -43,15 +44,23 @@ static inline void _dbg_unlock(void) {
...
     va_end(va);
-    line[len++] = '\n';
-    for (int i = 0; i < len; i++) {
-        ITM_SendChar(line[i]);
+
+    if (len < 0) {
+        _dbg_unlock();
+        return;
     }
+    len = MIN(len, DBG_MAXLINE - 1);
```
DRracer, line 58: I'd like to see unit tests for these fixes so that no other buffer overrun can occur in the future

PR #662, opened 2020-07-09
Error: New screen layout BFW-1082

espr14:


tests/unit/README.md
```diff
@@ -41,7 +41,7 @@ ctest .
 ```bash
 mkdir -p build_tests \
 && cd build_tests \
-&& rm -r * \
+&& if [ "$(ls -A $DIR)" ]; then rm -r *; fi \
```
DRracer, line 44: `rm -rf *` does the same and is shorter

PR #666, opened 2020-07-09
Add M601 & M602 GCODE

Chleba:
- added M601 - Pause & Park nozzle
- added M602 - Resume & Unpark nozzle

BFW-1053


Chleba:
Obsolete - after some discussion, we are going to handle this in different way.

PR #672, opened 2020-07-12
Fix removed and insert USB medium outside home screen with no filament in sensor

Chleba:
BFW-1107, BFW-1108, BFW-1111, BFW-1110


src/gui/screen_home.cpp
```diff
@@ -163,6 +163,10 @@ int screen_home_event(screen_t *screen, window_t *window, uint8_t event, void *p
         p_window_header_event_clr(&(pw->header), MARLIN_EVT_MediaError);
     }
 
+    if (p_window_header_event_clr(&(pw->header), MARLIN_EVT_MediaRemoved)) {
```
DRracer, line 166: Please walk through the history of this file to see if we reordered the checks for various MARLIN_EVT_errors here. I think we did ... I'd like to avoid breaking some other functionality

src/gui/screen_print_preview.cpp
```diff
@@ -313,13 +313,17 @@ static int screen_print_preview_event(screen_t *screen, window_t *window,
...
+        rect_ui16_t msg_box_size = gui_defaults.scr_body_no_foot_sz;
+        msg_box_size.y = 0;
+        msg_box_size.h += gui_defaults.header_sz.h;
         suppress_draw = true;
-        Sound_Play(eSOUND_TYPE_StandardAlert);
+
+        Sound_Play(eSOUND_TYPE_SingleBeep);
         const char *btns[3] = { N_("YES"), N_("NO"), N_("IGNORE") };
-        switch (gui_msgbox_ex(string_view_utf8::MakeNULLSTR(),
+        switch (gui_msgbox_ex(string_view_utf8::MakeRAM((const uint8_t *)gcode_file_name),
```
DRracer, line 323: document, that using MakeRAM in this case is safe, because gcode_file_name lives long enough
... I'm still thinking about renaming MakeRAM to something obscure like: `IKnowTheContainedPointerWillLiveLongEnoughWhenWrappingItIntoThisStringViewObject()` :)

src/gui/screen_print_preview.cpp
```diff
@@ -313,13 +313,17 @@ static int screen_print_preview_event(screen_t *screen, window_t *window,
...
             _("Filament not detected. Load filament now? Select NO to cancel, or IGNORE to disable the filament sensor and continue."),
             MSGBOX_BTN_CUSTOM3,
-            gui_defaults.scr_body_no_foot_sz,
+            msg_box_size,
```
DRracer, line 326: what is this change?

Chleba: After discussion with @RadekVana new gui default rect is added - scr_fullscreen with position and dimensions for a full display screen, so creating new rect as was before is obsolete now. Btw. this was fix for BFW-1108

src/gui/screen_print_preview.cpp
Chleba: We talked about this case with gcode_file_name and even though I love that new name, I don't think is short enough ... MakeRAM is a little bit confusing, I suggest "MakeText" or "WrapText" ?

src/gui/screen_home.cpp
Chleba: I've checked history around dozen commits back and we didn't changed order of anything, so It should be fine and not break anything.

DRracer: ok

src/gui/screen_print_preview.cpp
DRracer: yes, it should be something like FromRAM or similar ... still awaiting better suggestions ;) .

src/gui/screen_print_preview.cpp
DRracer: ok

PR #673, opened 2020-07-12
New FAN control with RPM measurement

XPila:
BFW-1106
Finished:
 PWM control
 RPM measurement
 display RPM in system menu + new marlin vars


src/common/fanctl.h
```diff
@@ -0,0 +1,123 @@
...
+// this structure contain variables for software pwm fan control with phase-shifting
+typedef struct _fanctl_pwm_t {
+    uint8_t pin : 8; // output pwm pin
+    uint8_t min : 8; // minimum pwm value
+    uint8_t max : 8; // maximum pwm value
+    union {
+        struct {              // flags:
+            bool ini : 1;     //  hw initialized
+            bool out : 1;     //  current pwm output state (0/1)
+            bool pha_ena : 1; //  phase shift enabled
```
DRracer, line 19: make sure you don't rely on the ordering of bits - the compiler is allowed to reorder the bits arbitrarily

src/common/fanctl.h
```diff
@@ -0,0 +1,123 @@
...
+    { return m_pwm.get_PWM(); }
+    inline uint16_t getActualRPM() // get actual (measured) RPM
+    { return m_ActualRPM; }
+    inline uint16_t getTargetRPM() // get target RPM
+    { return m_TargetRPM; }
+    // setters
+    void setPWM(uint8_t pwm);              // set PWM value - switch to non closed-loop mode
+    void setTargetRPM(uint16_t targetRPM); // set target RPM - switch to closed-loop mode
+private:
+    uint8_t m_pinTach;    // input tacho pin (set in constructor)
```
DRracer, line 93: when you are done make sure the members are ordered well (prevent getting alignment bytes stuffed between the members)

src/common/fanctl.h
XPila: I want just save RAM, maybe we can remove the union/flg and keep only the structure with bit addressed members.

src/common/fanctl.cpp
```diff
@@ -0,0 +1,182 @@
...
+    if (++cnt >= max_value) {
+        cnt = 0;
+        if (val != pwm) { // pwm changed
+            val = pwm;    // update cached value
+            pha = 0;      // reset phase
+            if ((val > 1) && (val <= (max_value / 2))) {
+                pha_max = max_value - val;
+                uint8_t steps = max_value / val; // calculate number of steps
+                if (steps < 3)
+                    steps = 3;               // limit steps >= 3
```
espr14, line 51:
```suggestion
                const uint8_t steps = std::max(3, max_value / val); // calculate number of steps (at least 3)
```

DRracer: @espr14 hmmm, this may produce unnecessarily larger and slower code on the 8-bit platform: std::max will take ints (aka 16bit) to compute the result, which is 2x more data than necessary.

XPila: I agree with @DRracer

src/common/fanctl.h
```diff
@@ -0,0 +1,145 @@
+// fanctl.h
+#ifndef _FANCTL_H
```
RadekVana, line 2: i would prefer `#pragma once`

src/common/marlin_client.c
```diff
@@ -21,8 +21,8 @@
 //#define DBG_EVT_MSK (MARLIN_EVT_MSK_ALL & ~MARLIN_EVT_MSK(MARLIN_EVT_Acknowledge))
 
 //trace variable change notifications (client side), to disable trace undef DBG_VAR_MSK
-//#define DBG_VAR DBG
-//#define DBG_VAR_MSK (MARLIN_VAR_MSK_ALL & ~MARLIN_VAR_MSK_TEMP_ALL)
+#define DBG_VAR     DBG
```
RadekVana, line 24: should not this be disabled?

PR #676, opened 2020-07-14
Refactor: display_helper BFW-1020

espr14:


src/guiapi/src/display_helper.cpp
```diff
@@ -197,57 +198,58 @@ void render_text_align(rect_ui16_t rc, string_view_utf8 text, const font_t *font
...
+
+    const rect_ui16_t rect_in = { rc_txt.x, rc_txt.y, uint16_t(rc_txt.w - unused_pxls), rc_txt.h };
+    fill_between_rectangles(rc, rect_in, clr_bg);
+    text.rewind();
+    // 2nd pass reading the string_view_utf8 - draw the text
+    render_text(rc_txt, text, font, clr_bg, clr_fg);
 }
 
-void render_icon_align(rect_ui16_t rc, uint16_t id_res, color_t clr0, uint16_t flags) {
+void render_icon_align(const rect_ui16_t &rc, uint16_t id_res, color_t clr_bg, uint16_t flags) {
```
DRracer, line 231: adding a const reference here means you cannot pass an immediate value into the function. This can cause some discomfort.
AFAIK rect_uin16t is just 8B long while the pointer to rc (or a reference) is 4B long.

PR #679, opened 2020-07-14
Add language singleton to set, get & store active language

Chleba:
- Updated EEPROM values with new sound Volume & Language
- Added convert function from old EEPROM version to new (4.1.0 -> 4.2.0)
- Added language EEPROM managing singleton class to set, get & store new active language set by user
- Prepared splash screen if language is not set to open the language setting screen (TODO)

BFW-1117

### DIFF
--- a/src/common/eeprom.cpp
+++ b/src/common/eeprom.cpp
@@ -66,6 +66,8 @@ typedef struct _eeprom_vars_t {
     char LAN_HOSTNAME[LAN_HOSTNAME_MAX_LEN + 1];
     int8_t TIMEZONE;
     uint8_t SOUND_MODE;
+    uint8_t SOUND_VOLUME;
+    uint16_t LANGUAGE;
     char _PADDING[EEPROM__PADDING];
     uint32_t CRC32;
 } eeprom_vars_t;
@@ -103,6 +105,8 @@ static const eeprom_entry_t eeprom_map[] = {
     { "LAN_HOSTNAME",    VARIANT8_PCHAR, LAN_HOSTNAME_MAX_LEN + 1, 0 }, // EEVAR_LAN_HOSTNAME
     { "TIMEZONE",        VARIANT8_I8,    1, 0 }, // EEVAR_TIMEZONE
     { "SOUND_MODE",      VARIANT8_UI8,   1, 0 }, // EEVAR_SOUND_MODE
+    { "SOUND_VOLUME",      VARIANT8_UI8,   1, 0 }, // EEVAR_SOUND_VOLUME
+    { "LANGUAGE",					VARIANT8_UI16,   1, 0 }, // EEVAR_LANGUAGE
     { "_PADDING",        VARIANT8_PCHAR, EEPROM__PADDING, 0 }, // EEVAR__PADDING32
     { "CRC32",           VARIANT8_UI32,  1, 0 }, // EEVAR_CRC32
 };
@@ -136,6 +140,8 @@ static const eeprom_vars_t eeprom_var_defaults = {
     "PrusaMINI",     // EEVAR_LAN_HOSTNAME
     0,               // EEVAR_TIMEZONE
     0xff,            // EEVAR_SOUND_MODE
+    0x64,            // EEVAR_SOUND_VOLUME
+    0xffff,            // EEVAR_LANGUAGE
     "",              // EEVAR__PADDING
     0xffffffff,      // EEVAR_CRC32
 };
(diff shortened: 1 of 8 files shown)


src/common/language_eeprom.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+/*!
+ * Setter & getter class to store language code into a EEPROM
+ * Constructor will check if language code is not set, then it will set
+ * a default value.
+ * If it's set, then Translator is called to start translating.
+ */
+LangEEPROM::LangEEPROM()
+    : _language(0) {
+    uint16_t _language = static_cast<uint16_t>(eeprom_get_var(EEVAR_LANGUAGE).ui16);
```
DRracer, line 13: Are you sure the eeprom is properly initialized when the constructor get called? I'm sure it is not

src/common/language_eeprom.cpp
```diff
@@ -0,0 +1,46 @@
...
+    if (_language == static_cast<uint16_t>(0xffff)) {
```
DRracer, line 14: default lang should be "en"

src/common/language_eeprom.cpp
```diff
@@ -0,0 +1,46 @@
...
+        setLanguage(Translations::MakeLangCode("cs"));
+    } else {
+        Translations::Instance().ChangeLanguage(_language);
```
DRracer, line 17: I'm not sure if the LangEEPROM class should do anything else than reading and saving the lang code - i.e. calling Translations::Instance here looks like an unnecessary dependency. I'd rather see this in the Translations class - explicitly obtaining the correct lang code from LangEEPROM

tests/unit/lang/CMakeLists.txt
```diff
@@ -1,2 +1,3 @@
 add_subdirectory(string_view_utf8)
 add_subdirectory(translator)
+add_subdirectory(eeprom)
```
DRracer, line 3: looks like the unit tests are missing from this PR

src/common/language_eeprom.cpp
Chleba: I am sure it is, because it's initialized first thing after eeprom class is inicialized.

src/common/language_eeprom.cpp
Chleba: sure, that was just a test - it's gonna be changed

src/common/language_eeprom.cpp
Chleba: This class after discussion will stay as a manager class with Translator dependency to change newly set language code right after it's stored into a EEPROM. This will avoid unneccessary explicit calling getting language and then call change language to translator.

tests/unit/lang/CMakeLists.txt
Chleba: Missing unit tests added

src/common/eeprom.cpp
```diff
@@ -103,6 +105,8 @@ static const eeprom_entry_t eeprom_map[] = {
     { "LAN_HOSTNAME",    VARIANT8_PCHAR, LAN_HOSTNAME_MAX_LEN + 1, 0 }, // EEVAR_LAN_HOSTNAME
     { "TIMEZONE",        VARIANT8_I8,    1, 0 }, // EEVAR_TIMEZONE
     { "SOUND_MODE",      VARIANT8_UI8,   1, 0 }, // EEVAR_SOUND_MODE
+    { "SOUND_VOLUME",      VARIANT8_UI8,   1, 0 }, // EEVAR_SOUND_VOLUME
+    { "LANGUAGE",					VARIANT8_UI16,   1, 0 }, // EEVAR_LANGUAGE
```
DRracer, line 109: why is this code format so weird?

tests/unit/lang/eeprom/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
+# define the test executable
+add_executable(language_eeprom_tests
+../../../../src/lang/translator.cpp
+# ../../../../src/common/eeprom.cpp
```
DRracer, line 4: if the eeprom is not needed (not even in the future), remove it from the cmake file

src/common/eeprom.cpp
Chleba: Clang format if in this file and I did not know about that - sorry

tests/unit/lang/eeprom/CMakeLists.txt
Chleba: it's gone

PR #694, opened 2020-07-17
Localization support - fine-tuning texts and features

DRracer:
BFW-1126


tests/unit/README.md
```diff
@@ -3,7 +3,8 @@
 ```bash
 # create a build folder and run cmake within it
 cd ../.. && mkdir build_tests && cd build_tests
-cmake .. -G Ninja
+cmake .. -G Ninja -DCUSTOM_COMPILE_OPTIONS:STRING=-Dstrlcpy=strncpy
```
espr14, line 6: Please, remove lines 15,16. And update line 51.

src/lang/string_view_utf8.hpp
```diff
@@ -141,11 +141,27 @@ class string_view_utf8 {
         rewind_impl(attrs);
     }
 
-    /// this
+    /// returns true if the string if of type NULLSTR - typically used as a replacement for nullptr or "" strings
```
espr14, line 144: if the string `if` - should be `is`?

PR #704, opened 2020-07-21
Configurable ramming

mkbel:


src/marlin_stubs/pause/pause.cpp
```diff
@@ -318,30 +318,23 @@ bool Pause::FilamentUnload() {
...
-        { 3, 800 },
-        { 2, 1200 },
-        { 2, 2200 },
-        { 2, 2600 }, // end of ramming
-        { -2, 2200 },
-        { -20, 3000 },
-        { -30, 4000 }, // end of pre-unload
-    };
+    static const RamUnloadSeqItem ramUnloadSeq[] = FILAMENT_UNLOAD_RAMMING_SEQUENCE;
+    typeof(RamUnloadSeqItem::e) ramUnloadLength = 0; //Sum of ramming distances starting from first retraction
```
DRracer, line 322: C++ language has no such thing as `typeof`. 
GCC's typeof is an extension used before C++11.
Use `decltype`

mkbel: fixed

PR #714, opened 2020-07-24
Spanish language, few fixes

eried:
Spanish is my mother language but I do not usually use anything on spanish because the horrible machine translations make devices hard to understand.

I think there is going to be a lot of spacing issues in spanish, I just tried to correct some of the most evident mistakes but I think it needs some inputs from spanish people:

1) Hotend - nozzle - extruder
Is interpreted as "boquilla", but I think the right definition would be that the whole setup is "Extrusor" (including the gearing) and the tip is "punta del extrusor"

2) Home, Parking, Purging, Ramming
This concepts work nice in english but they sound silly in spanish, I think that also needs discussion

3) Heatbed, Heating, Cooling
I think "base/cama" is the most proper spanish for things relating with heat in the printing area. In the original translation there is few differences, some saying "base calefactable", etc so the whole concept about this needs to be brainstormed. (Instead of heating and cooling "calentando" some phrases could play with temperature-words i.e.: "la temperatura es muy baja para extruir" -> the temperature is too low to extrude)

And **finally**, for fitting the string sizes I am not sure if you have an emulator to test the firmware.

### DIFF
--- a/src/lang/po/es/Prusa-Firmware-Buddy_es.po
+++ b/src/lang/po/es/Prusa-Firmware-Buddy_es.po
@@ -10,16 +10,16 @@ msgstr ""
 #: src/gui/screen_menu_version_info.cpp:90
 #, c-format
 msgid "\nBootloader Version\n%d.%d.%d\n\nBuddy Board\n%d.%d.%d\n%s"
-msgstr "Versión del cargador %d.%d.%d\nPlaca Buddy %d.%d.%d %s"
+msgstr "Versión de bootloader %d.%d.%d\nPlaca Buddy %d.%d.%d %s"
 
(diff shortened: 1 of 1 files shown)


DRracer:
@eried Could you please do the following:
- checkout the current master and look at the most recent Spanish translation PO file?
- fix it and update your PR here accordingly (or file a new PR with your fixes)
- try building the FW with your changes

**Please note** - the FW build now finally includes a full-featured build script for the languages from the PO files directly - so when you build your FW and upload to your MINI, you should see your changes. Please let us know if that works for you, this build system improvement is relatively new.

Also, there is another PR to improve the Spanish translation: #999 , please feel free to collaborate

eried:
Ah, yes I will check that updated PR

PR #715, opened 2020-07-24
Alphabet file sort

michalrudolf:
BFW-1131

Add variable to eeprom and connect it to already established file sort in screen_filebrowser


src/guiapi/include/MItem_tools.hpp
```diff
@@ -233,6 +233,17 @@ class MI_SOUND_TYPE : public WI_SWITCH_t<8> {
     virtual void OnChange(size_t old_index) override;
 };
 
+class MI_SORT_FILES : public WI_SWITCH_t<2> {
+    constexpr static const char *const label = "Sort files by";
```
DRracer, line 237: Prepare for translation and update the POT file accordingly

PR #718, opened 2020-07-26
Fix Octoprint

ojrik:
Exiting the tune menu, the extruder does not lift and does not go to the park position.


Chleba:
This PR is duplicit to PR#719

PR #719, opened 2020-07-26
Fix Octoprint

ojrik:
Exiting the tune menu, the extruder does not lift and does not go to the park position.


DRracer:
@Chleba @ojrik confirming the fix, thank you @ojrik for your PR, merging.

ojrik:
This error is also in the master

PR #748, opened 2020-08-04
Sdcard commands

XPila:
BFW-837


src/common/marlin_server.cpp
```diff
@@ -442,7 +443,8 @@ void marlin_server_print_pause(void) {
 void marlin_server_print_resume(void) {
     if (marlin_server.print_state == mpsPaused) {
         marlin_server.print_state = mpsResuming_Begin;
-    }
+    } else
+        marlin_server_print_start(0);
```
DRracer, line 447: please use nullptr in C++ code

src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
```diff
@@ -34,12 +52,24 @@ void GcodeSuite::M25() {
 
 // M26 - Set SD position
 void GcodeSuite::M26() {
-    //TODO
+    if ((media_get_state() == media_state_INSERTED) && parser.seenval('S'))
+        media_print_set_position(parser.value_long());
```
DRracer, line 56: this will fail with files >2GB ... use `parser.value_ulong()`

PR #749, opened 2020-08-04
PUT new command gpcf for configuring pins (input/output mode and pullup)

XPila:
BFW-729


src/common/putslave.c
```diff
@@ -420,6 +422,29 @@ int putslave_do_cmd_a_gpio(uartslave_t *pslave, char *pstr) {
...
+    int pull = 0;
+    int n = 0;
+    if (sscanf(pstr, "%d%n", &gpio, &n) != 1)
+        return UARTSLAVE_ERR_SYN;
+    if ((gpio < TPA0) || (gpio > TPE15))
+        return UARTSLAVE_ERR_OOR;
+    pstr += n;
+    if (sscanf(pstr, "%d%n", &mode, &n) != 1)
+        return UARTSLAVE_ERR_SYN;
+    if ((mode < 0) || (mode > 1))
```
espr14, line 437: `mode` cannot be negative because it would have to read extra `-` char => 2 chars returns `UARTSLAVE_ERR_SYN`

PR #761, opened 2020-08-06
BSOD when trinamic driver connection fails. BFW-1165

mkbel:
BFW-1165


RadekVana (approved):

danopernis:
This is already merged, see https://github.com/prusa3d/Prusa-Firmware-Buddy/blob/245e7e32c8282d338fb08fd3e4ccce0a95ef15d8/lib/Marlin/Marlin/src/feature/tmc_util.cpp#L1197-L1200

PR #774, opened 2020-08-11
G-CODE for another printer type warning

michalrudolf:
BFW-1093

When user hits **Print button** in Screen print preview, it will check for _printer_model_ comment in selected gcode. If it doesn't equal "**MINI**", then it pops up warning messagebox saying this gcode was created for another printer type.


src/gui/screen_print_preview.cpp
```diff
@@ -109,6 +111,12 @@ GCodeInfo::GCodeInfo() {
             sscanf(value_buffer, "%u", &filament_used_mm);
         } else if (name_equals("filament used [g]")) {
             sscanf(value_buffer, "%u", &filament_used_g);
+        } else if (name_equals("printer_model")) {
+            if (strncmp(value_buffer, "MINI", sizeof(value_buffer)) == 0) {
```
dragomirecky, line 115: /suggestion/ Instead of hardcoding the `"MINI"` value, add it to the printer's configuration (if it's not there already).

src/gui/screen_print_preview.cpp
```diff
@@ -134,10 +142,19 @@ GCodeInfoWithDescription::GCodeInfoWithDescription(window_frame_t *frame)
     }
 }
 
+static void print_button_press() {
+    if (!valid_printer_settings) {
+        MsgBoxTitle(_("WARNING:"), _("This G-CODE was set up for another printer type."),
+            Responses_Ok, 0, GuiDefaults::RectScreenBodyNoFoot);
+        Sound_Play(eSOUND_TYPE_SingleBeep);
```
dragomirecky, line 149: It seems to me, that the user is warned but they have no way to actually not-start the printing process. Right?

michalrudolf: Yes, that is the behaviour of MK3 in the same situation. Is something else requested?

src/gui/screen_print_preview.cpp
michalrudolf: commit `e5f0c48`

src/gui/screen_print_preview.cpp
dragomirecky: I think it might be nicer to give the user two options: „Ok“ and „Cancel“.

include/printers.h
```diff
@@ -10,4 +10,11 @@
...
     #error "macro PRINTER_TYPE not defined"
-#endif
+    #define PRINTER_MODEL "Not defined"
+#else
+    #if (PRINTER_TYPE == PRINTER_PRUSA_MINI)
+        #define PRINTER_MODEL "MINI"
+    #else
+        #define PRINTER_MODEL "Unknown"
+    #endif // (PRINTER_TYPE == PRINTER_PRUSA_MINI)
+#endif     // PRINTER_TYPE
```
dragomirecky, line 20: It took me some time to fully comprehend what is going on here :D, but I think we can simplify it to something like this

```
#if defined(PRINTER_TYPE) && PRINTER_TYPE == PRINTER_PRUSA_MINI
  #define PRINTER_MODEL "MINI"
#else
  #error "Unknown printer type"
#endif
```

michalrudolf: commit `f0485ef`

PR #784, opened 2020-08-12
Static analysis: fix 8 BFW-1178

espr14:


espr14:
This code should be safer but due to C language is not. Moreover, the binary code is bigger. So it does not make sense to merge.

PR #796, opened 2020-08-15
Window rect invalidations

RadekVana:
BFW-1176


src/guiapi/src/window.cpp
```diff
@@ -67,19 +71,42 @@ void window_t::SetCapture() {
...
+
+void window_t::ShowAfterDialog() {
+    if (flag_hidden_behind_dialog) {
+        flag_hidden_behind_dialog = false;
+        //must invalidate even when is not visible
         Invalidate();
     }
 }
 
+void window_t::HideBehindDialog() {
```
Chleba, line 99: ShowAfterDialog and HideBehindDialog doing the same stuff. Is this work in progress to invalidate those hidden parts which intersect with dialog rect ?

RadekVana: It is inversive operation

RadekVana: oh it ha same code, its a bug
thx

PR #804, opened 2020-08-17
Fix unit test run file dependency

marekmosna-prusa3d:
BFW-1201


marekmosna-prusa3d:
contain unmerged changes

marekmosna-prusa3d:
contain unmerged changes

PR #811, opened 2020-08-17
fix rendering

RadekVana:
BFW-1205


src/guiapi/src/display_helper.cpp
```diff
@@ -292,6 +300,7 @@ point_ui16_t font_meas_text(const font_t *pf, string_view_utf8 *str, uint16_t *n
             x += char_w;
         h = y + char_h;
     }
+    str->rewind();
```
DRracer, line 303: hmmm, do you really want to hide the rewind call in this place?

RadekVana: I want this function to "clean" after itself
added comment

PR #815, opened 2020-08-19
Printer type checking

michalrudolf:
BFW-1093

### DIFF
--- a/src/gui/screen_print_preview.cpp
+++ b/src/gui/screen_print_preview.cpp
@@ -10,11 +10,13 @@
 #include "DialogHandler.hpp"
 #include "ScreenHandler.hpp"
 #include "print_utils.hpp"
+#include "printers.h"
 
 #define DBG _dbg0
 
 static const char *gcode_file_name = NULL;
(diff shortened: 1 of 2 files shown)


src/gui/screen_print_preview.cpp
```diff
@@ -88,6 +90,7 @@ GCodeInfo::GCodeInfo() {
     filament_type[0] = 0;
     filament_used_mm = 0;
     filament_used_g = 0;
+    valid_printer_settings = false;
```
dragomirecky, line 93: /question/ Does this mean, that if the gcode does not contain the comment, we will report to the user that it is not „fit“ their printer?

michalrudolf: That this gcode was set up for anoter printer. Yes.

michalrudolf: It should only pop up if there is this comment and it says its a different printer right?

dragomirecky (approved):

dragomirecky (approved):

PR #823, opened 2020-08-19
move and rename some display functions

RadekVana:


src/guiapi/include/display_math_helper.h
```diff
@@ -0,0 +1,94 @@
...
+//basic mathematical operations for display
+#pragma once
+
+#include <inttypes.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif //__cplusplus
+
+__attribute__((used)) inline uint16_t swap_ui16(uint16_t val) {
```
DRracer, line 11: I don't like the non-standard attributes here, but the same code-magic was done in st77xxxxx before.

DRracer (approved):

PR #827, opened 2020-08-21
Error: refactor error screen

espr14:
BFW-1211


espr14:
Don't review ./lib/Prusa-Error-Codes except ./lib/Prusa-Error-Codes/12.

DRracer (approved):

DRracer (approved):

PR #835, opened 2020-08-24
Fix webServerTask bosd

marekmosna-prusa3d:
hotfix BFW-1207


DRracer:
This is a possible fix of BSOD crashes in FW 4.2.0. The correct cause is yet to be found, however this seems to help.

XPila (approved):

PR #844, opened 2020-08-27
fix bsod in variant_free/malloc_get_usable_size (reported as "webServerTask bsod")

XPila:
BFW-1207

### DIFF
--- a/src/common/marlin_client.c
+++ b/src/common/marlin_client.c
@@ -114,10 +114,11 @@ void marlin_client_shdn(void) {
     //TODO
 }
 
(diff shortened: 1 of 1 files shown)


src/common/marlin_client.c
```diff
@@ -114,10 +114,11 @@ void marlin_client_shdn(void) {
     //TODO
 }
 
+variant8_t client_msg[MARLIN_MAX_CLIENTS];
```
DRracer, line 117: this could have been static inside marlin_client_loop(void)

PR #854, opened 2020-09-01
Lang: Translate error codes' headers BFW-1210

espr14:


src/common/bsod.cpp
```diff
@@ -224,45 +227,41 @@ void general_error_run() {
...
+void draw_error_screen(const uint16_t error_code_short) {
+
+    const uint16_t error_code = ERR_PRINTER_CODE * 1000 + error_code_short;
+
+    /// search for proper text according to error code
+    const char *text_title;
+    const char *text_body;
+
+    uint32_t i = 0;
+    while (i < sizeof(error_list) && error_code_short != error_list[i].err_num) {
```
DRracer, line 239: ideally use something like std::find_if or similar - you are sequentially searching for an element in an array

src/guiapi/include/Rect16.h
```diff
@@ -18,11 +18,11 @@ enum class ShiftDir_t {
 
 ////////////////////////////////////////////////////////////////////////////////
 ///
-/// Class prepesents rectagle in graphics system where x rising to the right
+/// Class represents rectangle in graphics system where x rising to the right
```
DRracer, line 21: when improving EN wording, make sure you add a/the as well ;) .... (This) class represents A rectangle in A graphics system...

src/guiapi/src/MItem_tools.cpp
```diff
@@ -222,6 +223,78 @@ void MI_EE_SAVEXML::click(IWindowMenu & /*window_menu*/) {
     eeprom_save_xml_to_usb("eeprom.xml");
 }
 
+/*****************************************************************************/
+MI_ES_12201::MI_ES_12201()
+    : WI_LABEL_t(label, 0, true, false) {
```
DRracer, line 228: I know these classes are debug-only, but probably they could have been templated ;)

DRracer (approved):
there are minor comments what to improve in the future, otherwise the PR looks ok

PR #860, opened 2020-09-02
replace cvariant8 by variant8_t and isolate

marekmosna-prusa3d:
BFW-1241


src/common/marlin_server.hpp
```diff
@@ -87,8 +87,8 @@ class FSM_notifier {
 template <int VAR_ID, class T>
 class Notifier : public FSM_notifier {
 public:
-    Notifier(ClientFSM type, uint8_t phase, T min, T max, uint8_t progress_min, uint8_t progress_max)
```
RadekVana, line 90: do not change signature.
U can use partial template specialization for T as float

RadekVana (changes requested):

RadekVana (approved):

PR #865, opened 2020-09-04
isolated variant8_t unit tests

marekmosna-prusa3d:
BFW-1244


src/common/variant8.cpp
```diff
@@ -1,18 +1,64 @@
...
 #include <stdint.h>
 #include <malloc.h>
 
 #define VARIANT8_DBG_MALLOC
 
-#define _VARIANT8_TYPE(type, _8, _16, _32) ((variant8_t) { { _32 }, { _16 }, type, _8 })
+#define _VARIANT8_TYPE(type, _8, _16, _32) ((_variant8_t) { { _32 }, { _16 }, type, _8 })
 #define _VARIANT8_EMPTY()                  _VARIANT8_TYPE(VARIANT8_EMPTY, 0, .size = 0, .ui32 = 0)
 
+#pragma pack(push)
```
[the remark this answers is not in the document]
marekmosna-prusa3d: I can only agree...next steps will be focused on proper implementation of these functions

tests/unit/test_main.cpp
```diff
@@ -1,2 +1,7 @@
 #define CATCH_CONFIG_MAIN
+#define CATCH_CONFIG_ENABLE_BENCHMARKING
 #include "catch2/catch.hpp"
+
+TEST_CASE("type tests", "[system]") {
+    REQUIRE(sizeof(uint64_t) == 8);
```
DRracer, line 6: probably also add sizeof(char *) == 4 to make sure the tests only run in 32bit mode

marekmosna-prusa3d: Good opinion. Unfortunately I found out that testcases in this file run with every testfile due to linkage dependency of CATCH_CONFIG_MAIN. I will make separate system_tests.cpp where we will place such a useful type size tests.

tests/unit/test_main.cpp
```diff
@@ -1,2 +1,7 @@
 #define CATCH_CONFIG_MAIN
+#define CATCH_CONFIG_ENABLE_BENCHMARKING
```
marekmosna-prusa3d: I have to disable benchmarks due to unspecified behavior in other tests. It's really confusing from chatch2 to cause SEGFAULT in totally unrelated tests which are solid, well-formed and correct.

DRracer (approved):

PR #876, opened 2020-09-08
Add Menu Timeout and store it's value into the EEPROM

Chleba:
BFW-1255, BFW-1266


src/common/appmain.cpp
```diff
@@ -12,6 +12,7 @@
 #include "gpio.h"
 #include "sound.hpp"
 #include "language_eeprom.hpp"
+#include "gui.hpp"
```
DRracer, line 15: this is not good - appmain has no knowledge of GUI ... and shouldn't have

src/common/eeprom.cpp
```diff
@@ -13,7 +13,7 @@
 #include "wdt.h"
 #include "../Marlin/src/module/temperature.h"
 
-static const constexpr uint8_t EEPROM__PADDING = 1;
+static const constexpr uint8_t EEPROM__PADDING = 4;
```
DRracer, line 16: too bad we can't make an empty member, right?

src/common/appmain.cpp
```diff
@@ -87,6 +88,7 @@ void app_run(void) {
 
     uint8_t defaults_loaded = eeprom_init();
     LangEEPROM::getInstance();
+    menu_timeout_enabled = variant_get_ui8(eeprom_get_var(EEVAR_MENU_TIMEOUT));
```
DRracer, line 91: this has nothing to do with appmain - shall be moved to where it belongs

src/guiapi/include/gui.hpp
```diff
@@ -12,7 +12,7 @@ typedef void(gui_loop_cb_t)(void);
 
 extern gui_loop_cb_t *gui_loop_cb;
 
-extern int8_t menu_timeout_enabled;
+extern uint8_t menu_timeout_enabled;
```
DRracer, line 15: NO more extern variables!

DRracer (changes requested):
Move initialization to where it belongs, do not use extern variables (that's a very bad C-style habit completely unsafe)

src/common/eeprom.cpp
Chleba: too bad :_(

src/gui/guimain.cpp
```diff
@@ -128,14 +127,23 @@ extern "C" void gui_run(void) {
...
+    GuiDefaults::menu_timeout_enabled = variant_get_ui8(eeprom_get_var(EEVAR_MENU_TIMEOUT));
     while (1) {
         Screens::Access()->Loop();
         // show warning dialog on safety timer expiration
         if (marlin_event_clr(MARLIN_EVT_SafetyTimerExpired)) {
             MsgBoxInfo(_("Heating disabled due to 30 minutes of inactivity."), Responses_Ok);
         }
         gui_loop();
-        /*if (menu_timeout_enabled) {
+        /*if (marlin_message_received()) {
```
DRracer, line 138: what is this commented code?

src/common/support_utils.cpp
```diff
@@ -10,7 +10,8 @@
 #include "lang.h"
 #include "../../gui/wizard/selftest.h"
 #include "version.h"
-#include "eeprom.h"
+// #include "eeprom.h"
```
DRracer, line 13: remove the commented line

src/gui/guimain.cpp
Chleba: This is old pop up msg & menu timeout logic before GUI refactoring - new implementation is prepared to review

src/guiapi/include/GuiDefaults.hpp
```diff
@@ -20,4 +20,5 @@ struct GuiDefaults {
     static font_t *Font;                                                     //todo constexpr
     static font_t *FontBig;                                                  //todo constexpr
     static const uint32_t MAX_DIALOG_BUTTON_COUNT = 4;                       // maximum number of radio buttons
+    static uint8_t menu_timeout_enabled;
```
RadekVana, line 23: bool?

src/guiapi/include/window.hpp
```diff
@@ -47,22 +51,22 @@ class window_t {
...
-            bool flag_custom7 : 1;                        // 0F - this flag can be defined in parent
+            bool flag_visible : 1;                         // 00 - is visible
+            bool flag_enabled : 1;                         // 01 - is enabled (can be focused)
+            bool flag_invalid : 1;                         // 02 - content is invalid (draw)
+            bool flag_checked : 1;                         // 03 - is checked/selected
+            bool flag_timer : 1;                           // 04 - window has timers
+            is_dialog_t flag_dialog : 1;                   // 05 - window id dialog
+            is_closed_on_click_t flag_close_on_click : 1;  // 06 - window id dialog
+            bool flag_hidden_behind_dialog : 1;            // 07 - there is an dialog over this window
+            is_closed_on_timeout_t flag_timeout_close : 1; // 08 - menu timeout flag
```
RadekVana, line 62: comment it is meant to be used in window_frame_t

src/guiapi/include/GuiDefaults.hpp
RadekVana: constexpr

src/gui/ScreenHandler.cpp
```diff
@@ -164,6 +164,20 @@ void Screens::PushBeforeCurrent(const ScreenFactory::Creator screen_creator) {
 }
 
 void Screens::Loop() {
+    if (GuiDefaults::menu_timeout_enabled) {
+        window_frame_t *w = current.get();
+        if (w->IsTimeout()) {
+            gui_timeout_id = gui_get_menu_timeout_id();
+            if (gui_timer_expired(gui_timeout_id) == 1) {
+                close_all = true;
```
RadekVana, line 172: You cannot close all screens just because last screen can be closed on menu timeout.

src/gui/ScreenHandler.cpp
```diff
@@ -164,6 +164,20 @@ void Screens::PushBeforeCurrent(const ScreenFactory::Creator screen_creator) {
...
+                gui_timer_delete(gui_timeout_id);
```
RadekVana, line 173: I do not like this logic, creating gui_timer in gui.cpp and deleting it here
Also GuiDefaults::menu_timeout_enabled is redundand, because it was already checked in gui.cpp while creating gui_timer

In this case I would not use gui_timer. Would be better readable to handle entire logic inside this module.

RadekVana (changes requested):

src/gui/ScreenHandler.cpp
```diff
@@ -146,6 +146,16 @@ void Screens::CloseAll() {
     close_all = true;
 }
 
+void Screens::CloseSerial() {
+    /// serial close logic:
+    /// when serial printing screen (M876) is open, Screens::SerialClose() is
+    /// called and it will iterate all screens to close those that should be closed
+    while (current.get()->IsSerial() || stack_iterator != stack.begin()) {
```
DRracer, line 153: Are you absolutely sure the current always returns a valid pointer in this case?
The same applies to the stack_iterator and stack.begin()
... not saying it is wrong, just asking...

src/gui/ScreenHandler.cpp
```diff
@@ -163,7 +173,35 @@ void Screens::PushBeforeCurrent(const ScreenFactory::Creator screen_creator) {
     }
 }
 
+void Screens::ResetTimeout() {
+    if (GuiDefaults::menu_timeout_enabled) {
+        if (gui_get_menu_timeout_id() >= 0) {
+            gui_timer_reset(gui_get_menu_timeout_id());
+        } else {
+            gui_timer_create_timeout(current.get(), (uint32_t)MENU_TIMEOUT_MS);
```
DRracer, line 181: again - is current a valid pointer?

src/gui/ScreenHandler.cpp
```diff
@@ -163,7 +173,35 @@ void Screens::PushBeforeCurrent(const ScreenFactory::Creator screen_creator) {
...
+
 void Screens::Loop() {
+    /// menu timeout logic:
+    /// when timeout is expired on current screen,
+    /// we iterate thrue whole stack and close every screen thta should be closed
+    if (current.get()->IsTimeout()) {
+        gui_timeout_id = gui_get_menu_timeout_id();
+        if (gui_timer_expired(gui_timeout_id) == 1) {
+            gui_timer_delete(gui_timeout_id);
+            while (current.get()->IsTimeout() || stack_iterator != stack.begin()) {
```
DRracer, line 194: dtto. - current...

src/gui/ScreenHandler.hpp
```diff
@@ -18,6 +18,9 @@ class Screens {
 
     bool close;
     bool close_all;
+    bool close_serial;
```
DRracer, line 21: consider packing these flags into a single byte

src/gui/guimain.cpp
```diff
@@ -125,23 +125,17 @@ extern "C" void gui_run(void) {
     //Screens::Init(ScreenFactory::Screen<screen_splash_data_t>);
     Screens::Init(screen_initializer, screen_initializer + (sizeof(screen_initializer) / sizeof(screen_initializer[0])));
 
+    //TIMEOUT variable getting value from EEPROM when EEPROM interface is inicialized
+    GuiDefaults::menu_timeout_enabled = (bool)variant_get_ui8(eeprom_get_var(EEVAR_MENU_TIMEOUT));
```
DRracer, line 129: it is better to write `variant_get_ui8(eeprom_get_var(EEVAR_MENU_TIMEOUT)) != 0;` instead of an explicit C-style cast `(bool)`

src/gui/screen_printing_serial.cpp
```diff
@@ -17,7 +17,9 @@ screen_printing_serial_data_t::screen_printing_serial_data_t()
     : IScreenPrinting(string_view_utf8::MakeCPUFLASH((const uint8_t *)caption))
     , octo_icon(this, Rect16((240 - pt_ico().x) / 2, GuiDefaults::RectScreenBody.Top(), pt_ico().x, pt_ico().y), IDR_PNG_serial_printing)
     , last_tick(0)
-    , connection(connection_state_t::disconnected) {
+    , connection(connection_state_t::connected) {
+    IScreenPrinting::ClrMenuTimeoutClose();
```
DRracer, line 21: do you really need to fully qualify the method names?
`IScreenPrinting::something` instead of just `something`?

src/guiapi/src/gui.cpp
```diff
@@ -82,7 +84,7 @@ void gui_loop(void) {
             } else {
                 capturedWin->WindowEvent(capturedWin, WINDOW_EVENT_ENC_DN, (void *)-diff);
             }
-            gui_reset_menu_timer();
+            Screens::Access()->ResetTimeout();
```
DRracer, line 87: does Screen::Access always return a valid pointer?

src/guiapi/src/gui.cpp
```diff
@@ -97,7 +99,7 @@ void gui_loop(void) {
                 capturedWin->WindowEvent(capturedWin, WINDOW_EVENT_BTN_UP, 0);
                 Sound_Play(eSOUND_TYPE_ButtonEcho);
             }
-            gui_reset_menu_timer();
+            Screens::Access()->ResetTimeout();
```
DRracer, line 102: does Screen::Access always return a valid pointer?

src/guiapi/src/gui.cpp
```diff
@@ -124,17 +126,7 @@ void gui_loop(void) {
...
-}
-
-void gui_reset_menu_timer() {
-    if (menu_timeout_enabled) {
-        if (gui_get_menu_timeout_id() >= 0) {
-            gui_timer_reset(gui_get_menu_timeout_id());
-        } else {
-            //gui_timer_create_timeout((uint32_t)MENU_TIMEOUT_MS, (int16_t)-1);
-        }
+        Screens::Access()->ResetTimeout();
```
DRracer, line 129: does Screen::Access always return a valid pointer?

DRracer (dismissed):
please address my comments

src/gui/ScreenHandler.cpp
RadekVana: I think it is safe - now. But if something changes it could cause bsod.
So pls add check.

src/gui/ScreenHandler.cpp
RadekVana: again - pls add check.

src/guiapi/include/window.hpp
```diff
@@ -97,6 +101,8 @@ class window_t {
     bool IsCaptured() const;
     bool HasTimer() const;
     bool IsDialog() const;
+    bool IsTimeout() const;
```
RadekVana, line 104: use better name like `bool ClosesOnTimeout() const;`

src/guiapi/include/window.hpp
```diff
@@ -97,6 +101,8 @@ class window_t {
...
+    bool IsSerial() const;
```
RadekVana, line 105: use better name like `bool ClosesOnSerialPrint() const;`

src/gui/ScreenHandler.cpp
```diff
@@ -163,7 +173,35 @@ void Screens::PushBeforeCurrent(const ScreenFactory::Creator screen_creator) {
     }
 }
 
+void Screens::ResetTimeout() {
+    if (GuiDefaults::menu_timeout_enabled) {
+        if (gui_get_menu_timeout_id() >= 0) {
```
RadekVana, line 178: please double or triple check gui_timers before use
they were broken and I am not sure if they are fixed

in this case I would just check time stamps instead of use complicated gui_timers

src/guiapi/src/gui.cpp
RadekVana: Yes, this is safe.

RadekVana: Screen::Access calls bsod in case of invalid pointer

src/guiapi/src/gui.cpp
RadekVana: yes

src/guiapi/src/gui.cpp
RadekVana: yes

src/guiapi/src/window_msgbox.cpp
```diff
@@ -26,6 +26,7 @@ MsgBoxBase::MsgBoxBase(Rect16 rect, const PhaseResponses *resp, size_t def_btn,
     , text(this, getTextRect(), is_multiline::yes, is_closed_on_click_t::no, txt)
     , buttons(this, get_radio_button_size(rect), resp, labels)
     , result(Response::_none) {
+    window_frame_t::ClrMenuTimeoutClose();
```
RadekVana, line 29: This will not do anything.
Parent screen containing msgbox can still close.

src/marlin_stubs/pause/G27.cpp
```diff
@@ -33,9 +33,8 @@
  */
 void GcodeSuite::G27() {
     // Don't allow nozzle parking without homing first
-    if (axis_unhomed_error())
-        return;
-    nozzle.park(parser.ushortval('P'));
+    if (!axis_unhomed_error())
```
RadekVana, line 36: why did U edit this code?
It does same thing as before.

src/gui/guimain.cpp
```diff
@@ -125,23 +125,17 @@ extern "C" void gui_run(void) {
...
     gui_loop_cb = _gui_loop_cb;
-    //int8_t gui_timeout_id;
     while (1) {
         Screens::Access()->Loop();
         // show warning dialog on safety timer expiration
         if (marlin_event_clr(MARLIN_EVT_SafetyTimerExpired)) {
             MsgBoxInfo(_("Heating disabled due to 30 minutes of inactivity."), Responses_Ok);
         }
         gui_loop();
-        /*if (menu_timeout_enabled) {
```
RadekVana, line 138: pls also remove `screen_close_multiple.cpp` and `screen_close_multiple.h`
since it is rewritten

RadekVana (changes requested):

src/guiapi/include/gui.hpp
RadekVana: You reverted change. Please remove this extern.

src/guiapi/include/GuiDefaults.hpp
Chleba: cannot be constexpr, because it's value will change - it will be moved out from GuiDefaults

src/marlin_stubs/pause/G27.cpp
Chleba: I was just trying to get one bug and did not return it

src/gui/ScreenHandler.cpp
Chleba: They are working just fine. I had a look on them, made a little debug and did not see any potential issue.

src/gui/ScreenHandler.hpp
```diff
@@ -53,8 +59,11 @@ class Screens {
     static void RInit(const ScreenFactory::Creator *begin, const ScreenFactory::Creator *end); // init in reversed order, skips nullptr
 
     static Screens *Access();
+    static bool menu_timeout_enabled;
```
RadekVana, line 62: Make it private

RadekVana: also singleton should not have static member variables

src/gui/ScreenHandler.cpp
```diff
@@ -163,7 +177,35 @@ void Screens::PushBeforeCurrent(const ScreenFactory::Creator screen_creator) {
...
+            gui_timer_create_timeout(current.get(), (uint32_t)MENU_TIMEOUT_MS);
+        }
+    }
+}
+
 void Screens::Loop() {
+    /// menu timeout logic:
+    /// when timeout is expired on current screen,
+    /// we iterate thrue whole stack and close every screen thta should be closed
+    if (current.get()->ClosedOnTimeout()) {
```
RadekVana, line 194: Get() && Get()->ClosedOnTimeout()

src/gui/ScreenHandler.cpp
```diff
@@ -198,9 +240,16 @@ void Screens::Loop() {
                 }
             }
         }
+
+        /// when msgbox(IDialog) is closed on Event and in response is closed
```
RadekVana, line 244: msgbox is not screen, it is dialog - it is not stored here
it calls GuiLoop so events work, but it never gets into ScreenHandler
Also it auto restores capture when destroyed, so your comment is wrong

src/marlin_stubs/pause/G27.cpp
RadekVana: than please revert this, fewer changes == better

RadekVana (changes requested):

src/gui/ScreenHandler.cpp
Chleba: fixed

src/gui/ScreenHandler.hpp
Chleba: replace to having Enable/DisableMenuTimeout. It was just shorter to have Screens::menu_timeout = true/false

RadekVana (approved):

PR #882, opened 2020-09-10
Fanctl2 - state machine, fan start, random phase shifting

XPila:
BFW-918


src/common/fanctl.h
```diff
@@ -52,6 +50,13 @@ typedef struct _fanctl_tach_t {
 
 // class for software pwm control with phase-shifting
 class CFanCtlPWM : private fanctl_pwm_t {
+public:
+    enum PhaseShiftMode {
```
DRracer, line 54: probably `enum PhaseShiftMode : uint8_t {` , there are only few items, so we may save some RAM

src/common/fanctl.h
```diff
@@ -65,9 +70,11 @@ class CFanCtlPWM : private fanctl_pwm_t {
     inline uint8_t get_min_PWM() { return min_value; }
     inline uint8_t get_max_PWM() { return max_value; }
     inline uint8_t get_PWM() { return pwm; }
+    inline PhaseShiftMode get_PhaseShiftMode() { return (PhaseShiftMode)pha_mode; }
```
DRracer, line 73: const

src/common/fanctl.h
```diff
@@ -65,9 +70,11 @@ class CFanCtlPWM : private fanctl_pwm_t {
     inline uint8_t get_min_PWM() { return min_value; }
     inline uint8_t get_max_PWM() { return max_value; }
     inline uint8_t get_PWM() { return pwm; }
```
DRracer, line 72: const ... and the other methods as well...

src/common/fanctl.cpp
```diff
@@ -48,22 +49,31 @@ int8_t CFanCtlPWM::tick() {
...
+                if (pha >= pha_max) {
+                    pha_stp = -pha_stp;
+                    pha = pha_max;
+                } else if (pha < 0) {
+                    pha_stp = -pha_stp;
+                    pha = 0;
+                }
+                break;
+            case random:
+                pha = pha_max * ((float)rand() / RAND_MAX);
```
DRracer, line 74: the float probably won't be too healthy for the 8bit, but is ok here

XPila: you are right, for 8bit we cannot use float, i will try to find common solution with fixpoint aritmetic or somthing similar

src/common/fanctl.h
XPila: fixed. internaly it is stored in pha_mode (uint8_t).

src/common/fanctl.h
XPila: fixed

src/common/fanctl.h
XPila: fixed

src/common/fanctl.h
DRracer: almost, but ok...

RadekVana (approved):

PR #891, opened 2020-09-18
Integrate .po files compilation into the build process

dragomirecky:
This PR
1. introduces a new `utils/lang.py` utility, which 
  a) is able to dump .po files to .txt files (with easier-to-parse format suitable for tests), and
  b) is able to generate a hash table and other structures required by the „CPUFLASHTranslationProvider“ and dump them to C++ files.
2. Integrates the `utils/lang.py` utility with our build system, making the compilation of .po files automatic and part of the build process.

There is one missing piece before it can be merged – making necessary changes to the translations' tests suite (@DRracer).

BFW-1160


Chleba (dismissed):

dragomirecky:
I just rebased the history (to linearize it and solve conflicts with the master branch).

@DRracer changes in lang.py look good to me 👍

utils/lang.py
```diff
@@ -187,7 +187,7 @@ def cmd_generate_hash_tables(args):
     if not translations:
         logger.error('no translations found')
         return 1
-    buckets_count = 256
```
dragomirecky: Actually, do we want this?

This is the number of buckets we start checking for no collisions. Having it lower shouldn't break anything, it will just take a bit longer to generate the tables. 

So, in other words, are we sure that a non-colliding table will never have less than 340 buckets?

DRracer: Actually, I was thinking about the same - the less buckets we achieve the better. Right now the optimal number of buckets is 266, which is way less than 340 (which was chosen as a rule of thumb).

dragomirecky: And I didn't notice that your last commit changed it back to 256 :)

Nevertheless, based on our discussion on Slack, I changed it to 128 (commit c7ffcf3).

DRracer (approved):
Approving again based upon approval from @Chleba

PR #897, opened 2020-09-23
Event debug

RadekVana:
BFW-1287

### DIFF
--- a/src/guiapi/src/window.cpp
+++ b/src/guiapi/src/window.cpp
@@ -8,6 +8,90 @@
 
 extern osThreadId displayTaskHandle;
 
+constexpr bool GUI_event_IsKnob(GUI_event_t event) {
+    switch (event) {
+    case GUI_event_t::BTN_DN:
+    case GUI_event_t::BTN_UP:
+    case GUI_event_t::ENC_DN:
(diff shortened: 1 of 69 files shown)


src/guiapi/src/window.cpp
```diff
@@ -4,9 +4,95 @@
...
+}
+
+constexpr bool GUI_event_IsAnyButLoop(GUI_event_t event) {
+    return event != GUI_event_t::LOOP;
+}
+
+constexpr const char *GUI_event_prt(GUI_event_t event) {
+    switch (event) {
+    case GUI_event_t::BTN_DN:
+        return ("button down");
```
DRracer, line 46: Are you really returning an address of a local variable?
Technically, this may work - the string must be somewhere in CPU FLASH and the compiler might be smart enough to not-to load it into RAM and just return the address in FLASH, but the whole function is highly suspicios.

DRracer (changes requested):
Please improve GUI_event_prt

src/guiapi/src/window.cpp
RadekVana: cannot do it in constexpr function

DRracer (approved):

PR #901, opened 2020-09-24
Optimize string_view_utf8

DRracer:
This needs a bit of explanation - originally the string_view_utf8 was to be
a dynamically created container supporting various sources of data (text).
During the development it turned out we don't need that much of variability,
so we can skip the pointers to *_getbyte() and *_rewind() - these are purely
set up just once when the instance of string_view_utf8 is created and are not
changed ever after.

That said we can shave off 2 pointers inside the string_view_utf8 thus making it
8B shorter which results not only in savings of stack/RAM, but also ~2.5KB of code.

Performance-wise the new code is at least as efficient as the previous one -
string_view_utf8 still has to decide how to extract a byte at runtime.

The only downside of this optimization is probably a less nice code than before.
However the benefits prevail IMO.

### DIFF
--- a/src/lang/string_view_utf8.hpp
+++ b/src/lang/string_view_utf8.hpp
@@ -41,15 +41,6 @@ class string_view_utf8 {
     };
     Attrs attrs;
 
-    /// Extracts one byte from source media and advances internal read ptr, implementation defined in derived classes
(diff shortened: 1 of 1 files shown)


dragomirecky (approved):
Awesome! Looks good to me 👍

marekmosna-prusa3d (approved):

PR #904, opened 2020-09-25
C++: Add function pointer enhancement

marekmosna-prusa3d:
Please check added unit tests.


marekmosna-prusa3d:
increased binary code too much. It's not worth in such a cost.

PR #909, opened 2020-09-28
Bfw 1271

RadekVana:


src/marlin_stubs/G26.hpp
```diff
@@ -0,0 +1,165 @@
+// G26.hpp
+
+#pragma once
+
+/// Path of Manhattan snake
+/// Alternate X and Y coordinates.
+/// It cannot print diagonals etc.
+/// First X and Y are a starting point.
+static const constexpr float snake1[] = {
```
DRracer, line 9: this data structure is very hard to read and verify (even though optimal for the CPU)
ideally if you could at least distinguish x and y ... something like `float [][2] = { { x, y }, {x, y} ...`

src/marlin_stubs/G26.cpp
```diff
@@ -0,0 +1,173 @@
...
+            const float x = path[position - 1];
+            const float y = path[position];
+            return extrusion(x, y, x, last, layerHeight, threadWidth);
+        }
+    }
+
+    void print_snake(const float *snake, const size_t snake_size, const float speed) {
+
+        /// move to start
+        go_to_destination(snake[0], snake[1], 0); // Process X Y Z E F parameters
```
DRracer, line 105: this assumes snake has at least 2 items ... a static_assert or runtime check would be appropriate to add

DRracer (approved):
Besides the comments not directly related to this PR, I approve this PR

PR #911, opened 2020-09-29
Integer variable calculation correction when oscilating between 0-1

zoltan-l:
changing the calculation formula order the rounding error of the integer oscilating between 0-1 has been corrected


DRracer:
Fix code formatting

XPila (approved):

PR #912, opened 2020-09-29
GUI: Fix text drawing BFW-1290

espr14:


src/guiapi/src/display_helper.cpp
```diff
@@ -36,19 +44,21 @@ size_ui16_t render_text(Rect16 rc, string_view_utf8 str, const font_t *pf, color
     const uint16_t h = pf->h; //char height
     // prepare for stream processing
     unichar c = 0;
+    /// TODO define parent class for both below and use parent.character(str) instead (few lines below)
```
DRracer, line 47: _define parent class for both..._
That will only improve readability of the source code, it won't bring any simplification of the compiled code (the CPU must still decide at runtime which source to use, so there will be some conditional JMP). The only solution to this is to unroll the loop (e.g. template it) but at the cost of code size ...

src/guiapi/include/display_helper.h
```diff
@@ -38,7 +38,7 @@ struct txtroll_t {
         , px_cd(0) {}
 };
 
-extern size_ui16_t render_text(Rect16 rc, string_view_utf8 str, const font_t *pf, color_t clr_bg, color_t clr_fg, uint16_t flags);
+extern size_ui16_t render_text(Rect16 rc, string_view_utf8 str, const font_t *pf, color_t clr_bg, color_t clr_fg, uint16_t flags = 0);
```
DRracer, line 41: is the `flags=0` really necessary? The default value hasn't been here before

DRracer (approved):

PR #925, opened 2020-10-01
variant8_t C++ implementation

marekmosna-prusa3d:
BFW-1253

### DIFF
--- a/src/common/variant8.cpp
+++ b/src/common/variant8.cpp
@@ -54,23 +54,26 @@ struct t_variant8_t {
         pointer_t = 3,
         user_t = 4,
     };
-    template <class T>
-    t_variant8_t(T i, typename std::enable_if<std::is_integral<T>::value>::type = 0)
+    template <class T,
+        typename std::enable_if<std::is_integral<T>::value>::type * = nullptr>
+    t_variant8_t(T i)
         : storage_(
             (static_cast<uint64_t>(type_t::int_t) << type_shift) | (i << integral_shift))
         , packed_(false) {};
 
-    template <class T>
-    t_variant8_t(T f, typename std::enable_if<std::is_floating_point<T>::value>::type = 0)
-        : storage_(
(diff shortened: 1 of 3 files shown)


mkbel:
I am too lazy to read the code. Can you write some summary (or philosophy behind) of what is changed and why? Also I am big opponent of having dynamic memory allocation in time critical embedded software. Is this aspect of variant8_t somehow improved? (e.g. not allocating at least small types on heap)

marekmosna-prusa3d:
> I am too lazy to read the code. Can you write some summary (or philosophy behind) of what is changed and why? Also I am big opponent of having dynamic memory allocation in time critical embedded software. Is this aspect of variant8_t somehow improved? (e.g. not allocating at least small types on heap)

OK, I'll try. 
Short versions of the answers are:
* it changes entire C implementation
* I got the task to make C++ version
* up to 7 characters string  can be passed by the variant without any allocation as an user type

Philosophy is not such a straight. Basic visions are:
* as less code as possible => easy to maintain
* strongly typed implementation
* focused on minimize reinterpret casts

I hope it helps as a brief introduction

espr14 (approved):

DRracer (commented):
So far the biggest problem with variant (C-style) was the fact, that variants holding pointers didn't carry any information whether the memory pointed to is owned by the variant or not.
From that arose serious issues like duplicit deallocation attempts or memory leaking.
I don't see any improvement in this solution, even though I like the enable_if construct.

espr14:
> So far the biggest problem with variant (C-style) was the fact, that variants holding pointers didn't carry any information whether the memory pointed to is owned by the variant or not.
> From that arose serious issues like duplicit deallocation attempts or memory leaking.
> I don't see any improvement in this solution, even though I like the enable_if construct.

Duplicit deallocation can be avoided with these steps:
- set pointer to nullptr after deallocation
- avoid deallocating of nullptr
- thread safe deallocation

To define ownership of the data you can add a single bit flag.

PR #927, opened 2020-10-01
selftest3 - heater test

XPila:


src/common/selftest/selftest_MINI.cpp
```diff
@@ -234,10 +247,30 @@ void CSelftest::log_open() {
         serial_otp = 0;
     }
     char serial[32] = "unknown";
-    char fname[32] = "test_unknown.txt";
+    const char *suffix = "";
+    switch (m_Mask) {
+    case stmFans:
+        suffix = "_fans";
```
DRracer, line 253: I recommend using `static const char fans[] = "_fans"; suffix = fans;`
This makes sure the compiler won't try to load the "_fans" onto stack first, but uses the address into CPU FLASH

src/guiapi/include/MItem_tools.hpp
```diff
@@ -90,7 +90,15 @@ class MI_TEST_HEAT : public WI_LABEL_t {
 
 public:
     MI_TEST_HEAT();
+protected:
+    virtual void click(IWindowMenu &window_menu) override;
+};
 
+class MI_TEST_FANS_fine : public WI_LABEL_t {
+    static constexpr const char *const label = N_("Test FANs fine");
```
DRracer, line 98: the EN wording of this message is really strange, what is the intention? Does it mean, the test of the fans went ok?

DRracer (changes requested):

src/common/selftest/selftest_MINI.cpp
```diff
@@ -197,26 +203,33 @@ bool CSelftest::phaseAxis(const selftest_axis_config_t *pconfig_axis, CSelftestP
...
     m_pHeater_Bed = m_pHeater_Bed ? m_pHeater_Bed : new CSelftestPart_Heater(&Config_HeaterBed);
     m_pHeater_Nozzle->Loop();
     m_pHeater_Bed->Loop();
-    if (m_pHeater_Nozzle->IsInProgress() || m_pFan1->IsInProgress()) {
-        //int p0 = m_pHeater_Nozzle->GetProgress();
-        //int p1 = m_pHeater_Bed->GetProgress();
+    if (m_pHeater_Nozzle->IsInProgress() || m_pHeater_Bed->IsInProgress()) {
+        int p0 = m_pHeater_Nozzle->GetProgress();
+        int p1 = m_pHeater_Bed->GetProgress();
+        int p = _PF * ((p1 > p0) ? p0 : p1);
```
RadekVana, line 214: std::min

src/common/selftest/selftest_MINI.cpp
XPila: i do not see any benefits, all the strings constants are placed in the flash, nothing loaded onto stack.

src/guiapi/include/MItem_tools.hpp
XPila: this is fan test with 24 measured points (normal test has 5 points). This is only for Zoltan and we will remove this menuitem in RC.

DRracer (approved):

src/common/selftest/selftest_MINI.cpp
```diff
@@ -234,10 +247,30 @@ void CSelftest::log_open() {
...
+        break;
+    default:
+        break;
+    }
+    char fname[64] = "test_unknown.txt";
+    sprintf(fname, "test_unknown%s.txt", suffix);
     if (serial_otp) {
         sprintf(serial, "CZPX%.15s", serial_otp);
-        sprintf(fname, "test_CZPX%.15s.txt", serial_otp);
+        sprintf(fname, "test_CZPX%.15s%s.txt", serial_otp, suffix);
```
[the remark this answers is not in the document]
XPila: fixed

src/common/selftest/selftest_MINI.cpp
XPila: fixed

RadekVana (approved):

PR #934, opened 2020-10-02
GUI: Fix move over the hidden menu item

marekmosna-prusa3d:
BFW-1329

### DIFF
--- a/src/guiapi/src/window_menu.cpp
+++ b/src/guiapi/src/window_menu.cpp
@@ -140,22 +140,14 @@ void window_menu_t::windowEvent(EventLock /*has private ctor*/, window_t *sender
         if (item->IsSelected()) {
             invalid |= item->Decrement(value);
         } else {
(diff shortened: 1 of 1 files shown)


src/guiapi/src/window_menu.cpp
```diff
@@ -90,17 +90,24 @@ void window_menu_t::Increment(int dif) {
...
+                Sound_Play(eSOUND_TYPE::BlindAlert); // play sound at first or last index of menu
+                return 0;
+            }
+            if ((index + dif) >= GetCount()) {
+                Sound_Play(eSOUND_TYPE::BlindAlert); // play sound at first or last index of menu
+                return GetCount() - 1;
+            }
+            for (index = index + dif;
+                 GetItem(index) && GetItem(index)->IsHidden();
+                 index += dif)
```
RadekVana, line 105: i think it should be `index += 1`

RadekVana (approved):

PR #943, opened 2020-10-06
Jogwheel events

RadekVana:
BFW-1326 BFW-1327 BFW-1328

### DIFF
--- a/src/common/circle_buffer.hpp
+++ b/src/common/circle_buffer.hpp
@@ -2,7 +2,7 @@
 #pragma once
 
 #include <stdint.h>
-
+#include <stddef.h>
 /*****************************************************************************/
 // general circular buffer
 // you can never use entire size
--- a/src/common/circle_buffer.hpp
+++ b/src/common/circle_buffer.hpp
@@ -9,7 +9,7 @@ static constexpr const char *CircleBufferEmpty = "";
 // because write position (end) cannot be equal to begin
 // because begin == end == empty
 template <size_t SIZE, size_t MAX_LENGTH>
-class CircleBuffer {
+class CircleStringBuffer {
     char msg_data[SIZE][MAX_LENGTH];
     size_t begin; // position of first element
     size_t end;   // position behind last element == write position
@@ -18,13 +18,13 @@ class CircleBuffer {
     static void decrementIndex(size_t &index) { index = (index + SIZE - 1) % SIZE; }
 
 public:
-    CircleBuffer()
+    CircleStringBuffer()
         : begin(0)
         , end(0)
         , pushed(0) {}
 
-    void push_back(const char *const popup_msg);
-    void push_back_DontRewrite(const char *const popup_msg);
+    void push_back(const char *const msg);
+    void push_back_DontRewrite(const char *const msg);
     size_t Count() const { return (end + SIZE - begin) % SIZE; }
     bool IsEmpty() const { return begin == end; }
     size_t PushedCount() const { return pushed; }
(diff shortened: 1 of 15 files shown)


src/common/circle_buffer.hpp
```diff
@@ -3,14 +3,15 @@
 
 #include <stdint.h>
 
-static constexpr const char *CircleBufferEmpty = "";
-
+/*****************************************************************************/
+// general circle buffer
```
DRracer, line 7: circular

src/guiapi/include/Jogwheel.hpp
```diff
@@ -103,25 +108,26 @@ class Jogwheel {
...
+    void ChangeStateTo(BtnState_t new_state);
+
+    // variables are set in interrupt, volatile variables are read in GUI
+    // ordered by size, from biggest to smallest (most size-effective)
+    CircleBuffer<BtnState_t, 32> btn_events; //!< event buffer
+    uint32_t speed_traps[4];                 //!< stores previous encoder's change timestamp
+    uint32_t spin_speed_counter;             //!< counting variable for encoder_gear system
+    volatile int32_t encoder;                //!< jogwheel encoder
+    uint16_t hold_counter;                   //!< keep track of ms from button down
+    BtnState_t btn_state;                    //!< actual state of button, size uint8_t
```
DRracer, line 124: actual != current

src/common/circle_buffer.hpp
RadekVana: fixed

src/guiapi/include/Jogwheel.hpp
RadekVana: fixed

src/guiapi/src/Jogwheel.cpp
```diff
@@ -60,9 +61,15 @@ void Jogwheel::ReadInput(uint8_t &signals) {
     }
 }
 
-const Jogwheel::ButtonAction Jogwheel::GetButtonAction() {
-    ButtonAction ret = btn_action;
-    btn_action = ButtonAction::BTN_NO_ACTION;
+volatile bool Jogwheel::ConsumeButtonEvent(Jogwheel::BtnState_t &ev) {
+    static uint32_t last_read;          //cannot read it too often
+    const uint32_t min_read_dellay = 8; //ms
+    if ((last_read - HAL_GetTick()) < min_read_dellay) {
```
michalrudolf, line 67: I don't see where is "last_read" assigned to some value. Here on row:65 and on row:117, it seems to me like this variable is uninitialized. Am I wrong?

src/guiapi/src/gui.cpp
```diff
@@ -67,40 +67,45 @@ void gui_loop(void) {
...
 
     #ifdef GUI_JOGWHEEL_SUPPORT
-    Jogwheel::ButtonAction btn = jogwheel.GetButtonAction();
-    bool encoder_changed = jogwheel.EncoderChanged();
-    if (btn == Jogwheel::ButtonAction::BTN_PUSHED) {
+    Jogwheel::BtnState_t btn_ev;
+    volatile bool is_btn = jogwheel.ConsumeButtonEvent(btn_ev);
+    volatile int32_t encoder_diff = jogwheel.GetEncoderDiff();
+
+    // I do not know why this is separate. To play sound asap?
```
michalrudolf, line 74: It was separate because it was developed separately and BTN_PUSHED was a special event, but I think you can add it now to scope ant row:97...

michalrudolf (approved):
Looks clean.

DRracer (changes requested):
What do you need disable_IRQ for?
This is very suspicious and is potentially very dangerous. Use this only as a last resort when all other sychronization attempts fail. I'm not convinced this is the case.
Prefer using std::atomic

src/guiapi/src/Jogwheel.cpp
RadekVana: fixed

src/guiapi/src/gui.cpp
RadekVana: fixed

XPila (approved):

DRracer (approved):
This still needs a lot of cleanup, it looks unnecessarily too complex. Approving only not to block future work.

DRracer (approved):

PR #956, opened 2020-10-12
GUI: Refactor window_menu BFW-1355

espr14:


michalrudolf (commented):
Looks good to me:]

src/guiapi/src/window_menu.cpp
```diff
@@ -77,47 +80,111 @@ IWindowMenuItem *window_menu_t::GetActiveItem() {
     return GetItem(index);
 }
 
+bool window_menu_t::moveToNextVisibleItem(int steps) {
```
RadekVana, line 83: My personal opinion is: 3 methods would be better readable
`bool window_menu_t::moveToNextVisibleItem() ` move +1 step
`bool window_menu_t::moveToPrevVisibleItem() ` move -1 step
`bool window_menu_t::moveToVisibleItem(int steps) ` call Next and Prev methods inside this one

RadekVana (approved):
Drawing inside events is wrong
Approved only because we need a fix asap

PR #960, opened 2020-10-13
Allow easier switching of supported/compiled languages

DRracer:
This is necessary for debugging purposes.

### DIFF
--- a/src/lang/translator.cpp
+++ b/src/lang/translator.cpp
@@ -34,6 +34,9 @@ bool Translations::RegisterProvider(uint16_t langCode, const ITranslationProvide
 }
 
 bool Translations::ChangeLanguage(uint16_t langCode) {
+    if (langCode == 0) {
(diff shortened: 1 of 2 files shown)


src/lang/translator.cpp
```diff
@@ -34,6 +34,9 @@ bool Translations::RegisterProvider(uint16_t langCode, const ITranslationProvide
 }
 
 bool Translations::ChangeLanguage(uint16_t langCode) {
+    if (langCode == 0) {
+        return false; // requesting an invalid langCode to change to
+    }
```
dragomirecky, line 39: /question/ Why is this case handled here explicitly? If zero is not a valid code, no language should have it assigned and we should return `false` from line 46. Or am I missing something?

dragomirecky (approved):

michalrudolf (commented):
It fixes displayTask bug. Test successful.

RadekVana (approved):

PR #966, opened 2020-10-14
Allow X and Y enable to share the same pin.

mkbel:


DRracer (changes requested):
clang-formatting check failed

DRracer (approved):

PR #967, opened 2020-10-15
Jogwheel simplify

RadekVana:
BFW-1367

### DIFF
--- a/src/guiapi/include/Jogwheel.hpp
+++ b/src/guiapi/include/Jogwheel.hpp
@@ -9,7 +9,6 @@
 #pragma once
 
 #include <inttypes.h>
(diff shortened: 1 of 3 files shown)


src/guiapi/include/Jogwheel.hpp
```diff
@@ -38,6 +38,19 @@ class Jogwheel {
         Held
     };
 
+    //structure to be read in rtos thread (outside interrupt)
+    //size must be 32 bit to be atomic
```
DRracer, line 42: I'd recommend adding static_assert(sizeof(encoder_t)==4) somewhere when you rely on the exact size

DRracer (approved):

PR #972, opened 2020-10-16
SYSTEM: Fix eeprom padding

marekmosna-prusa3d:
BFW-1379


DRracer (approved):
The fact, that the eeprom cells must be allocated even if the feature may not be in use, reveals severe design/concept flaws which must be improved in the future.
This is sadly just a hacky workaround
Approving ...

PR #973, opened 2020-10-16
GUI: Fix window_menu redraw BFW-1366

espr14:


src/guiapi/src/window_menu.cpp
```diff
@@ -80,33 +83,37 @@ IWindowMenuItem *window_menu_t::GetActiveItem() {
...
+
             item = GetItem(index + moved);
-            if (item == nullptr)
+            if (item == nullptr) {
+                moveIndex = 0;
                 return false;
+            }
         } while (item->IsHidden());
-        SetIndex(index + moved); /// sets new cursor position (visible item)
+        SetIndex(uint8_t(index + moved)); /// sets new cursor position to a visible item
```
RadekVana, line 110: what about boundaries use `std::clamp` or `std::min` and `std::max` to check?

RadekVana (approved):

src/guiapi/src/window_menu.cpp
espr14: There's a check on line 98.

PR #979, opened 2020-10-19
Sound update

Chleba:
- added DEBUG Sound mode for BSOD debuggind purpouses
- updated CriticalAlert sound type as repeating sound
- added fixed volume for ERROR sounds

BFW-1387, BFW-1392

### DIFF
--- a/src/common/sound.cpp
+++ b/src/common/sound.cpp
@@ -18,6 +18,8 @@ const float Sound::volumes[eSOUND_TYPE::count] = {
     Sound::volumeInit, Sound::volumeInit, Sound::volumeInit, Sound::volumeInit,
     0.175F, 0.175F, Sound::volumeInit, Sound::volumeInit, Sound::volumeInit
 };
+/// forced types of sounds - mainly for ERROR sounds. Ignores volume settings.
+const bool Sound::forced[eSOUND_TYPE::count] = { false, false, false, true, false, false, false, false, false };
 
 /// array of usable types (eSOUND_TYPE) of every sound modes (eSOUND_MODE)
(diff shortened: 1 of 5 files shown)


src/common/sound.cpp
```diff
@@ -174,12 +174,16 @@ void Sound::_sound(int rep, float frq, int16_t dur, int16_t del, float vol /*, b
...
     }
 
-    /// store variables for timing method
+    /// store ACTIVE variables for timing method
     repeat = rep;
     frequency = frq;
     duration_set = dur;
     delay_set = del;
     volume = (vol * varVolume) * 0.3F;
+    /// for debugging BSOD
```
espr14, line 183: #ifdef _DEBUG

src/guiapi/include/MItem_tools.hpp
```diff
@@ -343,13 +343,19 @@ class MI_TIMEOUT : public WI_SWITCH_OFF_ON_t {
...
+#else
 class MI_SOUND_MODE : public WI_SWITCH_t<4> {
+#endif
     constexpr static const char *const label = N_("Sound Mode");
 
     constexpr static const char *str_Once = N_("Once");
     constexpr static const char *str_Loud = N_("Loud");
     constexpr static const char *str_Silent = N_("Silent");
     constexpr static const char *str_Assist = N_("Assist");
+    constexpr static const char *str_Debug = N_("Debug");
```
espr14, line 357:
```suggestion
    constexpr static const char *str_Debug = "Debug";
```

src/common/sound.cpp
Chleba: No need to put it in macro. Debug mode is added in sound_enums.cpp - so only macro that is staying is in MItem_tools.cpp.

DRracer (approved):

PR #1004, opened 2020-10-26
Fix not showing version info text when screen is up

Chleba:
BFW-1415

### DIFF
--- a/src/gui/screen_menu_version_info.cpp
+++ b/src/gui/screen_menu_version_info.cpp
@@ -30,6 +30,8 @@ public:
     std::array<char, VERSION_INFO_STR_MAXLEN> version_info_str;
     constexpr static const char *label = N_("VERSION INFO");
     ScreenMenuVersionInfo();
+protected:
+    virtual void windowEvent(EventLock /*has private ctor*/, window_t *sender, GUI_event_t event, void *param) override;
 };
 
 ScreenMenuVersionInfo::ScreenMenuVersionInfo()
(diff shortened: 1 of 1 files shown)


src/gui/screen_menu_version_info.cpp
```diff
@@ -91,8 +93,18 @@ ScreenMenuVersionInfo::ScreenMenuVersionInfo()
 
     // this MakeRAM is safe - version_info_str is allocated in RAM for the lifetime of ths
     help.SetText(string_view_utf8::MakeRAM((const uint8_t *)version_info_str.data()));
+    help.Invalidate();
+		gui_invalidate();
```
DRracer, line 97: indentation!

DRracer (changes requested):
clang-format was not run on these changes

src/gui/screen_menu_version_info.cpp
```diff
@@ -91,8 +93,18 @@ ScreenMenuVersionInfo::ScreenMenuVersionInfo()
...
 }
 
 ScreenFactory::UniquePtr GetScreenMenuVersionInfo() {
     return ScreenFactory::Screen<ScreenMenuVersionInfo>();
 }
+
+void ScreenMenuVersionInfo::windowEvent(EventLock /*has private ctor*/, window_t *sender, GUI_event_t event, void *param) {
+    help.Invalidate();
```
RadekVana, line 105: Do not call Invalidate in every event, it will redraw like 100x / sec

src/gui/screen_menu_version_info.cpp
RadekVana: `Invalidate();` cointains `gui_invalidate();`

RadekVana (changes requested):

Chleba:
This "quickfix" is irrelevant and not a good solution in general. @espr14 already have a good solution.

PR #1010, opened 2020-10-30
Fix python open files encoding

marekmosna-prusa3d:
BFW-1432

### DIFF
--- a/utils/lang.py
+++ b/utils/lang.py
@@ -140,19 +140,19 @@ def format_c_array(items, indent=4):
 
 def dump_hpp_array(path: Path, decl, items):
     """Write down a file with array definition."""
-    with open(path, 'w') as f:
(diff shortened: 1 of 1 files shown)


DRracer (changes requested):
Enforcing utf-8 encoding on all files produced by the lang.py script may look like a good idea, however a problem apears on the MS Windows platform, which still (almost 40 years) prefers a stone-age like solution of user/system codepages.
Therefore this commit should be followed by another one checking, that all the source code files are utf-8 encoded.

Chleba:
First we should have encoding for our Python script in header of our file:
```
#!/usr/bin/env python
# -*- coding: utf-8 -*-
```
And another commit will come to ensure that all files will be encoded in UTF-8 ... by the dark side of the Force

PR #1014, opened 2020-10-30
fixed bsod after restart

XPila:
1. initiaize w25x before reading flags
2. display error screen only for valid dump


DRracer (commented):
w25x_init() shall probably end up at one spot in appmain or similar, there are already several situations when this is being called. So far it looks ok to call w25x_init() multiple time.

DRracer (approved):

PR #1021, opened 2020-11-02
Seltest suffix

zoltan-l:
BFW-1428

### DIFF
--- a/src/common/selftest/selftest_MINI.cpp
+++ b/src/common/selftest/selftest_MINI.cpp
@@ -281,11 +281,11 @@ void CSelftest::log_open() {
     const char *suffix = "_no_suffix";
     if (m_Mask & stmFans)
         suffix = "_fans";
-    if (m_Mask & stmXYAxis)
+    else if (m_Mask & stmXYAxis)
         suffix = "_xyz";
-    if (m_Mask & stmXYZAxis)
+    else if (m_Mask & stmXYZAxis)
         suffix = "_xyz";
-    if (m_Mask & stmHeaters)
+    else if (m_Mask & stmHeaters)
         suffix = "_heaters";
 
     char fname[64];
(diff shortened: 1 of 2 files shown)


src/common/selftest/selftest_MINI.cpp
```diff
@@ -278,25 +278,16 @@ void CSelftest::log_open() {
         serial_otp = 0;
     }
     char serial[32] = "unknown";
-    const char *suffix = "";
-    switch (m_Mask) {
-    case stmFans:
+    const char *suffix = "_no_suffix";
```
DRracer, line 281: why is this initialization needed?

src/common/selftest/selftest_MINI.cpp
```diff
@@ -278,25 +278,16 @@ void CSelftest::log_open() {
...
+    if (m_Mask & stmFans)
         suffix = "_fans";
-        break;
-    case stmXYAxis:
-    case stmHome_XYAxis:
+    if (m_Mask & stmXYAxis)
```
DRracer, line 284: can we have multiple suffices? This code looks like we could (no "else if"), but the suffix pointer is always changed hard

DRracer (changes requested):
Please address my comments, I'm not sure about the intent and side effects.
Also, a PR consisting of Fix, format, format doesn't look good, please improve the commits by rebasing interactively

src/common/selftest/selftest_MINI.cpp
zoltan-l: for identification and testing purposes- when the further conditions do not meet, empty char does not provide any suffix

src/common/selftest/selftest_MINI.cpp
zoltan-l: else if will be used

src/common/selftest/selftest_MINI.cpp
DRracer: why do you need any suffix at all in such case?

DRracer (approved):

PR #1028, opened 2020-11-03
New GUI menu implementation

michalrudolf:
BFW-1279 (followed designs are in the task)
implementation is hidden, MINI GUI is not changed apart from:
 - txtroll bug fixed (solution: reinit txtroll when clicked)


src/guiapi/include/IWindowMenuItem.hpp
```diff
@@ -28,23 +29,26 @@ class IWindowMenuItem {
 protected:
     virtual void printIcon(IWindowMenu &window_menu, Rect16 rect, uint8_t swap, color_t color_back) const;
     void printLabel_into_rect(Rect16 rolling_rect, color_t color_text, color_t color_back, const font_t *font, padding_ui8_t padding, uint8_t alignment) const;
-    virtual void printText(IWindowMenu &window_menu, Rect16 rect, color_t color_text, color_t color_back, uint8_t swap) const;
+    virtual void printItem(IWindowMenu &window_menu, Rect16 rect, color_t color_text, color_t color_back, uint8_t swap) const;
     virtual void click(IWindowMenu &window_menu) = 0;
     virtual Rect16 getRollingRect(IWindowMenu &window_menu, Rect16 rect) const;
+    virtual std::array<Rect16, 2> getMenuRects(IWindowMenu &window_menu, Rect16 rect) const = 0;
```
DRracer, line 35: using std::pair or a tuple is more common in this case ... but never mind, the resulting code will be similar

src/guiapi/include/WindowMenuItems.hpp
```diff
@@ -42,14 +44,65 @@ class WI_SPIN_t : public AddSuper<IWiSpin> {
...
+            window_menu.padding,
+#else  // PRINTER TYPE != PRUSA_PRINTER_MINI
+            window_menu.secondary_font,
+            color_back, IsSelected() ? COLOR_ORANGE : color_text,
+            padding_ui8(0, 6, 0, 0),
+#endif // PRINTER TYPE == PRUSA_PRINTER_MINI
+            window_menu.GetAlignment());
+
+#if (PRINTER_TYPE != PRINTER_PRUSA_MINI)
+        uint8_t unit_padding_left = units[0] == '\177' ? 0 : 6;                                                                // °C are not separated with halfspace
```
DRracer, line 92: when playing with units, start thinking about internationalization as well (Celsius vs. Fahrenheit)

src/guiapi/include/WindowMenuItems.hpp
```diff
@@ -42,14 +44,65 @@ class WI_SPIN_t : public AddSuper<IWiSpin> {
...
+        render_text_align(rects[2], string_view_utf8::MakeRAM((const uint8_t *)units), window_menu.secondary_font, color_back, // is this MakeRAM safe?
```
DRracer, line 93: MakeRAM will be safe as long as the "units" scope is valid - you must make sure "units" is valid whenever this statement is called

src/guiapi/include/WindowMenuItems.hpp
```diff
@@ -146,12 +196,65 @@ class WI_SWITCH_t : public AddSuper<IWiSwitch> {
...
+    **/
+    std::array<Rect16, 4> getSwitchMenuRects(IWindowMenu &window_menu, Rect16 rect) const {
+        Rect16 base_rect = getRollingRect(window_menu, rect);
+        Rect16 switch_rect, bracket_start, bracket_end;
+
+        bracket_end = { int16_t(base_rect.Left() + base_rect.Width() - window_menu.secondary_font->w), base_rect.Top(), uint16_t(window_menu.secondary_font->w), base_rect.Height() };
+        base_rect -= bracket_end.Width();
+
+        string_view_utf8 localizedItem = _(get_item());
+        switch_rect = getCustomRect(window_menu, base_rect, localizedItem.computeNumUtf8CharsAndRewind() * window_menu.secondary_font->w);
```
DRracer, line 221: beware: computeNumUtf8CharsAndRewind will always run through the whole translated string to compute the number of glyphs - aka O(n)

src/guiapi/src/IWindowMenuItem.cpp
```diff
@@ -86,3 +83,17 @@ Rect16 IWindowMenuItem::getRollingRect(IWindowMenu &window_menu, Rect16 rect) co
     rect -= irc.Width();
     return rect;
 }
+
+// Returns custom width Rectangle, aligned intersection on the right of the base_rect
+Rect16 IWindowMenuItem::getCustomRect(IWindowMenu &window_menu, Rect16 base_rect, uint16_t custom_rect_width) {
+    Rect16 custom_rect = { base_rect.Left(), base_rect.Top(), custom_rect_width, base_rect.Height() };
+    custom_rect += Rect16::Left_t(base_rect.Width() - custom_rect.Width());
```
DRracer, line 90: this could be possibly rewritten into one expression, which will improve the chance of optimizing the whole function

src/guiapi/src/MItem_print.cpp
```diff
@@ -5,8 +5,8 @@
 /*****************************************************************************/
 //MI_NOZZLE
 MI_NOZZLE::MI_NOZZLE()
-    : WI_SPIN_U16_t(uint16_t(marlin_vars()->target_nozzle),
-        MenuVars::nozzle_range.data(), _(label), 0, true, false) {}
+    : WI_SPIN_U16_t(uint16_t(marlin_vars()->target_nozzle), "\177C",
```
DRracer, line 8: Declare a set of units and reuse them! I don't like the repeated strings here, do not rely on potential merging of duplicit strings

src/guiapi/src/window_menu.cpp
```diff
@@ -271,7 +275,7 @@ void window_menu_t::unconditionalDraw() {
 
 void window_menu_t::redrawWholeMenu() {
     const int item_height = font->h + padding.top + padding.bottom;
-    const size_t visible_available = rect.Height() / item_height;
+    const size_t visible_available = rect.Height() / (item_height + 1);
```
DRracer, line 278: what is this "+1"?

src/guiapi/include/Iwindow_menu.hpp
```diff
@@ -7,6 +7,7 @@ struct IWindowMenu : public window_aligned_t {
     color_t color_text;
     color_t color_disabled;
     font_t *font;
+    font_t *secondary_font;
```
DRracer, line 10: this is more like a "unit_font" than a "secondary_font"

src/guiapi/include/WindowMenuItems.hpp
```diff
@@ -42,14 +44,65 @@ class WI_SPIN_t : public AddSuper<IWiSpin> {
...
+        std::array<Rect16, 2> rects = IWiSpin::getMenuRects(window_menu, rect); // MINI GUI implementation
+#else                                                                           // PRINTER TYPE != PRUSA_PRINTER_MINI
+        std::array<Rect16, 3> rects = getSpinMenuRects(window_menu, rect);
+#endif                                                                          // PRINTER TYPE == PRUSA_PRINTER_MINI
+        //draw label
+        printLabel_into_rect(rects[0], color_text, color_back, window_menu.font, window_menu.padding, window_menu.GetAlignment());
+        //draw spin
+        // this MakeRAM is safe - temp_buff is allocated for the whole life of IWiSpin
+        render_text_align(rects[1], string_view_utf8::MakeRAM((const uint8_t *)temp_buff.data()),
+#if (PRINTER_TYPE == PRINTER_PRUSA_MINI)
```
DRracer, line 80: All these randomly scattered #ifdefs only show a poor design choice/principles. Remove them (use a derived class when some other functionality is needed)

DRracer (changes requested):
There are multiple issues with this PR:
- the whole PR looks like carpet bombing while the added functionality does't bring much
- changing the flags from `bool enabled = true` to `is_enabled_t enabled = is_enabled_t::yes` makes it even larger without any added value (only syntactic sugar)
- randomly scattered #ifdefs making the code very hard to read and verify
- some variables (font, unit text pointer) are being added - that surely enlarges the whole code, however no statement about how much more RAM and CPUFLASH is needed
- then there is some rect-fu-magic with item rectangles for drawing I'm unable to verify at the first look (not saying the code is wrong)
- beware: computeNumUtf8CharsAndRewind will always run through the whole translated string to compute the number of glyphs - aka O(n) - if you do it once, it's fine. Doing it for each paint call is wrong
- I personally don't like returning std::array<something, 2> from a method, but using std::pair or a tuple will (shall) result in the same code, so no harm done here.

src/guiapi/src/window_menu.cpp
michalrudolf: Fixed in force push.

.vscode/settings.json
```diff
@@ -20,4 +20,7 @@
     "cmake.cmakePath": "${workspaceFolder}/.dependencies/cmake-3.15.5/bin/cmake",
     "cmake.generator": "Ninja",
     "files.insertFinalNewline": true,
+    "files.associations": {
```
DRracer, line 23: what is this?

src/gui/screen_menu_lan_settings.cpp
```diff
@@ -178,7 +178,7 @@ class MI_LAN_IP_t : public WI_SWITCH_t<2> {
 
 public:
     MI_LAN_IP_t()
-        : WI_SWITCH_t<2>(Eth::IsStatic() ? 1 : 0, _(label), 0, true, false, str_DHCP, str_static) {}
+        : WI_SWITCH_t(Eth::IsStatic() ? 1 : 0, _(label), 0, is_enabled_t::yes, is_hidden_t::no, _(str_DHCP), _(str_static)) {}
```
DRracer, line 181: I doubt DHCP will ever receive any translation, therefore _(str_DHCP) doesn't make sense

src/guiapi/include/IWindowMenuItem.hpp
RadekVana: rewritten

src/guiapi/include/menu_spin_config_types.hpp
```diff
@@ -0,0 +1,93 @@
...
+};
+
+template <class T>
+size_t SpinConfig<T>::txtMeas(T val) {
+    return snprintf(NULL, 0, prt_format, val);
+}
+
+template <>
+inline size_t SpinConfig<float>::txtMeas(float val) {
+    return snprintf(NULL, 0, prt_format, (double)val);
```
DRracer, line 82: use nullptr in c++ code please

src/guiapi/src/gui.cpp
```diff
@@ -19,6 +19,8 @@ osThreadId gui_task_handle = 0;
 
 font_t *GuiDefaults::Font = 0;
 font_t *GuiDefaults::FontBig = 0;
+font_t *GuiDefaults::FontMenuItems = 0;
+font_t *GuiDefaults::FontMenuSpecial = 0;
```
DRracer, line 23: use nullptr please!

src/guiapi/src/window_menu.cpp
DRracer: I still don't see the "1" being changed into some meaningful constant

src/guiapi/src/WindowMenuSpin.cpp
```diff
@@ -0,0 +1,77 @@
...
+        padding.left = Utf8Char == '\177' ? 0 : unit__half_space_padding;                  //177oct (127dec) todo check
+        render_text_align(unit_rc, units, Font, color_back, COLOR_SILVER, padding, align); //render unit
+    }
+}
+
+Rect16::Width_t IWiSpin::calculateExtensionWidth(const char *unit, size_t value_max_digits) {
+    size_t ret = value_max_digits * Font->w;
+    if (unit) {
+        ret += 2 * (Padding.left + Padding.right);
+        ret += _(unit).computeNumUtf8CharsAndRewind() * GuiDefaults::FontMenuSpecial->w;
```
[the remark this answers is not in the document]
RadekVana: added fixed option

DRracer (approved):

PR #1029, opened 2020-11-03
tests: Fix tests on Windows BFW-1462

espr14:


DRracer (changes requested):
This PR is a set of random commits which are even reverted at some spot. Ideally, rebase/squash them, so that the whole sequence of commits makes sense.

Please check also PR#1010 which tries to do similar stuff when exporting txt files for unit tests from a more complex perspective.

DRracer (approved):

PR #1034, opened 2020-11-06
Fix basic formating

marekmosna-prusa3d:
BFW-1436
- [vscode utf8 encoding](https://stackoverflow.com/questions/30082741/change-the-encoding-of-a-file-in-visual-studio-code)
- [fix clrf](https://stackoverflow.com/questions/10418975/how-to-change-line-ending-settings)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,3 @@
 * text=auto
+* text eol=crlf
+* text eol=lf
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,5 @@
 * text=auto
 * text eol=crlf
 * text eol=lf
+*.png binary
+*.xlsx -text
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,11 +1,6 @@
 # See https://pre-commit.com for more information
 # See https://pre-commit.com/hooks.html for more hooks
 repos:
-- repo: https://github.com/dragomirecky/cmake-format-pre-commit-hook
-  rev: 'v0.6.0'
-  hooks:
-    - id: cmake-format  # cmake formatter
-      files: (CMakeLists.*|.*\.cmake|.*\.cmake.in)
 - repo: https://github.com/pre-commit/mirrors-yapf
   rev: 'v0.27.0'
   hooks:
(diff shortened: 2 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -1,11 +1,6 @@
 # See https://pre-commit.com for more information
 # See https://pre-commit.com/hooks.html for more hooks
 repos:
-- repo: https://github.com/dragomirecky/cmake-format-pre-commit-hook
```
DRracer, line 4: I don't agree with removal of cmake formatter

.gitattributes
```diff
@@ -1 +1,5 @@
 * text=auto
+* text eol=crlf
```
DRracer, line 2: Why are the text eol=something 2x here? Does it have a technical reason?

.vscode/settings.json
```diff
@@ -20,4 +21,5 @@
     "cmake.cmakePath": "${workspaceFolder}/.dependencies/cmake-3.15.5/bin/cmake",
     "cmake.generator": "Ninja",
     "files.insertFinalNewline": true,
+    "files.encoding": "utf8"
```
DRracer, line 24: this may cover most of the people developing on MS Windows ... not a rock solid solution, but a good start.

.vscode/settings.json
```diff
@@ -1,5 +1,6 @@
 {
     "cmake.configureOnOpen": true,
+    "editor.renderWhitespace": "all",
```
DRracer, line 3: Do we really want to add this? Is it somehow related to basic formatting?

DRracer (changes requested):

.gitattributes
marekmosna-prusa3d: one is for checkout and one for commit

.vscode/settings.json
marekmosna-prusa3d: I'm going to fix this configuration. You're right, It's not welcomed for everyone.

.pre-commit-config.yaml
marekmosna-prusa3d: IMHO, It's time to specify what formatting we don't want instead of how we see the best style.

PR #1040, opened 2020-11-09
Add metrics and improve memory management

dragomirecky:
- Adds metrics. Their gcodes are disabled by default (in order to disable them in production effectively).
- Static allocations for the CCRAM are now being handled by the linker. 
- To better utilize resources, some buffers were moved to CCRAM.
- Adds several checks for memory allocation failures and other possible issues.
- Unites memory pool for malloc from stdlib with FreeRTOS's memory pool.
- Makes malloc()/free() thread and interrupt safe.


src/marlin_stubs/M330.h
```diff
@@ -0,0 +1,12 @@
+#pragma once
+
+namespace PrusaGcodeSuite {
+
+/// Metric-related GCodes.
+/// TODO: This is WIP. Those Gcodes need redesign.
```
DRracer, line 6: As a proof of concept this is ok, however discussing this with teams developing Marlin, Repetier and others is necessary, ideally into FW 4.3 final

DRracer (dismissed):
Nice ;)

DRracer (approved):
approving again after solving the conflict

PR #1042, opened 2020-11-10
Add a new UART driver implementation + Fix TMC communication

dragomirecky:
This PR does the following
- It improves the robustness of communication with trinamics
- It introduces a new high-level BufferedSerial class 
  - which uses DMA in a circular mode for reception
  - and interrupts for non-blocking transmission
- Both writes and reads to the trinamics' registers are now checked for errors

In order to achieve the goals above, it also enables Timers and EventGroups (features of FreeRTOS). This slightly increases RAM consumption (as it runs a new task with minimal stack size). However, I believe those are very useful features we _should_ be using in other parts of the firmware too, and the benefit of using them outweighs the downside of the higher ram consumption.

One more note: the way `main_X.c` files interface the BufferedSerial class (and how BufferedSerial::uart2 is defined) shall be temporary. This was the cleanest way I found in order to make the main changes, but interfacing the main_X.c files is still very messy. I hope this will get better as soon as we fix the main_X.c files.


mkbel (approved):
Please put into test instructions that fastboot/putslave needs to be retested.

PR #1051, opened 2020-11-12
RAM: Unify use of PNG buffer BFW-1492

espr14:


DRracer (approved):
For future we must also add some locking mechanism for debugging, when different pieces of code try to use this scratch buffer for different purposes.

Issue #1052, opened 2020-11-13
[BUG] Printer Status on GUI does not update on Pause


JeffersGlass:
**Printer type** - [MINI]

**Printer firmware version**-  4.2.0

**Original or Custom firmware** Original

**Optional upgrades** - None

**USB drive or USB/Octoprint**
USB Drive

**Describe the bug**
When the printer is printing, clicking the "Pause" button on the GUI does not change the "Printing" header at the top of the GUI, nor the associated print icon. The header remains "Printing" with the triangular "printing" symbol next to it.

**How to reproduce**

1. Begin a new print from USB.
2. Once print has started, pause the print.
3. Observe that the header and icon still reflect the "printing" state rather than the paused state

**Expected behavior**
The header should change to "Paused" with the pause icon, or perhaps should change to "Pausing" and then "Paused"

**Video**
[IMG_2149 (1).zip](https://github.com/prusa3d/Prusa-Firmware-Buddy/files/5537783/IMG_2149.1.zip)

JohnnyDeer:
Hi @JeffersGlass thank you for reporting and also big thanks for your pull request :)

DRracer:
@JeffersGlass your PR has been approved and merged (will be in the upcoming FW 4.3-RC1), would you mind closing this issue?

JeffersGlass:
Can do! Thanks for integrating the fix.

PR #1053, opened 2020-11-13
Make Pause, Stop Buttons on Print Screen update State

JeffersGlass:
Addresses issue #1052  
https://github.com/prusa3d/Prusa-Firmware-Buddy/issues/1052

Previously, only the stopAction() on the printing screen updated the
state of the GUI (based on the state of the printer given by the
Marlin server. Now, both the Tune and Pause buttons will cause the
printer's state to update.

Signed-off-by: Jeff Glass <glass.jeffrey@gmail.com>

### DIFF
--- a/src/gui/screen_printing.cpp
+++ b/src/gui/screen_printing.cpp
@@ -60,6 +60,7 @@ void screen_printing_data_t::tuneAction() {
     case printing_state_t::PRINTING:
     case printing_state_t::PAUSED:
         Screens::Access()->Open(GetScreenMenuTune);
+        change_print_state();
         break;
     default:
         break;
@@ -73,12 +74,15 @@ void screen_printing_data_t::pauseAction() {
     switch (GetState()) {
     case printing_state_t::PRINTING:
         marlin_print_pause();
+        change_print_state();
         break;
     case printing_state_t::PAUSED:
         marlin_print_resume();
+        change_print_state();
         break;
     case printing_state_t::PRINTED:
         screen_printing_reprint();
+        change_print_state();
         break;
     default:
         break;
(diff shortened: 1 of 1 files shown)


src/gui/screen_printing.cpp
```diff
@@ -60,6 +60,7 @@ void screen_printing_data_t::tuneAction() {
     case printing_state_t::PRINTING:
     case printing_state_t::PAUSED:
         Screens::Access()->Open(GetScreenMenuTune);
+        change_print_state();
```
RadekVana, line 63: GUI supports only one screen at the time.
So `Screens::Access()->Open(GetScreenMenuTune);` will close (destroy) Print screen.
But `change_print_state()` could cause extra unwanted redraw just before it is closed.

Please remove it.

src/gui/screen_printing.cpp
```diff
@@ -91,6 +95,7 @@ void screen_printing_data_t::stopAction() {
     }
     switch (GetState()) {
     case printing_state_t::PRINTED:
+        change_print_state();
```
RadekVana, line 98: `change_print_state()` (with `Screens::Access()->Close();`) could cause extra unwanted redraw just before screen is closed.

Please remove it.

RadekVana (changes requested):
Calling `change_print_state();` with `Screens::Access()->Open` or `Screens::Access()->Close` has not meaning, but could cause extra unwanted redraw just before screen is.
Please remove it.

Other changes are fine.

RadekVana (approved):

DRracer:
@JeffersGlass thank you for contributing to the Buddy Firmware. We made the slight changes for you, now the PR is ready to get merged.

PR #1061, opened 2020-11-17
GUI: Refactor wizard

RadekVana:
BFW-1270, BFW-1438, BFW-1497, BFW-1514, BFW-1515

### DIFF
--- a/src/gui/wizard/selftest.cpp
+++ b/src/gui/wizard/selftest.cpp
@@ -5,6 +5,7 @@
 #include "selftest_MINI.h"
 #include "marlin_client.h"
 #include "DialogHandler.hpp"
+#include "eeprom.h"
 
 StateFncData StateFnc_SELFTEST_FAN(StateFncData last_run) {
     marlin_test_start(stmFans);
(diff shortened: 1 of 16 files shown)


src/gui/wizard/selftest.cpp
```diff
@@ -5,107 +5,65 @@
...
-}
-
-StateFncData StateFnc_SELFTEST_FAIL(StateFncData last_run) {
-    static const char en_text[] = N_("The selftest failed to finish. Double-check the printer's wiring and axes. Then restart the Selftest.");
-    string_view_utf8 translatedText = _(en_text);
-    MsgBox(translatedText, Responses_Next);
-    return StateFncData(WizardState_t::EXIT, WizardTestState_t::PASSED);
+WizardState_t StateFnc_SELFTEST_RESULT() {
+    static const char en_text_ok[] = N_("All tests finished successfully!");
+    static const char en_text_nok[] = N_("The selftest failed to finish. Double-check the printer's wiring and axes. Then restart the Selftest.");
```
ftrofin, line 42: This message is too generic and not very helpful. I get this message all the time on my MINI (I think it's because it doesn't cool off fast enough) and it's not useful to the user or customer support. It will be better to add an error code that specifies *what* failed during the self test so the users can search for a specific problem or they can pass that information to customer service which can diagnose where the problem is.  There are multiple steps during the self test, we should specify why it failed.

ftrofin (changes requested):

src/gui/wizard/selftest.cpp
DRracer: @ftrofin thank your for your comment - you are right, this is something we'd like to improve further - now the Wizard can distinguish what failed and report accordingly. Stay tuned.

DRracer (approved):
Approving the parts done by @RadekVana.
Please @espr14 review my commit and we can merge this.
@ftrofin we'll work on top of this PR to supply the user with more detailed data about what exactly failed.

espr14 (changes requested):
It's OK. Just the title does not reflect what / how much happened. What about `GUI: Refactor wizard`?

espr14 (approved):

PR #1063, opened 2020-11-18
Park the head after calibration.

### LINKED ISSUE #637: [ENHANCEMENT]First Layer Calibration Finish Position
Related to #50 and #242 . Prusa Mini. Running firmware 4.0.5.

When a typical print finishes, the bed returns to (x,y) 170,170.

When first layer calibration finishes, it returns to (170,0) This puts the bed at the rear-most 
position that makes clearing the filament off more difficult. I end up moving the axis forward and raising the z-height in order to clear off the calibration filament. 

Can the default after-calibration home position be set to (170,170) like it does for normal prints?

ftrofin:
Move the head higher on Z axis after calibration is done so the plate can be cleaned easily. Fixes #637.


ftrofin:
Who can review this and merge it? @DRracer do you know?

RadekVana (approved):

RadekVana:
We are building RC today, so i will merge it. Thx

PR #1078, opened 2020-11-23
HandleTimeoutInsideWindow

RadekVana:


src/guiapi/src/window.cpp
```diff
@@ -235,9 +238,24 @@ void window_t::draw() {
...
+        return;
+    unregisterSubWin(*win);
+    Screens::Access()->ResetTimeout();
+}
+
+bool window_t::registerSubWin(window_t &win) {
     return false;
 }
 
+void window_t::unregisterSubWin(window_t &win) {
```
Chleba, line 256: no return ?

Chleba: Sorry, bad reading. My mistake

Chleba (approved):

PR #1083, opened 2020-11-25
Add switch to anonymize QR code

Chleba:
BFW-1534

### DIFF
--- a/src/common/bsod.cpp
+++ b/src/common/bsod.cpp
@@ -40,6 +40,7 @@
     #include "../../lib/Marlin/Marlin/src/core/language.h"
     #include "../../lib/Marlin/Marlin/src/lcd/language/language_en.h"
     #include "scratch_buffer.hpp"
+    #include "eeprom.h"
 
     /* FreeRTOS includes. */
     #include "StackMacros.h"
@@ -272,7 +273,15 @@ void draw_error_screen(const uint16_t error_code_short) {
 
         /// draw QR
         char qr_text[MAX_LEN_4QR + 1];
-        error_url_long(qr_text, sizeof(qr_text), error_code);
+
+        /// switch for sending UID of printer or not
+        bool qr_privacy = variant_get_ui8(eeprom_get_var(EEVAR_QR_PRIVACY));
+        if (!qr_privacy) {
+            error_url_long(qr_text, sizeof(qr_text), error_code);
+        } else {
+            error_url_short(qr_text, sizeof(qr_text), error_code);
+        }
+
         constexpr uint8_t qr_size_px = 140;
         const Rect16 qr_rect = { 120 - qr_size_px / 2, 223 - qr_size_px / 2, qr_size_px, qr_size_px }; /// center = [120,223]
         window_qr_t win(nullptr, qr_rect);
--- a/src/common/eeprom.cpp
+++ b/src/common/eeprom.cpp
@@ -10,7 +10,7 @@
 #include "../Marlin/src/module/temperature.h"
 #include "cmath_ext.h"
 
-static const constexpr uint8_t EEPROM__PADDING = 3;
+static const constexpr uint8_t EEPROM__PADDING = 2;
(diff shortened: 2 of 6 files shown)


src/gui/screen_menu_settings.cpp
```diff
@@ -83,7 +83,7 @@ using Screen = ScreenMenu<EHeader::Off, EFooter::On, HelpLines_None, MI_RETURN,
     MI_LAN_SETTINGS,
     MI_TIMEZONE,
     #endif //BUDDY_ENABLE_ETHERNET
-    MI_SAVE_DUMP, MI_SOUND_MODE, MI_SOUND_VOLUME, MI_LANGUAGE>;
+    MI_SAVE_DUMP, MI_SOUND_MODE, MI_SOUND_VOLUME, MI_LANGUAGE, MI_QR_PRIVACY>;
```
DRracer, line 86: may be we put this before the language item

src/common/bsod.cpp
```diff
@@ -272,7 +273,15 @@ void draw_error_screen(const uint16_t error_code_short) {
 
         /// draw QR
         char qr_text[MAX_LEN_4QR + 1];
-        error_url_long(qr_text, sizeof(qr_text), error_code);
+
+        /// switch for sending UID of printer or not
+        bool qr_privacy = variant_get_ui8(eeprom_get_var(EEVAR_QR_PRIVACY));
```
DRracer, line 278: probably we shall call the vars anonymized_qr to make it consistent with the menu item

DRracer (commented):
Approving changes done by @Chleba

src/common/bsod.cpp
dragomirecky: `eeprom_get_var` calls FreeRTOS's `osSemaphoreWait`. Is it safe, considering FreeRTOS is not running at that point?

src/guiapi/include/MItem_menus.hpp
```diff
@@ -187,3 +187,11 @@ class MI_EEPROM : public WI_LABEL_t {
 protected:
     virtual void click(IWindowMenu &window_menu) override;
 };
+
+class MI_QR_PRIVACY : public WI_SWITCH_OFF_ON_t {
+    constexpr static const char *const label = N_("Anonymized QR");
```
dragomirecky, line 192: I, as a user, would be confused, by what this means. Because if I haven't seen any BSOD yet then I wouldn't know what QR is this talking about.

Do we have space on the screen for something like „Anonymized Crash Reports“, or something similar?

src/common/eeprom.cpp
```diff
@@ -195,7 +197,8 @@ static const eeprom_vars_t eeprom_var_defaults = {
     {"Custom2", FLT_MAX },
     {"Custom3", FLT_MAX },
     {"Custom4", FLT_MAX },
-	0,               // EEVAR_SELFTEST_RESULT
+	  0,               // EEVAR_SELFTEST_RESULT
```
dragomirecky, line 200: Extra tab here 🧐

src/common/bsod.cpp
Chleba: That's a very good question, but now BSOD screen is up right after reset, so it's ok and everything is running

src/common/eeprom.cpp
Chleba: clang format is off at this point :(

src/guiapi/include/MItem_menus.hpp
DRracer: This is only about removing the hash from QR codes, BSOD has no QR codes yet.
We may change the name of this item into 4.3 final, this the best name I could get from our Content team yesterday.

src/common/eeprom.cpp
DRracer: there are some edge cases of clang-format, I wouldn't block the PR because of this

src/common/bsod.cpp
DRracer: @Chleba is the eeprom itself already initialized at this point?

DRracer (approved):

PR #1085, opened 2020-11-25
tmp fix for blinking underscore

Chleba:
BFW-1426, BFW-1398


RadekVana:
no need for this workaround, it already works as it should

PR #1089, opened 2020-11-26
Unify display source

mkbel:


src/guiapi/include/st7789v.hpp
```diff
@@ -34,10 +36,6 @@ typedef struct _st7789v_config_t {
     uint8_t control;
 } st7789v_config_t;
 
-#ifdef __cplusplus
-extern "C" {
-#endif //__cplusplus
-
 __attribute__((used)) inline uint16_t color_to_565(uint32_t clr) {
```
DRracer, line 39: while switching to C++ this attribute(used) may also disappear

DRracer (approved):
Checked the original st7789v.c vs. the new cpp file - the changes look correctly transferred. The rest is just playing with c++ headers, approving this PR.

src/guiapi/include/st7789v.hpp
mkbel: fixed

RadekVana (approved):

PR #1098, opened 2020-11-27
Update USB Driver & HAL

marekmosna-prusa3d:
BFW-1457
Due to incompatibility between the latest USB Drivers and an older version of the STM HAL(Hardware Abstract Layer), PR is forced to update STM HAL as well to ensure proper functionality of the USB peripheral.


joshymjose (approved):

DRracer (approved):
approving again after PR's base change

PR #1099, opened 2020-11-27
Fix ccram being "loadable"

dragomirecky:


DRracer (commented):
I'll wait for the build ... 128MB being loadable... hmmm, if only we had so much memory :)

DRracer (approved):
now the bin/bbf size looks correct again, approving

PR #1100, opened 2020-11-27
remove not used marlin safety timer event handling

Chleba:


DRracer (approved):
reflects the new reporting, approving

PR #1101, opened 2020-11-27
Enable EEPROM upgrading from 4.2.1

DRracer:
The main trouble arose with the change of implementation of the CRC32 algorithm recently.
The new impl. uses the reverse CRC32, while the EEPROM checks used the normal CRC32.
This legacy behavior must be kept for backwards compatibility, thus there are now
both algorithms (both HW accelerated).

Kudos to @dragomirecky for help with the CRC32 HW chip.

Two issues remain to be solved/explained:
- `float active_z_offset = variant8_get_flt(eeprom_get_var(EEVAR_ZOFFSET));` will probably work, but is not entirely correct (the live-z value gets read when the eeprom block gets copied)
- EEPROM v8 probably never made it into the public, it is some pre-4.3-RC1 variant

BFW-1541

### DIFF
--- a/src/common/crc32.c
+++ b/src/common/crc32.c
@@ -63,6 +63,21 @@ static uint32_t crc32_hw(const uint32_t *buffer, uint32_t length, uint32_t crc) 
     return result;
 }
 
+uint32_t crc32_eeprom(const uint32_t *buffer, uint32_t length) {
+    // ensure nobody else uses the peripheral
+    osMutexWait(crc32_hw_mutex_id, osWaitForever);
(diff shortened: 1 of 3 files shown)


dragomirecky (dismissed):
Looks good to me 👍

DRracer:
After discussion with @marekmosna-prusa3d all the upgrading functions shall retain the same capability of copying the live-z value into the first active sheet. This will be added shortly to this PR

marekmosna-prusa3d (dismissed):

marekmosna-prusa3d (approved):

PR #1108, opened 2020-11-30
Reset dump in XFLASH when FW upgraded or EEPROM reset

DRracer:
In rare scenarios, the XFLASH dump could remain "unprocessed".
That would lead to not displaying a red error screen after printer reset.
All the safety features were working fine, but the user wouldn't get notified about the error that caused the reset.
This should be fixed with this PR - whenever there is a FW upgrade or the machine loads defaults into the EEPROM (usually CRC mismatch or similar problems), the XFLASH dump gets erased completely as well - since it becomes obsolete anyway.

### DIFF
--- a/src/common/appmain.cpp
+++ b/src/common/appmain.cpp
@@ -31,6 +31,7 @@
 #include "safe_state.h"
 #include "crc32.h"
 #include "ff.h"
+#include "dump.h"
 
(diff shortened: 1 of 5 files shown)


DRracer:
closing in favour of PR #1114 and #1118

PR #1111, opened 2020-12-01
Restoring Print Speed

michalrudolf:
BFW-1193
Motivation:
When printing starts executing G28 and G30 it saves and sets its own feedrate_percentage (100%). When user overwrites those settings during executing, at the end, old saved values will be restored. (This discards user defined feedrate value)

Solution:
The settings from before will be set back only if feedrate_percentage(100%) was not changed during G28/G30. Otherwise it keeps user defined values.


DRracer (commented):
This needs serious explanation, I have no idea what this may cause in Marlin

DRracer (approved):

PR #1114, opened 2020-12-01
XFLASH reset test for Red Screens

Chleba:
add xflash reset menu item, update condition to generate dump, add xflash reset function from PR #1108


DRracer (dismissed):
This seems to fix the real cause of red screens not popping up upon restart. PR #1108 was merely a workaround, but helped to identify the issue - the dump was not saved at all when the XFLASH was clean.

DRracer (approved):
approving based upon the successful tests

PR #1117, opened 2020-12-02
FAN selftest thresholds adjustment BFW-1557 

zoltan-l:
Fan speed thresholds adjustment and Fan names unification


DRracer (approved):
approving, we'll keep the range of valid RPM more vague in 4.3-RC

PR #1121, opened 2020-12-03
Wizard caption text

RadekVana:
BFW-1316


DRracer (commented):
approving into 4.3 final
changes in the POT file are just in comments, no need to attach it here

DRracer (approved):

PR #1132, opened 2020-12-07
Add WUI files script generation of hex data

Chleba:
Replaced manual conversion of static files from WUI with python script generating raw hex data into a one main file.

BFW-1478, BFW-1525


src/lang/CMakeLists.txt
```diff
@@ -12,6 +12,12 @@ add_custom_command(
   COMMENT "Generating C++ files out of .po files."
   VERBATIM
   )
+add_custom_command(
+  OUTPUT "${PROJECT_ROOT_DIR}/lib/WUI/resources/fsdata_wui_local.c"
+  COMMAND "${Python3_EXECUTABLE}" "${PROJECT_ROOT_DIR}/utils/wui.py"
+  COMMENT "Generating fsdata_wui_local raw data from WUI resources."
+  VERBATIM
+  )
```
dragomirecky, line 20: Why is it inside `src/lang`? 

/suggestion/ If I am not missing something, please move to a more appropriate location.

utils/wui.py
```diff
@@ -0,0 +1,220 @@
...
+        self.raw_data_file_tmp = Path(__file__).resolve().parent.parent / 'lib' / 'WUI' / 'resources' / str(self.main_file_name + '_tmp')
+
+        # start
+        self.getFiles()
+        self.gzipFiles()
+        self.generateRawDataFile()
+    #endef
+
+    # store paths to static wui files in array for later use
+    def getFiles(self):
```
dragomirecky, line 25: /suggestion/ Let's please follow the so-far only naming convention we have for Python in our repository and change the method names to snake_case

utils/wui.py
```diff
@@ -0,0 +1,220 @@
+#!/usr/bin/env python3
+
+import os
+from pathlib import Path
+import gzip
+import shutil
+
+class WUIFiles:
```
dragomirecky, line 8: /question/ Should this functionality be really associated with WUI itself? It seems to me, that it only takes some files, compresses them, and wraps/encodes them into C source files. It is a generic task that can be used by any other part of the firmware, not just WUI.

utils/wui.py
```diff
@@ -0,0 +1,220 @@
...
+        # path to main C file with raw data
+        self.main_file_name = 'fsdata_wui_local.c'
+        self.raw_data_file = Path(__file__).resolve().parent.parent / 'lib' / 'WUI' / 'resources' / self.main_file_name
+        self.raw_data_file_tmp = Path(__file__).resolve().parent.parent / 'lib' / 'WUI' / 'resources' / str(self.main_file_name + '_tmp')
+
+        # start
+        self.getFiles()
+        self.gzipFiles()
+        self.generateRawDataFile()
+    #endef
```
dragomirecky, line 22: 🧐

utils/wui.py
```diff
@@ -0,0 +1,220 @@
...
+
+    # store paths to static wui files in array for later use
```
dragomirecky, line 24: /suggestion/ Use python docstring for documentation

```Python
def getFiles(self):
    """ Store paths to static wui files in array for later use """
    ...
```

utils/wui.py
```diff
@@ -0,0 +1,220 @@
...
+        data.append('#endif\n')
+        data.append('#if FSDATA_FILE_ALIGNMENT == 2\n')
+        data.append('#include "fsdata_alignment.h"\n')
+        data.append('#endif\n\n')
+        #  data.append('')
+
+        # write data
+        self.writeTmpData(data)
+    #endef
+#enclass
```
dragomirecky, line 217: 😳

dragomirecky (commented):
Thanks for working on improving the build system!

I think we should consider making this tool more lightweight and making sure it does only one thing – that is converting files to includable C source files usable by WUI. 

It is the responsibility of the build system to translate all the necessary files using this tool. This tool doesn't have to know anything about WUI, how to find such files, or where to store the translated versions. 

Consider this tool just as another type of a compiler. It takes input files, translates them, and saves the output somewhere (where it is told to). The compiler doesn't have to know anything about the specific project it is being used for (in this case WUI). It is the job of the build-system to call this compiler with the right set of arguments in order to assemble the whole project.

utils/wui.py
Chleba: Sorry - old habits from projects I've done before. We used to mark end of each cycle, function, method, class, etc.. with these comments. I can delete them if you want.

utils/wui.py
Chleba: This script is made specifically for generating fsdata_wui_local.c file, that have some static code parts just for handling WUI files. I know what you mean, and generating raw data from files could be separated and used for different tasks (for example - image assets from png files for fw GUI).

utils/wui.py
Chleba: I will change that - sorry

src/lang/CMakeLists.txt
Chleba: My mistake. I will be moved into it's own folder. Sorry about that.

utils/wui.py
dragomirecky: I see you changed the way methods are documented. However, the docstrings have to be _under_ the function name (as in the example I gave you above). See https://www.python.org/dev/peps/pep-0257/.

lib/WUI/CMakeLists.txt
```diff
@@ -1,29 +1,29 @@
...
-  WUI
-  INTERFACE LwIP
+add_custom_command(
+  OUTPUT "${PROJECT_ROOT_DIR}/lib/WUI/resources/fsdata_wui_local.c"
+  COMMAND "${Python3_EXECUTABLE}" "${PROJECT_ROOT_DIR}/utils/wui.py"
+  COMMENT "Generating fsdata_wui_local raw data from WUI resources."
+  VERBATIM
   )
+
+add_custom_target(wui-genfile DEPENDS "${PROJECT_ROOT_DIR}/lib/WUI/resources/fsdata_wui_local.c")
```
dragomirecky, line 25: /question/ I might be missing something, but where is this target being used? Is it meant for direct invocation from CLI only? I would expect the WUI library to have it as a dependency.

utils/wui.py
```diff
@@ -0,0 +1,239 @@
...
+        # path to resource files directory
+        self.wui_path = Path(__file__).resolve(
+        ).parent.parent / 'lib' / 'WUI' / 'resources' / 'src_local'
+        # path to main C file with raw data
+        self.main_file_name = 'fsdata_wui_local.c'
+        self.raw_data_file = Path(__file__).resolve(
+        ).parent.parent / 'lib' / 'WUI' / 'resources' / self.main_file_name
+        self.raw_data_file_tmp = Path(__file__).resolve(
+        ).parent.parent / 'lib' / 'WUI' / 'resources' / str(
+            self.main_file_name + '_tmp')
```
dragomirecky, line 21: DRY 

``` Python
project_root = Path(__file__).resolve().parent.parent
wui_root = project_root / 'lib' / 'WUI'
self.wui_path = wui_root / 'resources' / 'src_local'
self.raw_data_file = wui_root / 'resources' / self.main_file_name
self.raw_data_file_tmp = wui_root / 'resources' / f'{self.main_file_name}.tmp'
```

lib/WUI/CMakeLists.txt
Chleba: This script is not meant to be used for WUI library, just for client side of FW. WUI library does not need any dependencies for this.

utils/wui.py
Chleba: That blame should be on yapf side, because previous build has failed on pre-commit check and yapf did that.

PR #1134, opened 2020-12-07
change wizard progressbar, add WizardDefaults

RadekVana:
BFW-1568


DRracer (approved):
Even though there are many changes, they are mostly related to GUI layout. Based on successful tests, I'm approving this PR.

PR #1138, opened 2020-12-09
Setltest heaters

RadekVana:
BFW-1516, BFW-1579, BFW-1578


src/common/selftest/selftest_MINI.cpp
```diff
@@ -387,7 +387,7 @@ CSelftestPart::CSelftestPart()
 CSelftestPart::~CSelftestPart() {
 }
 
-float CSelftestPart::GetProgress() const {
+float CSelftestPart::GetProgress() {
```
DRracer, line 390: why isn't this const anymore?

src/common/selftest/selftest_heater.cpp
```diff
@@ -62,33 +160,36 @@ bool CSelftestPart_Heater::Loop() {
...
+        break;*/
     }
     case spsMeasure:
-        if ((Selftest.m_Time - m_Time) < TEMP_MEASURE_CYCLE_DELAY) {
+        if (int(m_EndTime - Selftest.m_Time) > 0) {
+            m_Temp = getTemp();
+            return true;
+        }
+
+        /* if ((Selftest.m_Time - m_Time) < TEMP_MEASURE_CYCLE_DELAY) {
```
DRracer, line 176: what about this commented code? Is it still relevant?

src/common/selftest/selftest_MINI.cpp
RadekVana: because child changes interval variable

src/common/selftest/selftest_heater.cpp
RadekVana: it is, did not want to remove Roberts code

XPila (approved):

PR #1141, opened 2020-12-09
GUI: Fix red screen layout

marekmosna-prusa3d:
BFW-1574


DRracer (approved):
looks good, I thought it would have been something like this

PR #1142, opened 2020-12-09
Fix appendix status

marekmosna-prusa3d:
BFW-1580


DRracer (commented):
This looks like a very correct solution to not clear the bootloader-FW exchange space. Still, this must be checked thoroughly, so let's test this well into FW 4.3 final

RadekVana (approved):

PR #1156, opened 2020-12-14
Fix long words draw

marekmosna-prusa3d:
BFW-1417
In case that the word cannot fit into the radio button, the word will be cut to the longest possible suitable length.

### DIFF
--- a/src/gui/dialogs/DialogRadioButton.cpp
+++ b/src/gui/dialogs/DialogRadioButton.cpp
@@ -127,7 +127,14 @@ void RadioButton::draw_n_btns(const size_t btn_count) const {
     rect.HorizontalSplit(splits, spaces, btn_count, GuiDefaults::ButtonSpacing, ratio);
 
     for (size_t i = 0; i < btn_count; ++i) {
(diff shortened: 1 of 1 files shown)


src/gui/dialogs/DialogRadioButton.cpp
```diff
@@ -127,7 +127,14 @@ void RadioButton::draw_n_btns(const size_t btn_count) const {
     rect.HorizontalSplit(splits, spaces, btn_count, GuiDefaults::ButtonSpacing, ratio);
 
     for (size_t i = 0; i < btn_count; ++i) {
-        button_draw(splits[i], _((*texts)[i]), pfont, GetBtnIndex() == i && IsEnabled());
+        string_view_utf8 drawn = _((*texts)[i]);
+        char buffer[128];
+        if ((pfont->w * ratio[i]) > splits[i].Width()) {
+            size_t length = drawn.copyToRAM(buffer, splits[i].Width() / pfont->w);
+            buffer[length] = 0;
```
RadekVana, line 134: check buffer size

RadekVana (changes requested):

RadekVana (approved):

PR #1166, opened 2020-12-16
SYSTEM: Appendix status read from the PIN

marekmosna-prusa3d:
BFW-1580

### DIFF
--- a/src/common/support_utils.cpp
+++ b/src/common/support_utils.cpp
@@ -12,6 +12,7 @@
 #include "language_eeprom.hpp"
 #include "sha256.h"
 #include "crc32.h"
+#include "stm32f4xx_hal_gpio.h"
 
 #include "qrcodegen.h"
 #include "support_utils_lib.hpp"
@@ -67,7 +68,7 @@ void printerCode(char *str) {
     }
 
     /// appendix state
-    if (ram_data_exchange.model_specific_flags & APPENDIX_FLAG_MASK) {
+    if (appendix_exist()) {
         setBit((uint8_t *)hash, 6);
         //setBit(str[0], 6);
     }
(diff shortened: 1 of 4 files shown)


src/common/support_utils.cpp
```diff
@@ -164,3 +165,20 @@ void create_path_info_4service(char *str, const uint32_t str_size) {
...
+    const version_t *bootloader = (const version_t *)BOOTLOADER_VERSION_ADDRESS;
+    GPIO_InitTypeDef GPIO_InitStruct = { 0 };
+    GPIO_InitStruct.Pin = GPIO_PIN_13;
+    GPIO_InitStruct.Mode = GPIO_MODE_INPUT;
+    GPIO_InitStruct.Pull = GPIO_NOPULL;
+
+    if (bootloader->major >= 1 && bootloader->minor >= 1) {
+        return !(ram_data_exchange.model_specific_flags & APPENDIX_FLAG_MASK);
+    } else {
+        HAL_GPIO_Init(GPIOA, &GPIO_InitStruct);
```
RadekVana, line 179: something should revert changes made by `HAL_GPIO_Init(GPIOA, &GPIO_InitStruct);` after read

RadekVana (changes requested):

src/common/support_utils.cpp
RadekVana: also missing `#ifndef _DEBUG`

marekmosna-prusa3d: HAL_GPIO_Init doesn't force any state of the PIN by GPIO_NOPULL. There is only read of the PIN setup by hardware solution of the appendix.

RadekVana (approved):

PR #1177, opened 2020-12-17
Strong Dialogs - new design

Chleba:
BFW-1593


RadekVana:
nicer strong dialogs #1180

PR #1206, opened 2021-01-02
Game: snake BFW-1558

espr14:


vintagepc:
.. and this is why we love Prusa :rofl:

riodoro1:
Sounds more important than WiFi.

riodoro1:
> .. and this is why we love Prusa 🤣

A useless feature stuck in review while there are better things to do?

PR #1216, opened 2021-01-06
Show valid remaining time

RadekVana:
BFW-1645

### DIFF
--- a/src/gui/screen_printing.cpp
+++ b/src/gui/screen_printing.cpp
@@ -166,6 +166,7 @@ screen_printing_data_t::screen_printing_data_t()
     initAndSetIconAndLabel(btn_tune, res_tune);
     initAndSetIconAndLabel(btn_pause, res_pause);
     initAndSetIconAndLabel(btn_stop, res_stop);
+    change_etime();
 }
 
 #ifdef DEBUG_FSENSOR_IN_HEADER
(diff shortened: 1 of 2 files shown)


src/gui/screen_printing.cpp
```diff
@@ -261,12 +266,15 @@ void screen_printing_data_t::enable_tune_button() {
     btn_tune.ico.Invalidate();
 }
 
-void screen_printing_data_t::update_remaining_time(time_t rawtime, uint16_t print_speed) {
-    w_etime_value.color_text = rawtime != time_t(-1) ? GuiDefaults::COLOR_VALUE_VALID : GuiDefaults::COLOR_VALUE_INVALID;
-    if (rawtime != time_t(-1)) {
-        if (print_speed != 100)
+void screen_printing_data_t::update_remaining_time(uint32_t millis, uint16_t print_speed) {
+    bool is_time_valid = int(millis) > 0;
```
espr14, line 270: Is there a reason to convert to int?

espr14 (changes requested):

src/gui/screen_printing.cpp
DRracer: yes, the time can become negative, which results in displying incorrect 37d+ of print time

DRracer, line 270: `bool is_time_valid = millis > 0U;` ... if there was a sign comparison warning (which I doubt)?

DRracer: but probably the whole unsigned int is a problem - either compare millis > 2^31 or keep millis as signed int32_t

RadekVana: fixed

RadekVana: changed, must be shorter than 365 days

espr14 (approved):

DRracer (approved):

PR #1217, opened 2021-01-07
G-code input file filtering+buffering

DRracer:
This PR contains a generic and simple read buffering (RDbuf) and a generic filtering class (GCodeInputFilter).
Both classes need parametrization of important functionality:
- the instance of RDbuf needs to know how to fetch input data
- the instance of GCodeInputFilter needs to know how enqueue filtered G-code commands

Design concerns (over previous implementation of media_loop() in media.cpp):
- handle long comment lines correctly, even if they contain G-code commands -> filter these lines out, avoid sending them into Marlin
- detect long G-code lines and stop printing in such a case avoiding damage to the machine - need to discuss this, marked with '@@TODO'
- avoid FS operations for short chunks of data - ideally read the whole block/sector into RAM -> buffering, avoid f_gets()
- make the processing/filtering generic to enable unit testing

Compared to previous state of art this implementation needs an additional read buffer of 64B long (can be any number)

BFW-1264, BFW-1159, BFW-1161, BFW-1162

### DIFF
--- a/src/common/gcode_input_filter.hpp
+++ b/src/common/gcode_input_filter.hpp
@@ -1,0 +1,177 @@
+/// @brief G-code input filtering
+///
+/// Contains a generic and simple read buffering (RDbuf) and a generic filtering class (GCodeInputFilter).
+/// Both classes need parametrization of important functionality:
+/// - the instance of RDbuf needs to know how to fetch input data
+/// - the instance of GCodeInputFilter needs to know how enqueue filtered G-code commands
+///
+/// Design concerns (over previous implementation of media_loop() in media.cpp):
+/// - handle long comment lines correctly, even if they contain G-code commands -> filter these lines out, avoid sending them into Marlin
+/// - detect long G-code lines and stop printing in such a case avoiding damage to the machine
+/// - avoid FS operations for short chunks of data - ideally read the whole block/sector into RAM -> buffering, avoid f_gets()
+/// - make the processing/filtering generic to enable unit testing
+
+#pragma once
+
+#include <stdint.h>
+#include <limits>
+
+/// RDbuf handles buffering of input data stream - for now it shall sit on top of raw FATfs's f_read and f_eof functions.
+/// It may get replaced in the future when we have "normal" libc file interfaces
+///
+/// Memory requirements:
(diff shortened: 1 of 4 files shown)


src/common/gcode_input_filter.hpp
```diff
@@ -0,0 +1,177 @@
...
+    /// shall take the content of gcodebuff (zero terminated) and push it into Marlin (or any other processing engine)
+    /// new_file_offset represents the offset in the input file/stream AFTER the gcode is pushed into Marlin
+    /// - that's because we want to track which commands were successfully pushed into Marlin and processed
+    ///   to know exactly where to continue in case of an input file/stream removal/error
+    ///
+    /// @returns -1 in case of being "on hold" - i.e. was unable to enqueue the G-code into Marlin (can be called again to retry).
+    ///          Otherwise it returns a positive number of input bytes processed.
+    ///          Please note - if the number != rdbuf.size(), the filter has reached a line which couldn't fit into Marlin's internal line buffer
+    ///             which implies the line would have been truncated and could possibly damage the machine being misinterpreted!
+    template <typename RDBUF, typename ENQUEUE_GCODE>
```
marekmosna-prusa3d, line 124: Please try to cut the logic from the RDBuff and implement it into the filter. Therefore, we will be able to use any std container as a templated parameter. It radically increases the compatibility, usability and extendibility of the filter.

tests/unit/common/media_test.cpp
```diff
@@ -0,0 +1,329 @@
...
+
+enum EMediaLoop { OK,
+    ERROR,
+    ONHOLD };
+
+template <
+    typename ENQUEUEGCODE,
+    uint32_t RDBUFSIZE = 64,
+    uint32_t GCBUFSIZE = gcBuffSize>
+EMediaLoop media_loop_inner(ENQUEUEGCODE enq) {
```
marekmosna-prusa3d, line 49: Please try to cover the edge cases as much as possible by emplace the invalid characters into the defined strings. It allows us to produce more robust unit tests only for the logic of the filter as we need.

DRracer:
closing in favour of PR #1372

PR #1232, opened 2021-01-13
Remove cooldown in forced preheat

RadekVana:
Remove cooldown in forced preheat 
affected firstlayer calibration and print preview


DRracer (approved):
Approving, this was clearly a copy/paste error from the previous code.
Cooldown doesn't make sense when the user wants to heat up the extruder to insert fillament.

PR #1233, opened 2021-01-14
Filament autoload

Hagridpon3:


src/common/thread_measurement.cpp
```diff
@@ -21,20 +20,48 @@ void StartMeasurementTask(void const *argument) {
...
+        if (next_fs_cycle <= now) {
             fs_cycle();
             // call fs_cycle every ~50 ms
             next_fs_cycle = now + 50;
         }
 
+        // sample filament autoload
+        // call every ~100ms
+        marlin_update_vars(MARLIN_VAR_MSK_FS);
+        marlin_vars_t *vars = marlin_update_vars(MARLIN_VAR_SD_PRINT);
```
dragomirecky, line 42: /question/ I am not really familiar with marlin_client's api, but shouldn't it be possible to read this in one go?

```C
marlin_vars_t *vars = marlin_update_vars(MARLIN_VAR_SD_PRINT | MARLIN_VAR_MSK_FS);
```

src/common/thread_measurement.cpp
```diff
@@ -53,6 +80,6 @@ void StartMeasurementTask(void const *argument) {
             next_sg_cycle = now + next_delay;
         }
 
-        osDelay(checkTimestampsAscendingOrder(next_fs_cycle, next_sg_cycle) ? next_fs_cycle - now : next_sg_cycle - now);
+        osDelay(std::max({ next_fs_cycle, next_sg_cycle, next_fs_autoload_cycle }) - now);
```
dragomirecky, line 83: `checkTimestampsAscendingOrder` properly handled the case of uint32_t overflowing; this ignores the problem :)

Btw – shouldn't it be std::min? You want to wake up with the closest event, not on the one being the furthest in the future.

src/common/thread_measurement.cpp
dragomirecky: /question/ Why don't we check for the `next_fs_autoload_cycle` first and then check whether autoload is enabled? Updating marlin vars seems like more expensive operation we might not want to do on every cycle here.

src/common/thread_measurement.cpp
```diff
@@ -5,13 +5,12 @@
...
+#include "marlin_server.h"
 #include "trinamic.h"
 #include "stm32f4xx_hal.h"
+#include "eeprom.h"
 
-static inline bool checkTimestampsAscendingOrder(uint32_t a, uint32_t b) {
-    uint32_t u = (b - a);
-    return !(u & 0x80000000u);
-}
+static bool current_detect_filament_insert = false;
```
dragomirecky, line 13: /question/ Why isn't this a local variable? I don't see it being used anywhere else then in StartMeasurementTask().

src/common/thread_measurement.cpp
Hagridpon3: Solved.

src/common/thread_measurement.cpp
Hagridpon3: Solved

src/gui/screen_menu_settings.cpp
```diff
@@ -69,7 +69,12 @@ bool MI_FILAMENT_SENSOR::consumeNotConnected() {
 }
 
 void MI_FILAMENT_SENSOR::OnChange(size_t old_index) {
-    old_index == 1 ? fs_disable() : fs_enable();
```
RadekVana, line 72: unnecessary change, pls remove

src/common/thread_measurement.cpp
```diff
@@ -33,6 +40,29 @@ void StartMeasurementTask(void const *argument) {
             next_fs_cycle = now + 50;
         }
 
+        // sample filament autoload
```
RadekVana, line 43: move all changed code from this file into filament_sensor.cpp
create static function `autoload_loop` and call it inside `fs_cycle();` faster cycle 50ms for autoload is irrelevant

RadekVana (changes requested):

src/gui/screen_menu_settings.cpp
Hagridpon3: Solved

src/common/thread_measurement.cpp
Hagridpon3: Solved

RadekVana (approved):

Issue #1076, opened 2020-11-21
[FEATURE REQUEST] Treat USB media as SD card for serial control


compmstr:
It would be nice to be able to use the USB media plugged in to the Prusa Mini as an SD card for use with Octoprint.  If nothing else, it seems like it might alleviate issues with https://github.com/prusa3d/Prusa-Firmware-Buddy/issues/739, since you would be able to remotely initiate a USB print, which does appear to properly pause with the filament sensor.

If it makes sense, enter what type of printer or upgrade the feature request applies to.
Printer type - [MINI]
Optional upgrades - [Filament Runout Sensor]

**Is your feature request related to an existing problem? Please describe.**
I wanted to use up the last of some filament, and so I kicked off a print from OctoPrint.  The sensor never triggered the print to pause.  Luckily I was watching it, so I was able to manually pause and change the filament.  Researching the issue, it seems like printing from the USB is the only way to currently have the filament sensor properly work.
I then tried to use the 'SD' commands on octoprint (M20/M21, etc.), and they all appear to do nothing.  I'm assuming this is because the hardware doesn't have an SD card.  If the USB media storage could be treated like an SD card, this would allow us to remotely kick off and control USB prints from octoprint.

**Describe the expected functionality**
Have the SD card GCODE commands (M20, M21, M24, M25, etc.) use the USB media storage in place of a physical SD card slot.

JohnnyDeer:
Hi @compmstr fixing this commands is on our to do list.

pickworthi:
I would like to add another use case, given that the original reason for this issue has potentially been solved by later firmware releases.
The use case I am interested in is to use the M20 to M32 commands over the serial link to load/delete files on the Mini+ USB stick. This allows a local operator to fetch gcode files from a remote source, without having to physically walk there and back with the USB stick in his/her hand.

bschollnick:
> I would like to add another use case, given that the original reason for this issue has potentially been solved by later firmware releases.
> The use case I am interested in is to use the M20 to M32 commands over the serial link to load/delete files on the Mini+ USB stick. This allows a local operator to fetch gcode files from a remote source, without having to physically walk there and back with the USB stick in his/her hand.
> 
> 

I think this is an interesting idea.  But, I’m a bit concerned security wise, that this would allow any unvetted gcode file to erase files (or download files) to my system.  

Just the loading by itself, could potentially open the system up to bad actors (eg Malware, Virii, etc).

[...]

PR #1246, opened 2021-01-20
Add #define to support moving serial to UART1

vintagepc:
This adds an alternate wiring option to route the serial output from USB to UART1 (the TX/RX pins on the white 10 pin header)

To use:
- #define (uncomment) `USE_UART1_SERIAL` in `HardwareSerial.h`
- Change `SERIAL_PORT` from `-1` to `1` in `Configuration_A3ides_2209_MINI.h`

### DIFF
--- a/lib/Arduino_Core_A3ides/cores/arduino/HardwareSerial.h
+++ b/lib/Arduino_Core_A3ides/cores/arduino/HardwareSerial.h
@@ -5,6 +5,11 @@
 #include <inttypes.h>
 #include "Stream.h"
 
+// Changes the serial output to UART1 TX/RX.
+// Note you also need to change the marlin serial
+// port from -1 to 1 in the marlin config.
+//#define USE_UART1_SERIAL
+
 class HardwareSerial : public Stream {
 public:
     HardwareSerial(void *peripheral);
@@ -18,6 +23,23 @@ public:
     virtual size_t write(const uint8_t *buffer, size_t size);
     operator bool();
 };
+
+#ifdef USE_UART1_SERIAL
+// Route to USART1 instead of USB
+class HardwareSerial2 : public Stream {
+public:
+    HardwareSerial2(void *p) {};
+    void begin(unsigned long baud);
+    virtual int available(void);
+    virtual int peek(void) { return -1; };
+    virtual int read(void);
+    virtual void flush(void);
+    virtual size_t write(uint8_t);
+    virtual size_t write(uint8_t *buffer, size_t size);
+    operator bool() { return true; };
+};
+extern HardwareSerial2 SerialUART1;
+#endif
 
 extern HardwareSerial Serial3;
 
--- a/src/common/hardware_serial.cpp
+++ b/src/common/hardware_serial.cpp
@@ -1,6 +1,7 @@
 //HardwareSerial.cpp - A3ides/STM32
 #include <Arduino.h>
 #include "buffered_serial.hpp"
+#include "HardwareSerial.h"
 #include "cmsis_os.h"
 #include "bsod.h"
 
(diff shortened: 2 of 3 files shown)


Turro75:
Thanks for the commit, I'm trying adding it to 4.3.0 branch but no way to get TX working, I moved your code to usart6 to enable the esp8266 serial comms but I get the same, receive is fine while transmit does not work. Any idea?

vintagepc:
I can't say for sure without seeing your specific code changes, but the built-in Marlin is not equipped to handle the DMA receive interrupt or buffers by default (uartrxbuffer doesn't support the required functions that check whether data is available). Make sure you have disabled that for UART6, else the data will get shuttled away before it ever sees an RXNE flag.

Turro75:
I'm getting the same problem even on uart1 by using Your commit, does it works on your side?

vintagepc:
Hmmm... I have not run it on real hardware but I do know it works on simulated hardware.

Turro75:
Ah that's probably the reason, can You kindly address me a page that explains how to setup eclipse to debug this? Switching to another ide isn't a problem to me.

vintagepc:
Sorry, the hardware simulation is a (temporarily) private project that is not yet ready for public release. Eclipse's support for the STM32 platform is basically limited to GPIO and is not usable for this.

Turro75:
ok no problem, in case You want to send me some bbf to test on a real device I'm here.

Best Regards

vintagepc:
Thanks, I'll keep that in mind. My best guess why it many not work is probably some missing serial port/uart configuration since these ports are not used by default.

Turro75:
Well,

I understood what is the failure, setting in cutecom a delay of 2msec between each char allows the buddy board to successfully reading all chars. So this PR provides a functional but slow code, may be the switch to a buffer where data are appended by IT or DMA will be the solution then Marlin can pick one char at a time without the risk to skip something.

vintagepc:
That is possible but does require further work to uartrxbuffer to support checking the level.

Turro75:
I found a working code for uart6, anyway unless this [https://github.com/prusa3d/Prusa-Firmware-Buddy/issues/1076](https://github.com/prusa3d/Prusa-Firmware-Buddy/issues/1076) won't be implemented serial interface is quite useless

vintagepc:
> ok no problem, in case You want to send me some bbf to test on a real device I'm here.
> 
> Best Regards
 I'm happy to announce that said project being private is no longer the case, if you are still interested :)

https://github.com/vintagepc/MINI404

OJFord:
Sorry, this is a tangential question, but it seems like you'd know (and is potentially related in the sense that if 'yes', I'm also interested in configuring it differently): is the serial port hard coded to the micro USB on the Mini, in the released 4.3.1?

I have an ESP8266 device connected and seemingly working (spewing gcode from Octoprint if I plug it into my computer instead of the Mini) but it has full-size USB type A only, and in that port of the Mini it can't connect.

I then noticed the Octoprint guide mentions the Mini's micro USB port (which I hadn't even noticed it had until then, heh.) but doesn't say why / if that's the only one supported? The guide also expects you to be connecting a RPi directly, for which a USB A -> micro would be more natural I suppose, so I'm not sure if it's just an arbitrary choice; may as well leave the bigger port free for flash storage if you can sort of thing.

I don't know (because I don't know the physical mapping of USART6?) if this is the same issue @Turro75 mentioned?

Cheers.

vintagepc:
If I'm understanding your question correctly, then no, the full-size USB port does not support anything other than USB drives. It explicitly expects a mass storage device (can't even connect a hub) unless you make significant alterations to the firmware to swap the ports (and I'm not even sure if that would work given it lacks the 5th OTG sense pin to determine if the connection is supposed to be host or device-mode (in other words, who "owns" the bus)

You might have some success with some variant of  this though to adapt your ESP until such time as the internal port becomes usable:
https://www.amazon.com/UGREEN-Adapter-Samsung-Controller-Android/dp/B00N9S9Z0G/ref=sr_1_4?dchild=1&keywords=usb+a+to+micro&qid=1624115593&sr=8-4

OJFord:
Great, thanks. Of course, I would've just checked it if I could've found an adapter like that, but wanted to confirm that was the issue before buying (and waiting for) one. Many thanks!

vintagepc:
No problem. Hopefully we'll see some ESP movement soon given the hints about PrusaConnect they've teased in recent livestreams :)

teotwaki:
@vintagepc Is this PR still relevant? If not, could it be closed?

vintagepc:
Closing, the changes are stale.

PR #1265, opened 2021-01-27
print sheet profiles optimization

marekmosna-prusa3d:
BFW-1713

### DIFF
--- a/src/gui/status_footer.cpp
+++ b/src/gui/status_footer.cpp
@@ -49,7 +49,10 @@ void status_footer_t::windowEvent(EventLock /*has private ctor*/, window_t *send
 
     if (dynamic_cast<screen_home_data_t *>(GetParent()) != nullptr //is home_screen
         && sheet_number_of_calibrated() > 1) {                     // calibrated more profiles than 1
-        update_sheet_profile();
+        if (mseconds - last_timer_repaint_profile >= REPAINT_PROFILE_PERIOD) {
+            update_sheet_profile();
+            last_timer_repaint_profile = mseconds;
+        }
     } else if (mseconds - last_timer_repaint_z_pos >= REPAINT_Z_POS_PERIOD) {
         update_z_axis();
         last_timer_repaint_z_pos = mseconds;
@@ -252,6 +255,7 @@ status_footer_t::status_footer_t(window_t *parent)
     , last_timer_repaint_values(0)
     , last_timer_repaint_colors(0)
     , last_timer_repaint_z_pos(0)
+    , last_timer_repaint_profile(0)
     , print_speed(0) /// print speed in percents
     //, nozzle_state;
     //, heatbed_state;
--- a/src/gui/status_footer.cpp
+++ b/src/gui/status_footer.cpp
@@ -48,11 +48,8 @@ void status_footer_t::windowEvent(EventLock /*has private ctor*/, window_t *send
     }
 
     if (dynamic_cast<screen_home_data_t *>(GetParent()) != nullptr //is home_screen
-        && sheet_number_of_calibrated() > 1) {                     // calibrated more profiles than 1
-        if (mseconds - last_timer_repaint_profile >= REPAINT_PROFILE_PERIOD) {
-            update_sheet_profile();
-            last_timer_repaint_profile = mseconds;
-        }
+        && no_of_calibrated_sheets() > 1) {                        // calibrated more profiles than 1
+        update_sheet_profile();
     } else if (mseconds - last_timer_repaint_z_pos >= REPAINT_Z_POS_PERIOD) {
         update_z_axis();
         last_timer_repaint_z_pos = mseconds;
@@ -259,7 +256,8 @@ status_footer_t::status_footer_t(window_t *parent)
     , print_speed(0) /// print speed in percents
     //, nozzle_state;
     //, heatbed_state;
-    , show_second_color(false) {
+    , show_second_color(false)
+    , cached_no_of_calibrated_sheets(0) {
 
     wt_nozzle.font = resource_font(IDR_FNT_SPECIAL);
     wt_nozzle.SetAlignment(ALIGN_CENTER);
@@ -287,7 +285,7 @@ status_footer_t::status_footer_t(window_t *parent)
     update_temperatures();
     update_feedrate();
     update_filament();
-    if (sheet_number_of_calibrated() > 1)
+    if (no_of_calibrated_sheets() > 1)
         update_sheet_profile();
     else
         update_z_axis();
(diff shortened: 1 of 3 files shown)


RadekVana (approved):

RadekVana (dismissed):

dragomirecky (commented):
/question/ What if you call `sheet_calibrate()` before `sheet_number_of_calibrated()`? Looks to me like in `sheet_calibrate()`, the `no_of_calibrated_sheets` gets incremented to 1 and  `sheet_number_of_calibrated()` is then going to return 1 even if you have had some other sheets calibrated already.

/question/ Can you please explain how does the sheet counting work in general? I am not sure I fully understand why the counting starts with value of 1 & from the index 1 (skipping index 0)
```
    if (!no_of_calibrated_sheets) {
        ++no_of_calibrated_sheets;
        for (int8_t i = 1; i < MAX_SHEETS; ++i) {
            if (sheet_is_calibrated(i))
                ++no_of_calibrated_sheets;
        }
    }
    return no_of_calibrated_sheets;
```

/thinking out loud/ Is this the right place to cache the number of sheets? Until now, all the eeprom functions could be considered slow as they load/unload something from an external peripheral. With this change, all the functions are slow **except** `sheet_number_of_calibrated`. We introduced (undocumented) exceptions – all the eeprom functions aren't being cached, except one. I don't think this is the right way to go. All the functions should be cached, or none and the caching should be done (if needed) by the caller.

marekmosna-prusa3d:
My apologies, but I think I don't fully understand your comment @dragomirecky . Can you please post specification where is described the working process with slow peripherals in current architecture? Can I find anywhere caching proposal? Because involve such a mechanism after RC was released, I consider as highly dangerous.  I'm going to implement it gladly regarding to such documents that you've mentioned as undocumented.

dragomirecky:
That is a misunderstanding – I am not proposing to implement a caching layer on top of eeprom. I am saying that adding a caching to _one_ of the eeprom functions is not conceptual. All the functions should be cached or none (or if there are some exceptions, they should be at least documented). In the case of this PR and quick fix of the high cpu usage, I think reading the value only when needed or caching it within GUI would be more appropriate. But this is just a note regarding the architecture, the two first questions are more important in the case of this fix.

marekmosna-prusa3d:
Well, let me explain it a bit.

> In the case of this PR and quick fix of the high cpu usage, I think reading the value only when needed or caching it within GUI would be more appropriate.
The are several unrelated places in GUI where the content depends on how many print sheets is configured already, so that's the main reason why I consider `no_of_calibrated_sheets` as the only necessary value to cache and place it in eeprom module scope as global variable. 

>  What if you call sheet_calibrate() before sheet_number_of_calibrated()?
This situation could not happen due to application architecture. 

> Can you please explain how does the sheet counting work in general?
The most important information is: **There is always at least 1 configured print sheet from the manufacture**. So after the startup the 0 indicates uninitialized cache variable and counting the current calibrated sheets. There is a possibility of unwanted behavior described as bug `BFW-1680`, but it's not the part of this PR

I hope it helps you out.

dragomirecky:
> > In the case of this PR and quick fix of the high cpu usage, I think reading the value only when needed or caching it within GUI would be more appropriate.
> 
> The are several unrelated places in GUI where the content depends on how many print sheets is configured already, so that's the main reason why I consider `no_of_calibrated_sheets` as the only necessary value to cache and place it in eeprom module scope as global variable.

No, there is only one place in the code base which requires this function to be fast (status_footer.cpp:51), which is the problem. 

> > What if you call sheet_calibrate() before sheet_number_of_calibrated()?
> 
> This situation could not happen due to application architecture.

„Application architecture“ here meaning „the firmware now calls sheet_number_of_calibrated before sheet_calibrate“. Just because the situation can't happen now doesn't mean the code is not broken.

> > Can you please explain how does the sheet counting work in general?
> 
> The most important information is: **There is always at least 1 configured print sheet from the manufacture**. So after the startup the 0 indicates uninitialized cache variable and counting the current calibrated sheets. There is a possibility of unwanted behavior described as bug `BFW-1680`, but it's not the part of this PR

I think I am unsure about the part how the „1 configured print sheet from manufacture“ is handled in code. Is that the one saved at position 0? And because we know `sheet_is_calibrated(0)` would return true, we just increment the variable by one and skip the first index? Is that it?

marekmosna-prusa3d:
I apologize, but I get confused again. Can you please specify in details the requirements for BFW-1713 to make it clear? Till than there is no specification it's obvious that I cannot meet the expectations. 
Thank you

dragomirecky:
Ok, no problem. 

_The Issue_
When debugging why PUT does not work I found out that the reason is a very high CPU usage (usually 100 %). Then I found out that it is caused by a call to `sheet_number_of_calibrated()`(status_footer.cpp:51), which was recently added to a method that is being called dozen of times per second. The function `sheet_number_of_calibrated()` – as any other from the eeprom module – talks to the external eeprom, making the call quite expensive. This leads to the high cpu usage and needs to be fixed.

_The Implementation_
I am sorry if some of the above wasn't told to you. I think the issue itself is fixed by this PR (and in case it works can be considered a good hotfix). However we will want to merge this into master and that is why I am concerned about the quality of the solution.
- It introduces an issue to the eeprom module: If you don't call sheet_calibrate() and sheet_number_of_calibrated() in the right order, the module ends up in an invalid state
- It adds a relatively lot of code into the eeprom module to make the sheet_number_of_calibrated function cached (while all the others stay slow), instead of dealing with this issue at the (single!) call site

marekmosna-prusa3d:
Well. So I assume that the solution you're expecting will not care about other calls of `sheet_number_of_calibrated` and the "cache" will be expected only inside the class `status_footer`. Please correct me when I'm wrong. Underneath the new light I can redesign current PR.

dragomirecky:
Yes, as far as I know, this is the only problematic call site.

Btw, It doesn't have to be a cache. Maybe you will find a better solution (like updating this part of the display only when truly necessary).

marekmosna-prusa3d:
I walked across the code again and I unfortunately did not find the proper way how to accomplish your requirements. There is no mechanism how to pass changed state of print sheet profiles from any other screen, and till now, the entire print sheet api was stateless. If you have any idea how to meet your requirements please let me know. I will appreciate it the most.

src/gui/status_footer.cpp
```diff
@@ -49,7 +49,10 @@ void status_footer_t::windowEvent(EventLock /*has private ctor*/, window_t *send
 
     if (dynamic_cast<screen_home_data_t *>(GetParent()) != nullptr //is home_screen
         && sheet_number_of_calibrated() > 1) {                     // calibrated more profiles than 1
-        update_sheet_profile();
+        if (mseconds - last_timer_repaint_profile >= REPAINT_PROFILE_PERIOD) {
```
DRracer, line 52: OMG, this is brutal - that means, we are repainting the footer too fast ... so basically any form of slowing-down like you just did for the sheet will be benefical.

dragomirecky (approved):
Perfect, let's merge this 👍

PR #1270, opened 2021-01-30
make color_t enum class

RadekVana:
Main motivation is to be able to write `color_t::` and just select color with arrows.
Also removed `color_t` from `MockDisplay`, because it emulates st7789v display which is written in C (does not know `color_t`)

BFW-1718


DRracer (changes requested):
No reasonable commit/PR message - I'm missing the explanation of motivation for this change that affects 43 source code files (carpet bombing again without too much added value).
The new code look correct, however changes in unit tests (color_t -> uint32_t) look unnecessary (like return to previous untyped state).

RadekVana:
Mock display used in unit tests emulates st7789v display which is written in C.
And it "does not know" color_t.
But I can add conversion methods into Mock display to simplify unit tests.

danopernis:
Closing this, but feel free to reopen after fixing conflicts. Also note that we have new contribution guidelines regarding enums https://github.com/prusa3d/Prusa-Firmware-Buddy/blob/master/doc/contributing.md#code-style-cc-naming-conventions

PR #1271, opened 2021-01-30
disable safety timer in pause

RadekVana:
BFW-1650


src/marlin_stubs/feature/safety_timer.cpp
```diff
@@ -63,9 +63,13 @@ SafetyTimer::expired_t SafetyTimer::Loop() {
         pBoundPause->NotifyExpiredFromSafetyTimer(thermalManager.degTargetHotend(0), thermalManager.degTargetBed());
     }
     if (printingIsPaused()) {
+        last_reset = now;
```
DRracer, line 66: a brutal hack, but probably solves the problem

DRracer (approved):

PR #1273, opened 2021-01-31
Add support for STM32F429VI - to be able to debug without optimization

RadekVana:


dragomirecky (commented):
Thanks for adding support for this MCU! One thought regarding the way we tell CMake to use the STM32F429VI:

Currently, the PR adds a non-standard `CMAKE_BUILD_TYPE=NoOpt` option, which
1. switches the compilation from -Og to -O0, and
2. switches the MCU (linker script) to STM32F429VI

I see two problems here:
1. The `NoOpt` build type is essentially the same as `Debug`; however, by introducing this new non-standard build type, it will become harder to maintain the CMake code. Prove of this point is this PR itself – for example, you missed [this condition](https://github.com/prusa3d/Prusa-Firmware-Buddy/blob/ccb52ecf462f63688f6c69356d0f191a0d2d126d/CMakeLists.txt#L303) which would need to be changed to handle the new `NoOpt` build type. 
2. Just the fact that switching the build type to a "No Optimizations" one magically changes the MCU to another seems wrong. What if I want to turn off some flash-demanding features and use the -O0 on the standard MCU? I can't.

/suggestion/
Let's make the semantics clearer. We want to add support for a new MCU (as this PR's title says). So let's add a new cache variable (cache variables are "cmake options," e.g., our `MOTHERBOARD` or `GENERATE_BBF`) named `MCU`. Possible values for this `MCU` variable would be STM32F407VG and STM32F429VI (the former being the default). The linker script would change based on this variable.

Then, if `CMAKE_BUILD_TYPE=Debug` and `MCU=STM32F429VI`, we can automatically switch to `-O0` as we know we sure have enough of FLASH. 

Also, adding the `MCU` option might be useful in the future in case we decide to support even more MCUs :)

CMakeLists.txt
```diff
@@ -271,10 +279,14 @@ add_compile_options(-DconfigUSE_NEWLIB_REENTRANT=1)
...
-  add_compile_options(-Og)
+if(MCU STREQUAL "STM32F429VI")
+  add_compile_options(-O0)
 else()
-  add_compile_options(-Os)
+  if(CMAKE_BUILD_TYPE STREQUAL "Debug")
+    add_compile_options(-Og)
+  else()
+    add_compile_options(-Os)
+  endif()
```
dragomirecky, line 289: /question/ Is this intentional? I would expect that a configuration with MCU=STM32F429VI and CMAKE_BUILD_TYPE=Release would produce a binary with `-Os` (as that is usually expected from a release configuration) and `-O0` would be used for that MCU only when CMAKE_BUILD_TYPE is Debug. In other words, something like

```
if (build type is Debug)
  -O0 if mcu is F429VI else -Og
else
  -Os
```

dragomirecky (approved):
Looks great 👍

PR #1277, opened 2021-02-01
Remove unnecessary formatting from texts/translations

DRracer:
Matches the new functionality of correct text wrapper and should improve the user experience of rendered text. However, some typographical rules might have been broken since there is no stable support for nonbreakable spaces yet.

Additionally, some more had to be done:
- fix CZ formatting string
- adjust formatting buffer size for first layer calibration dialog


dragomirecky (approved):
👍

PR #1278, opened 2021-02-01
Fix high CPU usage in WUI task

dragomirecky:
BFW-1684


DRracer (approved):
approving to unblock PUT

PR #1290, opened 2021-02-05
fix bug in safety timer while printing

RadekVana:
fix bug in safety timer while printing
fix not opening safety timer in "insert filament" state of load sequence


DRracer (approved):
approving based on successful feature tests

PR #1304, opened 2021-02-13
inherit filament sensor from abstract base class

RadekVana:
inherit filament sensor from abstract base class
to be able to run unit tests easier and avoid copying of code in private


Hagridpon3 (approved):
Looks good for my

PR #1338, opened 2021-03-19
timed dialog (automatic show after X ms)

RadekVana:
BFW-1799

### DIFF
--- a/src/gui/dialogs/DialogTimed.hpp
+++ b/src/gui/dialogs/DialogTimed.hpp
@@ -1,0 +1,32 @@
+/**
+ * @file DialogTimed.hpp
+ * @author Radek Vana
+ * @brief Base of dialog which opens after period without knob event and closes on knob event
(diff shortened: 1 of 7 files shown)


src/gui/dialogs/DialogTimed.hpp
```diff
@@ -0,0 +1,32 @@
...
+ * @file DialogTimed.hpp
+ * @author Radek Vana
+ * @brief Base of dialog which opens after period without knob event and closes on knob event
+ * @date 2021-03-05
+ */
+#pragma once
+
+#include "IDialog.hpp"
+
+class TimedDialog : public AddSuperWindow<IDialog> {
```
dragomirecky, line 11: Shouldn't this be „DialogTimed“ to have the same name as the file (and as other dialogs)?

dragomirecky (dismissed):

dragomirecky (approved):

PR #1357, opened 2021-04-07
Fix executing gcode inside long comments

milosfec:
When line was langer than 96 chars, remaining part of the line was
parsed as new line.
If the line included long comment and after 96 chars there was a
gcode, that gcode was executed.

Throw away rest of the line to avoid the issue.

Solves BFW-1264

### DIFF
--- a/src/common/media.cpp
+++ b/src/common/media.cpp
@@ -236,6 +236,15 @@ float media_print_get_percent_done(void) {
     return 0;
 }
 
+void media_no_data() {
+    if (f_eof(&media_print_fil)) //we need check eof also after read operation
+        media_print_stop();      //stop on eof
+    else {
+        set_warning(WarningType::USBFlashDiskError);
+        media_print_pause(); //pause in other case (read error - media removed)
+    }
+}
+
 void media_loop(void) {
     _usbhost_reenum();
     if (media_print_state == media_print_state_PRINTING) {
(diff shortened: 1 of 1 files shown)


src/common/media.cpp
```diff
@@ -272,6 +274,11 @@ void media_loop(void) {
                     }
                     break;
                 }
+                while (len == MAX_CMD_SIZE && buffer[MAX_CMD_SIZE - 1] != '\n') {
+                    // Read until the end of the line
+                    f_gets(buffer, sizeof(buffer), &media_print_fil);
```
dragomirecky, line 279: /nitpick/ In the case that this `f_gets` call starts to fail, we might end up in an infinite cycle. Right?

milosfec: Good point, thanks

src/common/media.cpp
```diff
@@ -272,6 +275,13 @@ void media_loop(void) {
                     }
                     break;
                 }
+                while (len == MAX_CMD_SIZE && buffer[MAX_CMD_SIZE - 1] != '\n') {
+                    // Read until the end of the line
+                    if (!f_gets(buffer, sizeof(buffer), &media_print_fil)) {
```
DRracer, line 280: And what will be the state of `media_current_position` when the line is too long? Could you please add some unit tests so that we can verify these potential edge cases?

src/common/media.cpp
```diff
@@ -272,6 +275,13 @@ void media_loop(void) {
...
+                        goto no_data;
```
espr14, line 281: I know this is already an if-hell suitable for refactor. However, I would avoid `goto`. You can create `media_no_data()`. Then you'd place here:
```
media_no_data();
return;
```
and this instead of `no_data:` block:
```
} else {
    media_no_data();
    return;
}
```

milosfec:
Replaced by https://github.com/prusa3d/Prusa-Firmware-Buddy/pull/1372

PR #1363, opened 2021-04-09
operator == for string view

RadekVana:


src/lang/string_view_utf8.hpp
```diff
@@ -221,4 +221,27 @@ class string_view_utf8 {
         s.type = EType::NULLSTR;
         return s;
     }
+
+    /// string view has same resource
+    constexpr bool operator==(const string_view_utf8 &other) const {
+        if (type != other.type)
+            return false; // type missmatch
```
DRracer, line 228: mismatch (one 's')

src/lang/string_view_utf8.hpp
```diff
@@ -221,4 +221,27 @@ class string_view_utf8 {
         s.type = EType::NULLSTR;
         return s;
     }
+
+    /// string view has same resource
```
DRracer, line 225: has THE same resource

src/lang/string_view_utf8.hpp
```diff
@@ -221,4 +221,27 @@ class string_view_utf8 {
...
+        if (type != other.type)
+            return false; // type missmatch
+
+        switch (type) {
+        case EType::RAM:
+        case EType::CPUFLASH:
+            return attrs.cpuflash.utf8raw == other.attrs.cpuflash.utf8raw;
+        case EType::FILE:
+            return (attrs.file.f == other.attrs.file.f) && (attrs.file.startOfs == other.attrs.file.startOfs);
+        case EType::SPIFLASH:
```
DRracer, line 236: SPIFLASH and USBFLASH types are equal by default as well? That's not true

DRracer (changes requested):
please fix:
- some typos in comments
- comparison of usb and spiflash string views (even though there are no string views like this yet)

src/lang/string_view_utf8.hpp
RadekVana: added assert

src/lang/string_view_utf8.hpp
RadekVana: fixed

src/lang/string_view_utf8.hpp
RadekVana: fixed

DRracer (approved):

PR #1366, opened 2021-04-12
Fix executing gcode inside long comments

milosfec:
When line was langer than 96 chars, remaining part of the line was
parsed as new line.
If the line included long comment and after 96 chars there was a
gcode, that gcode was executed.

Throw away rest of the line to avoid the issue.

Solves BFW-1264


milosfec:
Replaced by https://github.com/prusa3d/Prusa-Firmware-Buddy/pull/1372

PR #1368, opened 2021-04-16
FW uploader + GUI cmds

Chleba:
BFW-1815


Chleba:
Old implementation via library lwESP using DMA UART with GUI. Not finished, closing because second implementation using serial flasher from espressif. See - #1369 for more info.

PR #1369, opened 2021-04-18
Serial flasher

Chleba:
BFW-1815, BFW-1812


src/gui/screen_menu_esp_update.cpp
```diff
@@ -0,0 +1,332 @@
...
+#include "dbg.h"
+#include "string.h"
+
+#include "cmsis_os.h"
+#include "stm32f4xx_hal.h"
+#include "stm32_port.h"
+#include "esp_loader.h"
+// #include "example_common.h"
+#include "esp_flasher.h"
+#define HIGHER_BAUDRATE 115520
```
DRracer, line 20: Isn't this a typo? The standard baudrate is 115200

src/gui/screen_menu_esp_update.cpp
```diff
@@ -0,0 +1,332 @@
...
+    char tmp_str1[] = "- ESP not connected\n";
+    WriteStr(tmp_str1);
+}
+void ESP_FLASHER::Connect() {
+    char tmp_str[ESP_DESC_SIZE];
+    if (connect_to_target() == ESP_LOADER_SUCCESS) {
+        // msg = Msg::ConnectOK;
+        esp_connected = true;
+        char tmp_str1[] = "- ESP Connected\n";
+        strcpy(tmp_str, tmp_str1);
```
DRracer, line 60: Why do you copy the string? Why don't you just swap pointers around?

src/gui/screen_menu_esp_update.cpp
```diff
@@ -0,0 +1,332 @@
...
+        i++;
+        d = str[i];
+    }
+    return lines;
+}
+void ESP_FLASHER::WriteStr(char *str) {
+    memset(plan_str, 0, ESP_DESC_SIZE);
+    strcpy(plan_str, str);
+
+    /*     char tmp_str1[ESP_DESC_SIZE]; */
```
DRracer, line 132: Does this commented code have some meaning (e.g. for future improvements)? If not it should be removed from the PR

src/main.cpp
```diff
@@ -190,13 +190,15 @@ char uart6slave_line[32];
 static volatile uint32_t minda_falling_edges = 0;
 uint32_t get_Z_probe_endstop_hits() { return minda_falling_edges; }
 
+static uint8_t is_flashing = 0;
```
DRracer, line 193: I assume this is a very brutal piece of development code ... is_flashing variable should be probably hidden somewhere in a more appropriate place related to the code being used in.

DRracer (changes requested):
- lib/WUI/resources/fsdata_wui_local.c disappeared and a .c.bak is being added - looks like an unintended accidental change
- as a prototype this PR is probably enough, my main complaint is about the flashing being done in the UI thread. This cannot be merged into the FW in its current state.

src/gui/screen_menu_esp_update.cpp
Chleba: It's not used any more. It will be deleted.

marekmosna-prusa3d:
development moved due to @Chleba relocation

PR #1380, opened 2021-04-26
serial flash provided by lib from espressif

marekmosna-prusa3d:
1st prototype of flash ESP from the fw
- it is bypassing entire lwesp
- changes of the origin sources are placed beneath the prusa folder
- the port for UART communication has to be changed to manage flash mode beside ordinary run

BFW-1815


vladamatena:
With files from this package: https://www.espressif.com/sites/default/files/ap/ESP8266_NonOS_AT_Bin_V1.7.4.zip on the USB drive it flashed the ESP. The ESP does not seem to work afterwards. Did I used wrong firmware files?

marekmosna-prusa3d:
I'm using the same ones. 
You must be very careful to use it. There is no proper GUI yet and you maybe interrupt the flashing process. Currently I only use SWO debug output to see the progress and afterward I even did not test the functionality of the ESP itself. I gonna test the validity of the flash today. I was really happy that I managed the UART communication beside lwesp library.

vladamatena:
> With files from this package: https://www.espressif.com/sites/default/files/ap/ESP8266_NonOS_AT_Bin_V1.7.4.zip on the USB drive it flashed the ESP. The ESP does not seem to work afterwards. Did I used wrong firmware files?

Sorry, this seems false positive. I didn't managed to fix the ESP yet, but the flash drive I used seems to have corrupted content:

ESP read data 313328
ESP flash write FAIL
ESP read data 314352
ESP read data 315376

vladamatena (approved):

marekmosna-prusa3d:
As you've mentioned the packet loss over UART I think this is the case...I made several runs and sometimes it's without any errors and sometimes with. We have to balance it properly.

joshymjose (changes requested):
There are changes which are not supposed to be in the WUI5. I suggest to add these are tests

marekmosna-prusa3d:
You are right @joshymjose . Entire ESP flash doesn't suppose to be part of the WUI.

PR #1381, opened 2021-04-27
footer settings menu

RadekVana:
Add menu to modify footer settings
BFW-1807


src/gui/footer/footer_item_union.hpp
```diff
@@ -27,4 +27,12 @@ enum class items {
     count_
 };
 
+constexpr void *EncodeItemForEvent(items item) {
+    return (void *)int(item);
```
marekmosna-prusa3d, line 31: can you please use c++ casting? I guess reinterpret_cast is necessary for this

src/gui/footer/footer_items_heaters.cpp
```diff
@@ -27,11 +36,11 @@ string_view_utf8 FooterItemHeater::static_makeViewIntoBuff(int value, std::array
     const uint current = std::clamp(int(temps.current), 0, 999);
     const uint target_or_display = std::clamp(int(temps.target_or_display), 0, 999);
 
-    int printed_chars = snprintf(buff.data(), buff.size(), left_aligned ? left_aligned_str : const_size_str, current, target_or_display);
+    int printed_chars = snprintf(buff.data(), buff.size(), draw_type == footer::ItemDrawType::Static ? const_size_str : left_aligned_str, current, target_or_display);
```
marekmosna-prusa3d, line 39: Be very careful about such a construct. Please close the ternary operator into the parenthesis. I don't even think that it works as you expect.

src/gui/footer/footer_item_union.hpp
RadekVana: ok,
I used C-style since using void* to pass "anything" is "ugly C approach"

marekmosna-prusa3d (changes requested):

src/gui/footer/footer_items_heaters.cpp
RadekVana: it does what I meant, but I agree parenthesis are more readable

marekmosna-prusa3d (approved):

PR #1384, opened 2021-04-28
Runtime switching between flash mode and operating mode

marekmosna-prusa3d:
BFW-1873


joshymjose (approved):

vladamatena (approved):
UartBufferThread runs out of memory, seems ok after sparing some RAM elsewhere.

PR #1388, opened 2021-04-29
Add filament runout for serial printing

RadekVana:
BFW-1883

### DIFF
--- a/src/common/non_file_printing_counter.cpp
+++ b/src/common/non_file_printing_counter.cpp
@@ -1,0 +1,26 @@
+/**
+ * @file non_file_printing_counter.cpp
+ * @author Radek Vana
+ * @date 2021-04-29
+ */
+
+#include "non_file_printing_counter.hpp"
+#include "rtos_api.hpp"
+
+unsigned NonFilePrintingCounter::counter = 0;
+
+void NonFilePrintingCounter::increment() {
+    CriticalSection C;
+    ++counter;
+}
+
+void NonFilePrintingCounter::decrement() {
+    CriticalSection C;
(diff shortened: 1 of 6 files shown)


src/common/non_file_printing_counter.cpp
```diff
@@ -0,0 +1,26 @@
...
+ * @date 2021-04-29
+ */
+
+#include "non_file_printing_counter.hpp"
+#include "rtos_api.hpp"
+
+unsigned NonFilePrintingCounter::counter = 0;
+
+void NonFilePrintingCounter::increment() {
+    CriticalSection C;
```
DRracer, line 13: Is the critical section really necessary? How about std::atomic? The MCU should have atomic instruction support.

src/common/non_file_printing_counter.hpp
```diff
@@ -0,0 +1,19 @@
...
+ * @file non_file_printing_counter.hpp
+ * @author Radek Vana
+ * @brief
+ * @date 2021-04-29
+ */
+
+#include <cstdint>
+
+class NonFilePrintingCounter {
+    static unsigned counter;
```
DRracer, line 11: This looks like a singleton, but that probably wasn't the intent. From the usage of this class it looks more like a guard combined with the guard variable into one piece. Looks overengineered to me.

DRracer (approved):
approving, but there are a few comments worth solving in the future

src/common/non_file_printing_counter.hpp
RadekVana: this is not singleton

PR #1397, opened 2021-05-05
Walkaround ESP UART DMA restart

vladamatena:
This fixes DMA just filling the buffer with some data and then stop
receiving. This is a dirty patch but lets the lwESP to initialize itself.


vladamatena:
Walkaround already integrated, lets close this one.

PR #1401, opened 2021-05-10
Esp 1.6.0

marekmosna-prusa3d:
* supporting esp v1.6
* entire lwesp folder isn't cloned as subrepo


vladamatena (approved):
- I can confirm this works both netconn TCP client and server working in wifi STA mode. 
- I had issues with stack overflow. Fixed by `osThreadDef(UartBufferThread, StartUartBufferThread, osPriorityNormal, 0, 512);`
- Connection to the AP takes quite some time `No access points available with preferred SSID` is printed ~ 30 times before it actually connects.
- Looks like we can push ~10KiB/s (this seems to be limitation of 115200 baud rate)
- Sending lots of data to the open server connection from the client side seems to break the communication. `cat /dev/urandom | pv | nc 10.42.0.137 5000` breaks but this works `while true; do echo -e "data data data \n"; sleep 0.01;  done | pv | nc 10.42.0.137 5000`

PR #1404, opened 2021-05-11
v4.3.2: Fix CDC BSOD and refactor the CDC interface

dragomirecky:
Originally, my intention today was to fix the CDC BSOD only. However, when I dug into the implementation we used, it was a 🤯 moment.

The STM's generated implementation, which our version was based on, was super sketchy. For example, it defines excessively large buffers which it doesn't use. Another issue is that sending data over the CDC interface could result in data corruption. Considering that almost every line of the implementation had some issue, I decided to create a new custom implementation.

1. It fixes the BSOD. It was caused by the update of the STM32 HAL Library in commit 5346d4ba73e0261. As mentioned in release notes, a new member to the `USBD_CDC_ItfTypeDef` structure was introduced, but our code did not define it.
2. It saves more than 3 kilobytes of RAM.
3. It fixes possible corruption of data being sent over the CDC interface.
4. It provides a new cleaner implementation of the `usbd_cdc_if.h/c`. However, this PR does not aim to cleanup implementation of the USBSerial class or other related modules.

### DIFF
--- a/include/usb_device/usb_device.h
+++ b/include/usb_device/usb_device.h
@@ -90,6 +90,7 @@ extern "C" {
 /** USB Device initialization function. */
 void MX_USB_DEVICE_Init(void);
 
+extern USBD_HandleTypeDef hUsbDeviceFS;
 /**
   * @}
   */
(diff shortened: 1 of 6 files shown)


DRracer (approved):
The proposed changes make sense to me, approving

PR #1411, opened 2021-05-12
Simplify mapping Marlin pins to hwio.

mkbel:
After this change, it is not needed to enumerate all ordinary IO pins
in hwio_a3ides_2209_02.cpp anymore.
Runtime sanity checks are preserved in DEBUG build.

Unrelated polishing added to this PR.

### DIFF
--- a/src/common/hwio_a3ides_2209_02.cpp
+++ b/src/common/hwio_a3ides_2209_02.cpp
@@ -643,11 +643,6 @@ void digitalWrite(uint32_t marlinPin, uint32_t ulVal) {
     }
 }
 
-void digitalToggle(uint32_t ulPin) {
-    digitalWrite(ulPin, !digitalRead(ulPin));
-    // TODO test me
-}
-
 uint32_t analogRead(uint32_t ulPin) {
     if (HAL_ADC_Initialized) {
         switch (ulPin) {
--- a/src/common/hwio_a3ides_2209_02.cpp
+++ b/src/common/hwio_a3ides_2209_02.cpp
@@ -1,5 +1,7 @@
-//----------------------------------------------------------------------------//
-// hwio_a3ides.c - hardware input output abstraction for a3ides board
+/**
+ * @file
+ * @brief hardware input output abstraction for Buddy board
+ */
 
 #include <inttypes.h>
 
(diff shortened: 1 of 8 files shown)


src/common/hwio_a3ides_2209_02.cpp
```diff
@@ -550,80 +564,89 @@ void hwio_arduino_error(int err, uint32_t pin32) {
...
+        return;
+    default:
+#if _DEBUG
+        if (buddy::hw::isOutputPin(marlinPin)) {
+#endif //_DEBUG
+            gpio_set(marlinPin, ulVal ? 1 : 0);
+#if _DEBUG
+        } else {
+            hwio_arduino_error(HWIO_ERR_UNDEF_DIG_WR, marlinPin); //error: undefined pin digital write
         }
```
dragomirecky, line 647: /suggestion/ what about a small change to make it more readable?

```
#if _DEBUG
    if (!isOutputPin)
        hwio_arduino_error(...);
#endif
gpio_set(...)
```

src/common/hwio_a3ides_2209_02.cpp
```diff
@@ -550,80 +564,89 @@ void hwio_arduino_error(int err, uint32_t pin32) {
...
+#if _DEBUG
+        } else {
+            hwio_arduino_error(HWIO_ERR_UNDEF_DIG_WR, marlinPin); //error: undefined pin digital write
+            return 0;
         }
-    } else
-        hwio_arduino_error(HWIO_ERR_UNINI_DIG_RD, ulPin); //error: uninitialized digital read
-    return 0;
+#endif //_DEBUG
+    }
```
dragomirecky, line 597: 🤔

what about something like

```C
#if _DEBUG
if (!HAL_GPIO_Initialized)
    error
if (!physicalPinExist(pin))
   error
#endif

return gpio_get(pin)
```

src/common/hwio_a3ides_2209_02.cpp
mkbel: Fixed as suggested.

src/common/hwio_a3ides_2209_02.cpp
mkbel: Partially fixed as suggested.

dragomirecky (approved):

PR #1414, opened 2021-05-13
Lwesp ll dma issue

joshymjose:
Test DMA buffer for UART RX
Uses the existing uartrxbuff implementation
Cmake cache variable "WUI_ESP_DMA_TEST" enables test
PUT functionality disabled when ESP is enabled
Ping test (www.google.com) with custom AT command parsing instead lwESP

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -87,6 +87,10 @@ set(WUI
     "YES"
     CACHE BOOL "Enable Web User Interface"
     )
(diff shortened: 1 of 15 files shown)


joshymjose:
The change of the lwesp to older version make this PR unusable. But the test is still valid. If you guys need just resolve the conflicts and continue. No plans from me to fix this ;,)

PR #1447, opened 2021-05-31
bsod: Fix task name not being presented on stack overflow

dragomirecky:
Cherry-picking #1295 to 4.3.2

### DIFF
--- a/src/common/bsod.cpp
+++ b/src/common/bsod.cpp
@@ -484,7 +484,7 @@ static signed char *tsk_name = 0;
 extern "C" void vApplicationStackOverflowHook(TaskHandle_t xTask, signed char *pcTaskName) {
     tsk_hndl = xTask;
     tsk_name = pcTaskName;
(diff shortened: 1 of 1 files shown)


src/common/bsod.cpp
```diff
@@ -484,7 +484,7 @@ static signed char *tsk_name = 0;
 extern "C" void vApplicationStackOverflowHook(TaskHandle_t xTask, signed char *pcTaskName) {
     tsk_hndl = xTask;
     tsk_name = pcTaskName;
-    if (pcTaskName != nullptr && strlen((const char *)pcTaskName) > 20)
+    if (pcTaskName != nullptr && strlen((const char *)pcTaskName) < 20)
```
DRracer, line 487: I'd recommend strnlen to avoid buffer overrun in case of pcTaskName pointing to something bad

DRracer (approved):

PR #1468, opened 2021-06-14
Add /api/job endpoint

marekmosna-prusa3d:
BFW-1950


lib/WUI/wui_REST_api.c
```diff
@@ -12,82 +12,169 @@
...
+        "\"job\":{"
+        "\"estimatedPrintTime\":%ld,"
+        "\"file\":{"
+        "\"date\":null,"
+        "\"name\":\"%s\","
+        "\"origin\":\"USB\","
+        "\"path\":\"%s\","
+        "\"size\":%ld"
+        "}"
+        "},"
```
dragomirecky, line 159: /suggestion/ Use proper tool for json serialization. It is only a matter of time until this introduces a nasty bug/security issue.

It might be actually very easy – the only issue I am aware of is string serialization (`%s`), where we need to escape `"` characters. As we have our own `[vsn]printf()` implementation (see `printf.c`), we can easily add a new one implementing the escaping (something line `jsonprintf(...)`).

lib/WUI/http/httpd_structs.h
```diff
@@ -28,7 +28,9 @@ static const char *const g_psHTTPHeaderStrings[] = {
     "Connection: keep-alive\r\nContent-Length: ",
     "Server: " HTTPD_SERVER_AGENT "\r\n",
     "\r\n<html><body><h2>404: The requested file cannot be found.</h2></body></html>\r\n",
-    "HTTP/1.0 500 Internal Server Error\r\n"
+    "HTTP/1.0 500 Internal Server Error\r\n",
+    "HTTP/1.0 401 Unauthorized\r\nWWW-Authenticate: ApiKey realm=\"401\"",
```
dragomirecky, line 32: /question/ Shouldn't `WWW-Authenticate: ApiKey realm=` be handled separately from the standard 401 status code?

dragomirecky (changes requested):

dragomirecky:
Btw, the PRs #1467 and #1466 seem to be redundant to this one or am I missing something?

lib/WUI/wui_REST_api.c
marekmosna-prusa3d: Implement function as you proposed is not part of this PR. There must be specified what we want/need from such a method and at least set of unit tests for it. It should be separate task and unfortunately there isn't any method to use to compose json response.

lib/WUI/http/httpd_structs.h
marekmosna-prusa3d: The possibility of the deployed web server are significantly limited so there must be a choice how to balanced new functionality with complexity of rework/rewrite of the existing web server.

XPila (approved):

PR #1481, opened 2021-06-20
Update metrics to use microseconds resolution for timestamps

dragomirecky:


teotwaki:
@dragomirecky I don't think this can be applied as such. Do you think this is something you will pick up again, or should we close the PR? If the latter, can I ask you to do so?

PR #1487, opened 2021-06-28
Bfw 1939 ini file changes

joshymjose:
Major changes
1. ini file read and parsed from file (before the whole file was copied to a char array and parsed)
2. There is a limit on the line length (now set to 200 chars) and data after this limit on any line is ignored
3. Added new EEPROM read and write functions for manipulating with string entries
4. The original inhi library was modified (some static functions made public)

### DIFF
--- a/src/common/ini_handler.c
+++ b/src/common/ini_handler.c
@@ -97,7 +97,7 @@ buddy_ini_error ini_save_file(const char *ini_save_str) {
 buddy_ini_error ini_load_file(void *user) {
     FIL fileObject;
     buddy_ini_error error = BUDDY_INI_OK;
-    char line[BUDDY_INI_LINE_SIZE];
+    char line[INI_MAX_LINE];
     char *start;
     char *end;
     char *name;
@@ -117,7 +117,7 @@ buddy_ini_error ini_load_file(void *user) {
 
     /* Scan through stream line by line */
     while ((f_eof(&fileObject) == 0) && (f_error(&fileObject) == 0)) {
-        if (NULL == f_gets(line, BUDDY_INI_LINE_SIZE, &fileObject)) {
+        if (NULL == f_gets(line, INI_MAX_LINE, &fileObject)) {
             f_close(&fileObject);
             return BUDDY_INI_FILE_READ;
         }
--- a/src/common/ini_handler.c
+++ b/src/common/ini_handler.c
@@ -20,10 +20,10 @@ static int ini_handler_func(void *user, const char *section, const char *name, c
     ETH_config_t *tmp_config = (ETH_config_t *)user;
 
     if (ini_string_match(section, "eth::ipv4", name, "type")) {
-        if (strncmp(value, "DHCP", 4) == 0 || strncmp(value, "dhcp", 4) == 0) {
+        if (strncmp(value, "DHCP", 4) == 0) {
             CHANGE_LAN_TO_DHCP(tmp_config->lan.flag);
             tmp_config->var_mask |= ETHVAR_MSK(ETHVAR_LAN_FLAGS);
-        } else if (strncmp(value, "STATIC", 6) == 0 || strncmp(value, "static", 6) == 0) {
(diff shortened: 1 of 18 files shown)


ondratu (approved):

src/common/ini_handler.c
```diff
@@ -18,44 +19,72 @@ static int ini_handler_func(void *user, const char *section, const char *name, c
 
     ETH_config_t *tmp_config = (ETH_config_t *)user;
 
-    if (ini_string_match(section, "lan_ip4", name, "type")) {
-        if (strncasecmp(value, "DHCP", 4) == 0) {
+    if (ini_string_match(section, "eth::ipv4", name, "type")) {
+        if (strncmp(value, "DHCP", 4) == 0 || strncmp(value, "dhcp", 4) == 0) {
```
DRracer, line 23: `dhcp` does not correspond to description of task BFW-1939 - beware upper/lower case, this has always been a major source of confusion while using these INI files. I'd recommend doing `stricmp` to making the configuration less pain/error prone

The other keywords here are dtto.

src/common/ini_handler.c
```diff
@@ -65,21 +94,136 @@ uint8_t ini_save_file(const char *ini_save_str) {
...
+        }
+        // check if the line was completely read
+        if (NULL == strchr(line, '\n')) {
+            skip_next_read = 1;
+        }
 
-    if (file_init || file_read || file_close) {
-        return 0;
-    }
+#if INI_ALLOW_REALLOC && !INI_USE_STACK
```
DRracer, line 134: Was this piece of code copied from the `inih` library? There are multiple unnecessary #defines like INI_ALLOW_REALLOC and INI_USE_STACK. 

Of course we want to avoid allocating dynamic memory and yes, we'd like to do the stuff on stack only.

DRracer (changes requested):

:+1: for updating also all the language files

src/common/ini_handler.c
joshymjose: understand. From discussion with @ondratu we keep such "constants" upper case

src/common/ini_handler.c
joshymjose: please have a look at the eeprom.c/.h file changes, the existing variant based function had issue with reading writing string to EEPROM

src/common/ini_handler.c
DRracer: BFW-1939 will be updated accordingly?
What will be the backwards compatibility with existing ini files?

src/common/ini_handler.c
DRracer: When you copied a piece of code from the INIH library (which is weird and suspicious on its own), please prune the meaningless compile-time options.

src/common/ini_handler.c
joshymjose: example ini files were always with upper case. if to keep complete backward compatibility both upper and lower case must be kept

src/common/ini_handler.c
joshymjose: I had strategy to keep the whole thing which worked, if you remove something it could have some influence.

DRracer (changes requested):
Please note we now have unified filesystem (so FILE * and fopen work as intended)
Please remove all direct calls to FATfs's functions.

Moreover, the INIH library has a FILE* loading capability, so the code may get pruned nicely.

joshymjose:
with reference to the current master branch, I try to use these functions but fopen does not open file,
I do not see it is used anywhere else (example), even the system call "open_r" is not invoked

joshymjose:
There is a new PR https://github.com/prusa3d/Prusa-Firmware-Buddy/pull/1530
It was more effective and clean to create a new PR because
1. Better fix for EEPROM issue is on the latest master
2. use of POSIX function fopen() enables us to use inih lib functions

joshymjose:
There is a new PR #1530
It was more effective and clean to create a new PR because

PR #1493, opened 2021-07-02
Fix menu item naming

marekmosna-prusa3d:
To use such a setting screen on only for LAN but also for WIFI the more relevant name will be "Network Settings"

BFW-1977


DRracer (changes requested):
Please update the language translation files accordingly

danopernis:
Menu has been already reworked several times since this PR, closing.

PR #1496, opened 2021-07-02
Fix startup config synchronization

marekmosna-prusa3d:
Synchronize network setting only in case that networking is enabled

BFW-1980

### DIFF
--- a/lib/WUI/wui.c
+++ b/lib/WUI/wui.c
@@ -72,13 +72,14 @@ static void update_eth_changes(void) {
 }
 
 void StartWebServerTask(void const *argument) {
+    wui_eth_config.var_mask = ETHVAR_EEPROM_CONFIG;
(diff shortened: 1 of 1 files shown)


lib/WUI/wui.c
```diff
@@ -72,13 +72,14 @@ static void update_eth_changes(void) {
 }
 
 void StartWebServerTask(void const *argument) {
+    wui_eth_config.var_mask = ETHVAR_EEPROM_CONFIG;
+    load_eth_params(&wui_eth_config);
```
DRracer, line 76: Looking at main.cpp and order of initialization (esp. EEPROM) it looks like the strange osDelay(1000) can be finally removed and loading data from EEPROM is safe here.

DRracer (approved):

PR #1498, opened 2021-07-09
GUI: Fix filenames with diacritic not showing

MistrKernnunos:
- added filter for shortened names by fatFS when character outside
current code page is present
- switched fatFS code page to 852, fatFS can load czech filenames, but
in code page 852 encoding and it is not compatible with utf8 encoding,
thus the name on screen is shown with * or ? intead of the characte


dragomirecky (approved):
Nice work 👍

PR #1501, opened 2021-07-12
Littlefs for SPI Flash

milosfec:


src/common/w25x.c
```diff
@@ -10,6 +10,7 @@
 #include <w25x.h>
 #include "main.h"
 
+#define PAGE_SIZE 256
```
DRracer, line 13: please avoid using #defines if not absolutely necessary

src/littlefs.c
```diff
@@ -0,0 +1,110 @@
...
+    w25x_wait_busy();
+    return 0;
+}
+
+lfs_t* littlefs_init() {
+    int err = lfs_mount(&lfs, &cfg);
+
+    // reformat if we can't mount the filesystem
+    // this should only happen on the first boot
+    if (err) {
```
DRracer, line 98: Personally, I'd allow the caller to choose what to do with the partition in case of an error. For debugging purposes, it may be beneficial to not format the XFLASH area but allow it to dump somewhere for investigation.
Please consider adding a parameter saying something like "format_on_error" ...

src/littlefs.c
```diff
@@ -0,0 +1,110 @@
+#include "littlefs.h"
+#include "w25x.h"
+
+#define BLOCK_SIZE 4096
```
DRracer, line 4: again - please avoid #defines if not necessary

lib/Middlewares/Third_Party/littlefs/CMakeLists.txt
```diff
@@ -1,5 +1,15 @@
 add_library(littlefs lfs.c lfs_util.c)
 
+# Required printfs for debug print is not available, therefore disable all prints
```
DRracer, line 3: we may want to replace the LFS's debug/warn/error output with our logging subsystem - to be considered in the future

include/filesystem.h
```diff
@@ -14,6 +16,21 @@ extern "C" {
 
 void filesystem_init();
 
+static inline const char *process_path(const char *path, const char *dev_name) {
+    unsigned int dev_name_len = strlen(dev_name);
+    const char *device_path = path;
+    while (*device_path == '/') {
+        // Skip leading space
```
DRracer, line 23: misleading comment: '/' is not a space, or am I missing something?
... and how about other characters at the beginning of the path? What else should we skip/cope with?

src/filesystem_littlefs.c
```diff
@@ -0,0 +1,517 @@
...
+#include <time.h>
+#include <string.h>
+#include <sys/iosupport.h>
+#include <sys/stat.h>
+#include <sys/syslimits.h>
+#include "filesystem.h"
+#include "filesystem_littlefs.h"
+#include "littlefs.h"
+
+#define RESULT result < 0 ? -1 : 0;
```
DRracer, line 13: Are you sure the `result` will always get expanded correctly into this ternary operator? I'd suggest adding parentheses at the very least...

src/filesystem_littlefs.c
```diff
@@ -0,0 +1,517 @@
...
+    return stat_r(r, file, st);
+}
+
+static int utimes_r(struct _reent *r, const char *filename, const struct timeval times[2]) {
+    // Timestamps not implemented for littlefs
+    return 0;
+}
+
+static const devoptab_t devoptab_littlefs = {
+    .name = "internal",
```
DRracer, line 474: What is this name used for?

src/filesystem_littlefs.c
```diff
@@ -0,0 +1,517 @@
...
+        r->_errno = EINVAL;
+        return -1;
+    }
+
+    memset(buf, 0, sizeof(struct statvfs));
+
+    buf->f_frsize = lfs->cfg->block_size;
+    buf->f_bfree = lfs->free.size;
+    buf->f_bavail = buf->f_bfree;
+    buf->f_files = 0; // TODO: Count all inodes
```
DRracer, line 439: Do we need the number of inodes yet for something? Or is it just a nice-to-have option?

src/filesystem_littlefs.c
```diff
@@ -0,0 +1,517 @@
...
+    path = process_path(path, devoptab_littlefs.name);
+
+    result = lfs_remove(lfs, path);
+    r->_errno = get_errno(result);
+
+    return RESULT;
+}
+
+static int chdir_r(struct _reent *r, const char *path) {
+    // chdir not implemented for littlefs
```
DRracer, line 308: no support for changing current directory, that's sad :smile: (just joking)

DRracer (changes requested):
Please address my comments in the code.
Also, it looks like the PR cannot be built due to some unused function in littleFS. Since it looks like it is inside the littleFS library directly, I'd recommend removing the -Werror.... flag from the compilation parameters for littleFS:

```
../../lib/Middlewares/Third_Party/littlefs/lfs.c:429:13: error: 'lfs_mlist_isopen' defined but not used [-Werror=unused-function]

 static bool lfs_mlist_isopen(struct lfs_mlist *head,

             ^~~~~~~~~~~~~~~~

cc1: all warnings being treated as errors
```

src/common/w25x.c
milosfec: In this case it's possible, fixed in the next push.

src/littlefs.c
milosfec: It would work in C++, but not possible in C (initializer element is not constant when setting the value to const struct)

src/littlefs.c
milosfec: This object is for initialization of the littlefs, meaning this is the caller that is deciding what to do. For debugging purpose this is the place where we can place a code with a different logic.

lib/Middlewares/Third_Party/littlefs/CMakeLists.txt
milosfec: Yes, we can replace this with something like
```
  PROPERTY COMPILE_DEFINITIONS
            LFS_DEBUG print_to_serial("LFS DEBUG: " fmt, __VA_ARGS__)
```

include/filesystem.h
milosfec: Thanks, it's a "typo", should be "Skip leading slashes"

src/filesystem_littlefs.c
milosfec: It's used only as `return RESULT;` and adding anything before or after this macro would break it on compile time.

src/filesystem_littlefs.c
milosfec: It's used as a "mount point", you can access SPI flash under path `/internal/`.
Name chosen after discussion with @dragomirecky

src/filesystem_littlefs.c
milosfec: Not really. It could be used maybe to check if we can create another file for this filesystem but I don't expect we will get even close to the limit.

src/filesystem_littlefs.c
milosfec: As littlefs doesn't support `chdir` and I don't think we will needed it, I didn't spend time implementing it :) If needed, it would require just storing the path set in this function and modifying `process_path` function.

.pre-commit-config.yaml
```diff
@@ -38,4 +38,5 @@ exclude: |
     ^lib/jsmn/|
     ^lib/Catch2/|
     ^lib/libsysbase/
+    ^lib/Middlewares/Third_Party/littlefs/|
```
dragomirecky, line 41: This regex does not look right 🧐 (see the line endings…)

src/filesystem_littlefs.c
```diff
@@ -0,0 +1,536 @@
...
+        __attribute__((unused)) const struct timeval times[2]) {
+
+    // Timestamps not implemented for littlefs
+    r->_errno = ENOTSUP;
+    return -1;
+}
+
+static const devoptab_t devoptab_littlefs = {
+    .name = "internal",
+    .structSize = sizeof(lfs_file_t),
```
dragomirecky, line 494: /question/ Isn't this the size preallocated for `FIL_EX`? If so, `FIL_EX` is bigger than `lfs_file_t`...

src/littlefs.c
```diff
@@ -0,0 +1,110 @@
...
+#define BLOCK_COUNT 2048
+#define FLASH_SIZE (BLOCK_SIZE * BLOCK_COUNT)
+
+// 512 kB for dump (128 x 4kB)
+#define DUMP_BLOCK_COUNT 128
+
+// Address offset to skip area reserved for the dump
+#define ADDR_OFFSET (BLOCK_SIZE * DUMP_BLOCK_COUNT)
+
+lfs_t lfs;
```
dragomirecky, line 14: /suggestion/ It seems to me that this and all other definitions in this file are private to this file (except `littlefs_init`). Make them `static` in the case that is true

dragomirecky (commented):
See my comments. Looks great otherwise 👏

.pre-commit-config.yaml
milosfec: Thanks :)

src/filesystem_littlefs.c
milosfec: You are right! At the beginning I didn't use FIL_EX, but now it's needed and the structSize has to be updated.

src/littlefs.c
milosfec: Makes sense, I'm going to update.

DRracer (approved):
all my comments and questions have been resolved, approving

PR #1503, opened 2021-07-14
Fix watchdog reset while printing from Octoprint

MistrKernnunos:
problem with serial error message from Marlin kill comand, will be fixed with new logging
framework from Alan

BFW-1158


DRracer (changes requested):
Please do not change formatting of lib/* source codes - we keep the external libs intact.

DRracer (approved):

PR #1511, opened 2021-07-20
Improve how board and its version is handled by CMake

dragomirecky:


CMakeLists.txt
```diff
@@ -89,12 +93,41 @@ set(WUI
...
+  endif()
+endif()
+
+# set MCU to its default if not specified
+if(${MCU} STREQUAL "<default>")
+  set(MCU "STM32F407VG")
+endif()
+
+# parse board version into its components
+string(REGEX MATCH "^([0-9]+)\\.([0-9]+)\\.([0-9]+)" BOARD_VERSION_MATCH ${BOARD_VERSION})
```
DRracer, line 126: :+1:

include/board.h
```diff
@@ -1,11 +1,50 @@
 #pragma once
+//
```
DRracer, line 2: please prefer doxygen format `///` when writing comments

DRracer (changes requested):
please prefer doxygen format `///` when writing comments, otherwise ok

include/board.h
dragomirecky: Changed to a doxygen-styled comment

DRracer (approved):

PR #1512, opened 2021-07-22
Rewrite timing using 32bit timer 5

RadekVana:


DRracer (approved):
I don't see any potential issues, approving

src/common/timing.h
```diff
@@ -4,6 +4,21 @@ extern "C" {
...
 
+/// convert ticks to nano seconds
+///
+/// @return uint64_t nanoseconds, cannot overflow
+uint64_t ticks_to_ns(uint32_t cnt);
+
+/// converts miliseconds to ticks
+///
+/// @return uint64_t ticks, cannot overflow
+uint64_t ms_to_ticks(uint32_t ms);
```
dragomirecky, line 15: The naming of those newly-introduced functions is misleading/not correct.

Look at the original timing.h API (without this change): 
- You have ticks and timestamps. 
- Ticks are overflowing and with undefined beginning. 
- Timestamps begin at 0 and never overflow.
- You have different functions to get ticks/timestamps with different unit: `ticks_ms()` gives you ticks with a unit of milliseconds. `ticks_us` gives you ticks with the unit of microseconds. Etc. In other words – both ticks and timestamps always have some unit (ms, us, s, ns) and the unit is given by the function you used to get the ticks/timestamps.

Given the above, what should `ticks_to_ns` do? What ticks does it convert? Ticks returned by `ticks_ns()` or ticks returned by `ticks_ms()`? 

Clearly, it is meant for some "raw ticks" of the underlying timers, that the user of the `timing` API does not even have access to.

/suggestion/ Rename the functions appropriately and make them private to the `timing` module.

src/common/timing.h
```diff
@@ -41,8 +56,9 @@ int32_t ticks_diff(uint32_t ticks_a, uint32_t ticks_b);
 /// Given the datatype, overflows every ~584 years
 uint64_t timestamp_ns();
 
-/// To be called every 1 ms from TIM12 IRQ
-void tick_ms_irq();
+/// Sys timer's overflow interrupt callback
+///
+void TICK_TIMER_PeriodElapsedCallback();
```
dragomirecky, line 61: /question/ Why the uppercase letters? Why to change it to a name which does not fit the naming style of this module?

/suggestion/ I believe something like `tick_irq()` or `tick_timer_callback_irq()` would be more appropriate here.

lib/Drivers/STM32F4xx_HAL_Driver/Src/stm32f4xx_hal.c
```diff
@@ -173,7 +173,8 @@ HAL_StatusTypeDef HAL_Init(void)
   HAL_NVIC_SetPriorityGrouping(NVIC_PRIORITYGROUP_4);
 
   /* Use systick as time base source and configure 1ms tick (default clock after Reset is HSI) */
-  HAL_InitTick(TICK_INT_PRIORITY);
+  //HAL_InitTick(TICK_INT_PRIORITY);
+  //already initialized at this point, second initialization is causing problems
```
dragomirecky, line 177: This does not look right. 

I believe we should either a) not initialize the tick timer before this is called (to prevent the second initialization), or b) make HAL_InitTick detect that the timer has been initialized already and on subsequent calls do nothing. This way, we don't have to change the code in the HAL library.

dragomirecky (commented):
Little late to the party...

PR #1530, opened 2021-08-03
Bfw 1939

joshymjose:
ini file changes

### DIFF
--- a/src/common/ini_handler.c
+++ b/src/common/ini_handler.c
@@ -2,9 +2,8 @@
 
 #include "ini_handler.h"
 #include "wui_api.h"
-#include "ff.h"
 #include "eeprom.h"
-#include "strings.h"
+#include "string.h"
 
 #define MAX_UINT16 65535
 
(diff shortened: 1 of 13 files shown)


src/common/ini_handler.c
```diff
@@ -53,9 +81,9 @@ uint8_t ini_save_file(const char *ini_save_str) {
     UINT written_bytes = 0;
     FIL ini_file;
 
-    f_unlink(network_ini_file_name);
+    f_unlink(ini_file_name);
```
DRracer, line 84: when you are already using the POSIX FILE* for the INI parsing, please update this old FatFs code into the POSIX file interface

src/common/ini_handler.c
```diff
@@ -18,32 +18,60 @@ static int ini_handler_func(void *user, const char *section, const char *name, c
 
     ETH_config_t *tmp_config = (ETH_config_t *)user;
 
-    if (ini_string_match(section, "lan_ip4", name, "type")) {
-        if (strncasecmp(value, "DHCP", 4) == 0) {
+    if (ini_string_match(section, "eth::ipv4", name, "type")) {
+        if (strncmp(value, "DHCP", 4) == 0) {
```
DRracer, line 22: I still don't agree with enforcing just the uppercase `DHCP` and `STATIC`. An ordinary user will have more trouble making it run. What was wrong with accepting case-insensitive `dhcp` vs. `DHCP`?

src/common/ini_handler.c
joshymjose: fixed

src/common/ini_handler.c
joshymjose: ok

src/common/ini_handler.c
```diff
@@ -18,68 +17,89 @@ static int ini_handler_func(void *user, const char *section, const char *name, c
 
     ETH_config_t *tmp_config = (ETH_config_t *)user;
 
-    if (ini_string_match(section, "lan_ip4", name, "type")) {
-        if (strncasecmp(value, "DHCP", 4) == 0) {
+    if (ini_string_match(section, "eth::ipv4", name, "type")) {
+        if ((strcmp(value, "DHCP") == 0) || (strcmp(value, "dhcp") == 0)) {
```
DRracer, line 21: ideally `strncasecmp(value, "DHCP", 4)` just like it was before

DRracer (changes requested):
Almost there ;) ... please just use the original strncasecmp for DHCP and STATIC, that's my last remark, then I'll approve.

src/common/ini_handler.c
joshymjose: done

DRracer (approved):
Now it's ok, thank you

PR #1555, opened 2021-08-30
Fix write error during the flash process

vladamatena:
Solves ESP flashing instability. Bootloader flashing kept skipped, this one can be fixed by patching the flash params in bootloader binary.

BFW-2052


marekmosna-prusa3d (approved):
The entire conflict file became obsolete.

marekmosna-prusa3d (approved):

PR #1557, opened 2021-08-31
selftest: Make Z selftest functional with variable axis length.

RadekVana:


src/common/selftest/selftest_MINI.cpp
```diff
@@ -54,7 +54,7 @@ static const selftest_axis_config_t Config_XAxis = { .partname = "X-Axis", .leng
 
 static const selftest_axis_config_t Config_YAxis = { .partname = "Y-Axis", .length = 185, .fr_table = XYfr_table, .length_min = 179, .length_max = 189, .axis = Y_AXIS, .steps = 4, .dir = 1 };
 
-static const selftest_axis_config_t Config_ZAxis = { .partname = "Z-Axis", .length = 185, .fr_table = Zfr_table, .length_min = 181, .length_max = 191, .axis = Z_AXIS, .steps = 1, .dir = 1 };
+static const selftest_axis_config_t Config_ZAxis = { .partname = "Z-Axis", .length = get_z_max_pos_mm(), .fr_table = Zfr_table, .length_min = get_z_max_pos_mm() - 4, .length_max = get_z_max_pos_mm() + 6, .axis = Z_AXIS, .steps = 1, .dir = 1 };
```
DRracer, line 57: Do you really think the global static variables will get ALWAYS initialized correctly, especially when the init of the EEPROM happens in main()? I'd say it is a pure concidence that it worked now...

Global static vars are usually inited BEFORE entering the main()...

src/main.cpp
```diff
@@ -258,6 +247,18 @@ int main(void) {
     if (irq == 0)
         __disable_irq();
 
+    buddy::hw::BufferedSerial::uart2.Open();
```
DRracer, line 250: this can be a problem in the future, the uart2 probably uses something from RTOS which is not inited yet ... therefore the move after eeprom init

DRracer (approved):

PR #1558, opened 2021-08-31
Avoid ESP reset at GPIO init

vladamatena:
This fixes sporadic ESP reset combined with random ESP boot mode enter on STM
GPIO init.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -958,20 +958,25 @@ static void MX_GPIO_Init(void) {
     GPIO_InitStruct.Speed = GPIO_SPEED_FREQ_LOW;
     HAL_GPIO_Init(USB_EN_GPIO_Port, &GPIO_InitStruct);
 #ifdef USE_ESP01_WITH_UART6
-    /*Configure GPIO pins : ESP_RST_Pin */
-    GPIO_InitStruct.Pin = ESP_RST_Pin;
-    GPIO_InitStruct.Mode = GPIO_MODE_OUTPUT_PP;
-    GPIO_InitStruct.Pull = GPIO_NOPULL;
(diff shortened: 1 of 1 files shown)


src/main.cpp
```diff
@@ -958,20 +958,25 @@ static void MX_GPIO_Init(void) {
...
     GPIO_InitStruct.Mode = GPIO_MODE_OUTPUT_PP;
     GPIO_InitStruct.Pull = GPIO_PULLUP;
     GPIO_InitStruct.Speed = GPIO_SPEED_FREQ_LOW;
-    HAL_GPIO_Init(GPIOE, &GPIO_InitStruct);
     HAL_GPIO_WritePin(GPIOE, GPIO_PIN_6, GPIO_PIN_SET);
+    HAL_GPIO_Init(GPIOE, &GPIO_InitStruct);
+    /* Configure GPIO pins : ESP_RST_Pin */
+    GPIO_InitStruct.Pin = ESP_RST_Pin;
+    GPIO_InitStruct.Mode = GPIO_MODE_OUTPUT_PP;
+    GPIO_InitStruct.Pull = GPIO_NOPULL;
```
marekmosna-prusa3d, line 976: question: Does here assign GPIO_PULLUP do the same job as write GPIO_PIN_SET before HAL_GPIO_Init?

vladamatena: In theory the pull-up just gives the pin a bias using a few KOhm resistor while setting the pin should keep it high directly. As we configure the pin as output the pull-up resistor is either entirely disabled or irrelevant as we initially drive the pin directly.

marekmosna-prusa3d (approved):

PR #1562, opened 2021-09-01
Bulk "A3ides" files/boards/reference change to "Buddy"

wavexx:
This PR changes all references of A3ides to Buddy in file names, board definitions and code comments, which makes much more sense. A3ides was the internal board name, but that was changed to Buddy.

We also change the printer configuration file from ``Configuration_A3ides_[printer].h`` to ``Configuration_[printer].h`` (essentially, only ``Configuration_MINI/_adv.h``. The printer model defines the internal board and drivers, not the other way around.


dragomirecky (approved):
👏

PR #1573, opened 2021-09-03
GUI: Prusa Link api key

marekmosna-prusa3d:
- X-Api_Key contains 15 random characters
- Add ability to generate random number in the firmware
- Prusa Link can be enabled or disabled on startup

BFW-1945
BFW-2066


vladamatena (approved):
In general I approve this. I just have two remarks:

First, I assume the EEPROM data are not supposed to be backward compatible. It is ok the eeprom settings was lost. If ts is not the case the code needs to be fixed.

Second, I think this PR could be separated into multiple commits:
- RNG HW support
- variant vs. variant8
- Prusa Link settings

marekmosna-prusa3d:
> In general I approve this. I just have two remarks:
> 
> First, I assume the EEPROM data are not supposed to be backward compatible. It is ok the eeprom settings was lost. If ts is not the case the code needs to be fixed.
> 

Before release we increment the version of EEPROM and implement the convert script that solves exactly what you assumed. It's not welcomed to increment EEPROM version on each modification.

> Second, I think this PR could be separated into multiple commits:
> 
> * RNG HW support
> * variant vs. variant8
> * Prusa Link settings

I absolutely agree with you. Next time I will take a time and do not make PR so invasive.

PR #1574, opened 2021-09-03
Eeprom init

RadekVana:
Initialization of eeprom from startup script
BFW-2089

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -199,11 +199,8 @@ extern "C" void EepromSystemInit() {
     int irq = __get_PRIMASK() & 1;
     __enable_irq();
 
-    eeprom_init_status_t status = eeprom_init();
-    if (status == EEPROM_INIT_Defaults || status == EEPROM_INIT_Upgraded) {
-        // this means we are either starting from defaults or after a FW upgrade -> invalidate the XFLASH dump, since it is not relevant anymore
-        dump_in_xflash_reset();
-    }
+    eeprom_init();
+
     if (irq == 0)
         __disable_irq();
 }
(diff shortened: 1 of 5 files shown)


src/startup/startup_stm32f407xx.s
```diff
@@ -108,6 +108,8 @@ LoopFillZerobss:
 
 /* Call the clock system intitialization function.*/
   bl  SystemInit
+/* Initialize necessary minimum for eeprom.*/
+  bl  EepromSystemInit
```
DRracer, line 112: :+1:

src/main.cpp
```diff
@@ -182,30 +186,38 @@ void main_preinit() {
...
+    MX_I2C1_Init();
+    crc32_init();
+
+    int irq = __get_PRIMASK() & 1;
+    __enable_irq();
 
+    eeprom_init_status_t status = eeprom_init();
+    if (status == EEPROM_INIT_Defaults || status == EEPROM_INIT_Upgraded) {
+        // this means we are either starting from defaults or after a FW upgrade -> invalidate the XFLASH dump, since it is not relevant anymore
+        dump_in_xflash_reset();
```
DRracer, line 205: have you checked that `dump_in_xflash_reset` is low-level enough to be left here? I think the SPI XFLASH is being inited in main(), which is AFTER `EepromSystemInit`.

DRracer (changes requested):
The following questions must be answered/explained before approval:
- where do we upgrade the EEPROM structure and is it safe to be done at that spot in terms of the new EEPROM init sequence?
- Have you checked that dump_in_xflash_reset is low-level enough to be left here? I think the SPI XFLASH is being inited in main(), which is AFTER EepromSystemInit.

src/main.cpp
RadekVana: it was initialized internally in this function too, but you are right it is better to do it in main

RadekVana:
> The following questions must be answered/explained before approval:
> 
> * where do we upgrade the EEPROM structure and is it safe to be done at that spot in terms of the new EEPROM init sequence?
> * Have you checked that dump_in_xflash_reset is low-level enough to be left here? I think the SPI XFLASH is being inited in main(), which is AFTER EepromSystemInit.

eeprom is upgraded at first call of eeprom init at startup, and yes it is safe
dump_in_xflash_reset was initializing spi so it worked, but i moved it to main. It was my intention to do so, but I somehow missed it

DRracer (approved):

PR #1585, opened 2021-09-12
UI: Change blinking blue to azure when cooling

wavexx:
Blue on black is ugly and unreadable. Switch it to a brighter shade of azure:

2021-09-12T125559


wavexx:
On a side note, pr #1270 seemed to have defines for logical colors (COLOR_COOLING, etc), which would have been nice to keep:

``  constexpr auto COOLING_COLOR = COLOR_AZURE;``

so that we can consistently use the same logical color. I had to search for blue explicitly to find this undocumented array of color states..

DRracer (approved):

PR #1586, opened 2021-09-12
Tests: Fixup ini file name change in translation tests

wavexx:
Commit 5d285eb696071fcae9a4ce683b734bfcf964e9ff renamed lan_settings.txt, changing the po file translations, but didn't update the test files making them fail.

Do so here.


wavexx:
Looks like this is the wrong fix, the unit tests are broken.

wavexx:
Closing, as this is not the right fix.

PR #1589, opened 2021-09-14
Change FatFs to UTF-8

MistrKernnunos:
BFW-2104


DRracer:
@vorner good question, but:
- font rendering is limited, so if the character is not found in the font bitmaps, there is a fail-safe '?'
- if the filename contains some wchar (is that UCS-16 what they use in FAT32?) which doesn't fit into any UTF-8 character, the file will probably be inaccessible (classical situation on systems running MS Windows with different system locale)
- we must test this change properly

PR #1590, opened 2021-09-14
Add warning not to change provider test files

MistrKernnunos:


wavexx:
Did you find a way to avoid the copy into the build path?

PR #1604, opened 2021-09-17
Esp fixes

vladamatena:


include/esp/esp_config.h
```diff
@@ -112,4 +112,56 @@
...
+/**
+ * This configures LwESP, a library that talks to stock Espressif AT firmware
+ * using UART providing TCP connection abstraction. This seems to work,
+ * somehow, but there are some downsides.
+ *
+ * 1) There is no flow control (either HW or SW) used on the UART. Provided we
+ * can only buffer limited amount of data this requires us to process (write
+ * to flash) any incoming data in limited time. This requires us to keep HTTP
+ * request handling a bounded time algorithm. Also this forces us to keep UART
+ * baudrate low to avoid buffer overlow in worst case scenario.
```
vorner, line 124: :question: What happens if it overflows? Does everything break completely, or we just lose a packet? If the latter, then it's OK-ish (lost packets provide a form of natural flow control on the network level).

marekmosna-prusa3d (approved):

include/esp/esp_config.h
vladamatena: Actually, both. 

i) We might "just" loss a packet. Unfortunately the lost packet was already acknowledged by the ESP firmware to TCP will not save the situation. Particular TCP connection will just miss some data in the middle of the stream. In case of gcode upload this results in corrupted file being stored and HTTP server eventually closing the connection after not receiving rest of the file for too long.

ii) If we loss a command reply from ESP, the whole LwESP breaks waiting for the response with buffers full of enqued commands.

I think the UART buffer must not overflow.

PR #1606, opened 2021-09-20
ADC: Rewrite to use template class

Hagridpon3:


src/common/adc.h
```diff
@@ -3,17 +3,57 @@
...
-extern uint16_t get_adc_channel_value(ADC_HandleTypeDef *adc, uint32_t channel);
+template <ADC_HandleTypeDef &adc, size_t channels>
+class AdcDma {
+public:
+    AdcDma()
+        : m_data() {};
+    void init() {
+#ifndef SIM_HEATER
+        HAL_ADC_Init(&adc);
+        HAL_ADC_Start_DMA(&adc, (uint32_t *)m_data, channels); // Start ADC in DMA mode and
```
mkbel, line 18: Suggest: Use c++ style cast (`static_cast`) instead of c-style cast.

mkbel:
Maybe it would be worth to rename adc.h to adc.hpp, as it can not be used in C.

src/common/adc.h
Hagridpon3: Solved

mkbel (dismissed):

mkbel:
What positive effect is gained by added complexity of interface class?

RadekVana (approved):

mkbel (approved):

RadekVana:
I don't like `SIM_HEATER` macro and would prefer not having it in header files and in future replaced by just different sources in cmake.

PR #1608, opened 2021-09-22
Fix printer data in /api/version

marekmosna-prusa3d:
BFW-1947


lib/WUI/wui_REST_api.h
```diff
@@ -17,6 +17,17 @@
 extern "C" {
 #endif
 
+#define PL_VERSION_MAJOR    2
+#define PL_VERSION_MINOR    0
+#define PL_VERSION_REVISION 0
```
vorner, line 22: :question: The API version just went from 0.1 to 2.0.0. What does the version signify and where are the versions defined? Is it related to `doc/openapi.yaml` (which is on 1.0.0)?

Or did you take the old _server_ version and put it into _api_ by mistake?

vorner (changes requested):

vorner (approved):
:+1: Thanks (I assume the 2.0.0 is therefore on purpose, not a mistake)

PR #1609, opened 2021-09-22
Fix reconfigure UART on click

marekmosna-prusa3d:
BFW-2103


vorner (approved):
✔ Makes sense. Might be good to include the info or link to the ticket in the commit message next time, though 😇.

PR #1615, opened 2021-09-25
Use symbolic names for interrupt priorities

wavexx:
When setting interrupt priorities, two special values pop up:

5: configLIBRARY_MAX_SYSCALL_INTERRUPT_PRIORITY
15: configLIBRARY_LOWEST_INTERRUPT_PRIORITY

These are for 5) interrupts that need to call back into FreeRTOS code,
and 15) the lowest priority.

Do not hardcode such values. Use symbolic names, which makes the
priority meaningful.


dragomirecky (approved):
👍

Hagridpon3 (approved):

PR #1616, opened 2021-09-25
Introduce "config_features.h"

wavexx:
Getting configured printer features is currently tricky, as it needs to
be done by including the appropriate local and Marlin configuration,
without including any other Marlin global or header.

Create a new "config_features.h" file which exposes this info by including
the minimal set of appropriate files.


dragomirecky (approved):
👏

PR #1623, opened 2021-10-01
Add functionality to long press

MistrKernnunos:


src/common/marlin_server.cpp
```diff
@@ -370,6 +370,14 @@ void marlin_server_do_babystep_Z(float offs) {
     babystep.task();
 }
 
+extern void marlin_server_move_axis(float pos, size_t axis) {
+    xyze_float_t position = current_position;
+    position[axis] = pos;
+    current_position[axis] = pos;
+    static std::array<int, 4> feedrates = { MANUAL_FEEDRATE };
```
wavexx, line 377: Can we avoid the array, or use a constexpr?

MistrKernnunos: Yes, the feedrate will be added as argument to the function

src/gui/dialogs/DialogMoveZ.cpp
```diff
@@ -0,0 +1,134 @@
...
+    infoText.font = GuiDefaults::FontMenuSpecial;
+    infoText.SetAlignment(Align_t::Center());
+    //  close text
+    Rect16 closeText_rc(infoText_rc, ShiftDir_t::Bottom, item_height * 1.5);
+    closeText.SetRect(closeText_rc);
+    closeText.SetPadding(padding);
+    closeText.font = GuiDefaults::FontMenuSpecial;
+    closeText.SetAlignment(Align_t::Center());
+    // icon
+    Rect16 icon_rc(Width() / 2 - 40, (2 * Height() / 3) - 27, 81, 55);
```
michalrudolf, line 43: This constructor is very long and it contains complex calculations combing relative positioning with magic numbers.
Please define used rectangles like this with magic numbers: 
```
static constexpr Rect16 axisTextRect = Rect16(x, ,x ,x ,x);
static constexpr Rect16 infoTextRect = Rect16(x, ,x ,x ,x);
...etc..
```
Then assign them directly in the initializer list. It will save us time in the future.

src/gui/dialogs/DialogMoveZ.cpp
```diff
@@ -0,0 +1,134 @@
...
+    }
+    if (event == GUI_event_t::LOOP) {
+
+        marlin_update_vars(MARLIN_VAR_MSK(MARLIN_VAR_PQUEUE));
+        if (marlin_vars()->pqueue <= len) {
+            int difference = value - lastQueuedPos;
+            uint8_t freeSlots = len - marlin_vars()->pqueue;
+            if (difference > 0) {
+                // move up and queue steps
+                for (uint8_t i = 0; i < freeSlots && lastQueuedPos != value; i++) {
```
michalrudolf, line 106: Reduce to
```
if (difference) {
    for (uint8_t i = 0; i < freeSlots && lastQueuedPos != value; i++) {
        if (difference > 0) { 
            lastQueuedPos++;
        } else if (difference < 0) {
            lastQueuedPos--;
        }
        marlin_move_axis(lastQueuedPos, MenuVars::GetManualFeedrate()[2], 2);
    }
}
```

michalrudolf (dismissed):

michalrudolf (approved):

PR #1624, opened 2021-10-01
Fix static vs dhcp ip obtain

marekmosna-prusa3d:
Fix switching between static and dynamic IP address obtaining 

BFW-2088

### DIFF
--- a/lib/WUI/netdev.c
+++ b/lib/WUI/netdev.c
@@ -222,22 +222,34 @@ netdev_status_t netdev_get_status(uint32_t netdev_id) {
     }
 }
 
-netdev_ip_obtained_t netdev_get_ip_obtained_type() {
-    return IS_LAN_STATIC(ETH_CONFIG().lan.flag) ? NETDEV_STATIC : NETDEV_DHCP;
+netdev_ip_obtained_t netdev_get_ip_obtained_type(uint32_t netdev_id) {
+    return IS_LAN_STATIC(wui_netdev_config[netdev_id].lan.flag) ? NETDEV_STATIC : NETDEV_DHCP;
 }
 
 uint32_t netdev_set_dhcp(uint32_t netdev_id) {
+    ETH_config_t *pConfig = NULL;
+    err_t res = ERR_OK;
+
     if (netdev_id == NETDEV_ETH_ID) {
-        err_t res;
         res = netifapi_dhcp_start(&eth0);
-        CHANGE_FLAG_TO_DHCP(wui_netdev_config[netdev_id].lan.flag);
-        wui_netdev_config[netdev_id].var_mask = ETHVAR_MSK(ETHVAR_LAN_FLAGS);
-        save_eth_params(&wui_netdev_config[netdev_id]);
-        wui_netdev_config[netdev_id].var_mask = 0;
-        return res;
-    } else {
-        return ERR_OK;
+        pConfig = &wui_netdev_config[netdev_id];
+    } else if (netdev_id == NETDEV_ESP_ID) {
+        esp_sta_quit(NULL, NULL, 1);
+        esp_sta_join(ap.ssid, ap.pass, NULL, 0, NULL, NULL, 0);
+        pConfig = &wui_netdev_config[netdev_id];
+        res = ERR_OK;
     }
+
+    if (pConfig) {
+        CHANGE_FLAG_TO_DHCP(pConfig->lan.flag);
+        pConfig->var_mask = ETHVAR_MSK(ETHVAR_LAN_FLAGS);
+        save_eth_params(pConfig);
+        pConfig->var_mask = 0;
+    } else {
+        res = ERR_IF;
+    }
+
+    return res;
 }
 
 uint32_t netdev_set_up(uint32_t netdev_id) {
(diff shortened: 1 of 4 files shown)


lib/WUI/netdev.c
```diff
@@ -222,22 +222,34 @@ netdev_status_t netdev_get_status(uint32_t netdev_id) {
...
         res = netifapi_dhcp_start(&eth0);
-        CHANGE_FLAG_TO_DHCP(wui_netdev_config[netdev_id].lan.flag);
-        wui_netdev_config[netdev_id].var_mask = ETHVAR_MSK(ETHVAR_LAN_FLAGS);
-        save_eth_params(&wui_netdev_config[netdev_id]);
-        wui_netdev_config[netdev_id].var_mask = 0;
-        return res;
+        pConfig = &wui_netdev_config[netdev_id];
+    } else if (netdev_id == NETDEV_ESP_ID) {
+        esp_sta_quit(NULL, NULL, 1);
+        esp_sta_join(ap.ssid, ap.pass, NULL, 0, NULL, NULL, 0);
```
vorner, line 238: Could you add a comment why we do this quit-join dance? I know it now (to reset it so it gives up potential static IP), but a future reader few years from now will rightfully wonder what might be happening.

lib/WUI/netdev.c
```diff
@@ -222,22 +222,34 @@ netdev_status_t netdev_get_status(uint32_t netdev_id) {
...
+        pConfig = &wui_netdev_config[netdev_id];
+        res = ERR_OK;
+    }
+
+    if (pConfig) {
+        CHANGE_FLAG_TO_DHCP(pConfig->lan.flag);
+        pConfig->var_mask = ETHVAR_MSK(ETHVAR_LAN_FLAGS);
+        save_eth_params(pConfig);
```
vorner, line 246: :question: Will the EEPROM have the _same_ slot for both ethernet and wifi? Because right now no matter which one we save, it goes to the same place and doesn't allow us to have the future „both“ mode.

lib/WUI/netdev.c
```diff
@@ -268,21 +280,32 @@ uint32_t netdev_set_down(uint32_t netdev_id) {
...
+        res = netifapi_netif_set_addr(&eth0, &pConfig->lan.addr_ip4, &pConfig->lan.msk_ip4, &pConfig->lan.gw_ip4);
         res = netifapi_dhcp_inform(&eth0);
-        CHANGE_FLAG_TO_STATIC(wui_netdev_config[netdev_id].lan.flag);
-        wui_netdev_config[netdev_id].var_mask = ETHVAR_MSK(ETHVAR_LAN_FLAGS);
-        save_eth_params(&wui_netdev_config[netdev_id]);
-        wui_netdev_config[netdev_id].var_mask = 0;
-        return res;
+    } else if (netdev_id == NETDEV_ESP_ID) {
+        pConfig = &wui_netdev_config[netdev_id];
+        esp_sta_setip((esp_ip_t *)&pConfig->lan.addr_ip4, (esp_ip_t *)&pConfig->lan.gw_ip4, (esp_ip_t *)&pConfig->lan.msk_ip4, 1, NULL, NULL, 1);
```
vorner, line 295: No DNS in case of Wifi? Would it be better to put both the `dns_setserver` into the `if (pConfig)` section?

lib/WUI/netdev.c
```diff
@@ -268,21 +280,32 @@ uint32_t netdev_set_down(uint32_t netdev_id) {
...
+        res = ERR_OK;
+    }
+
+    if (pConfig) {
```
vorner, line 299: I've seen the practice of treating pointers as bools frowned upon at various places. So, just to check, is this accepted practice here?

lib/WUI/netdev.h
```diff
@@ -126,8 +126,12 @@ netdev_status_t netdev_get_status(uint32_t);
 ////////////////////////////////////////////////////////////////////////////
 /// @brief Retrive method used to obtain ip address on active network device
 ///
+/// @param[in] dev_id device ID. One of
+///             - #NETDEV_ETH_ID
+///             - #NETDEV_ESP_ID
+///             - #NETDEV_NODEV_ID
```
vorner, line 132: :bomb: I think passing the `NETDEV_NODEV_ID` will result in out of bounds access on the array which has only 2 items.

src/gui/screen_menu_lan_settings.cpp
```diff
@@ -55,7 +55,7 @@ class MI_NET_IP_t : public WI_SWITCH_t<2> {
 public:
     MI_NET_IP_t()
         : WI_SWITCH_t(0, string_view_utf8::MakeCPUFLASH((const uint8_t *)label), 0, is_enabled_t::yes, is_hidden_t::no, string_view_utf8::MakeCPUFLASH((const uint8_t *)str_DHCP), string_view_utf8::MakeCPUFLASH((const uint8_t *)str_static)) {
-        this->index = netdev_get_ip_obtained_type() == NETDEV_STATIC
+        this->index = netdev_get_ip_obtained_type(netdev_get_active_id()) == NETDEV_DHCP
```
vorner, line 58: Can this code be reached with networking turned off (set to `None`)? If so, this'll blow up.

vorner (changes requested):

vorner (approved):
:+1: Thank you, I think this can be merged.

PR #1625, opened 2021-10-01
Slowdown print moves only.

mkbel:


wavexx:
For context, this is intended to avoid slowing down ahead of time when travel moves are being queued.

If these are service moves (parking & similar), slowing down doesn't make sense since motion will halt anyway. Obeying slowdown there is a mistake.

However even when we're about to plan extrusions, it still doesn't make sense to slow-down travel moves. Travels are generally long straight segments and will provide enough time to replenish the buffer.

wavexx (approved):
One word changes are the best

wavexx:
But actually, can we do better than that? What about slowing down forward extrusions only? Any side effect?

PR #1636, opened 2021-10-11
Fix simultaneous gcode upload

marekmosna-prusa3d:
Fix review remarks
In case of more upload simultaneously the printer reply status code 503

BFW-2160

### DIFF
--- a/lib/WUI/http/httpd_post.c
+++ b/lib/WUI/http/httpd_post.c
@@ -66,7 +66,7 @@ typedef struct {
 static uint8_t http_post_uri_file_index = 0;
 static uint32_t http_post_content_len = 0;
 #define HTTP_POST_URI_NUM 2
-const char *a[HTTP_POST_URI_NUM] = {
+const char *endpoints[HTTP_POST_URI_NUM] = {
     "/api/files/sdcard",
     "/api/files/local"
 };
@@ -104,11 +104,11 @@ int read_header_name(multipart_parser *p, const char *at, size_t length) {
 #ifdef HTTPD_DEBUG
     _dbg("Key Name: %s\n", key_name);
 #endif
-    if (strncmp(key_name, "print", 5) == 0) {
+    if (lwip_strnicmp(key_name, "print", 5) == 0) {
         upload.phase = UPLOAD_PROCESS_PRINT;
-    } else if (strncmp(key_name, "path", 4) == 0) {
+    } else if (lwip_strnicmp(key_name, "path", 4) == 0) {
         upload.phase = UPLOAD_PROCESS_PATH;
-    } else if (strncmp(key_name, "file", 4) == 0) {
+    } else if (lwip_strnicmp(key_name, "file", 4) == 0) {
         upload.phase = UPLOAD_PROCESS_FILE;
         char *filename = strstr(key_name + 4 + 3, "filename");
         const char *end_filename = strtok(filename + 8 + 2, "\"");
@@ -134,7 +134,7 @@ int read_part_data(multipart_parser *p, const char *at, size_t length) {
 #endif
     switch (upload.phase) {
     case UPLOAD_PROCESS_PRINT:
-        if (strncmp(at, "true", 4) == 0) {
+        if (lwip_strnicmp(at, "true", 4) == 0) {
             upload.start_print = 1;
         }
         break;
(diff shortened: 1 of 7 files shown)


lib/WUI/http/httpd_post.c
```diff
@@ -229,27 +229,34 @@ err_t httpd_post_begin(void *connection,
...
+        uri = "/401";
         goto invalid;
     } else {
         const char *api_key = wui_get_api_key();
         uint32_t token_length = strlen(api_key);
-        if (memcmp(api_key, api_key_start + api_key_tag_length, token_length) != 0) {
-            uri = "/401\0";
+
+        if (lwip_strnicmp(api_key, api_key_start + api_key_tag_length, token_length) != 0
+            || (api_key_start[api_key_tag_length + token_length] != '\r' && api_key_start[api_key_tag_length + token_length + 1] != '\n')) {
```
vorner, line 245: Is the newline in-range for sure? Could our buffer end up with the end of the token?

lib/WUI/wui.c
```diff
@@ -122,7 +122,11 @@ void StartWebServerTask(void const *argument) {
 }
 
 const char *wui_get_api_key() {
+#ifdef _DEBUG
+    return "PL-mini";
```
vorner, line 126: I understand the motivation, but this seems problematic for two reasons:

* It is quite reasonable to expect developers or even some tinkerer users to run a development version of the firmware and use the printer the usual way, some kind of background testing. Having a well-known password that's impossible to change looks like a serious security issue, especially when evil GCode can lead to HW damage…
* This is surprising. If I read it correctly, when user clicks on the „Generate API Key“ to change it, it generates one, stores it into EEPROM, shows it, but will still use the old one.

If your motivation is to make it easier to repeatedly reflash & lose the eeprom content, I'd suggest we allow loading API key from the config file. When you reflash, you need to set the SSID/Password from there anyway. It'll also be a nice feature for people running multiple printers, so they can set the same API key on all of them if they like.

Should I create a jira ticket for that?

vorner (changes requested):
There are few accesses to places that could be after the end of a buffer (or, at least it is not clear why it can't be after the end).

I don't think we want that hard-coded API key. Besides, if we did, that one should really go into a separate commit.

lib/WUI/http/httpd_post.c
vorner: I have a safer fix (I hope) in the branch where I'm fixing UBs, so let's not dwell on these. The question if we want to authenticate case-sensitive or insensivite still stands, though.

lib/WUI/wui_api.c
```diff
@@ -386,7 +388,7 @@ uint32_t wui_upload_finish(const char *old_filename, const char *new_filename, u
     }
 
 clean_temp_file:
-    remove(tmp_filename);
+    result = remove(tmp_filename);
```
vorner, line 391: Is the `result` used anywhere? It seems not…

lib/WUI/http/httpd_post.c
marekmosna-prusa3d: perfect

lib/WUI/wui.c
marekmosna-prusa3d: perfect...I'm looking forward for the USB configuration usage

lib/WUI/http/httpd_structs.h
```diff
@@ -31,8 +31,9 @@ static const char *const g_psHTTPHeaderStrings[] = {
     "HTTP/1.0 500 Internal Server Error\r\n",
     "HTTP/1.0 401 Unauthorized\r\nWWW-Authenticate: ApiKey realm=\"401\"",
     "HTTP/1.0 304 Not Modified\r\n"
-    "HTTP/1.0 409 Conflict \r\n"
+    "HTTP/1.0 409 Conflict\r\n"
     "HTTP/1.0 415 Unsupported Media Type\r\n"
+    "HTTP/1.0 503 Service Temporarily Unavailable\r\n"
```
vorner, line 36: The strings are missing commas between them, so they'll form one long string instead and the indexing will be out of bounds.

vorner (approved):
After fixing the problem of commas between string constants this can go in :+1:

PR #1642, opened 2021-10-15
Add scale library

RadekVana:


src/common/algorithm_scale.hpp
```diff
@@ -0,0 +1,93 @@
...
+ * @tparam T
+ * @param num number to be scaled
+ * @param min original minimum range
+ * @param max original maximum range
+ * @param scaled_min scaled minimum range
+ * @param scaled_max scaled maximum range
+ * @return T constexpr
+ */
+template <class T>
+T constexpr scale_with_overflow_detection(T num, T min, T max, T scaled_min, T scaled_max) {
```
espr14, line 73: `detection` or `prevention`? There's no output giving the overflow flag.

src/common/algorithm_scale.hpp
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    num = std::clamp(num, min, max);
+
+    if (scaled_min > scaled_max) {
+        std::swap(scaled_min, scaled_max);
+        num = num - min;
+        num = max - num;
+    }
+    return (scaled_max - scaled_min) * (num - min) / (max - min) + scaled_min;
```
espr14, line 43: Can be dangerous in `int` arithmetic because of division. At least, should be stated in comments that integers will give different results from floats.

espr14 (changes requested):

src/common/algorithm_scale.hpp
RadekVana: name fixed

src/common/algorithm_scale.hpp
RadekVana: max == min is checked
and it will first multiply, than divide to not loose precision

PR #1643, opened 2021-10-15
ESP backport manual receive BFW-2166

vladamatena:
- Port ESP 1.x FW support to recent enough LwESP supporting TCP manual receive
- Various LwESP integration fixes
- Fixes two ALTCP - LwESP integration layer bugs

### DIFF
--- a/lib/ESP/prusa/include/esp_tcp.h
+++ b/lib/ESP/prusa/include/esp_tcp.h
@@ -46,7 +46,8 @@
 #include "esp/esp_config.h"
 #include "sockets/lwesp_sockets_priv.h"
 
-#define EPCB_POOL_SIZE 10
+// It would be better to obtain this from ESP but there seems to be no way
+#define ESP_TCP_MSS 536
 
 #if ESP_ALTCP /* don't build if not configured for use in lwipopts.h */
 
--- a/lib/ESP/prusa/src/system/esp_tcp.c
+++ b/lib/ESP/prusa/src/system/esp_tcp.c
@@ -712,11 +712,11 @@ static err_t altcp_esp_output(struct altcp_pcb *conn) {
 }
 
 static u16_t altcp_esp_mss(struct altcp_pcb *conn) {
-    return 536; // Minimal required MSS, TODO: Implement properly
+    return ESP_TCP_MSS;
 }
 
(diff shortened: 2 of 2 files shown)


lib/ESP/lwesp/.gitrepo
```diff
@@ -4,9 +4,9 @@
 ; git-subrepo command. See https://github.com/git-commands/git-subrepo#readme
 ;
 [subrepo]
-	remote = https://github.com/MaJerle/lwesp.git
-	branch = v0.5.0
-	commit = 4dd5399fabebedf601042d5dad4c29f9e6c24abe
-	parent = 04f8547400d0f5493dbd8f690b24146b5d144999
+	remote = https://github.com/vladamatena/lwesp.git
```
vorner, line 7: Question about this subrepo. I assume it contains some of the backports. Are we going to maintain them, or will they be eventually pushed upstream? If the former, should it live under your account, or should it move to `prusa3d`?

lib/WUI/netdev.c
```diff
@@ -184,7 +200,7 @@ uint32_t netdev_init() {
 }
 
 uint32_t netdev_init_esp() {
-    esp_hard_reset_device();
+    // esp_hard_reset_device();
```
vorner, line 203: Any reason why we are not resetting the device here, to make sure it starts in known state?

lib/ESP/prusa/src/system/lwesp_ll_buddy.c
```diff
@@ -87,10 +87,11 @@ void StartUartBufferThread(void const *arg) {
     }
 }
 
-void esp_hard_reset_device() {
+uint8_t esp_hard_reset_device(uint8_t state) {
```
vorner, line 90: Is `state` used for anything?

vorner (commented):
Is there a way to see what changes you did and which were introduced by pulling from some upstream?

lib/WUI/netdev.c
vladamatena: Leaving commented out code behind was a mistake. Integrated upstream changes in LwESP implement hard reset as part of init sequence. We just provide a function pointer to do it.

PR #1650, opened 2021-10-19
Fix retrieve MAC from ESP

marekmosna-prusa3d:
BFW-2154

### DIFF
--- a/lib/WUI/netdev.c
+++ b/lib/WUI/netdev.c
@@ -88,16 +88,15 @@ void netdev_get_ipv4_addresses(uint32_t netdev_id, lan_t *config) {
     }
 }
 
-void netdev_get_MAC_address(uint32_t netdev_id, uint8_t mac[6]) {
+void netdev_get_MAC_address(uint32_t netdev_id, uint8_t mac[OTP_MAC_ADDRESS_SIZE]) {
     if (netdev_id == NETDEV_ETH_ID) {
-        memcpy(mac, (void *)OTP_MAC_ADDRESS_ADDR, 6);
+        memcpy(mac, (void *)OTP_MAC_ADDRESS_ADDR, OTP_MAC_ADDRESS_SIZE);
     } else if (netdev_id == NETDEV_ESP_ID) {
         esp_mac_t tmp;
-        espr_t result = esp_sta_getmac(&tmp, NULL, NULL, 1);
-        result += 10;
-        memcpy(mac, (void *)&tmp, 6);
+        esp_sta_getmac(&tmp, NULL, NULL, 1);
+        memcpy(mac, (void *)&tmp, OTP_MAC_ADDRESS_SIZE);
     } else {
-        memset(mac, 0, 6);
+        memset(mac, 0, OTP_MAC_ADDRESS_SIZE);
     }
 }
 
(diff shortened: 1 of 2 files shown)


lib/WUI/netdev.c
```diff
@@ -69,21 +70,34 @@ static void tcpip_init_done_callback(void *arg) {
...
-        config->lan.gw_ip4.addr = 0;
+        config->addr_ip4.addr = 0;
+        config->msk_ip4.addr = 0;
+        config->gw_ip4.addr = 0;
+    }
+}
+
+void netdev_get_MAC_address(uint32_t netdev_id, uint8_t mac[6]) {
+    if (netdev_id == NETDEV_ETH_ID) {
+        memcpy(mac, (void *)OTP_MAC_ADDRESS_ADDR, 6);
```
vorner, line 93: That 6 sounds like it could either be a constant or we could use typedef + sizeof combination. It's repeated here 4 times.

lib/WUI/netdev.c
```diff
@@ -69,21 +70,34 @@ static void tcpip_init_done_callback(void *arg) {
...
+    } else if (netdev_id == NETDEV_ESP_ID) {
+        esp_mac_t tmp;
+        espr_t result = esp_sta_getmac(&tmp, NULL, NULL, 1);
+        result += 10;
```
vorner, line 97: You assign to the result and you add 10, but then don't read it. Can you either fix it or provide an explaining comment?

src/gui/screen_menu_lan_settings.cpp
```diff
@@ -109,13 +111,31 @@ ScreenMenuLanSettings::ScreenMenuLanSettings()
...
+
+    netdev_get_ipv4_addresses(netdev_id, &config);
+    netdev_get_MAC_address(netdev_id, mac);
+
+    ip4addr_ntoa_r(&(config.addr_ip4), addr, IP4_ADDR_STR_SIZE);
+    ip4addr_ntoa_r(&(config.msk_ip4), msk, IP4_ADDR_STR_SIZE);
+    ip4addr_ntoa_r(&(config.gw_ip4), gw, IP4_ADDR_STR_SIZE);
+
+    snprintf(mac_str, MAC_ADDR_STR_LEN, "%02x:%02x:%02x:%02x:%02x:%02x",
+        *mac, *(mac + 1), *(mac + 2), *(mac + 3), *(mac + 4), *(mac + 5));
```
vorner, line 128: Just a style opinion: wouldn't `mac[0], mac[1], …` be more readable than this?

src/gui/screen_menu_lan_settings.cpp
```diff
@@ -109,13 +111,31 @@ ScreenMenuLanSettings::ScreenMenuLanSettings()
     msg_shown = false;
 }
 
+uint32_t ScreenMenuLanSettings::StringifyNetworkDevice(uint32_t netdev_id, lan_descp_str_t buffer) {
```
vorner, line 114: Style consistency: Methods here seem to start with lower case letter, upper case is for types/clases.

What does the return type mean and is it in any way useful? The only caller below doesn't use it, shall it be `void` then? Besides, it is wrong ‒ `snprintf` returns `int`.

vorner (approved):

PR #1655, opened 2021-10-21
Fix ESP Ip address BOOM inside lan setting screen

marekmosna-prusa3d:
Network devices have got their own status which is cached inside lan setting screen and redraw and even retrieve ip details only when change appeared

BFW-2150


jrsall92:
Does this mean that ESP wifi modules support is coming?

src/gui/screen_menu_lan_settings.cpp
```diff
@@ -100,69 +99,67 @@ ScreenMenuLanSettings::ScreenMenuLanSettings()
...
 
 /*****************************************************************************/
 //non static member function definition
-void ScreenMenuLanSettings::refresh_addresses() {
-    if (netdev_get_status(netdev_get_active_id()) == NETDEV_NETIF_UP) {
+void ScreenMenuLanSettings::refresh_addresses(uint32_t netdev_id) {
+    netdev_status_t status = netdev_get_status(netdev_id);
+    if (status == cached_network_status) {
+        return;
+    }
```
vorner, line 116: Is this the right place to cache it? Because the current IP may be used by other parts of the code.

Also, this introduces a reverse problem ‒ the IP not being refreshed enough. I think this could happen in two cases:

* I've both wifi and ethernet configured, with signal, cable connected ‒ therefore both are connected, one of them is active. I switch from one to another, in both cases the state is `NETDEV_NETIF_UP`, so the IP does not get refreshed.
* I take one cable out, plug another one in, it changes the IP, goes down and up. This happens during one refresh cycle, so the down is not seen by the GUI at all, and shows the old IP.

vorner (changes requested):
I think this introduces the reverse bug. If you explain why it can't happen, then OK…ish.

Nevertheless, I wonder if the cure isn't worse than the disease here. It adds more complexity into the code. Does it matter that the IP is being requested over and over? Is there an actual downside to that?

PR #1660, opened 2021-10-28
BFW-1996: Do a proper httpd reinit

vorner:
Reinitialize the http server on network changes, so the API is properly available.

Make the httpd_reinit actually fully reinit (including ability to
accommodate an interface change).

Make the said httpd_reinit thread safe by locking around the httpd
socket.

Add ability to shut down the listening socket too.

Call the reinit on various network setting changes.

# Reviewers

I'm tagging @vladamatena as well, as @marekmosna-prusa3d is on vacation r.n. If you don't have time for this or feel this is not in your area of interest any more, it can probably wait.

### DIFF
--- a/lib/WUI/http/httpd.c
+++ b/lib/WUI/http/httpd.c
@@ -111,12 +111,15 @@
 #include <string.h> /* memset */
 #include <stdlib.h> /* atoi */
 #include <stdio.h>
-#include "stdbool.h"
+#include <stdbool.h>
+#include <FreeRTOS.h>
+#include <semphr.h>
 
 /*******   Customization ***************************************/
 #include "wui_REST_api.h"
 #include "wui_api.h"
 #include "wui.h"
+#include <netdev.h>
 
 #include "dbg.h"
 #define WUI_API_ROOT_STR_LEN 5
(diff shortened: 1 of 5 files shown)


lib/WUI/http/httpd.c
```diff
@@ -2467,20 +2474,53 @@ void httpd_init(void) {
...
     httpd_reinit();
 }
 
+// The inner part of httpd_close. Assumes the mutex is already locked.
+static void httpd_close_locked() {
+    // According to docs, the close can fail in case there's not enough memory.
+    // That's likely because usual sockets still wait for some more data to
+    // arrive (?). Hopefully this is not the case for listening sockets.
+    //
+    // Using altcp_abort would be better in theory (as it just wipes it and
```
vladamatena, line 2486: LwESP does not seem to have abort counterpart. Unfortunately every ESP action allocates message structure -> can fail, also the ESP module may fail to fulfill the request -> there are multiple reasons this may fail.

In case of the ESP it may be possible to reset the device and reinitialize the LwESP. Anyway we may be thinking of deinit and reinit the library to save some RAM. Anyway, as of now I am fine with proposed solution using altcp_close.

vladamatena (approved):
Looks good.

For me it fails to switch from ethernet to wifi. The wifi does not connect. The wifi only works when the device boots up with wifi enabled. The other way round it works as expected, nice. 
The bug regarding wifi not connecting after boot with ethernet seems to be already present in master so I do not consider it a stopper for this pull request.

vorner:
> The bug regarding wifi not connecting after boot with ethernet seems to be already present in master

Is it reported to Jira, or shall I create a ticket? Can you reproduce it reliably? Also, I might be using some older version of firmware inside the ESP itself…

Anyway, what I see is that I click on the button to move to wifi, it freezes for ~1-2s and then it blinks blue and connects.

Anyway, I'll merge this and may have a look at the connecting when booted with ethernet. I might have booted with ESP every time and kept switching (which maybe works?).

PR #1665, opened 2021-11-03
esp: Distinguish between no password and empty one

vorner:
The esp_sta_join function asks us to provide NULL in case no password is
to be used, previously we were only able to provide "" (which might or
might not have worked).


vladamatena (approved):
Good point, empty password is not the same as no security.

PR #1678, opened 2021-11-10
Fix 100% complete print

marekmosna-prusa3d:
BFW-2144

### DIFF
--- a/lib/WUI/wui_REST_api.c
+++ b/lib/WUI/wui_REST_api.c
@@ -20,6 +20,8 @@
 extern uint32_t start_print;
 extern char *filename;
 
+static marlin_print_state_t cached_state = mpsIdle;
+
 void get_printer(char *data, const uint32_t buf_len) {
     marlin_vars_t *vars = marlin_vars();
     const char *filament_material = get_selected_filament_name();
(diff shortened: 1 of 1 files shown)


vorner (commented):
I can't see a description how the bug happened to behave and what was the cause. But reading the changes, I don't have the impression this could be the right solution to any kind of bug, because:

We have two API endpoints, implemented by these two functions. Now they share a global variable and their results depend on in which order they get called, how many times, etc. This is wrong on several levels:

* These are GET requests. According to the HTTP standard, GET requests are idempotent. That means they have _no side effects_ to the server state. So this is breaking the http standard. Besides breaking the letter of the standard, this may break in presence of caches or other such things, as they are allowed to make fewer or extra GET requests as they see fit.
* Functions that previously were (AFAIK) thread safe are no longer so.
* If you fixed some problem in one of these endpoints, someone calling just one of them will not get the fix. Moreover, if this is not really a fix in one of the endpoints, but a workaround for some specific client (what I suspect), this actually breaks it for other clients.
* Even if we accepted that there's just that one client and that it really must always, even in future versions, call the API endpoints the right number of times and in the right order to work correctly, this breaks once a client opens two instances of the client, maybe one on their computer and another on their phone.

vorner:
As a follow-up. You seem to claim this is the best solution right now. Nevertheless, it is really unclear from the code alone why this weird behavior is needed, what is its purpose and why.

Therefore, both to help the future readers understand and avoid re-introducing the problem back (for example when someone will have to solve some of the above outlined problems) and for me to understand what's happening and be therefore able to give it an approval, can you put a comment in there explaining the behavior of the bug (what exactly happens)?

vorner:
Due to the inactivity here, I've went and had a look myself on what the problem is.

The frontend repeatedly calls `/api/printer`. If the printer is printing, it also calls `/api/job` to see how long it'll be printing, etc. Once the printer is no longer printing, it doesn't call the second API endpoint to save resources. I've checked manually and we correctly state that the progress is `0%` (since there's no job in progress) after it is finished.

So the frontend knows it is done, but still shows the outdated progress bar.

1. This is _not a solution_, this is introducing another bug (because now you break the API for someone who never calls `/api/job` ‒ remember, we are supposed to be octoprint compatible to allow users re-use octoprint applications). The new bug is roughly inverse in the behavior, so they sometimes cancel each other out, but not always. This is making the situation *worse*.
2. The bug clearly lies in the frontend. Once it knows it is not printing, it shall probably hide the progress bar.
3. The current bug is minor and we are expecting to get a newer version of the web page anyway.

All in all, I'd suggest dropping this and pursue the proper fix in the web page; until then, we can live with this. There's like 5 different bugs in the web page anyway, all of similar severity.

vorner:
Ups, I didn't intent do close this. Seems like this is automatic by github, as the target branch got used up. Please check your other branches, what they did.

PR #1680, opened 2021-11-11
Rewrite bin2cc and add font generation script

MistrKernnunos:


utils/font.py
```diff
@@ -0,0 +1,102 @@
...
+        return 1
+    if not src_png_path.exists():
+        logger.error('no %s found', src_png_path.name)
+        return 1
+
+    # not using with because I want to use charset outside of its scope
+    file = open(non_ascii_chars_path.resolve())
+    chars = file.read()
+    char_set = chars.split()
+    file.close
```
dragomirecky, line 27: 1. not calling `file.close()`
2. and you can use with, as `with` does not introduce a new scope 😊 (in Python, indentation != scope)
```Python
with open(non_ascii_chars_path, 'r') as file:  # btw, no need to resolve, I believe
     char_set = file.read().split()
print(char_set)  # it works! :)

File: utils/font.py [hunk 2]
```diff
@@ -0,0 +1,102 @@
...
+
+    with Image.open(src_png_path.resolve(), formats=None) as srcPng:
+        src_png_mode = srcPng.mode
```
dragomirecky, line 30: srcPng -> src_png

utils/font.py
```diff
@@ -0,0 +1,102 @@
...
+                file.write("{")
+                file.write(" {}, {}, {}".format(hex(charInt), x, y))
+                file.write("}\n")
+
+                x += 1
+                if (x >= 16):
+                    x = 0
+                    y += 1
+            output_image.save(dst_png_path.resolve())
+            file.write("};\n")
```
dragomirecky, line 76: The magic value 6 and 16 repeats all the time. What about giving it a name? ☺️

utils/font.py
```diff
@@ -0,0 +1,102 @@
...
+                                          (y + 1) * char_height))
+
+        x = 0
+        y = 6
+        with open(str(ipp_path.resolve()), "w") as file:
+            file.write("{\n")
+            for char in char_set:
+                charBytes = char.encode("utf-32", "little")
+                charBytes = charBytes.strip(codecs.BOM_UTF32)
+                charInt = int.from_bytes(charBytes, "little")
```
dragomirecky, line 54: char_bytes, char_int

dragomirecky (approved):

PR #1684, opened 2021-11-12
BFW-2171: More accurate HTTP response headers

vorner:
Provide an internal URL with an extension, to hint the automagic
routines in the http server to more accurate/descriptive headers.

(Yes, I'm aware there'll be a collision around #1682, but the solution is obvious)


DRracer (commented):
Ideally use a `static const rsp[] = "response.json"` and use that variable, i.e. not rely on platform's unified memory model / compiler optimization to prevent loading the string into RAM (which probably won't happen).

I know the rest of the code uses the same (bad) principles...

Otherwise, ok

marekmosna-prusa3d (approved):

marekmosna-prusa3d (approved):

PR #1686, opened 2021-11-12
Fix getting current IP address

vorner:
* Use the relevant function to get it from the driver instead of relying
  on a variable.
* Fix the comment describing the said variable, to make it clear it is
  not actually the active state of the interface.

Follow-up of BFW-2240.


DRracer (dismissed):

vorner:
Renamed things, as requested on the call.

marekmosna-prusa3d (approved):

PR #1697, opened 2021-11-18
Prusa link fixes

vorner:


src/common/config.h
```diff
@@ -13,7 +13,7 @@
 //--------------------------------------
 //marlin api config
 enum {
-    MARLIN_MAX_CLIENTS = 3,    // maximum number of clients registered in same time
+    MARLIN_MAX_CLIENTS = 5,    // maximum number of clients registered in same time
```
dragomirecky, line 16: Why does this PR increase this by two, but adds one client only?

lib/WUI/http/httpd.c
```diff
@@ -2734,6 +2734,8 @@ static err_t http_find_file(struct http_state *hs, const char *uri, int is_09) {
         } else {
             uri = "401";
         }
+    } else {
+        uri = "404";
```
dragomirecky, line 2738: Not sure how this works. I guess 404 is the response's status code, but why URI? And why a string?

src/common/config.h
vorner: I'm not entirely sure (I'm not the author). I vaguely remember asking @marekmosna-prusa3d the same thing at the original review, some months ago and he said something he wanted one extra for some reason.

I can clean it up and remove the one extra. Should I?

lib/WUI/http/httpd.c
vorner: Peculiarities of the current web server implementation. Yes, we want to change the web server.

* The web server has some built-in „pages“. Some of them are specials and also invoke special status codes (like the 404 one). In fact, if you do `curl -v ip-of-the-printer/404`.
* By changing the URI and returning without providing the response, the web server loops back and uses that special page.

So, this is the very weird way the server works right now. Another reason to try to get rid of it ASAP.

dragomirecky (approved):

src/common/config.h
dragomirecky: In case we don't think it is necessary, then yes, please remove the extra one.

vorner: I've grepped the code and I'm quite confident there are just 4 being created (two wui, one in guimain and one in thread_measurement).

dragomirecky (approved):

PR #1698, opened 2021-11-18
Esp8266 fixes

vorner:


src/gui/screen_menu_lan_settings.cpp
```diff
@@ -88,6 +88,8 @@ class ScreenMenuLanSettings : public AddSuperWindow<screen_t> {
 public:
     ScreenMenuLanSettings();
 
+    uint32_t StringifyNetworkDevice(uint32_t, lan_descp_str_t);
```
dragomirecky, line 91: Don't we want to add – at least – the names of the arguments (considering there is no docstring)? By looking at this signature, I had no idea what the parameters are or what they do.

dragomirecky (dismissed):

src/gui/screen_menu_lan_settings.cpp
vorner: Added.

dragomirecky (approved):

PR #1708, opened 2021-11-24
Refactor httpd ‒ split to content & http parts

vorner:
This splits the httpd related handling into the http parts and content related parts.

The primary intention is to unblock unit & fuzz tests on the http part (which is exposed to the outside without authentication). It also draws the boundary between parts in somewhat thicker line and will make our life a little bit easier once we start replacing the HTTP server itself… but the boundary is probably in a suboptimal place and will eventually move ‒ this is what the code was doing now and I didn't want to introduce big or functional changes. In other words, this is supposed to be gradual improvement but is far from the ideal.

### DIFF
--- a/lib/WUI/http/handler_defaults.c
+++ b/lib/WUI/http/handler_defaults.c
@@ -1,0 +1,44 @@
+#include "handler.h"
+
+#include "../wui_REST_api.h"
+#include "../wui_api.h"
+
+// The get_* functions we use do exactly what we need, except they take fewer
+// params. Generate the wrappers to throw them out without too much
+// copy-pasting.
+#define GET_WRAPPER(NAME)                                                                       \
+    static void handler_##NAME(struct HttpHandlers *unused, char *buffer, size_t buffer_size) { \
+        (void)unused;                                                                           \
+        NAME(buffer, buffer_size);                                                              \
+    }
+GET_WRAPPER(get_printer);
+GET_WRAPPER(get_version);
+GET_WRAPPER(get_job);
+GET_WRAPPER(get_files);
+#undef GET_WRAPPER
+
+static const struct GetDescriptor get_handlers[] = {
+    {
+        .uri = "/api/printer",
+        .handler = handler_get_printer,
+    },
+    {
+        .uri = "/api/version",
+        .handler = handler_get_version,
+    },
+    {
+        .uri = "/api/job",
+        .handler = handler_get_job,
+    },
+    {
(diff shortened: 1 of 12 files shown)


lib/WUI/http/handler_defaults.c
```diff
@@ -0,0 +1,60 @@
...
+#define GET_WRAPPER(NAME)                                                                       \
+    static void handler_##NAME(struct HttpHandlers *unused, char *buffer, size_t buffer_size) { \
+        (void)unused;                                                                           \
+        NAME(buffer, buffer_size);                                                              \
+    }
+GET_WRAPPER(get_printer);
+GET_WRAPPER(get_version);
+GET_WRAPPER(get_job);
+GET_WRAPPER(get_files);
+#undef GET_WRAPPER
```
dragomirecky, line 16: /question/ As I understand it, the `lib/WUI/http` directory contains implementation of a generic http client/server (which shouldn't have any knowledge of our specific endpoints). In such case, wouldn't it be cleaner to move this file under `lib/WUI` and name it – for example – `wui_handlers.c`?

lib/WUI/http/handler_defaults.c
```diff
@@ -0,0 +1,60 @@
...
+        .uri = "/api/job",
+        .handler = handler_get_job,
+    },
+    {
+        .uri = "/api/files",
+        .handler = handler_get_files,
+        .prefix = true, // Any uri starting with /api/files is fine (for now).
+    },
+    {}, // Sentinel
+};
```
dragomirecky, line 49: Just curious – do we use the `anonymous` flag somewhere? Is it the intention that it exists and isn't used anywhere (from what I've seen)

dragomirecky (dismissed):

lib/WUI/http/handler_defaults.c
vorner: Well, the file structure around WUI is its own chapter (some parts are in `src`, some in `lib`…) and it's not even fully true that the http server would be fully agnostic of the content (the POST part ☹). But I guess this is as good time as any to help it a bit.

lib/WUI/http/handler_defaults.c
vorner: Currently, no, not yet. I expect to use it in tests and mocks. We also have some files as static files and these are anonymous, but that doesn't go through this mechanism yet.

dragomirecky (approved):

PR #1713, opened 2021-11-26
One click upload

vorner:
Disabling the start-after-upload feature temporarily, replacing by showing a one-click-print dialog (the latter is not temporary), as agreed.

The dialog appears, but usually shows the wrong file. This is caused by not setting the date/time of the new file correctly. I've tried @dragomirecky 's patch for teaching fatfs to set it correctly, but it doesn't seem to work for me:

* Only one of the three times (ctime, mtime, atime) is set, the rest is still at 1.1. 1980.
* The one that _is_ set is set to 1.1.2098.

Which doesn't actually fix the right-file-to-show problem anyway. I'm leaving investigation of that part for another task.

BFW-2300

### DIFF
--- a/lib/WUI/wui_api.c
+++ b/lib/WUI/wui_api.c
@@ -22,6 +22,7 @@
 #include <time.h>
 #include <string.h>
 #include <stdio.h>
+#include <stdatomic.h>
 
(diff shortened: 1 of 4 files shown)


dragomirecky (approved):
looks good 👍

PR #1717, opened 2021-11-26
Rewrite png2cc script

MistrKernnunos:


utils/png2cc.py
```diff
@@ -0,0 +1,119 @@
...
+        dst_file.write("};\n")
+
+
+def bincc_folder(src_folder: Path, dst_folder: Path):
+    if not dst_folder.exists():
+        logger.error('no {} found', dst_folder.name)
+        return 1
+
+    for filename in os.listdir(src_folder.resolve()):
+        if filename.endswith(".png"):
```
vorner, line 32: If I wanted to be an evil user, I'd either create a filename with uppercase `.PNG` or a directory with `.png` name. But as this is a development utility, it doesn't have to be perfect and bullet proof, so it can be left as it is, it's good enough ‒ saying it just so you are aware.

utils/png2cc.py
```diff
@@ -0,0 +1,119 @@
...
+        logger.error(
+            "To convert folder of pngs the --folder flag must point to output folder"
+        )
+        return 1
+
+    if Path(args.src).is_dir() and Path(args.folder).is_dir():
+        tmp_name = "tmp"
+        x = 0
+        while Path(tmp_name + str(x)).exists():
+            x += 1
```
vorner, line 98: Would there be a better name than `x`? Also, is that variable used anywhere?

utils/png2cc.py
```diff
@@ -0,0 +1,119 @@
...
+        tmp_dir = Path(tmp_name + str(x)).resolve()
+        os.mkdir(tmp_dir)
```
vorner, line 100: This sounds a bit like the job of https://docs.python.org/3/library/tempfile.html

vorner (changes requested):
There are some minor things in the code, but considering the purpose of the tool, they are probably fine ‒ you can have a look at them and fix or not :shrug: 

But I'd prefer if you could remove that stray „fix USB“ commit from the branch first before merging.

vorner (approved):

PR #1721, opened 2021-12-01
Httpd tests

vorner:
More tests will come in future pull requests.

There are some shortcuts and such, some of them are not exactly nice, but I hope this is not such a problem in test code.

If someone knows what better to do about the cmake thing, I'm all ears.


lib/WUI/http/httpd_post.c
```diff
@@ -354,7 +354,7 @@ void httpd_post_finished(void *connection,
     if (connection != NULL && status_code == 200) {
         strncpy(response_uri, endpoints[http_post_uri_file_index], response_uri_len);
     } else {
-        snprintf(response_uri, response_uri_len, "/%ld", status_code);
+        snprintf(response_uri, response_uri_len, "/%" PRIu32, status_code);
```
DRracer, line 357: this is a nice example of formatting macros, please report to the rest of the team :+1:

tests/stubs/cmsis_os.h
```diff
@@ -0,0 +1,8 @@
+#pragma once
```
DRracer, line 1: Do you think this stub is usable for other than unit-tests? You propose not to place it into the `tests/unit` directory subtree.

lib/WUI/http/httpd_post.c
vorner: Report where? Meeting? Some channel? Yes, I'd like to set up some kind of quality guild, but I haven't gotten around that yet.

tests/stubs/cmsis_os.h
vorner: It's the nearest `stubs` directory I've found. I could create `tests/unit/stubs`… on the other hand, once we have the fuzzers, they are not exactly unit tests and they would probably like to reuse it.

DRracer: ok, fair enough

DRracer (approved):

PR #1722, opened 2021-12-01
New prusa link

vorner:
This brings newer version of the PrusaLink web pages.

# Stuff in the whole branch

I'm not sure how the original files were created, but I've found an ages old branch that solves the packaging. I've had to make few changes/fixes.

However:
* I'm not sure about the history of the branch. Should I squash it (in some way preserving authors) or should I at least remove the merge commits?
* I have a feeling the CMake changes don't work properly. I think it doesn't always pick up changes. Can someone suggest what's wrong?

# Changes in the web

I've cut down the things as much as possible (things we don't support yet; I'm not sure where the current configuration comes from). There still are some things that we don't support; I'd like to merge it this way now to easily experiment with it and negotiate the best ways to remove things that we don't want.

But the added form to upload gcode works :tada:


dragomirecky (commented):
I am quite unsatisfied by the state of the `wui.py` script. It does a simple thing in quite an unreadable way. 

I know you are not the author – I remember doing a code review for this script a long time ago, having some reservations which weren't fixed, and the PR wasn't merged. And now it's back 😁

My central reservation is that it tries to do too much. I imagine a simpler script that takes an input file and an output file—translating the generic input file into a `.c` output file. Then we can simply use this script from CMake to translate all the files we need. 

It is comparable to a compiler. We don't give the compiler the whole folder with our project. We give it a file to compile and get back an object file. Another program is responsible for orchestrating the build of the whole project.

Having said that, I am not going to block the PR, as this is most likely going to be a temporary solution (will be replaced by having the WUI files on a filesystem on the external flash).

utils/wui.py
```diff
@@ -0,0 +1,95 @@
...
+    print('#ifndef FSDATA_ALIGN_POST')
+    print('#define FSDATA_ALIGN_POST')
+    print('#endif')
+    print('#if FSDATA_FILE_ALIGNMENT == 2')
+    print('#include "fsdata_alignment.h"')
+    print('#endif\n')
+
+
+def to_hex(data):
+    return ','.join(map(lambda b: hex(b), data)) + ','
```
dragomirecky, line 37: No need for this. See [bytes.hex()](https://docs.python.org/3/library/stdtypes.html#bytes.hex)

dragomirecky (approved):
looks much better now 👍

utils/wui.py
vorner: I'm afraid not. This thing also adds the separators (the commas) between the values. The `sep` support got added in python 3.8 and we claim in README.md that we support python 3.6.

dragomirecky (approved):

PR #1726, opened 2021-12-02
BFW-2284: Avoid generating short api keys

vorner:
* Avoid \0 in the generated API-key (which effectively made it shorter)
* Avoid certain similar-looking pairs of characters (more readable API
  key would be better, but this is very little work).


DRracer (approved):
Approving.
- I like the fact, that similarly looking characters have been removed
- I don't like the fact that some of the special characters have been removed as well, but that will make life easier for users suffering from touch screen equipped hand held computers (some people call them smart phones)

PR #1727, opened 2021-12-02
Footer items

RadekVana:


src/gui/footer/footer_item_axis.cpp
```diff
@@ -0,0 +1,73 @@
...
+FooterItemAxisZ::FooterItemAxisZ(window_t *parent)
+    : AddSuperWindow<IFooterItemAxis>(parent, IDR_PNG_x_axis_16x16, static_readValue) {
+}
+float FooterItemAxisZ::static_readValue() {
+    return marlin_vars()->pos[2];
+}
+
+//TODO this should reflect MBL, aso precision must be higher
+string_view_utf8 FooterItemZHeigth::static_makeView(float value) {
+    static std::array<char, 7> buff; // the buffer must be static, because string_view does not copy data inside itself
```
DRracer, line 55: Is it safe to have the buffer static? I assume it is as there is only one Z-height in the whole printer...

src/gui/footer/footer_item_axis.cpp
```diff
@@ -0,0 +1,73 @@
...
+}
+
+FooterItemAxisZ::FooterItemAxisZ(window_t *parent)
+    : AddSuperWindow<IFooterItemAxis>(parent, IDR_PNG_x_axis_16x16, static_readValue) {
+}
+float FooterItemAxisZ::static_readValue() {
+    return marlin_vars()->pos[2];
+}
+
+//TODO this should reflect MBL, aso precision must be higher
```
DRracer, line 53: MBL? Really? Users complain that we display raw Z-height instead of the logical one (i.e. MBL shall NOT be considered here)

src/gui/footer/footer_item_fans.cpp
```diff
@@ -0,0 +1,38 @@
...
+ */
+#include "footer_item_fans.hpp"
+#include "marlin_client.h"
+#include "resource.h"
+#include <algorithm>
+
+static constexpr int max_rpm = 99999;
+
+string_view_utf8 IFooterItemFan::static_makeView(int value) {
+    static std::array<char, sizeof("99999rpm")> buff; // the buffer must be static, because string_view does not copy data inside itself
```
DRracer, line 14: `sizeof("99999rpm")` interesting construct :+1:

src/gui/footer/footer_item_fans.cpp
```diff
@@ -0,0 +1,38 @@
...
+    buff.fill('\0');
+    int value_clamped = std::clamp(value, 0, max_rpm);
```
DRracer, line 16: uint?

src/gui/footer/footer_item_fans.cpp
```diff
@@ -0,0 +1,38 @@
...
+    snprintf(buff.data(), buff.size(), "%drpm", value_clamped);
```
DRracer, line 17: if uint, then a better format specifier shall be used (%u AFAIK)

src/gui/footer/footer_item_axis.cpp
```diff
@@ -0,0 +1,73 @@
...
+string_view_utf8 FooterItemZHeigth::static_makeView(float value) {
+    static std::array<char, 7> buff; // the buffer must be static, because string_view does not copy data inside itself
+    float value_clamped = std::clamp((float)value, (float)MenuVars::GetMaximumZRange()[0], (float)MenuVars::GetMaximumZRange()[1]);
+    int printed_chars = snprintf(buff.data(), buff.size(), "%.2f", (double)value_clamped);
+
+    if (printed_chars < 1) {
```
DRracer, line 59: What is this? Can't we achieve the same (safe) behavior by tuning the format string correctly? Like %-4.2f?
Moreoveor - are you sure %f never produces an exponential representation of the float value?

src/gui/footer/footer_item_live_z.hpp
```diff
@@ -14,5 +14,6 @@ class FooterItemLiveZ : public AddSuperWindow<FooterIconText_IntVal> {
     static int static_readValue();
 
 public:
+    static string_view_utf8 GetName() { return _("LiveZ"); }
```
DRracer, line 17: Live Z (use a space)

DRracer (changes requested):

src/gui/footer/footer_item_axis.cpp
RadekVana: yes, and if it is not they would share same value

src/gui/footer/footer_item_axis.cpp
RadekVana: we have 2 menu items for Z axis, 1 should be with MBL, 1 without
Michal also mad 2 different icons

src/gui/footer/footer_item_axis.cpp
RadekVana: I think NAN will produce 0 bytes written
I just copied code from another item

RadekVana: it fills rest of the buffer with zeroes
I don't remember why is it important

src/gui/footer/footer_item_axis.cpp
DRracer: ok

src/gui/footer/footer_item_axis.cpp
DRracer: ok

DRracer (approved):
I still don't like the way Z-height value is being padded with zeros

PR #1731, opened 2021-12-02
Puts SNTP client back to WUI

michalrudolf:
BFW-2315
Add sntp_client_step back to WUI cycle -> real time will be updating regulary when connected to internet.


vorner (approved):
Currently, the SNTP client is not capable of running over WiFi/ESP. Nevertheless, this definitely improves the current situation, so let's put it in. The proper fix will be a bit more work.

PR #1746, opened 2021-12-03
BFW-2223: At least try aborting connections

vorner:
In case someone asks for aborting an ESP connection, at least try
closing it instead of doing nothing at all. That's still not perfect, as
closing might fail, but a better situation than now.

We still need some better architectural solution.


vladamatena (approved):
Do provozu asi lepsi nez zadna implementace. Ale neni to upne v pohode, uz proto ze ten close allokuje data (jako vsechny LwESP funkce).

vorner:
jj, to tam píšu, jak zde v ticketu, tak v poznámce.

PR #1748, opened 2021-12-03
Esp fault fix

vladamatena:
Partial fix. This seems to make the situation a bit better, but still there are some crashes.


include/esp/lwesp_ll_buddy.h
```diff
@@ -68,7 +68,7 @@ extern espr_t esp_flash_initialize();
 // be copied from the RX buffer to ESP mem before the buffer overflows.
 #define RX_BUFFER_LEN 0x1000
 #if !defined(ESP_MEM_SIZE)
-    #define ESP_MEM_SIZE 0x1000
+    #define ESP_MEM_SIZE 0x2000
```
vorner, line 71: Is that the size of the ESP's heap? Can we afford another 4k of RAM to be given to ESP? @DRracer :question:

lib/ESP/prusa/include/esp_tcp.h
```diff
@@ -49,6 +49,9 @@
 // It would be better to obtain this from ESP but there seems to be no way
 #define ESP_TCP_MSS 536
 
+// #define ALTCP_ESP_DEBUG      1
+#define ALTCP_ESP_EPCB_MAGIC 0xdeadf00d
```
vorner, line 53: Should we enable debug by default now, or is there a downside to that (apart from little overhead)?

lib/ESP/prusa/src/system/esp_tcp.c
```diff
@@ -127,25 +125,57 @@ static err_t espr_t2err_t(const espr_t err) {
...
+static void ASSERT_EPCB(esp_pcb *epcb) {
+    #ifdef ALTCP_ESP_DEBUG
+    if (epcb->magic_start != ALTCP_ESP_EPCB_MAGIC) {
+        ALTCP_ESP_DEBUG_FN("Failed: CORRUPTED EPCB START MAGIC");
+    }
+
+    if (epcb->magic_end != ALTCP_ESP_EPCB_MAGIC) {
+        ALTCP_ESP_DEBUG_FN("Failed: CORRUPTED EPCB END MAGIC");
+    }
+    #endif
```
vorner, line 137: Should we actually assert so we can catch the problem closer to when it happened, instead of just writing it to logs?

vorner (changes requested):
I have few questions/suggestions (mostly if we could be a bit more aggressive about ensuring the consistency).

Also, can you check the CI? It's red (seems like formatting).

lib/ESP/prusa/include/esp_tcp.h
vladamatena: Overhead is the only problem. I just didn't want the overhead to become part of the system. Maybe we can enable this as debugging is bound to quite heave prints. This limits chances it will stay enabled forever.

include/esp/lwesp_ll_buddy.h
vladamatena: Separated this into a commit for easier revert.

lib/ESP/prusa/src/system/esp_tcp.c
vladamatena: Checks turned into asserts and modified to avoid false positives.

vorner (approved):
I probably should add the define to my own compilation script, so my firmware does the checking.

Anyway, thank you :+1:

PR #1750, opened 2021-12-04
Add Integration tests

dragomirecky:
First version of integration tests running on simulator. 

It adds M505 gcode, which enables us to set the EEPROM content in a easy-to-use way (it's not an ideal solution, but should be good enough until we replace the eeprom implementation)


tests/integration/test_basic_examples.py
```diff
@@ -0,0 +1,46 @@
...
+
+async def test_max_temp_error_on_bed(printer: Printer):
+    await screen.wait_for_text(printer, 'HOME')
+    await temperature.set(printer, Thermistor.BED, 230)
+    await screen.wait_for_text(printer, 'MAXTEMP ERROR')
+
+
+async def test_web_interface_is_accessible(printer: Printer,
+                                           wui_client: aiohttp.ClientSession):
+    await screen.wait_for_text(printer, 'HOME')
```
vorner, line 29: That might not be enough :-(. The home screen might appear sooner than the whole DHCP handshake happens.

Well, maybe not on tests, hopefully.

tests/integration/test_basic_examples.py
```diff
@@ -0,0 +1,46 @@
...
+    response = await wui_client.get('/')
+    assert response.ok
+
+
+async def test_filebrowser_shows_files(printer_factory, printer_flash_dir):
+    # prepare files on the flash drive
+    filenames = ['first file.gcode', 'second file.gcode']
+    for filename in filenames:
+        (printer_flash_dir / filename).write(b'')
```
vorner, line 38: Huh? This actually works in python? :-O

src/common/eeprom.cpp
```diff
@@ -469,6 +469,16 @@ const char *eeprom_get_var_name(enum eevar_id id) {
...
 
+bool eeprom_find_var_by_name(const char *name, enum eevar_id *var_id_out) {
+    for (int i = 0; i < (int)EEPROM_VARCOUNT; i++) {
+        if (strcmp(eeprom_map[i].name, name) == 0) {
+            *var_id_out = (enum eevar_id)i;
+            return true;
+        }
+    }
+    return false;
+}
```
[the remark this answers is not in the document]
dragomirecky: The eeprom's interface is written in C...

src/common/eeprom.cpp
```diff
@@ -494,6 +504,30 @@ int eeprom_var_format(char *str, unsigned int size, enum eevar_id id, variant8_t
...
+    case EEVAR_LAN_IP4_DNS1:
+    case EEVAR_LAN_IP4_DNS2:
+    case EEVAR_WIFI_IP4_ADDR:
+    case EEVAR_WIFI_IP4_MSK:
+    case EEVAR_WIFI_IP4_GW:
+    case EEVAR_WIFI_IP4_DNS1:
+    case EEVAR_WIFI_IP4_DNS2: {
+        uint8_t bytes[4];
+        sscanf(str, "%hhu.%hhu.%hhu.%hhu", &bytes[0], &bytes[1],
+            &bytes[2], &bytes[3]);
```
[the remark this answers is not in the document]
dragomirecky: Yep, good idea, fixed

tests/integration/test_basic_examples.py
dragomirecky: Yep, might not be. Currently the best (and only) channel „fw->tests“ we have here is the OCR of the text on the screen. So let's hope it is gonna be enough 🙂

tests/integration/test_basic_examples.py
dragomirecky: Python 3.4 – [object oriented filesystem](https://docs.python.org/3/library/pathlib.html) ☺️

tests/integration/test_basic_examples.py
vorner: Seems like this is fine. I guess that the DHCP exchange is fast enough in the simulator, and in addition, the client will try retransmitting the ARP/Syn packet for a while before giving up.

src/common/eeprom.cpp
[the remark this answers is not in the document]
vorner: I see. That's sad. I've also hit few places where it would really benefit by turning stuff to more C++-style, but that would ripple down to several still C places that use it :-(.

utils/simulator/simulator.py
```diff
@@ -0,0 +1,198 @@
...
+        raise NotImplementedError()
+
+    async def encoder_release(self):
+        raise NotImplementedError()
+
+    async def encoder_rotate_left(self):
+        await self.command(f'encoder-input::Twist(1)')
+
+    async def encoder_rotate_right(self):
+        await self.command(f'encoder-input::Twist(1)')
```
vorner, line 156: Isn't it strange that both left and right are the same `Twist(1)`? Shouldn't one of them be `-1`?

utils/simulator/simulator.py
```diff
@@ -0,0 +1,198 @@
...
+    #
+
+    async def encoder_click(self):
+        await self.command('encoder-input::Push()')
+
+    async def encoder_push(self):
+        raise NotImplementedError()
+
+    async def encoder_release(self):
+        raise NotImplementedError()
```
vorner, line 150: Are these two harder to implement, or is it worth just throwing them in while we are at it?

utils/simulator/simulator.py
```diff
@@ -0,0 +1,198 @@
...
+        temp_str = await self.command(f'thermistor{thermistor_idx}::GetTemp()',
+                                      readline=True)
+        return float(temp_str)  # type: ignore
+
+    #
+    # network primitives
+    #
+
+    def network_ip_address_get(self, interface: NetworkInterface) -> str:
+        raise NotImplementedError()
```
vorner, line 195: This might actually be easier to just set to static in eeprom than to _get_ it later on.

utils/simulator/simulator.py
dragomirecky: Good catch :) Fixed

utils/simulator/simulator.py
dragomirecky: Yes, there is no support in the simulator yet. My intention is to have them here to clearly differentiate push/release and click.

utils/simulator/simulator.py
dragomirecky: Actually, let's delete this for now. Originally this was something I worked on but then I decided to take another path...

dragomirecky: Deleted...

vorner (approved):
OK, let's get it in :+1: 

I'll follow with a separate pull request with few tests for the PrusaLink soon after.

PR #1754, opened 2021-12-06
BFW-2295: Silence runtime warning about unknown ESP event.

laloch:
PR #1729 made the FW to restore ESP configuration to the factory defaults using "AT+RESTORE" command. The command triggers ESP_EVT_RESTORE which in turn triggers scary debug warning about unknown/unhadled event. This adds a dummy handler to silence the warning.

### DIFF
--- a/lib/WUI/netdev.c
+++ b/lib/WUI/netdev.c
@@ -158,6 +158,10 @@ esp_callback_func(esp_evt_t *evt) {
         _dbg("ESP_EVT_RESET");
         break;
     }
+    case ESP_EVT_RESTORE: {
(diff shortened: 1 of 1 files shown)


vorner (approved):
I'm OK with silencing it, that's fine.

But, thinking about that, don't we want to wait for the confirmation of restore before doing further things? Like, wait for that before sending more commands?

laloch:
> I'm OK with silencing it, that's fine.
> 
> don't we want to wait for the confirmation of restore before doing further things? Like, wait for that before sending more commands?

Yup, the same applies to reset command. It should be possible to execute as a blocking command, but it doesn't work. It's already on my to-do list.

PR #1765, opened 2021-12-08
Fix ESP ALTCP close

vladamatena:
- Fix close of listening connections.
- Fix false positive report of broken network stack links
  for listening connections.

### DIFF
--- a/lib/ESP/prusa/src/system/esp_tcp.c
+++ b/lib/ESP/prusa/src/system/esp_tcp.c
@@ -142,12 +142,15 @@ static void ASSERT_EPCB_VALID(esp_pcb *epcb) {
 }
 
 static void ASSERT_EPCB(esp_pcb *epcb) {
+    #ifdef ALTCP_ESP_DEBUG
     ASSERT_EPCB_VALID(epcb);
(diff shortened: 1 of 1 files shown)


lib/ESP/prusa/src/system/esp_tcp.c
```diff
@@ -669,21 +697,15 @@ static err_t altcp_esp_close(struct altcp_pcb *conn) {
...
-        ALTCP_ESP_DEBUG_FN("Failed to close connection: %d", err);
-        return espr_t2err_t(err);
+    if (epcb->listen_port) {
+        return close_listen_mode(epcb);
     }
 
-    epcb->closing = 1;
-
-    return ERR_OK;
+    return close_normal_mode(epcb);
```
vorner, line 708: Minor style nitpick: I'd prefer ` else { ` here instead of relying on that return above exiting the function.

vorner (changes requested):
Just some readability nitpick

lib/ESP/prusa/src/system/esp_tcp.c
vladamatena: Fixed, I admit it looks much better with explicit else. (One more point for disabling no-else-return check in pylint)

vorner (approved):

PR #1769, opened 2021-12-10
BFW-2268: Don't return crap from the /api/*

vorner:
* Don't try to make things up there based on heuristics. While the data
  is sometimes invalid, it needs to be fixed at the source.
* Don't include made-up values or values we don't have (according to the
  spec, all the fields are optional anyway).
* Fix types of returned values (bools are bools, not ints).
* Fix print progress (percentage according to docs) and
  estimatedPrintTime (which is a total time).
* Get rid of unused fields that are not present at the spec and make no
  sense in that position.

Tests:
* Integration test for this, including one during a "print"
* It sometimes fails, due to crap being returned in the time_to_end
  variable.


vorner:
Note that I've changed the `completion` field from fragment (number between 0..1) to percents. This seems to be correct according to the spec:

https://github.com/prusa3d/Prusa-Link-Web/blob/master/spec/openapi.yaml#L1369-L1373
https://docs.octoprint.org/en/master/api/datamodel.html#sec-api-datamodel-jobs-progress

Nevertheless, the PrusaLink page expects the fragment and shows stuff like 500% and such…. I'm trying to investigate.

tests/integration/test_prusa_link.py
```diff
@@ -67,6 +69,91 @@ async def test_auth(printer: Printer, wui_client: aiohttp.ClientSession):
         await response.json()  # Tries to parse and throws if fails decoding
 
 
+async def test_status_responses(printer_factory, printer_flash_dir, data_dir):
```
dragomirecky, line 72: Generally, the goal is to make test cases as small as possible. So that when one part (test case) fails, the others are still gonna be tested.

Consider splitting this into several test cases (eg `test_api_version_endpoint`, `test_api_printer_endpoint`, etc).

dragomirecky (approved):
looks great 👍

dragomirecky (approved):

PR #1770, opened 2021-12-10
BFW-2357: Workaround for 2nd packet not delivered

vorner:
Makes sure to send only one packet at a time, not having a second packet
at all (not until it is ACKed).

Considered a temporary workaround until the root cause is found (both
for why the packets are reordered and why the second ACK confuses the
printer).


vladamatena (approved):
Za me ok, zvlast kdyz to vyresi pomoci defines a nemusi se to hackovat nekam do kodu.

PR #1771, opened 2021-12-13
Fix font.py script for non-ascii 

RunZeJustin:
Some unicode will be stripped by mistake.
e.g.
一 b'\xff\xfe\x00\x00\x00N\x00\x00', should be b'\x00N\x00\x00', but stripped as b'N'
？ b'\xff\xfe\x00\x00\x1f\xff\x00\x00', should be b'\x1f\xff\x00\x00', but stripped as b'\x1f'


MistrKernnunos (approved):
Thanks for catching this bug, I never came across this one.

PR #1777, opened 2021-12-16
Upload state

vorner:
This rewrites handling of the upload with tracking state as an automaton and assuming anything can be broken up between packets at any time.

This fixes several bugs (see the commit message of the last commit).

It's still far from being perfect. There's some amount of working around the bad design in the http server and other awkward things. It still doesn't handle multiple parallel uploads and uses global variables :cry:. It also introduces another dynamic allocation to the upload path (there was one, now there are two). Also, eventually, I'd like to do something about the `multipart_parser.c` thing ‒ it _almost_ works, but doesn't accept all the things that are valid according to the standard. I've stopped myself there, as they are valid but highly _unlikely_, so it's probably OKish for now.

I'm open to suggestions how to improve it, what other scenarios to test (either in unit tests or integration tests), etc.

### DIFF
--- a/lib/WUI/http/upload_state.cpp
+++ b/lib/WUI/http/upload_state.cpp
@@ -84,13 +84,15 @@ const constexpr size_t MAX_SUBTOKEN = 19;
  */
 class Accumulator {
 private:
-    size_t len = 0;
     string_view looking_for;
-    bool sensitive = true;
+    uint8_t len : 7;
+    bool sensitive : 1;
     array<char, MAX_SUBTOKEN> buffer;
 
 public:
-    Accumulator() {
+    Accumulator()
+        : len(0)
+        , sensitive(true) {
         // Initialize to avoid UB due to reading/moving uninit values.
         memset(buffer.begin(), 0, buffer.size());
     }
@@ -135,7 +137,7 @@ public:
             len += 1;
         }
 
-        const size_t suffix_len = std::min(len, looking_for.size());
+        const size_t suffix_len = std::min(static_cast<size_t>(len), looking_for.size());
         string_view suffix(buffer.end() - suffix_len, suffix_len);
 
         if (suffix == looking_for) {
@@ -186,17 +188,20 @@ private:
      * This helps tracking if we start a new token or if this is a continuation
      * of the same one.
      */
-    enum class TokenType {
+    // Note: Use of 'enum class' generates a warning at the place of the
+    // bitfield usage below, even though it is big enough.
+    enum TokenType {
         NoToken,
         HeaderField,
         HeaderValue,
-        Data,
+        DataToken,
+        // Warning: Check the bitfield sizes if adding values.
     };
 
     /*
      * The states of the automaton.
      */
-    enum class State {
+    enum State {
         NoState,
         CheckHeaderIsCDisp,
         IgnoredHeader,
@@ -205,6 +210,7 @@ private:
         ReadPartName,
         Data,
         Done,
+        // Warning: Check the bitfield sizes if adding values.
     };
 
     /*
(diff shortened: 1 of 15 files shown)


lib/WUI/http/upload_state.cpp
```diff
@@ -0,0 +1,649 @@
...
+    }
+    /*
+     * Access the internal buffer.
+     *
+     * This can be used to recycle the buffer for different purposes (to save
+     * memory) while not being in used. The buffer or the functionality of
+     * looking for something are mutually exclusive (use one or the other, not
+     * both at the same time).
+     */
+    array<char, MAX_SUBTOKEN> &borrow_buf() {
```
DRracer, line 180: reusing memory is a good principle, but I'd recommend to secure the access to the memory area with some simple locking mechanism, which would assert in debug - so that the accidental usage of this buffer from multiple spots at once gets detected at least at runtime in debug mode if not at compile time.

lib/WUI/http/upload_state.cpp
```diff
@@ -0,0 +1,649 @@
...
+    Part part : 2;
+    bool have_valid_filename : 1;
+    /*
+     * The gcode_start hook has been called and we need to eventually call the finish hook too.
+     *
+     * This may get reset back to false in case the finish hook is called or
+     * when there's an error from hook (the hooks should know they erorred out
+     * and clean up their state as needed).
+     */
+    bool init_done : 1;
```
DRracer, line 269: Have you checked that the compiler is capable of reordering the member variables/bits to minimze the wasted mem space? And, yes, I read your intro design description, that you "hope" for it.

DRracer (commented):

lib/WUI/http/upload_state.cpp
vorner: I've added something, but please check if you like it. I'm not entirely sure about it, since:

* The scope in which this class is used is local and rather small. So it's possible to check manually. Furthermore, messing it up would very likely break tests.
* The current solution adds some complexity and consumes more memory (not sure how much, alignment issues are in play).

In case one would get it wrong, it would throw an exception. I assume this would terminate the program with something in our case?

I was also thinking about a bool tracking „Am I borrowed?“ and returning some kind of RAII object to unlock it on drop. But then the caller would need to _store that RAII object_, which is problematic ‒ currently, it doesn't store the borrow, but pointers into it.

This is the place where I miss Rust's enums (that are somewhat close to the std::variant thing, but they have language syntax support to make them comfortable and can store the index of currently active variant in „niches“ ‒ bits that are unused in the actual stored values; besides the order of fields is unspecified, unlike C/C++, which gives the compiler the freedom to optimize by reordering them ‒ the thing I'm hoping to give the compiler by hiding the implementation comes for free there).

lib/WUI/http/upload_state.cpp
vorner: I haven't and I don't think there's a reasonably easy way to do so.

What I'm sure of is, if I put it into a header file, the compiler is denied even the chance, because order of variables in structs is _guaranteed_ in C/C++ and if it's in a header file, there might be some other code that relies on it. By hiding the implementation, I'm giving the compiler more freedom, because it now knows „nobody is watching“, so it can „break the rules“. I would _like_ it to do the optimization with optimizing the members (for example, I'm pretty sure not all combination of states/tokens/parts is possible), but there's likely a huge number of other optimizations it may unlock (or not).

I'm still trying to provide the bit fields and order the members according to their size/alignment so even the „right“ representation would be somewhat OK.

DRracer: OK, good enough for me :wink:

lib/WUI/http/upload_state.cpp
DRracer: :thinking: the variant solution looks too heavy. Morover, we have exceptions disabled, so who knows what it does at runtime. But the principle is there - having some kind of handle which checks if the count of simultaneous requests for the handle is <= 1.

DRracer (approved):
approving, all my concernsof this PR have been explained/improved.

PR #1780, opened 2021-12-17
Add dns name resolving to syslog

MistrKernnunos:


src/syslog/syslog.c
```diff
@@ -11,12 +12,10 @@ static void report_error(syslog_transport_t *transport, const char *message_pref
     }
 }
 
-bool syslog_transport_open(syslog_transport_t *transport, const char *ip_address, int port) {
+bool syslog_transport_open_ip4(syslog_transport_t *transport, ip_addr_t ip_address, int port) {
```
dragomirecky, line 15: This isn't part of the public interface so it should be static :)

src/syslog/syslog.c
```diff
@@ -64,3 +63,22 @@ void syslog_transport_close(syslog_transport_t *transport) {
     }
     transport->is_open = false;
 }
+void dns_found_cb(const char *name, const ip_addr_t *ipaddr, void *callback_arg) {
```
dragomirecky, line 66: /nitpick/ Empty line between functions please 😁

dragomirecky: And I don't see the function being used anywhere. I guess you've forgotten to remove it :)

src/syslog/syslog.c
```diff
@@ -64,3 +63,22 @@ void syslog_transport_close(syslog_transport_t *transport) {
...
+        // ip address already cached or the name was valid ip address
+        return syslog_transport_open_ip4(transport, addr, port);
+    }
+    if (res == ERR_INPROGRESS) {
+        // dns lookup queued, wait for it to finish
+        return false;
+    }
+    // failed return false
+    return false;
+}
```
dragomirecky, line 84: What would be -very- useful are logs for events like 
- DNS failed to resolve
- DNS resolved to X

src/syslog/syslog.c
```diff
@@ -64,3 +63,28 @@ void syslog_transport_close(syslog_transport_t *transport) {
...
+            transport->last_resolve_state = Resolved;
+        }
+        return syslog_transport_open_ip4(transport, addr, port);
+    }
+    if (res == ERR_INPROGRESS) {
+        // dns lookup queued, wait for it to finish
+        log_info(Syslog, "DNS resolving for host %s was enqueued", host);
+        transport->last_resolve_state = Progress;
+        return false;
+    }
```
vorner, line 85: I wonder if this will always converge to success. I assume you try to look again from time to time and rely on having a working DNS cache. Nevertheless:

* Do we _have_ the cache? How large?
* What if other tasks also resolve and kick this one out?
* What if the DNS record has TTL 0 and forbids caching?

MistrKernnunos: Yes we have the cache iand it stores two IPs. 
If the IP gets kicked out, it will be queried once more and kickout another IP
If it has forbiden caching we wont get the IP.

vorner: I guess I won't block it over these concerns… this is debugging aid, not really production code. But the fact we abuse the cache this way (eg. try again later instead of using the callback) seems a bit wrong.

vorner (approved):

PR #1784, opened 2021-12-20
GUI: Change sequence and options of dialogs at print preview

espr14:
BFW-2375

### DIFF
--- a/src/gui/screen_print_preview.cpp
+++ b/src/gui/screen_print_preview.cpp
@@ -19,37 +19,97 @@ const uint16_t menu_icons[2] = {
     IDR_PNG_stop_58px,
 };
 
-static void print_button_press() {
-    bool approved = true;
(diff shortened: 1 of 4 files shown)


src/gui/screen_print_preview.cpp
```diff
@@ -19,37 +19,97 @@ const uint16_t menu_icons[2] = {
...
+}
+
+static void print_button_pressed() {
+    GCodeInfo &gcode = GCodeInfo::getInstance();
+    if (!check_printer_type(gcode)
+        || !check_filament_presence(gcode)
+        || !check_filament_type(gcode)) {
+        Sound_Play(eSOUND_TYPE::SingleBeep);
+        Screens::Access()->Close();
+        return;
```
MistrKernnunos, line 98: Do we want to check the if the print is possible when user clicks on the prin button?

MistrKernnunos (approved):

PR #1785, opened 2021-12-20
GUI: Change dynamic button size algorithm

espr14:
Radio button sizes were set according to text size ratio. Very short text had very small button width. New algorithm sets the same padding size for all buttons which looks more natural.
BFW-2377

### DIFF
--- a/src/guiapi/include/Rect16.h
+++ b/src/guiapi/include/Rect16.h
@@ -586,9 +586,9 @@ public:
     /// @param[out] splits[] buffer to fill of splitted Rect16
     /// @param[out] spaces[] buffer to fill of spaces between Rect16 splits
     /// @param[in] count number of splits
-    /// @param[in] spacing with of spaces between rectangle's splits (optional = 0)
-    /// @param[in] ratio[] ratio of wanted splits (optional = nullptr)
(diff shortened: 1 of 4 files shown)


src/guiapi/include/Rect16.h
```diff
@@ -586,7 +586,7 @@ class Rect16 {
     /// @param[out] splits[] buffer to fill of splitted Rect16
     /// @param[out] spaces[] buffer to fill of spaces between Rect16 splits
     /// @param[in] count number of splits
-    /// @param[in] spacing with of spaces between rectangle's splits (optional = 0)
+    /// @param[in] spacing width of spaces between rectangle's splits (optional = 0)
     /// @param[in] ratio[] ratio of wanted splits (optional = nullptr)
     void HorizontalSplit(Rect16 splits[], Rect16 spaces[], const size_t count, const uint16_t spacing = 0, uint8_t ratio[] = nullptr) const;
```
michalrudolf, line 591: Rename `ratio[]` to `text_width[]` like in the func definition

michalrudolf (approved):

michalrudolf (approved):

PR #1790, opened 2021-12-29
New http server

vorner:
This is a first (unfinished) shot at replacing the HTTP server. I didn't find anything that would provide what I had in mind, so it's in-house implementation. Still, I hope it's worth it. I'm putting it here to share it, as request for comments/means for further discussion. But it probably doesn't need a thorough review-in-depth yet, more like the higher level.

This solves several architectural problems we had:

* Adhoc „parsing“ of the old implementation by strstr and similar, that just Doesn't Work (tm).
* No sane way to return errors, headers, etc ‒ everything was solved by rewriting the URL and hoping some heuristic in future stage does something about it.
* No way to generate content in chunks.
* Shared global output buffer (that got rewritten from time to time by another request).
* No sane way to handle a post and then generate a response based on the data (only by abusing the URL rewriting and passing data in globals)
* No way to support multiple different post handlers.

The current code has all the functionality of the old one, but I don't consider it „finished“ or „ready for merge“. In particular:

- [x] The CMake generation of the automata is broken. For some reason, the test ones do work, but the same/similar fragment for the real one doesn't. Once I generate it once by hand, it then gets updated next time... I have no idea why and would be happy for help.
- [ ] While the implementation _allows_ for solving a lot of problems in the code (some global variables, …), it yet _doesn't do so_. I just plugged the old code in in any way possible; refactoring and cleaning up is necessary.
- [ ] Some parts probably need better names, moved somewhere, etc.
- [x] Documentation of classes/code/etc. In the mean time, I'd point you to the readme's, they should give the overview and the rest can probably be guessed by names.
- [x] Tests. I have tests for the automata, but I've killed the old http tests (they don't compile now) and haven't updated/rewritten them for the new implementation yet.
- [ ] Some work on sizes ‒ both code and memory. I have some optimizations in mind.


dragomirecky:
Hey @vorner, nice work! (once again ☺️)

I looked into the CMake issue you showed me today at the office and found the problem. Short answer: the `http_req_automaton.cpp` file does not have the `GENERATED` property in the required scope.

_Long answer:_ 
At configuration time, CMake does a lot of sanity checks. One of them is that it checks that source files of a target exist. So for example `add_library(awesome sourcefile.cpp)` fails at configuration time in case the `sourcefile.cpp` does not exist.

When a source file is expected to be created at build time, it has to have its GENERATED property set to 1. This way, CMake knows that it should not check its presence at configure time. This GENERATED property is set automatically by commands like `add_custom_target`.

In case it's set automatically, why is there an issue? Well, the GENERATED property is set within the scope of the current directory (`lib/WUI`). However, the WUI library is an INTERFACE library – it does not really exist (it's not compiled as a standalone thing, but its files get merged to the target, which is dependent on it). So, this library is really built as part of the `firmware` target, which is defined in the root `CMakeLists.txt`. See the problem? In that context (the root directory), the GENERATED property is not set for that source file. Remember how I told you the WUI library shouldn't be an INTERFACE library today? 🤓

_Solutions:_
CMake version 3.20 addresses this issue and changes the behavior so that the GENERATED property is always global. However, our project is configured [for compatibility with CMake 3.15](https://github.com/prusa3d/Prusa-Firmware-Buddy/blob/9bbe5cb1502b7d39dfaf6a07c41224956b8ed362/CMakeLists.txt#L1), so even if you have a CMake >= 3.20, you will have the issue. I don't really want to raise the required CMake version just because of this, so let's look for some other solution.

CMake 3.18 adds a way to set a source property with a custom directory scope manually. That would be a solution, but again, we are using CMake 3.15.

With CMake 3.15, I believe our best shot is to set the property from the root CMakeList. Not ideal, not nice, but it does not require us to raise the required version of CMake. The line should look something like this
```
set_property(SOURCE ${CMAKE_BINARY_DIR}/lib/WUI/http_req_automaton.cpp PROPERTY GENERATED 1)
```
Please add a comment, that it can be removed after we raise the required CMake version (and ideally permalink to this comment).

-----
One more thing. I noticed paths like `${CMAKE_BINARY_DIR}/http_req_automaton.cpp` in `lib/WUI/CMakeLists.txt`. This evaluates to something like `<build_dir>/http_req_automaton.cpp` polluting the root of the build folder. Please consider using the `CMAKE_CURRENT_BINARY_DIR` (counterpart to the `CMAKE_CURRENT_SOURCE_DIR`).

vorner:
Thank you, this has fixed it. I would never have figured this out by myself…

lib/WUI/http_lifetime.cpp
```diff
@@ -0,0 +1,196 @@
...
+public:
+    virtual optional<ConnectionState> accept(const RequestParser &parser) const override {
+        char filename[MAX_URL_LEN + 1];
+
+        if (!parser.uri_filename(filename, sizeof(filename))) {
+            return nullopt;
+        }
+
+        if (strcmp(filename, "/") == 0) {
+            strncpy(filename, "/index.html", MAX_URL_LEN);
```
DRracer, line 41: please prefer the non-standard `strlcpy` over `strncpy`

lib/WUI/automata/core.cpp
```diff
@@ -0,0 +1,143 @@
...
+        if (transitions[i].matches(byte)) {
+            const auto new_state = transitions[i].target_state;
+            return TransitionResult { new_state, gen_event(old.state, new_state, byte), transitions[i].fallthrough };
+        }
+    }
+
+    return nullopt;
+}
+
+Execution::~Execution() {}
```
DRracer, line 103: use `= default` to define a trivial constructor/destructor if possible

lib/WUI/nhttp/server.cpp
```diff
@@ -0,0 +1,555 @@
...
+             * available).
+             *
+             * Yes, we do _not_ free the pbuf.
+             */
+            return ERR_MEM;
+        }
+    }
+
+    assert(is_active_slot(base_slot));
+    Slot *slot = static_cast<Slot *>(base_slot);
```
DRracer, line 405: this looks suspicious ... a `dynamic_cast` should be a more appropriate construct at this spot (downcasting `Server::BaseSlot*` to `Server::Slot*`)

lib/WUI/nhttp/req_parser.cpp
```diff
@@ -0,0 +1,177 @@
...
+        // No API key -> can't authenticate.
+        if (!api_key) {
+            auth_status = false;
+        }
+        // The first character, just starting.
+        if (holds_alternative<monostate>(auth_status)) {
+            auth_status = static_cast<uint8_t>(0);
+        }
+        if (uint8_t *idx = get_if<uint8_t>(&auth_status); idx != nullptr) {
+            const size_t key_len = strlen(api_key);
```
DRracer, line 78: Is it possible for the `api_key` to be `nullptr` in some cases? You have a check for it above, but that doesn't protect this line from being executed.

lib/WUI/nhttp/headers.cpp
```diff
@@ -0,0 +1,140 @@
...
+        pos += snprintf(buf + pos, buffer_len - pos, "Transfer-Encoding: chunked\r\n");
+        pos = std::min(buffer_len, pos);
+    }
+    for (; extra_hdrs && *extra_hdrs; extra_hdrs++) {
+        size_t copy = std::min(buffer_len - pos, strlen(*extra_hdrs));
+        memcpy(buf + pos, *extra_hdrs, copy);
+        pos += copy;
+    }
+
+    buffer_len += 2;
```
DRracer, line 109: `buffer_len` is not read after this line. Am I missing something?

lib/WUI/nhttp/headers.cpp
```diff
@@ -0,0 +1,140 @@
...
+    memcpy(buf + pos, "\r\n", 2);
```
DRracer, line 110: the result of memcpy may not be null-terminated, is that ok?

lib/WUI/nhttp/gen_once.cpp
```diff
@@ -0,0 +1,83 @@
...
+            progress = Progress::Done;
+            return Step { 0, written, Terminating::for_handling(connection_handling) };
+        }
+        if (buffer_size - written < last_chunk_len) {
+            return Step { 0, written, Continue() };
+        }
+        // Fall through: the last chunk does fit.
+    case Progress::EndChunk:
+        assert(buffer_size - written >= last_chunk_len);
+        memcpy(buffer + written, LAST_CHUNK, last_chunk_len);
```
DRracer, line 73: the result from calling `memcpy` may not be null-terminated, is it ok here?

lib/WUI/nhttp/handler.h
```diff
@@ -0,0 +1,177 @@
...
+
+/**
+ * \brief All the different handlers currently supported by the server.
+ *
+ * For now, we list all the possible handlers the server may contain here. This
+ * allows the server to know the sizes at compile time. But it also introduces
+ * dependencies between types and headers, which is slightly uncomfortable and
+ * annoying. Better solution (one where we still reserve the space, with
+ * compiler-computed size, but don't list it upfront) would be nice. However,
+ * we don't want to template the whole server eather (because we would have to
```
DRracer, line 108: eather -> either

lib/WUI/http_lifetime.cpp
```diff
@@ -0,0 +1,196 @@
...
+             * The data of non-custom fs files are embedded inside the program
+             * as C arrays. Therefore, we can get the data and close the file
+             * right away.
+             */
+            /*
+             * We guess the content type by a file extension. Looking inside
+             * and doing "file magic" would be better/more reliable, but don't
+             * want the code complexity of that.
+             */
+            static const char *extra_hdrs[] = { "Content-Encoding: gzip\r\n", nullptr };
```
DRracer, line 63: For now we only have gzipped static data, so the extra header seems to be fine. What happens once someone tries to return a file which is not gzipped?

lib/WUI/http_lifetime.cpp
```diff
@@ -0,0 +1,196 @@
...
+            SendStaticMemory send(string_view(file.data, file.len), guess_content_by_ext(filename), parser.can_keep_alive(), extra_hdrs);
```
DRracer, line 64: ... also in case of sending something else than a static string from CPUFLASH, the `string_view(file.data, file_length)` would have to be changed.

lib/WUI/nhttp/req_parser.cpp
dragomirecky: - When `!api_key`, it sets auth_status to some `bool`. 
- The line 75 won't get executed, as the condition on line 74 isn't true
- `get_if` returns `nullptr`, as `auth_status` still holds `bool` from the first step. Therefore, this line is protected from such condition.

It is getting late and I am starting to be desperate. What am I missing? 😭

dragomirecky (commented):
Great work Michal, enjoyed reading it

Unfortunately, I did not manage to go through everything (mostly missing some server.cpp and some handlers). Maybe @DRracer  reviewed that part already?

lib/WUI/nhttp/req_parser.cpp
DRracer: TBH, clang analyzer pointed this out, not sure if such a scenario can be real:
```
req_parser.cpp:30:5: 'Default' branch taken. Execution continues on line 51
req_parser.cpp:51:5: Control jumps to 'case XApiKey:'  at line 68
req_parser.cpp:70:13: Assuming field 'api_key' is null
req_parser.cpp:70:9: Taking true branch
req_parser.cpp:74:13: Assuming the condition is false
req_parser.cpp:74:9: Taking false branch
req_parser.cpp:77:9: Taking true branch
req_parser.cpp:78:36: Null pointer passed as 1st argument to string length function
```

lib/WUI/nhttp/req_parser.cpp
```diff
@@ -0,0 +1,177 @@
...
+        }
+        *content_length = 10 * *content_length + (event.payload - '0');
+        return ExecutionControl::Continue;
+    case Names::Version:
+        switch (event.payload) {
+        case '.':
+            version_major = version_minor;
+            version_minor = 0;
+            return ExecutionControl::Continue;
+        case '0' ... '9':
```
DRracer, line 100: I assume you know the case's range syntax (0 ... 9) is a GCC extension.

lib/WUI/nhttp/req_parser.cpp
```diff
@@ -0,0 +1,177 @@
...
+        error_code = Status::BadRequest;
+        done = true;
+    }
+
+    if (done) {
+        /*
+         * Note: we assume the array of selectors is terminated by one that
+         * handles unhandled/unknown requests ‒ one that returns 404 or so.
+         */
+        for (const Selector *const *selector = server->selectors();; selector++) {
```
DRracer, line 145: I'd recommend adding some comment between the two `;;` to avoid confusion (that it is not a typo but you really meant it that way)

lib/WUI/nhttp/req_parser.cpp
```diff
@@ -0,0 +1,177 @@
...
+    return (connection == Connection::KeepAlive) || (version_major == 1 && version_minor >= 1 && connection != Connection::Close);
+}
+
+bool RequestParser::uri_filename(char *buffer, size_t buffer_size) const {
+    // FIXME: Parsing and _validation_ of the URL:
+    // * Stop at any `&`
+    // * Make sure there are no `..` or such in there.
+    // * Demangling?
+    if (buffer_size > url_size) {
+        memcpy(buffer, url.begin(), url_size);
```
DRracer, line 165: consider using `strlcpy` instead (makes sure the last character is `\0`)

DRracer (commented):
Generally, I like the proposed design :+1: . Having a chain of "selectors" and trying to match the incoming data resembles much of an input scanner/parser. In this case the proposed implementation seems to be easily understandable and extendible. Also, I like the fact, that requests are being parsed using finite automata which is the correct way of doing so. 

However, for extending these automata a programmer must look into the python generating scripts and tweak their implementation, which may not look straight forward (not a complaint but probably some note in C++ sources would help navigate someone else).

Std::variants and infrastructure around them is a bit new to me (never used it before that much), but their implementation doesn't seem to be inefficient in some important way. Moreover, on the semantics level, using std::variant over classical unions brings some code-safety benefits.

In terms of syntax/semantics, please prefer:
- `nullptr` over `NULL` if possible
- strlcpy,strlcat over strncpy/strncat (sometimes even over memcpy especially when followed by setting the last byte to \0)

Some of the data structures (like a RequestParser) are quite large in terms of memory footprint, but you already noted that as one of the TODO items (to shave off some bytes), so I'm looking forward to it.

Also, I've noticed some potential issues in the code (marked with my comments), please address them.

lib/WUI/nhttp/server.cpp
vorner: That won't bring any advantage here. Dynamic cast uses vtables to check which type it is and this one doesn't have a vtable ‒ no virtual tables. Or, at least that's the theory I know.

lib/WUI/nhttp/req_parser.cpp
vorner: So, Alan is right that it can't (couldn't) happen. His analysis is, however, more complex than what clang analyzer can do ‒ he takes how it behaved in the previous loop into account, while the clang analyzer doesn't reason in such a complex way.

Nevertheless, the fact that one needs to do such complex analysis to prove it can't happen is not really reader-friendly, so the best way is to just throw another `return ...` statement in there just to make sure.

lib/WUI/nhttp/headers.cpp
vorner: You're right. This is leftover from moving the code around. Apparently, it was there to return it back to the previous value after line 81 temporarily made it smaller by the „reserved“ to bytes. Nevertheless, once this got into a function of its own, it is no longer necessary.

lib/WUI/nhttp/headers.cpp
vorner: Yes, that's OK. Nothing in around here expects null-terminated strings. The design always works with „byte buffers“, not strings (that's also why it works with `uint8_t *` and not `char *`) and always manipulate lengths instead. Because, well, one may want to send binary data over the wire.

lib/WUI/http_lifetime.cpp
vorner: These specific „static files“ are for serving PrusaLink pages and they are all gzipped. We'll be getting rid of this soon (hopefully) in favor XFlash.

Anyway, if we ever want to also return some other (non-gzipped) „static files“, then we would create a different selector, or have a whitelist/blacklist for the urls that doesn't pass these extra headers :shrug:

lib/WUI/http_lifetime.cpp
vorner: In that case we will have a different „Handler“ than `SendStaticMemory`. I have a branch that adds `SendFile`, that gets a `File *` and sends that.

lib/WUI/CMakeLists.txt
```diff
@@ -9,16 +9,23 @@ target_sources(
...
+            automata/core.cpp
+            nhttp/common_selectors.cpp
+            nhttp/gcode_upload.cpp
+            nhttp/gen_once.cpp
+            nhttp/handler.cpp
+            nhttp/headers.cpp
+            nhttp/req_parser.cpp
+            nhttp/server.cpp
+            nhttp/static_mem.cpp
+            nhttp/status_page.cpp
```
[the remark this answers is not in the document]
vorner: In general, I'm in favor of that. Only, I probably don't want to do it right now, because:

* I tend to work in a way that chunk of related functionality develops itself somewhere and then is torn off once it stabilizes a bit. I don't think this is stabilized enough for that.
* This PR is already too big and I'd prefer to get it in sooner than later.
* There currently are some reverse dependencies. Part of them will disappear when refactoring the link between the current http code and the gcode upload code tailored for the previous implementation, part will need a bit of design to do right.

So it kind of is my longer-term plan from the beginning, that this should eventually be separated. But right now I'd prefer to make it work first.

lib/WUI/automata/core.h
```diff
@@ -0,0 +1,356 @@
...
+#pragma once
+
+#include <cstdbool>
+#include <cstdint>
+#include <optional>
+#include <string_view>
+#include <tuple>
+#include <variant>
+
+namespace automata {
```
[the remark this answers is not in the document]
vorner: I kind of expect I'd use the same thing for http client (connect is going with something existing for now, but it seems to be only slightly better than what shipped with LwIP as their server, uses quite bug buffers, etc… so long-term, I'd prefer to replace it with something), possibly proper json parsing (we have a json parser library around, but that one doesn't support "incremental" parsing, so the whole input needs to be buffered ☹).

So, right now, `nhttp` is the only user, but I don't think that's the final state.

On the other hand, I'd probably prefer to long-term extract the automata as a completely separate library, similar as nhttp. Again, right now is probably not the right time 😇

lib/WUI/automata/core.cpp
```diff
@@ -0,0 +1,143 @@
...
+ExecutionControl Execution::feed(uint8_t byte) {
+    const auto new_state = automaton->transition(current_state, byte);
+
+    if (new_state.has_value()) {
+        current_state = new_state->new_state;
+        if (new_state->emit_event.has_value()) {
+            const auto result = event(*new_state->emit_event);
+            if (result != ExecutionControl::Continue) {
+                return result;
+            }
```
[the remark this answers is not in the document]
vorner: I think it might be used in tests (or maybe not).

Anyway, that's kind-of future-proof design. I was thinking about other use cases than just http and if we wanted to eg. parse the multipart forms (which is not that terrible right now, so it definitely isn't on the table right now), there are cases where the upper level may want to just abort the parsing on the semantic level instead of the syntactic one. Like, if the user wants to upload a file but didn't provide the file name, or the file name contains forbidden characters, we want to give up at that point without having to worry about file not being ready when dealing with the payload.

lib/WUI/automata/core.cpp
```diff
@@ -0,0 +1,143 @@
...
+        return ExecutionControl::NoTransition;
+    }
+}
+
+std::tuple<ExecutionControl, size_t> Execution::consume(std::string_view data) {
+    size_t consumed = 0;
+    for (const uint8_t b : data) {
+        const ExecutionControl control = feed(b);
+        if (control != ExecutionControl::Continue) {
+            return std::make_tuple(control, consumed);
```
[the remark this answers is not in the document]
vorner: Well, `Stop` is kind of meant as „I give up, throw the whole thing out of the window“, but sure, it makes sense to be consistent.

lib/WUI/CMakeLists.txt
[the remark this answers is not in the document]
dragomirecky: sure, no problem. Can be done later :)

lib/WUI/nhttp/server.cpp
DRracer: ok

lib/WUI/nhttp/headers.cpp
DRracer: ok

lib/WUI/http_lifetime.cpp
DRracer: ok

lib/WUI/http_lifetime.cpp
DRracer: ok

lib/WUI/automata/core.cpp
```diff
@@ -0,0 +1,143 @@
...
+        }
+        consumed++;
+    }
+
+    return std::make_tuple(ExecutionControl::Continue, consumed);
+}
+
+ExecutionControl Execution::feed(std::string_view data) {
+    return std::get<0>(consume(data));
```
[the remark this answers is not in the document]
vorner: Won't try convincing. That's a leftover from how the code evolved, I've noticed I need to know how much is consumed much later on and it got preserved. Removed now.

utils/gen-automata/parts.py
```diff
@@ -0,0 +1,145 @@
...
+def constant(s, nocase=False):
+    """
+    Automaton accepting/consuming a string constant.
+    """
+    assert len(s) >= 2  # Not implemented yet
+    auto = Automaton()
+    start = auto.start()
+    # We go through a single "real" transition, otherwise the
+    # automaton acts a bit unpredictable when joining together with
+    # others.
```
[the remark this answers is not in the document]
vorner: Not only that, it was acting a bit unpredictable even without paths.

I have to admit that the automata are a bit quirky to work with and I would like to improve their ergonomics… but so far the balance (of how many automata I've needed vs. the effort to improve the tooling) didn't tip it over.

lib/WUI/automata/core.cpp
```diff
@@ -0,0 +1,143 @@
...
+                    gen_event(old.state, new_state.state, byte),
+                    false,
+                };
+            } else {
+                return TransitionResult { old.path_step(), nullopt, false };
+            }
+        }
+    }
+
+    const State &sentinel = states[old.state + 1];
```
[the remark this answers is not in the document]
vorner: I know it looks a bit weird and it sounds strange, but I actually need this behavior. Let me explain why.

The reason why paths even exist is to save on number of states when there's a constant string (I'd like to use them for tries eventually too, and save even more). Otherwise, the number of states would be bigger and the flash consumption higher. This is to recognize things like known headers or known methods.

Now, when there's an known header, it can _start_ as something we know, but get out of that path in the middle. In that case, I can't just give up the whole parsing. I have to go to some kind of „Unknown header“ state. So I have to check for that state.

The way these things are used is that the path state has some wildcard/fallback transitions.

Maybe we should have some kind of states from the state and states from the path. I don't know if I want to introduce them right now, though.

lib/WUI/nhttp/req_parser.cpp
vorner: Not really (now I do). Should I replace it? I guess we don't really compile under anything else.

vorner:
I've tried to address most of the things. They are sent there as fixup commits, so I don't rewrite the history in the middle of review. I'll squash them into the relevant commits before merge.

I got stuck with the one with python f-strings, I've send a request for help to @dragomirecky to see how that one goes.

> However, for extending these automata a programmer must look into the python generating scripts and tweak their implementation, which may not look straight forward

It is not straightforward and it is a weak point of this implementation :-(. I would like to improve that one eventually, to make them easier to use and debug.

> nullptr over NULL if possible

Generally, I try to (unless it's inside a C file). Did I miss some?

> strlcpy,strlcat

Tests make that hard, they are not available on my system :-(.

> Some of the data structures (like a RequestParser) are quite large in terms of memory footprint,

I already have few commits dealing with that elsewhere. I can share them if you want, but I didn't want to inflate this pull request more than necessary.

lib/WUI/http_lifetime.cpp
```diff
@@ -124,9 +124,14 @@ class PrusaLinkApi final : public Selector {
...
-                 * type separately. Seems like extending variants is beoynd
-                 * even C++ template madness to do automagically.
+                 * So, we have a "smaller" variant (eg. variant<A, B, C>) and
+                 * want a "bigger" variant<A, B, C, D, E>. C++ templates can't
+                 * do the "upgrade" automatically. But it can upgrade A into
+                 * the bigger one, it can upgrade B into the bigger one and can
+                 * upgrade C...
+                 *
+                 * Therefore we use the visit to take the first one apart and
+                 * then convert each type separately into the bigger one.
```
DRracer, line 134: :+1: wow

lib/WUI/nhttp/req_parser.cpp
DRracer: yes, there is an additional implementation elsewhere (copied from BSD) just for the unit tests.

lib/WUI/nhttp/req_parser.cpp
DRracer: Generally we prefer writing standard-compliant code. Probably clang understands this construct as well as QtCreator's code model was not complaining when I was reviewing the code.
In this case, I'd keep your syntax. I just wanted to make sure we agree that this is a potentially non-portable extension.

utils/gen-automata/common.py
```diff
@@ -0,0 +1,317 @@
...
+        output += "namespace %s {\n" % self.__namespace
+        output += "enum Names {\n"
+        for name in self.__names:
+            output += "    %s = %d,\n" % (name, self.__names[name])
+        output += "};\n"
+        output += "extern const automata::StrPath *paths;\n"
+        output += "extern const automata::Transition *transitions;\n"
+        output += "extern const automata::State *states;\n"
+        output += "};\n"
+        return output
```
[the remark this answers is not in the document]
dragomirecky: Sorry, I missed the for cycle. The f expression might look something like

```
...
{ '\n'.join(' = '.join([name, str(value)]) for name, value in self.__names.items()) }
```

lib/WUI/automata/core.cpp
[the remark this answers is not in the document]
dragomirecky: Ok, get it. Those are a bit rough edges rn, but something we can focus on later :)

And when the time comes, I would also be down to use another name for it (not "path"). Paths already exist in graph theory, and it took me a while to get my head around the fact that they mean here something slightly different :). But I didn't come up with a good name yet (it is something like "multi transition" here, not a great name either 😁)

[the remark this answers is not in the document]
dragomirecky: Btw, one great (and super simple to implement) feature would be outputting the graphs in DOT lang for easy visualisation with Graphviz 🤓

DRracer (approved):

PR #1794, opened 2022-01-03
WUI: Rename temporary upload file

vorner:
Don't give it a .gcode extension. This'll prevent the GUI from ever
showing it. This previously could happen in various corner cases like:

* When the printer got restarted during upload, the file got left in
  place (at least until overwritten by next upload) and the user selects
  it in the menu.
* When the user opens the print dialog during upload.
* When there are two uploads close after one another ‒ the first one
  signals the GUI to show the single-click print dialog, but if second
  upload starts before it is shown, the dialog would then open the
  temporary file of the second upload (and explode spectacularly, as the
  same file got opened for both reading and writing at the same time).

Naming it differently makes the GUI ignore such file.

### DIFF
--- a/tests/unit/lib/WUI/http/multipart_upload.cpp
+++ b/tests/unit/lib/WUI/http/multipart_upload.cpp
@@ -230,11 +230,11 @@ TEST_CASE("One Big Chunk") {
 
     const auto &log = test.log();
     REQUIRE(log.start.has_value());
-    REQUIRE(log.start->filename == "tmp.gcode");
+    REQUIRE(log.start->filename == "upload.tmp");
     REQUIRE(log.data.size() == 1);
     REQUIRE(log.data[0] == GCODE_LINE1);
     REQUIRE(log.finish.has_value());
-    REQUIRE(log.finish->tmp_filename == "tmp.gcode");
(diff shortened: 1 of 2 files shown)


tests/unit/lib/WUI/http/multipart_upload.cpp
```diff
@@ -230,11 +230,11 @@ TEST_CASE("One Big Chunk") {
 
     const auto &log = test.log();
     REQUIRE(log.start.has_value());
-    REQUIRE(log.start->filename == "tmp.gcode");
+    REQUIRE(log.start->filename == "upload.tmp");
```
DRracer, line 233: may be declare a constant (other than DEFAULT_UPLOAD_FILENAME) for the unit tests as well

DRracer (approved):

tests/unit/lib/WUI/http/multipart_upload.cpp
vorner: I'll merge it this way. I know this is a bit of slacky, but that's a bit of a reminder about another TODO ‒ these names need to be generated dynamically in the future to allow parallel uploads.

PR #1797, opened 2022-01-04
Use std::optional

espr14:


src/common/marlin_server.cpp
```diff
@@ -1557,9 +1555,9 @@ void fsm_destroy(ClientFSM type) {
 
 void _fsm_change(ClientFSM type, fsm::BaseData data) {
     {
-        static int previous_type = -1;
-        static int previous_phase = -1;
-        if (previous_type != static_cast<int>(type) || previous_phase != static_cast<int>(data.GetPhase())) {
+        static std::optional<int> previous_type;
+        static std::optional<int> previous_phase;
+        if (previous_type.value_or(-1) != static_cast<int>(type) || previous_phase.value_or(-1) != static_cast<int>(data.GetPhase())) {
```
vorner, line 1560: The value_or is useful if you actually have a fallback value. That -1 doesn't really make sense here.

Nevertheless, if your only intention is to make sure `nullopt` is different than the value, then I believe this Just Works ‒ `previous_type != static_cast(...)` will make it different if it's not there.

src/common/marlin_server.cpp
```diff
@@ -1594,7 +1592,8 @@ FSM_notifier::FSM_notifier(ClientFSM type, uint8_t phase, variant8_t min, varian
     s_data.progress_min = progress_min;
     s_data.progress_max = progress_max;
     s_data.var_id = var_id;
-    s_data.last_progress_sent = -1;
+    // TODO: use std::optional
```
vorner, line 1595: Use optional where? Leftover todo?

src/common/marlin_server.cpp
```diff
@@ -1624,7 +1623,7 @@ void FSM_notifier::SendNotification() {
         progress = s_data.progress_max;
 
     // after first sent, progress can only rise
-    if ((s_data.last_progress_sent == uint8_t(-1)) || (progress > s_data.last_progress_sent)) {
+    if (!s_data.last_progress_sent.has_value() || (progress > s_data.last_progress_sent.value())) {
```
vorner, line 1626: According to the docs, this might also be covered by just the comparison. But here I don't mind the explicitness of this current solution.

src/common/marlin_server.hpp
```diff
@@ -27,22 +29,21 @@ void fsm_change(ClientFSM type, T phase, fsm::PhaseData data) {
...
     //call inside marlin server on received response from client
     static void SetResponse(uint32_t encoded_bt) {
         server_side_encoded_response = encoded_bt;
     }
     //return response and erase it
-    //return -1 if button does not match
     template <class T>
     static Response GetResponseFromPhase(T phase) {
-        uint32_t _phase = server_side_encoded_response >> RESPONSE_BITS;
+        uint32_t _phase = server_side_encoded_response.value_or(-1) >> RESPONSE_BITS;
```
vorner, line 42: Again, the point of optional is not to have these special values and going back to them is kind of against that. Is it possible to restructure this function? Maybe first check there is something and if not, return Response::_none right away?

vorner (changes requested):
This improves on things, but I'd suggest we go the whole way and not have these `-1` in handling anywhere 😇

src/common/marlin_server.cpp
```diff
@@ -1557,9 +1555,10 @@ void fsm_destroy(ClientFSM type) {
 
 void _fsm_change(ClientFSM type, fsm::BaseData data) {
     {
-        static int previous_type = -1;
-        static int previous_phase = -1;
-        if (previous_type != static_cast<int>(type) || previous_phase != static_cast<int>(data.GetPhase())) {
+        static std::optional<int> previous_type;
+        static std::optional<int> previous_phase;
+        // no value: comparison returns true
+        if (previous_type.value() != static_cast<int>(type) || previous_phase.value() != static_cast<int>(data.GetPhase())) {
```
vorner, line 1561: Oops, I probably wasn't clear previously :-|.

The `.value()` is extracting the held value of type `T` and if it's not there it throws an exception (which will likely result in termination, since we have exceptions disabled). That's not what we want.

But there's an overloaded `optional<T> != T` operator out there (eg. if you _don't_ unpack it, just `previous_type != static_cast<int>(type)`, without unpacking.

src/common/marlin_server.cpp
```diff
@@ -1624,7 +1623,8 @@ void FSM_notifier::SendNotification() {
         progress = s_data.progress_max;
 
     // after first sent, progress can only rise
-    if ((s_data.last_progress_sent == uint8_t(-1)) || (progress > s_data.last_progress_sent)) {
+    // no value: comparison returns true
+    if (progress > s_data.last_progress_sent.value()) {
```
vorner, line 1627: Similar…

vorner (changes requested):

vorner (approved):

PR #1806, opened 2022-01-07
Check bed temp at resume

espr14:
If you pause print at the beginning to clean the sheet the temperature could drop. FW not checking the bed so you could continue with a too cold bed.
BFW-2382

### DIFF
--- a/src/common/marlin_server.cpp
+++ b/src/common/marlin_server.cpp
@@ -480,16 +480,25 @@ void marlin_server_print_resume(void) {
 }
 
 void marlin_server_print_reheat_start(void) {
-    if ((marlin_server.print_state == mpsPaused) && (marlin_server_print_reheat_ready() == 0)) {
+    if ((marlin_server.print_state == mpsPaused) && marlin_server_print_reheat_ready()) {
         thermalManager.setTargetHotend(marlin_server.resume_nozzle_temp, 0);
+        // No need to set bed temperature because we keep it on all the time.
     }
 }
 
-int marlin_server_print_reheat_ready(void) {
-    if (marlin_server.vars.target_nozzle == marlin_server.resume_nozzle_temp)
-        if (marlin_server.vars.temp_nozzle >= (marlin_server.vars.target_nozzle - 5))
-            return 1;
-    return 0;
+// Fast temperature recheck.
+// Does not check stability of the temperature.
+bool marlin_server_print_reheat_ready() {
+    // check nozzle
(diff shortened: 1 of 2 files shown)


src/common/marlin_server.cpp
```diff
@@ -480,16 +480,24 @@ void marlin_server_print_resume(void) {
 }
 
 void marlin_server_print_reheat_start(void) {
-    if ((marlin_server.print_state == mpsPaused) && (marlin_server_print_reheat_ready() == 0)) {
+    if ((marlin_server.print_state == mpsPaused) && marlin_server_print_reheat_ready()) {
         thermalManager.setTargetHotend(marlin_server.resume_nozzle_temp, 0);
     }
```
vorner, line 485: I think I understand why it _doesn't_ set the bed target temperature. But a comment explaining that would still be useful.

src/common/marlin_server.cpp
```diff
@@ -480,16 +480,24 @@ void marlin_server_print_resume(void) {
...
+// Does not check stability of the temperature.
+bool marlin_server_print_reheat_ready() {
+    // check nozzle
+    if (marlin_server.vars.target_nozzle != marlin_server.resume_nozzle_temp
+        || marlin_server.vars.temp_nozzle < (marlin_server.vars.target_nozzle - TEMP_HYSTERESIS)) {
+        return false;
+    }
+    // check bed
+    if (marlin_server.vars.temp_bed < (marlin_server.vars.target_bed - TEMP_BED_HYSTERESIS))
+        return false;
```
vorner, line 498: • I know it was here to begin with, but that `!=` seems extra (it could be incorporated as `<=` in the main condition).
• It's a bit odd that these checks are slightly different from each other.
• Does it have to go over the temperature, shouldn't the hysteresis work symmetrically to both sides?

vorner (changes requested):
It mostly minor nitpicking, but still :shrug:

src/common/marlin_server.cpp
espr14: 1. There are 3 variables:
1.1. target_nozzle
1.1. temp_nozzle
1.1. resume_nozzle_temp
They can't be mixed.
1. Looks odd but nozzle can be turned off so we need to set the printing temperature back. No need for the bed.
2. Mathematically should. But there's no issue if bed has higher temperature. There are drawbacks if nozzle has higher temp. but it cools down very quickly once the nozzle is moving and extruding and/or with fan on. Heating is slow so we need to wait.

vorner (approved):

PR #1811, opened 2022-01-10
BFW-2387: ESP flashing UX improvements

vorner:
• Don't time out the menu in the middle.
• Show some progress to look alive.

This probably isn't the final version, we want nice colorful progress
bar or something. But it's good enough for further testing.

### DIFF
--- a/src/gui/screen_menu_esp_update.cpp
+++ b/src/gui/screen_menu_esp_update.cpp
@@ -5,6 +5,8 @@
 #include "ScreenHandler.hpp"
 #include "RAII.hpp"
 
+#include <basename.h>
+
 #ifdef __cplusplus
 extern "C" {
 #endif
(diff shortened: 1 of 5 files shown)


src/gui/screen_menu_esp_update.cpp
```diff
@@ -84,6 +85,9 @@ class ScreenMenuESPUpdate : public AddSuperWindow<screen_t> {
     esp_entry *current_file;
     uint32_t readCount;
     loader_stm32_config_t loader_config;
+    std::unique_ptr<char[]> progress_buffer;
```
RadekVana, line 88: use `std::array` instead , there is filename length limit somewhere in fatfs
there is no dynamic allocation in GUI thread and I would like to stay that way
dynamic allocation is causing RAM fragmentation and you must be able to handle worse case scenario anyway

src/gui/screen_menu_esp_update.cpp
```diff
@@ -128,6 +132,28 @@ ScreenFactory::UniquePtr GetScreenMenuESPUpdate() {
...
+     * FIXME: Is there a better way?
+     *
+     * We apparently need to keep the source of the text around, not have it on stack.
+     * We also can't reuse the buffer between calls because it
+     * doesn't redraw (the SetText checks for _pointer_
+     * equality).
+     *
+     * Is there a way to get rid of that "translation" here?
+     */
+    help.SetText(_(buffer));
```
RadekVana, line 152: use `stringview::MAKERAM` instead `_()`
`_()` would try to translate text, which cannot be done with dynamical texts

Also if you use `std::array` than you must manually call `help.Invalidate()`, because text will have same address, in this case `help.SetText` call should be only in ctor

src/gui/screen_menu_esp_update.cpp
```diff
@@ -153,6 +179,15 @@ void ScreenMenuESPUpdate::windowEvent(EventLock /*has private ctor*/, window_t *
...
             progress_state = esp_upload_action::Connect;
+            /*
+             * Don't leave on timeout when flashing, which would cancel the
+             * flashing (but be willing to time out before that if the user
+             * just wonders into the menu.
+             *
+             * This flag is not reset. We want the user to be able to see the
+             * result.
+             */
+            flags.timeout_close = is_closed_on_timeout_t::no;
```
RadekVana, line 190: move this to ctor

RadekVana (changes requested):

src/gui/screen_menu_esp_update.cpp
vorner: This was here kind of on purpose ‒ the intention was that if the user doesn't start flashing, the menu times out the usual way.

Should I still move it?

RadekVana (approved):

PR #1813, opened 2022-01-11
eeprom: Fix initialization

RadekVana:
eeprom: Fix initialization, read eeprom only once to avoid need of multiple crc checks
disable unused test, because they currently does not fit on screen
BFW-2380

### DIFF
--- a/src/common/eeprom.cpp
+++ b/src/common/eeprom.cpp
@@ -716,11 +716,8 @@ static bool eeprom_convert_from(eeprom_vars_t &eevars) {
 static bool eeprom_check_crc32(const eeprom_vars_t &eevars) {
     if (eevars.DATASIZE > EEPROM_MAX_DATASIZE)
         return false;
-#if 1 //simple method
     uint32_t crc = crc32_eeprom((uint32_t *)(&eevars), (eevars.DATASIZE - 4) / 4);
     return eevars.CRC32 == crc;
-#else //
-#endif
 }
 
 static void eeprom_update_crc32() {
--- a/src/common/eeprom.cpp
+++ b/src/common/eeprom.cpp
@@ -331,14 +331,12 @@ static inline void eeprom_unlock(void) {
 }
 
 // forward declarations of private functions
-
+static void eeprom_write_vars(eeprom_vars_t &vars);
 static uint16_t eeprom_var_size(enum eevar_id id);
-static uint16_t eeprom_var_addr(enum eevar_id id);
-//static void eeprom_print_vars(void);
-static int eeprom_convert_from_v2(void);
-static int eeprom_convert_from(uint16_t version, uint16_t features);
+static uint16_t eeprom_var_addr(enum eevar_id id, uint16_t addr = EEPROM_ADDRESS);
+static bool eeprom_convert_from(eeprom_vars_t &eevars);
 
-static int eeprom_check_crc32(void);
+static bool eeprom_check_crc32(const eeprom_vars_t &eevars);
 static void eeprom_update_crc32();
 
 static uint16_t eeprom_fwversion_ui16(void);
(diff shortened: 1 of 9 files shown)


src/common/eeprom.cpp
```diff
@@ -623,120 +627,127 @@ static int eeprom_convert_from_v2(void) {
...
+    eevars = vars;
+}
+
+/**
+ * @brief conversion function for new version format (features, firmware version/build)
+ * does not change crc, it is changed automatically by write function
+ *
+ * @param eevars eeprom struct in RAM
+ * @return true updated (changed)
+ * @return false not changed
```
mkbel, line 677: suggest @retval instead of @return

src/common/eeprom.cpp
```diff
@@ -348,35 +345,38 @@ eeprom_init_status_t eeprom_init(void) {
...
     status = EEPROM_INIT_Normal;
     osSemaphoreDef(eepromSema);
     eeprom_sema = osSemaphoreCreate(osSemaphore(eepromSema), 1);
     st25dv64k_init();
 
-    version = variant8_get_ui16(eeprom_get_var(EEVAR_VERSION));
-    features = (version >= 4) ? variant8_get_ui16(eeprom_get_var(EEVAR_FEATURES)) : 0;
-    if ((version >= EEPROM_FIRST_VERSION_CRC) && !eeprom_check_crc32())
+    //read entire eeprom at once
+    eeprom_vars_t &eevars = eeprom_startup_vars(); //initialize internal static variable, first call must be done now - to ensure interrupt is enabled
```
mkbel, line 354: Are those comments here necessary? Are they understandable?

I can imagine:
```
initialize internal static variables
read entire eeprom at once
```
would go to `eeprom_startup_vars()` documentation.

And I completely don't understand:
`first call must be done now - to ensure interrupt is enabled`
Can you make it more clear? Or delete it?

src/common/eeprom.cpp
```diff
@@ -348,35 +346,40 @@ eeprom_init_status_t eeprom_init(void) {
...
+
+    const uint16_t version = eevars.VERSION;
+    const uint16_t features = (version >= 4) ? eevars.FEATURES : 0;
```
mkbel: Can you name that magic number?

src/common/eeprom.cpp
```diff
@@ -348,35 +346,40 @@ eeprom_init_status_t eeprom_init(void) {
...
 }
 
-void eeprom_defaults(void) {
-    eeprom_vars_t vars = eeprom_var_defaults;
-    vars.FWBUILD = project_build_number;
-    vars.FWVERSION = eeprom_fwversion_ui16();
+static void eeprom_write_vars(eeprom_vars_t &vars) {
     eeprom_lock();
     // calculate crc32
     vars.CRC32 = crc32_eeprom((uint32_t *)(&vars), (EEPROM_DATASIZE - 4) / 4);
```
mkbel, line 382: This probably requires EEPROM_DATASIZE % 4 == 0, but there is no static_assert anywhere.

src/common/eeprom.cpp
```diff
@@ -623,120 +627,127 @@ static int eeprom_convert_from_v2(void) {
...
 #if 1 //simple method
-    uint8_t data[EEPROM_MAX_DATASIZE];
-    uint32_t crc2;
-    st25dv64k_user_read_bytes(EEPROM_ADDRESS, data, datasize);
-    crc2 = crc32_eeprom((uint32_t *)data, (datasize - 4) / 4);
-    return (crc == crc2) ? 1 : 0;
+    uint32_t crc = crc32_eeprom((uint32_t *)(&eevars), (eevars.DATASIZE - 4) / 4);
+    return eevars.CRC32 == crc;
 #else //
 #endif
```
mkbel, line 712: Maybe you can remove
```
#if 1
#else //
#endif
```

src/gui/test/screen_menu_eeprom_test.cpp
```diff
@@ -0,0 +1,216 @@
+/**
+ * @file screen_menu_eeprom_test.cpp
```
mkbel, line 2: It is redundant to put screen_menu_eeprom_test.cpp there, if the documentation is in the same file as file documented.

src/gui/test/screen_test.cpp
```diff
@@ -22,64 +23,38 @@ static volatile void recursive(uint64_t i) {
...
+    , tst_safety_dlg(
+          this, this->GenerateRect(ShiftDir_t::Bottom), []() { Screens::Access()->Open(ScreenFactory::Screen<screen_test_dlg_data_t>); }, string_view_utf8::MakeCPUFLASH((const uint8_t *)"test dialog"))
+#if 0
+    , tst_graph(this, this->GenerateRect(ShiftDir_t::Bottom), []() { /*screen_open(get_scr_test_graph()->id);*/ }, string_view_utf8::MakeCPUFLASH((const uint8_t *)"temp graph"))
+    , tst_temperature(this, this->GenerateRect(ShiftDir_t::Bottom), []() { /*screen_open(get_scr_test_temperature()->id);*/ }, string_view_utf8::MakeCPUFLASH((const uint8_t *)"temp - pwm"))
+    , tst_heat_err(this, this->GenerateRect(ShiftDir_t::Bottom), []() { /*("TEST BED ERROR", "Bed", 1.0, 2.0, 3.0, 4.0);*/ }, string_view_utf8::MakeCPUFLASH((const uint8_t *)"HEAT ERROR"))
+    , tst_disp_memory(this, this->GenerateRect(ShiftDir_t::Bottom), []() { /*screen_open(get_scr_test_disp_mem()->id);*/ }, string_view_utf8::MakeCPUFLASH((const uint8_t *)"Disp. R/W"))
+#endif // 0
```
mkbel, line 45: This change is not stated in commit message.

src/common/eeprom.cpp
RadekVana: it is generated automatically this way

RadekVana: via _Doxygen Documentation Generator_
vscode plugin

src/common/eeprom.cpp
RadekVana: it is not may code, I just simplified original expression
it is eeprom version 4 - (value stored in eeprom to know if eeprom should be updated) 
we had versions 2,4,6,7,8,9,10, I really don't know what happened with 1,3,5
I also do not understand why we need second variable features

RadekVana: this code had been erased with commit removing version 2 support

src/common/eeprom.cpp
RadekVana: there is padding variable making so .. not my code
sadly it is different than code in private

src/common/eeprom.cpp
RadekVana: it corresponds with some code in private and would create conflict

src/gui/screen_test.hpp
```diff
@@ -8,15 +8,18 @@
...
     window_text_button_t tst_term;
     window_text_button_t tst_msgbox;
     window_text_button_t tst_wizard_icons;
     window_text_button_t tst_safety_dlg;
+#if 0
     window_text_button_t tst_graph;
     window_text_button_t tst_temperature;
     window_text_button_t tst_heat_err;
     window_text_button_t tst_disp_memory;
+#endif // 0
```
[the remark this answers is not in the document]
RadekVana: added to PR message

src/gui/test/screen_menu_eeprom_test.cpp
RadekVana: I know, it is automatically generated and I use it for text searching because other methods like go to definition sometime does not work in vscode

src/gui/test/screen_test.cpp
RadekVana: it corresponds to same change in header

src/common/eeprom.cpp
```diff
@@ -551,48 +561,38 @@ static uint16_t eeprom_var_size(enum eevar_id id) {
     return 0;
 }
 
-static uint16_t eeprom_var_addr(enum eevar_id id) {
-    uint16_t addr = EEPROM_ADDRESS;
+static uint16_t eeprom_var_addr(enum eevar_id id, uint16_t addr) {
```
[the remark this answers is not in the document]
RadekVana: I agree that intptr_t would be better, but I wanted to make as few changes as possible, because if I started to change parts of eeprom that I do not like, I would rewrite it all

mkbel (approved):

PR #1817, opened 2022-01-12
prusa_link: Don't set busy when printing

vorner:
The busy flag reportedly means something like changing filament and
similar ‒ times when the printer doesn't want to talk to outside world.

The bigger problem is it disables the controls on the web, which we
don't want during the whole print.


DRracer (approved):
I agree - _busy_ is something blocking Marlin's loop

PR #1827, opened 2022-01-14
Add Support for USB MSC & Logging over USB CDC

dragomirecky:
This PR includes the following changes
- Replaces the STM's USB Device stack with TinyUSB stack
- It implements a single[1] USB Communications Data Class line working under two possible modes:
  1. The standard Marlin serial port (when configured with traditional settings, like 115200 baud rate),
  2. And logging output (when configured with baud rate = 57600).
- It implements USB Mass Storage Class
  - It is disabled by default. It can be enabled in settings (dev-only option).
  - This MSC provides raw access to the part of the external flash reserved for littlefs.
  - Essentially, it turns the printer into (a slow) USB flash drive.
  - We are using littlefs, so most operating systems won't know how to mount it. However, this can be easily fixed with a user-space implementation of littlefs: https://github.com/littlefs-project/littlefs-fuse.
- It modifies the driver for the external flash to support DMA and basic error reporting. Also, it makes its interface a wee bit safer. My goal was to make only necessary changes to the driver, but I think I can paraphrase the greatest Czech here: after a few minor modifications, it became an entirely different driver.

[1] Why a single CDC interface and switch based on the baud rate instead of two separate interfaces? The USB peripheral is limited by the number of endpoints it implements, and we don't have enough of them to add another separate CDC interface.


mkbel:
I would suggest to use some standart baud rate, as in lot off uart terminals won't let you set non-standart baud rate.
I would not select 115200 as this is most comon.
I would not select 57600 as this is used to reset arduino project cpu.
Probably would not selelect 9600 as this is commonly used by bootloaders/boot monitors.

src/main.cpp
```diff
@@ -278,30 +281,38 @@ int main(void) {
...
-                 * should apply to any network communication ‒ one probably shall
-                 * not eg. start a print from there.
-                 */
-                block_networking = true;
-            }
+    /*
+     * Checking this first, before starting the GUI thread. The GUI thread
+     * resets/consumes the dump as a side effect.
+     */
+    if (false && dump_in_xflash_is_valid() && !dump_in_xflash_is_displayed()) {
```
vorner: Is that false some kind of leftover?

src/usb_device_msc.c
```diff
@@ -0,0 +1,169 @@
+#include <inttypes.h>
+#include "usb_device.h"
+#include "usbh_msc.h"
+#include "log.h"
+#include "littlefs.h"
+#include "scratch_buffer.hpp"
```
vorner: :question: Do we use `h` for C++ headers, or `.hpp`? I've seen both 😈.

vorner (commented):
I've read through it and noticed few aesthetic/style things. But honestly, this is not area I'd have any experience and I'm not sure if my review brought anything particularly useful.

lib/Arduino_Core_Buddy/cores/arduino/USBSerial.h
```diff
@@ -1,27 +1,39 @@
-//USBSerial.h - Buddy/STM32
-#ifndef _USBSERIAL_H
-#define _USBSERIAL_H
+#pragma once
 
 #include "Arduino.h"
 #include "Stream.h"
+#include <array>
```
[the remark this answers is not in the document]
dragomirecky: Arduino_Core_Buddy is our code, not 3rd party

yep, at this time, I would prefer it to be located within the `src/` directory...

src/common/eeprom_current.hpp
```diff
@@ -31,11 +31,12 @@ struct vars_body_t : public eeprom::v10::vars_body_t {
     char WIFI_HOSTNAME[LAN_HOSTNAME_MAX_LEN + 1];
     char WIFI_AP_SSID[WIFI_MAX_SSID_LEN + 1];
     char WIFI_AP_PASSWD[WIFI_MAX_PASSWD_LEN + 1];
+    uint8_t USB_MSC_ENABLED;
 };
 
 #pragma pack(pop)
 
-static_assert(sizeof(vars_body_t) == sizeof(eeprom::v10::vars_body_t) + sizeof(uint8_t) * 3 + PL_API_KEY_SIZE + LAN_HOSTNAME_MAX_LEN + 1 + WIFI_MAX_SSID_LEN + 1 + WIFI_MAX_PASSWD_LEN + 1 + sizeof(uint32_t) * 5, "eeprom body size does not match");
+static_assert(sizeof(vars_body_t) == sizeof(eeprom::v10::vars_body_t) + sizeof(uint8_t) * 3 + PL_API_KEY_SIZE + LAN_HOSTNAME_MAX_LEN + 1 + WIFI_MAX_SSID_LEN + 1 + WIFI_MAX_PASSWD_LEN + 1 + 1 + sizeof(uint32_t) * 5, "eeprom body size does not match");
```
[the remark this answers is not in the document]
dragomirecky: Well, this somehow checks whether the eeprom's body size is correct. I've added one one-byte variable, so I had to „add one“ here. 

I am not the author of this assert. I am just using it the way (I believe) it was designed to be used :)

src/common/w25x.h
```diff
@@ -1,35 +1,69 @@
...
+/// Return the number of available sectors
+extern uint32_t w25x_get_sector_count();
+
+/// Read data from the flash.
+/// Errors can be checked (and cleared) using w25x_fetch_error()
 extern void w25x_rd_data(uint32_t addr, uint8_t *data, uint16_t cnt);
+
+/// Write data to the flash (the sector has to be erased first)
+/// Errors can be checked (and cleared) using w25x_fetch_error()
 extern void w25x_page_program(uint32_t addr, const uint8_t *data, uint16_t cnt);
```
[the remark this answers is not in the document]
dragomirecky: Two different things :)

You write by pages. You erase by sectors.

src/main.cpp
dragomirecky: It is a leftover 😭 Thanks 🙏

src/common/w25x.h
```diff
@@ -1,35 +1,69 @@
...
-extern void w25x_enable_wr(void);
-extern void w25x_disable_wr(void);
-extern uint8_t w25x_rd_status_reg(void);
-extern void w25x_wr_status_reg(uint8_t val);
+/// Return the number of available sectors
+extern uint32_t w25x_get_sector_count();
+
+/// Read data from the flash.
+/// Errors can be checked (and cleared) using w25x_fetch_error()
 extern void w25x_rd_data(uint32_t addr, uint8_t *data, uint16_t cnt);
```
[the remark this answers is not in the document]
dragomirecky: Returning the error was my very first take on adding error handling support when working on this PR.

There are low-level functions of the driver, that produce the errors (and those return them). But, having the same error handling in the layers above made the code very verbose and unreadable. While not adding a much of a value (the API wasn't any easier, nor safer, to work with). 
Also, we don't need to act on the errors immediately. Therefore, I decided for this error-handling design.

src/common/eeprom_current.hpp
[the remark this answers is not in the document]
mkbel: I am not sure if you don't need to create new eeprom version instead of adding something to current version. Or eeprom_current means not released yet?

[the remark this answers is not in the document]
dragomirecky: I believe it is the unreleased one.

src/common/scratch_buffer.cpp
```diff
@@ -1,4 +1,18 @@
...
 #include "scratch_buffer.hpp"
+#include "cmsis_os.h"
 
+osMutexDef(scratch_buffer_mutex);
+osMutexId(scratch_buffer_mutex_id);
 uint8_t scratch_buffer[SCRATCH_BUFFER_SIZE] __attribute__((section(".ccmram")));
+
+uint8_t *scratch_buffer_acquire() {
+    if (scratch_buffer_mutex_id == NULL)
+        scratch_buffer_mutex_id = osMutexCreate(osMutex(scratch_buffer_mutex));
```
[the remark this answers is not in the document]
dragomirecky: Even though I also plan to set this rule, I – personally, until the rule is set – enjoy the readability of the single-statement conditional statements in such simple cases, as is this 😈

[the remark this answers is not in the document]
mkbel: I like to have possibility to put single statement after if condition. But I hate that our current style enforces to put it on next line.

`if (condition) single statement; // Is OK for me, can not easily break during merge conflict resolution.`

```
if (condition)
    single statement; // Pure evil, which can be broken easily during merge conflict resolution.
```

src/common/eeprom_current.hpp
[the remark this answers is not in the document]
mkbel: ok

src/common/scratch_buffer.hpp
```diff
@@ -1,18 +1,24 @@
...
-/// - QR code generator
+/// The scratch buffer isn't safe-to-use outside of the FreeRTOS context.
 ///
 /// TODO: use it for
 /// - SHA256 encoding
+#pragma once
+
+#ifdef __cplusplus
+extern "C" {
+#endif
```
mkbel, line 13: This suggest it is C header, so name of the file is supposed to be scratch_buffer.h according to our code style. *.hpp is reserved for header incompatible with C.

src/common/scratch_buffer.hpp
```diff
@@ -1,18 +1,24 @@
...
+
+#include <inttypes.h>
+
+#define SCRATCH_BUFFER_SIZE 49152
 
-#define SCRATCH_BUFFER_SIZE 49152 ///< ugly but constexpr int does not work
+uint8_t *scratch_buffer_acquire();
```
mkbel, line 19: Why uint8_t * is returned instead of uint8_t[SCRATCH_BUFFER_SIZE]?

mkbel: Or if this is c++ header you can return reference to fixed size array.

src/common/scratch_buffer.hpp
```diff
@@ -1,18 +1,24 @@
...
+void scratch_buffer_release(uint8_t *buffer);
```
mkbel, line 20: Parameter buffer is unused.
If it would be used I would suggest to use pointer to fixed size array or reference to fixed size array (if it would be C++ header)

src/common/w25x.c
```diff
@@ -1,130 +1,135 @@
...
+    W25X_STATUS_SRP = 0x80,
+} w25x_status_t;
+
+static inline void cs_low() {
+    HAL_GPIO_WritePin(GPIOD, GPIO_PIN_7, 0);
+}
+
+static inline void cs_high() {
+    HAL_GPIO_WritePin(GPIOD, GPIO_PIN_7, 1);
+}
```
mkbel, line 56: Pin is configured in hwio_pindef.h with symbolic name, but here hardcoded pin number is used.

src/main.cpp
```diff
@@ -1019,12 +1036,6 @@ static void MX_GPIO_Init(void) {
     HAL_GPIO_Init(GPIOC, &GPIO_InitStruct);
     HAL_GPIO_WritePin(GPIOC, ESP_RST_Pin, GPIO_PIN_RESET);
 #endif
-    /*Configure GPIO pins : FLASH_CSN_Pin */
-    GPIO_InitStruct.Pin = FLASH_CSN_Pin;
-    GPIO_InitStruct.Mode = GPIO_MODE_OUTPUT_PP;
-    GPIO_InitStruct.Pull = GPIO_NOPULL;
-    GPIO_InitStruct.Speed = GPIO_SPEED_FREQ_LOW;
-    HAL_GPIO_Init(GPIOD, &GPIO_InitStruct);
```
mkbel: Initialization of pin removed, but pin definition

`#define FLASH_CSN_Pin               GPIO_PIN_7`
`#define FLASH_CSN_GPIO_Port         GPIOD`

was not removed, also 

```
/*Configure GPIO pin Output Level */
    HAL_GPIO_WritePin(GPIOD, FLASH_CSN_Pin, GPIO_PIN_RESET);
```
was not removed

src/common/scratch_buffer.hpp
dragomirecky: What? I wasn't aware I added the file 🙈

dragomirecky: Ah, I see. I made it usable from C...

dragomirecky: Fixed by removing all changes to the scratch_buffer. They weren't need in the last version this PR.

src/common/scratch_buffer.hpp
dragomirecky: Fixed by removing all changes to the scratch_buffer. They weren't need in the last version this PR.

src/common/scratch_buffer.hpp
dragomirecky: Fixed by removing all changes to the scratch_buffer. They weren't need in the last version this PR.

src/common/w25x.c
dragomirecky: Fixed

src/main.cpp
dragomirecky: Fixed

PR #1829, opened 2022-01-18
Filament sensor fixes

espr14:
Repairs auto load and filament run out, avoids deadlock, optimizes code and performance.

### DIFF
--- a/src/common/filament_sensor.cpp
+++ b/src/common/filament_sensor.cpp
@@ -26,6 +26,7 @@ FSensor::FSensor()
 /*---------------------------------------------------------------------------*/
 //debug functions
 bool FSensor::WasM600_send() {
+    CriticalSection C;
     return status.M600_sent;
 }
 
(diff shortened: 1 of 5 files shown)


src/common/filament_sensor.cpp
```diff
@@ -165,59 +166,50 @@ void FSensor::InitNever() {
...
+        status.Autoload_sent = true;
+        PrintProcessor::InjectGcode("M1400 S65"); //load with return option
     }
+}
 
-    return has_filament ? event::HasFilament : event::NoFilament; // state changed, but NO edge detected, most likely change from init state
+FSensor::event FSensor::generateEvent(fsensor_t previous_state) const {
+    const bool has_filament = fsensor_t::HasFilament == state;
+    const bool had_filament = fsensor_t::HasFilament == previous_state;
+    if (has_filament != had_filament) {
```
RadekVana, line 197: I would not use nested if

RadekVana (dismissed):

RadekVana (approved):

PR #1831, opened 2022-01-20
BFW-2332: Link: Show current gcode name

vorner:
* The long file name inside marlin was not set (apparently lost when
  transitioning to "common" file API). Return the setting back.
* Adjust the API to contain all the relevant names (in the best
  approximation we can do).
* Enabling the relevant test.

### DIFF
--- a/src/common/media.cpp
+++ b/src/common/media.cpp
@@ -2,6 +2,7 @@
 
 #include "media.h"
 #include "log.h"
+#include "lfn.h"
 #include "ff.h"
 #include "usbh_core.h"
 #include "../Marlin/src/gcode/queue.h"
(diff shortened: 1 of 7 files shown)


src/common/media.cpp
```diff
@@ -123,12 +124,14 @@ void media_get_SFN_path(char *sfn, uint32_t sfn_size, char *filepath) {
 
 void media_print_start(const char *sfnFilePath) {
     if (media_print_state == media_print_state_NONE) {
```
espr14, line 126:
```suggestion
    if (media_print_state != media_print_state_NONE)
        return();
```

src/common/media.cpp
```diff
@@ -123,12 +124,14 @@ void media_get_SFN_path(char *sfn, uint32_t sfn_size, char *filepath) {
...
+        if (sfnFilePath) { // null sfnFilePath means use current filename media_print_SFN_path
             strlcpy(media_print_SFN_path, sfnFilePath, sizeof(media_print_SFN_path));
+            get_LFN(media_print_LFN, sizeof(media_print_LFN), media_print_SFN_path);
+        }
         struct stat info = { 0 };
-        stat(sfnFilePath, &info);
+        int result = stat(sfnFilePath, &info);
 
-        if (stat(sfnFilePath, &info) == 0) {
+        if (result == 0) {
```
espr14, line 134:
```suggestion
        if (result != 0)
            return();
```

espr14 (changes requested):

espr14 (approved):

PR #1844, opened 2022-01-31
Sendfile

vorner:


lib/WUI/nhttp/send_file.cpp
```diff
@@ -0,0 +1,71 @@
...
+    , can_keep_alive(can_keep_alive) {
+    struct stat finfo;
+    if (stat(path, &finfo) == 0) {
+        if (can_keep_alive) {
+            connection_handling = ConnectionHandling::ContentLengthKeep;
+        }
+        content_length = finfo.st_size;
+    } else if (can_keep_alive) {
+        connection_handling = ConnectionHandling::ChunkedKeep;
+    }
```
MistrKernnunos, line 22: Add else for closed connection

lib/WUI/link_content/usb_files.cpp
```diff
@@ -0,0 +1,53 @@
...
+optional<ConnectionState> UsbFiles::accept(const RequestParser &parser) const {
+    const string_view uri = parser.uri();
+    const constexpr string_view prefix = "/usb/";
+
+    // We claim the /usb/ namespace.
+    if (uri.size() < prefix.size() || uri.substr(0, prefix.size()) != prefix) {
+        return nullopt;
+    }
+
+    // Content of the USB drive is only for authenticated, don't ever try anything without it.
```
MistrKernnunos, line 23: Should we move the authentication to the top of the function? Before checking if the path is correct.

lib/WUI/link_content/prusa_link_api.cpp
```diff
@@ -82,36 +79,6 @@ optional<ConnectionState> PrusaLinkApi::accept(const RequestParser &parser) cons
              */
             return std::visit([](auto upload) -> ConnectionState { return std::move(upload); }, std::move(upload));
         } else {
-            char fname[FILE_PATH_BUFFER_LEN + 10];
```
MistrKernnunos: Please merge with this commit 425bd5dc53cb1ad70f1c847583c1752f50f0963b

lib/WUI/link_content/usb_files.cpp
vorner: Reading it again (and changing it and running tests), no.

The thing above isn't checking for correctness (if the file exists). It checks if the URL starts with `/usb`. If not (eg. it's `/api/` or `/index.html`), it doesn't „belong“ to this selector and it can't really say anything about it -> it returns `nullopt` and „refuses“ the URL.

MistrKernnunos (approved):

PR #1847, opened 2022-02-02
Park nozzle further from side while loading

MistrKernnunos:
Move the parking point of nozzle further from end of the axis to lower the strain on the PTFE tube

### DIFF
--- a/src/marlin_stubs/pause/M701_M702.cpp
+++ b/src/marlin_stubs/pause/M701_M702.cpp
@@ -49,8 +49,11 @@
 #include "fs_event_autolock.hpp"
 
 #define DO_NOT_RESTORE_Z_AXIS
-static const constexpr uint8_t Z_AXIS_LOAD_POS = 40;
-static const constexpr uint8_t Z_AXIS_UNLOAD_POS = 20;
+static const constexpr size_t Z_AXIS_LOAD_POS = 40;
+static const constexpr size_t Z_AXIS_UNLOAD_POS = 20;
+static const constexpr size_t PARK_POINT[] = NOZZLE_PARK_POINT_M600;
+static const constexpr size_t X_AXIS_LOAD_POS = PARK_POINT[0];
+static const constexpr size_t X_AXIS_UNLOAD_POS = PARK_POINT[0];
 
 using Func = bool (Pause::*)(); //member fnc pointer
 static Pause &pause = Pause::Instance();
@@ -58,7 +61,7 @@ static Pause &pause = Pause::Instance();
 /**
  * Shared code for load/unload filament
  */
-static void load_unload(LoadUnloadMode type, Func f_load_unload, uint32_t min_Z_pos) {
+static void load_unload(LoadUnloadMode type, Func f_load_unload, uint32_t min_Z_pos, uint32_t X_pos) {
     const int8_t target_extruder = GcodeSuite::get_target_extruder_from_command();
     if (target_extruder < 0)
         return;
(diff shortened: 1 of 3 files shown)


yodax (commented):
Wouldn’t X_MAX_POS - X_MAX_POS always be zero?

MistrKernnunos:
It is X_MAX_POS - X_MAX_POS / 4, so it should stop in 3/4 of the axis

yodax:
> It is X_MAX_POS - X_MAX_POS / 4, so it should stop in 3/4 of the axis

Ah ofc. Sorry, I’d say it’s Monday, but that’s not true either.

src/marlin_stubs/pause/M701_M702.cpp
```diff
@@ -51,14 +51,16 @@
 #define DO_NOT_RESTORE_Z_AXIS
 static const constexpr uint8_t Z_AXIS_LOAD_POS = 40;
 static const constexpr uint8_t Z_AXIS_UNLOAD_POS = 20;
+static const constexpr uint8_t X_AXIS_LOAD_POS = X_MAX_POS - X_MAX_POS / 4;
```
RadekVana, line 54: use `size_t`, `uint8_t` is nearly too small

RadekVana: if it should be same as `NOZZLE_PARK_POINT_M600[0]` make it so (do not copy that calculation)
use something like
```
static constexpr size_t park_point[] = NOZZLE_PARK_POINT_M600;
static constexpr size_t  X_AXIS_LOAD_POS = park_point[0];
```

src/marlin_stubs/pause/M701_M702.cpp
```diff
@@ -78,6 +80,10 @@ static void load_unload(LoadUnloadMode type, Func f_load_unload, uint32_t min_Z_
         static const float Z_max = get_z_max_pos_mm();
         park_position.z = std::min(std::max(current_position.z, float(min_Z_pos)), Z_max);
     }
+    if (X_pos > 0) {
```
RadekVana, line 83: use `float `function parameter since it is passed to float anyway
and instead testing `if (X_pos > 0) {` use `isfinite` because 0 is a valid park position

RadekVana (changes requested):

RadekVana (approved):

src/marlin_stubs/pause/M701_M702.cpp
```diff
@@ -199,7 +206,7 @@ static PreheatStatus::Result M1400_no_parser(uint32_t val) {
     case PreheatMode::Unload:
     case PreheatMode::Change_phase1:
         Pause::Instance().SetUnloadLength(NAN);
-        load_unload(LoadUnloadMode::Unload, &Pause::FilamentUnload, Z_AXIS_UNLOAD_POS);
+        load_unload(LoadUnloadMode::Unload, &Pause::FilamentUnload, Z_AXIS_UNLOAD_POS, X_AXIS_LOAD_POS);
```
RadekVana, line 209: why has unload X_AXIS_LOAD_POS?

src/marlin_stubs/pause/pause.cpp
```diff
@@ -579,6 +589,13 @@ void Pause::unpark_nozzle_and_notify() {
     const float Z_len = current_position.z - resume_pos.z; // sign does not matter, does not check Z max val (unlike park_nozzle_and_notify)
     const float XY_len = begin_pos - end_pos;              // sign does not matter
 
+    //home the axis if it is not homed
```
RadekVana, line 592: Just comment:
Unparking is after parking, so axis should be already homed

RadekVana (changes requested):

src/marlin_stubs/pause/M701_M702.cpp
MistrKernnunos: Fixed

src/marlin_stubs/pause/pause.cpp
MistrKernnunos: Beacuse we can move to parking point in one axis only and the resume point could need to move in the other axis.

RadekVana (approved):

PR #1849, opened 2022-02-03
Previews

vorner:


lib/WUI/link_content/previews.cpp
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    uint16_t width;
+    uint16_t height;
+
+    if (uri[prefix.size()] == 's') {
+        width = 16;
+        height = 16;
+    } else if (uri[prefix.size()] == 'l') {
+        // TODO: Can we say "something bigger than 16" instead of hardcoding?
```
DRracer, line 37: I think we can - ideally return the largest thumbnail availble ... but the gcode filter is not ready to list all of the thumbnails and return the largest one.

vorner: That means we have to go through the file twice, which is unfortunate :-(.

Anyway, as it needs some kind of extension to the filter, this might want to go to a separate branch some time later?

lib/WUI/nhttp/gcode_preview.h
```diff
@@ -0,0 +1,32 @@
...
+
+class GCodePreview {
+private:
+    class FileDeleter {
+    public:
+        void operator()(FILE *f) {
+            fclose(f);
+        }
+    };
+    std::unique_ptr<FILE, FileDeleter> gcode;
```
DRracer, line 20: :+1:

DRracer (commented):
:warning: Beware - before merging, all the instances of GCodeThumbDecoder must be verified if they fit into the stack of each task where they originate from!

Previously, there was just one static instance of GCodeThumbDecoder used from the GUI task. It is not a small data structure.

lib/WUI/link_content/previews.cpp
DRracer: I think we can definitely implement something like that, but since the thumbnails can be stored in arbitrary order, you'd have to read the file twice anyway (detect the list of thumbnails and then pick the largest one). I can think of several optimizations, but let's leave that for later.

vorner:
> warning Beware - before merging, all the instances of GCodeThumbDecoder must be verified if they fit into the stack of each task where they originate from!

Is it enough that I've tested manually on a Mini printer (both showing preview from the print dialog and requested the preview over network) and that it passes the integration tests that also show the preview before print & download them? Or should I do something more?

DRracer:
@vorner hopefully yes ... if you can't crash it from the print preview screen and from the web server, it is hopefully safe. Could you check the stack usage watermark?
I've seen a few stack overflow situations in debug mode with the local GCodeThumbDecoder instances, therefore I'm pointing it out. It could have been a coincidence.

DRracer (approved):
approving changes done by @vorner

PR #1851, opened 2022-02-03
Store time in UTC+0

michalrudolf:
Until now time was recalculated every timezone change. This PR fix this bad approach and stores time in UTC+0.


dragomirecky (commented):
I see this reflects the change in the `Print will end` label. Does it also updates how the time is calculated for the time shown in the screen's header?

michalrudolf:
This PR is a duplicate, issue is resolved

PR #1853, opened 2022-02-04
Do not ask for temp if it is high enough

MistrKernnunos:
Do not ask for filament to preheat if purging or unloading if the target nozzle temperature is higher than minimal extrude temperature.


dragomirecky:
It seems that preheating the printer and then selecting „Unload Filament“ does nothing.

Also, please make the code easier to understand. The conditions seem to duplicate code and look very error prone.

MistrKernnunos:
Fixed the problem witn preheating and than uloading.

PR #1860, opened 2022-02-07
Remember pos

RadekVana:
BFW-2442
Define screen init variant
Add methods for screen to work with init variant
Modify screen handler to handle screen init variant
Remember position when returning to home screen
Remember position when nesting in menus
Save some RAM in menus via bit fields

### DIFF
--- a/src/gui/ScreenHandler.cpp
+++ b/src/gui/ScreenHandler.cpp
@@ -3,31 +3,30 @@
 
 Screens *Screens::instance = nullptr;
 
-Screens::Screens(const ScreenFactory::Creator screen_creator)
-    : stack({ { nullptr } })
-    , stack_iterator(stack.begin())
+Screens::Screens(screen_node screen_creator)
+    : stack_iterator(stack.begin())
     , current(nullptr)
-    , creator(screen_creator)
+    , creator_node(screen_creator)
     , close(false)
     , close_all(false)
     , close_serial(false)
     , timeout_tick(0) {
 }
 
-void Screens::Init(const ScreenFactory::Creator screen_creator) {
+void Screens::Init(screen_node screen_creator) {
     static Screens s(screen_creator);
     instance = &s;
 }
 
 Screens::iter Screens::find_enabled_node(iter begin, iter end) {
-    return std::find_if(begin, end, [](const ScreenFactory::Creator &node) { return node != nullptr; });
+    return std::find_if(begin, end, [](const screen_node &node) { return node.creator != nullptr; });
 }
 
 Screens::r_iter Screens::rfind_enabled_node(r_iter begin, r_iter end) {
-    return std::find_if(r_iter(end), r_iter(begin), [](const ScreenFactory::Creator &node) { return node != nullptr; });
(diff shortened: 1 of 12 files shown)


src/gui/ScreenHandler.hpp
```diff
@@ -25,18 +32,20 @@ class Screens {
...
-    Screens(const ScreenFactory::Creator screen_creator);
+    Screens(screen_node screen_creator);
     Screens(const Screens &) = delete;
     static Screens *instance;
 
 public:
     void Loop(); //call inside guiloop
 
-    void Open(const ScreenFactory::Creator screen_creator); //remember creator and create later
+    void Open(const ScreenFactory::Creator screen_creator); //remember creator with and create later with default initialization parameter (like selected item in menu)
```
espr14, line 42:
```suggestion
    void Open(const ScreenFactory::Creator screen_creator); //remember creator and create later with default initialization parameter (like selected item in menu)
```

src/gui/screen_home.cpp
```diff
@@ -216,3 +216,25 @@ bool screen_home_data_t::moreGcodesUploaded() {
     lastUploadCount = total;
     return result;
 }
+
+void screen_home_data_t::InitState(screen_init_variant var) {
+    if (!var.GetPosition())
+        return;
+
+    size_t pos = *(var.GetPosition());
+    if (pos >= sizeof(w_buttons) / sizeof(w_buttons[0]))
```
espr14, line 225: What about `std::array`? Then we can use safer `w_buttons.size()`.

src/guiapi/include/window_menu.hpp
```diff
@@ -2,16 +2,18 @@
...
 class window_menu_t : public IWindowMenu {
-    uint8_t index;    /// index of cursor
-    int8_t moveIndex; /// accumulator for cursor changes
-    bool redrawAll;   /// triggers whole menu redraw
-    bool clicked;     /// triggers click redraw (item->Click invalidates whole menu, but we need to know what happend to redraw only nessessary items)
+    uint8_t index;      /// index of cursor
+    int8_t moveIndex;   /// accumulator for cursor changes
+    uint8_t top_index;  /// offset of currently shown item
+    bool redrawAll : 1; /// triggers whole menu redraw
+    bool clicked : 1;   /// triggers click redraw (item->Click invalidates whole menu, but we need to know what happend to redraw only nessessary items)
```
espr14, line 16:
```suggestion
    bool clicked : 1;   /// triggers click redraw (item->Click invalidates whole menu, but we need to know what happened to redraw only necessary items)
```

espr14 (changes requested):

src/gui/ScreenHandler.cpp
```diff
@@ -210,9 +213,10 @@ void Screens::Loop() {
 }
 
 void Screens::InnerLoop() {
+    screen_init_variant variant;
```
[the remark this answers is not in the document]
RadekVana: ok

espr14 (approved):

PR #1866, opened 2022-02-09
PrusaLink: Support the projects tab

vorner:
* File listing support
* Enable in the web page
* TODO: Missing support for recursive listing (it returns an answer, but
  doesn't recurse).
* TODO: Listing subdirectories.


lib/WUI/link_content/prusa_link_api.cpp
```diff
@@ -89,18 +89,52 @@ optional<ConnectionState> PrusaLinkApi::accept(const RequestParser &parser) cons
...
+                static const char *const roots[] = {
+                    "",
+                    "/",
+                    "/local",
+                    "/local/",
+                    "/sdcard",
+                    "/sdcard/",
+                };
+
+                for (size_t i = 0; i < sizeof roots / sizeof *roots; i++) {
```
DRracer, line 125: nitpick: I prefer seeing:
`sizeof(roots) / sizeof(roots[0])`
but ideally add a template for such a construct like this:
```
template <typename T> const size_t arrsz(const T &aObj){     
    return sizeof(aObj)/sizeof aObj[0];
}
```

lib/WUI/link_content/prusa_link_api.cpp
```diff
@@ -89,18 +89,52 @@ optional<ConnectionState> PrusaLinkApi::accept(const RequestParser &parser) cons
...
+                    if (strcmp(fname_real, roots[i]) == 0) {
```
DRracer, line 126: in this case `strcmp` is safe (`strncmp` not necessary) because we are comparing to statically defined strings in CPUFLASH

lib/WUI/nhttp/file_info.cpp
```diff
@@ -23,19 +23,116 @@ FileInfo::FileInfo(const char *filename, bool can_keep_alive, bool after_upload)
...
+         * We already have a directory listing running. Try to pull another file out of there.
+         */
+        struct dirent *ent;
+        while ((ent = readdir(dir.get()))) {
+            // Check this is GCODE, not some other thing. Skip the others.
+            const char *last_dot = strrchr(ent->d_name, '.');
+            if (last_dot == nullptr) {
+                continue;
+            }
+            if (strcasecmp(last_dot, ".gcode") != 0 && strcasecmp(last_dot, ".gco") != 0) {
```
DRracer, line 58: there are may-be some other supported suffices as well ... `.g`? Ideally if we could unify this with the lazy file list impl.

lib/WUI/nhttp/file_info.cpp
```diff
@@ -23,19 +23,116 @@ FileInfo::FileInfo(const char *filename, bool can_keep_alive, bool after_upload)
...
+                continue;
+            }
+
+            size_t written = render_chunk(handling, output, output_size, [&](uint8_t *output, size_t output_size) {
+                size_t written = 0;
+                if (need_comma) {
+                    assert(output_size > 0);
```
DRracer, line 65: What does the assert do in the FW? I know what it does in unit tests :wink:

lib/WUI/nhttp/file_info.cpp
```diff
@@ -23,19 +23,116 @@ FileInfo::FileInfo(const char *filename, bool can_keep_alive, bool after_upload)
...
+            written += last_chunk_len;
+        }
+        return Step { 0, written, Terminating::for_handling(handling) };
+    }
+
+    /*
+     * Not having a directory listing running _yet_. So we try that. If it's
+     * not a directory, this'll fail and we will try a file.
+     */
+    if (DIR *dir_attempt = opendir(filename); dir_attempt) {
```
DRracer, line 119: nice :+1: (declaration inside an if ... local to if code block)

lib/WUI/nhttp/file_info.cpp
```diff
@@ -23,19 +23,116 @@ FileInfo::FileInfo(const char *filename, bool can_keep_alive, bool after_upload)
...
+                const char *name = ent->d_name;
+                JSONIFY_STR(name);
+#ifdef UNITTESTS
+                const char *long_name = ent->d_name;
+#else
+                const char *long_name = ent->lfn;
+#endif
+                JSONIFY_STR(long_name);
+                written += snprintf(reinterpret_cast<char *>(output + written), output_size - written,
+                    "{"
```
DRracer, line 81: What happens if the snprintf only manages to store a part of the whole string? Are we able to continue in the next packet? How?

DRracer (approved):

lib/WUI/link_content/prusa_link_api.cpp
vorner: I'll change it (though the `*` variant seems more nuanced to `arbitrary element` over `first element`; but let's not go into that discussion 😈).

I'd be a bit vary about the template; I don't remember the exact rules when an array decays to a pointer when passed as a parameter, in which case it could specialize as `arrsz(const Whatever &*aObj)`, which would always return 1… maybe it doesn't, in this case, but...

lib/WUI/link_content/prusa_link_api.cpp
vorner: `strncmp` would actually _not work_. I want exact match of the filename, not a prefix (in either direction). These are the paths that are „special“ in the octoprint API, but the sub-paths are never generated by us and I believe the client shall not make them up (let's see about that).

lib/WUI/nhttp/file_info.cpp
vorner: Similar to abort, if it fails, it kills the thread (which can be observed by the thread getting stuck). Then the backtrace is particularly handy, because it points to the exact location where it happened. Much more usable than bluescreen outputs and dumps, actually.

lib/WUI/nhttp/file_info.cpp
vorner: No, we are not able to continue in the next packet. In such case we would send a bogus answer (but not overwrite any memory or such). We assume packet buffers passed to these functions are „large enough“ (they may get split into smaller packets for the network transmission as necessary).

Being able to continue seems to be a somewhat hard problem that's not yet solved (and I'm not sure if it'll ever need systematic solving).

Nevertheless, it seems the usual answers are small enough and fit comfortably. I was actually thinking about optimizing and being able to put multiple files into the same buffer.

lib/WUI/link_content/prusa_link_api.cpp
DRracer: ok

lib/WUI/link_content/prusa_link_api.cpp
DRracer: I meant since one of the strings is constant, there is no danger of running past the buffer

lib/WUI/nhttp/file_info.cpp
DRracer: oh, it has moved to `src/gui/file_raii.hpp`
`static bool FnameMatchesPattern(const char *fname)`

probably already usable...

lib/WUI/nhttp/file_info.cpp
DRracer: ok

lib/WUI/nhttp/file_info.cpp
DRracer: ok

DRracer (dismissed):

DRracer (dismissed):

DRracer (approved):

PR #1868, opened 2022-02-09
Ask for temp when trying to move E axis

MistrKernnunos:
Ask for temperature when trying to move E axis and the nozzle is too cold

### DIFF
--- a/src/gui/screen_menu_move.cpp
+++ b/src/gui/screen_menu_move.cpp
@@ -5,6 +5,8 @@
 #include "marlin_client.h"
 #include "WindowMenuItems.hpp"
 #include "menu_spin_config.hpp"
+#include "WindowItemFormatableLabel.hpp"
+#include "window_dlg_load_unload.hpp"
 
 template <size_t INDEX, int8_t LONG_SEG, uint8_t BUFFER_LEN>
 class MI_AXIS : public WiSpinInt {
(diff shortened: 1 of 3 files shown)


src/gui/screen_menu_move.cpp
```diff
@@ -64,35 +66,62 @@ class MI_AXIS_E : public MI_AXIS<3, 5, 8> {
...
+                dummy->Hide();
+            }
+        } else {
+            if (!spinner->IsHidden()) {
+                spinner->Hide();
+            }
+            if (dummy->IsHidden()) {
+                dummy->Show();
+            }
+        }
```
espr14, line 114:
```suggestion
        if (temp_ok) {
            spinner->Show();
            dummy->Hide();
        } else {
            spinner->Hide();
            dummy->Show();
        }
```
If we had `SetVisibility(bool)` then we can have just:
```
spinner->SetVisibility(temp_ok);
dummy->SetVisibility(!temp_ok);
```
No `if` needed.

espr14 (changes requested):

src/gui/screen_menu_move.cpp
MistrKernnunos: Changed

espr14 (dismissed):

espr14 (approved):

PR #1877, opened 2022-02-14
nhttp: Declutter the gcode upload

vorner:
Make the gcode upload more C++, straighten few knots resulting from
previous C->C++ migration. Besides nicer code, this:

* Provides a little bit more information in error states.
* Saves one allocation by embedding the upload state inline; this should
  also save some memory during the upload because the variant is already
  largish and was mostly unused at that point.
* Allows parallel gcode uploads (BFW-2216).


lib/WUI/nhttp/upload_state.h
```diff
@@ -4,139 +4,283 @@
...
+     *
+     * Not owned by us.
+     */
+    UploadHooks *hooks = nullptr;
+
+    std::unique_ptr<multipart_parser, MultiparserDeleter> multiparter;
+
+    /*
+     * When reading a string separated by quotes, this points to the current
+     * poisition and an end of a buffer to store it. This makes it possible to
```
espr14, line 187:
```suggestion
     * position and an end of a buffer to store it. This makes it possible to
```

lib/WUI/nhttp/upload_state.h
```diff
@@ -4,139 +4,283 @@
...
+
+    TokenType type : 2;
+    State state : 3;
+    Part part : 2;
+    bool have_valid_filename : 1;
+    /*
+     * The gcode_start hook has been called and we need to eventually call the finish hook too.
+     *
+     * This may get reset back to false in case the finish hook is called or
+     * when there's an error from hook (the hooks should know they erorred out
```
espr14, line 227:
```suggestion
     * when there's an error from hook (the hooks should know they errored out
```

lib/WUI/nhttp/upload_state.h
```diff
@@ -4,139 +4,283 @@
...
+    void feed(const std::string_view &data);
+    /// \brief Returns the last error state, if any.
+    UploadHooks::Result get_error() const {
+        return error;
+    }
+    /// \brief Returns the file name as specified in the form (if already known).
+    const char *get_filename() const {
+        return filename.begin();
+    }
+    /// \brief Returns if the whole form was processed.
```
espr14, line 280:
```suggestion
    /// \brief Returns true if the whole form was processed.
```

lib/WUI/wui_api.c
```diff
@@ -365,46 +358,15 @@ void add_time_to_timestamp(int32_t secs_to_add, struct tm *timestamp) {
...
-        return 0;
-    }
-}
+void wui_uploaded_gcode(const char *filename, bool start_print) {
+    uploaded_gcodes++;
 
-uint16_t wui_upload_finish(const char *old_filename, const char *new_filename, bool start) {
-    uint16_t error_code = 0;
+    (void)filename;
+    (void)start_print;
```
espr14, line 369:
```suggestion
void wui_uploaded_gcode(const char */*filename*/, bool /*start_print*/) {
    uploaded_gcodes++;
```

espr14 (changes requested):

tests/unit/lib/WUI/nhttp/missing_functions.c
```diff
@@ -22,6 +22,16 @@
 
 size_t strlcpy(char *, const char *, size_t);
 
+size_t strlcat(char *dst, const char *src, size_t size) {
```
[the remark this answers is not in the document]
vorner: That would probably mean I have to solve copyright issues with that copy, which wasn't exactly worth it.

lib/WUI/wui_api.c
vorner: Technically, that's not valid `C`. Not having a variable name is valid in C++, but C afaik doesn't allow it (but compilers sometimes accept it).

espr14: `uint16_t wui_upload_finish(const char *, const char *, bool);` was used before in `wui_api.h`.

espr14 (approved):

PR #1878, opened 2022-02-15
nhttp: Prefer textual error pages

vorner:
This is what PrusaSlicer expects (it can handle the extended
Content-Location too, but we are content with the minimal support for
now).

### DIFF
--- a/lib/WUI/nhttp/status_page.cpp
+++ b/lib/WUI/nhttp/status_page.cpp
@@ -23,12 +23,19 @@ Step StatusPage::step(std::string_view, bool, uint8_t *output, size_t output_siz
     if (status == Status::NoContent) {
         content_buffer[0] = 0;
     } else {
-        snprintf(content_buffer, sizeof(content_buffer), "<html><body><h1>%u: %s</h1><p>%s", static_cast<unsigned>(status), text.text, extra_content);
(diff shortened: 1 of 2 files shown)


lib/WUI/nhttp/status_page.cpp
```diff
@@ -23,12 +23,19 @@ Step StatusPage::step(std::string_view, bool, uint8_t *output, size_t output_siz
     if (status == Status::NoContent) {
         content_buffer[0] = 0;
     } else {
-        snprintf(content_buffer, sizeof(content_buffer), "<html><body><h1>%u: %s</h1><p>%s", static_cast<unsigned>(status), text.text, extra_content);
+        snprintf(content_buffer, sizeof(content_buffer), "%u: %s\n\n%s\n", static_cast<unsigned>(status), text.text, extra_content);
```
DRracer, line 26: nitpick: printf format macro :wink: ? Something like PRIu8 or similar based on `status`?

DRracer (approved):
consider the printf formatting macro, otherwise ok

lib/WUI/nhttp/status_page.cpp
vorner: Well, `status` is of type `Status` and there's no `PRI_webstatus` thing around. It does have the values set to the right numeric values (here it's one place where it IMO actually makes sense).

Anyway, I need to cast it to _something_ that's printable and in that case it as well may be `unsigned` as well as anything else.

PR #1879, opened 2022-02-15
Fix HardFault in Metrics::RecordRuntimeStats

laloch:
Make sure to only treat addresses within the heap memory range as heap-allocated while calculating free stack space for a task.

### DIFF
--- a/src/common/app_metrics.cpp
+++ b/src/common/app_metrics.cpp
@@ -23,7 +23,11 @@ void Buddy::Metrics::RecordRuntimeStats() {
     if (ticks_diff(ticks_ms(), last_recorded_ticks) > 3000) {
         int count = uxTaskGetSystemState(task_statuses, sizeof(task_statuses) / sizeof(task_statuses[1]), NULL);
         for (int idx = 0; idx < count; idx++) {
-            size_t s = malloc_usable_size(task_statuses[idx].pxStackBase);
+            const char *stack_base = (char *)task_statuses[idx].pxStackBase;
+            size_t s = 0;
+            /* We can only report free stack space for heap-allocated stack frames. */
+            if (mem_is_heap_allocated(stack_base))
+                s = malloc_usable_size((void *)stack_base);
             const char *task_name = task_statuses[idx].pcTaskName;
             if (strcmp(task_name, "Tmr Svc") == 0)
                 task_name = "TmrSvc";
--- a/src/common/heap.h
+++ b/src/common/heap.h
@@ -7,6 +7,8 @@ extern "C" {
 uint32_t heap_total_size;
 uint32_t heap_bytes_remaining;
 
+uint32_t mem_is_heap_allocated(const void *ptr);
+
 #ifdef __cplusplus
 }
 #endif //__cplusplus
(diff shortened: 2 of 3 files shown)


src/common/app_metrics.cpp
```diff
@@ -23,7 +23,11 @@ void Buddy::Metrics::RecordRuntimeStats() {
     if (ticks_diff(ticks_ms(), last_recorded_ticks) > 3000) {
         int count = uxTaskGetSystemState(task_statuses, sizeof(task_statuses) / sizeof(task_statuses[1]), NULL);
         for (int idx = 0; idx < count; idx++) {
-            size_t s = malloc_usable_size(task_statuses[idx].pxStackBase);
+            char *stack_base = (char *)task_statuses[idx].pxStackBase;
```
DRracer, line 26: `const char *`

src/common/app_metrics.cpp
```diff
@@ -23,7 +23,11 @@ void Buddy::Metrics::RecordRuntimeStats() {
...
+            size_t s = 0;
+            /* We can only report free stack space for heap-allocated stack frames. */
+            if (stack_base >= heap_base && stack_base < heap_limit)
+                s = malloc_usable_size(stack_base);
```
DRracer, line 30: What do we do with the statically allocated stacks? The metrics provides vital information about stack usage, or better to say - unused stack space which can be reclaimed for some more important stuff.

src/common/heap.h
```diff
@@ -7,6 +7,9 @@ extern "C" {
 uint32_t heap_total_size;
 uint32_t heap_bytes_remaining;
 
+extern const char *heap_base;
+extern const char *heap_limit;
```
dragomirecky, line 11: I am not sure I like this API design. It exposes some internals of the heap which would be nicer to keep private.

What about adding a simple function like `bool mem_is_heap_allocated(void *ptr)`? (currently, it would have the super simple implementation, where it would only check whether `ptr` is within the heap's bounds)

src/common/app_metrics.cpp
laloch: OK, I'll do it, but I still need to cast the `const` out later in the call to `malloc_usable_size(...)` since it accepts `void *`. That's why it's not const now.

src/common/heap.h
laloch: I don't really understand why this should be kept internal, but OK, no problem.

dragomirecky: The `mem_is_heap_allocated` gives user „less power“ and keeps things better encapsulated. 

For example, I can imagine that in future, we might want to have two memory blocks reserved for heap. In such case, we would just reflect that in the `mem_is_heap_allocated` implementation. With two exposed variables `heap_base` and `heap_limit`, we would have to refactor all its usages, which might get much more messy.

src/common/app_metrics.cpp
laloch: Done.

src/common/heap.h
laloch: Removed in favor of `mem_is_heap_allocated`.

PR #1881, opened 2022-02-16
WUI: Don't forget AP security

vorner:
When the save_net_params is called without an ap, but with saving flags,
it would forget the AP security. Make sure it is always called with the
ap when called on the wifi iface.

This is not the bestest and nicest code I've ever written, but there's a plan to do an overhaul of networking threads, init, setup, etc anyway (and it should come soon-ish), so I'm doing somewhat minimal patch to make it work in the meantime.


laloch (approved):
Good catch @vorner. Thanks!

PR #1883, opened 2022-02-17
PrusaLink: Delete file support

vorner:
This allows deleting files on the USB drive (this is technically not
limited to gcode only, but the others are not listed so the user needs
to guess the filename; this is kind of on purpose, same as it is
possible to download arbitrary USB file ‒ the user knows the API key).

Unfortunately, the PrusaLink page currently generates some other,
unrelated request, so it doesn't work in the UI ‒ this is only the API
for it.


DRracer (approved):
looks ok to me

PR #1884, opened 2022-02-17
File/thumbnail caching in the browser

vorner:


lib/WUI/nhttp/req_parser.h
```diff
@@ -48,6 +48,23 @@ class RequestParser final : public automata::Execution {
...
+     * on the file metadata. To save space, we use a numeric representation
+     * (part of a hash). The chance of the file having the same name, same FS
+     * metadata and a different content is pretty low, same as the chance of
+     * collision on that 32-bit value and the damage would be small.
+     *
+     * As we produce the etag in the first place, we can decode it when
+     * receiving. A client can give us a wrong one, which would produce garbage
+     * ‒ but that can lead only to either false hit or false miss and that
+     * hurts mostly the client, not us ‒ so they deserve it for the
+     * misbehaviour.
```
DRracer, line 65: :smile:

lib/WUI/nhttp/headers.cpp
```diff
@@ -139,4 +147,33 @@ ContentType guess_content_by_ext(const char *filename) {
...
+uint32_t compute_etag(const struct stat &stat) {
+    /*
+     * We need something that hopefully changes whenever someone replaces a
+     * file with a different file but the same name. Hashing a bit of metadata
+     * seems like a good candidate.
+     */
+    mbedtls_sha256_context ctx;
+    mbedtls_sha256_init(&ctx);
+    mbedtls_sha256_starts_ret(&ctx, false);
+    mbedtls_sha256_update_ret(&ctx, (const uint8_t *)&stat.st_ino, sizeof stat.st_ino);
```
DRracer, line 159: Have you checked, that our POSIX filesystem layer is actually able to provide some meaningful inode number from the underlying FATfs and littleFS?

lib/WUI/nhttp/headers.cpp
```diff
@@ -139,4 +147,33 @@ ContentType guess_content_by_ext(const char *filename) {
...
+    mbedtls_sha256_update_ret(&ctx, (const uint8_t *)&stat.st_size, sizeof stat.st_size);
+    mbedtls_sha256_update_ret(&ctx, (const uint8_t *)&stat.st_mtime, sizeof stat.st_mtime);
```
DRracer, line 161: ... also, for G-code files ordering in the filebrowser we have to use both creation time and modification time (which :window: s sets separately unlike linux), so I'd better use both times for computing hash.

lib/WUI/nhttp/file_info.cpp
```diff
@@ -106,6 +106,7 @@ Step FileInfo::step(std::string_view, bool, uint8_t *output, size_t output_size)
             memcpy(output + written, LAST_CHUNK, last_chunk_len);
             written += last_chunk_len;
         }
+        dir.reset();
```
DRracer, line 109: Why do you need resetting the `dir` here? Just a precaution/saving stack space before creating the returned structures?

lib/WUI/nhttp/gcode_preview.cpp
```diff
@@ -54,6 +67,7 @@ Step GCodePreview::step(string_view, bool, uint8_t *buffer, size_t buffer_size)
             } else {
                 // The decoder doesn't distinguish between error or end, so we handle both as end.
                 instruction = Terminating::for_handling(handling);
+                gcode.reset();
```
DRracer, line 70: again - please explain the `reset`

lib/WUI/nhttp/headers.cpp
vorner: Good tip. It's not worth picking the newest, but taking both is better ‒ better chance a different file won't provide the same values.

lib/WUI/nhttp/file_info.cpp
vorner: Dir is (and always was) a member variable. It holds the position in listing file entries between sending individual files/packets. And I suspect there's some dynamic allocation (in `opendir`) that I don't know how to avoid :-(

It's not 100% necessary (it would get closed in the destructor soon), but I wanted to return the memory sooner if possible, as the integration tests often fail on out of memory (I'm not sure this helps, though).

Should I put some explanation into the code, or is this OK?

DRracer: explained enough :+1:

DRracer (approved):

PR #1891, opened 2022-02-22
Remove unneded marlin client

vorner:
We got rid of one of the threads that were running the web server (now
there's just one, not two) ‒ that means we don't need the extra marlin
client.


zoltan-l (approved):
Ok two threads were confusing

PR #1895, opened 2022-02-24
Turn off AlTCP

vorner:
We don't need it now, we have both interfaces directly on LwIP. Tests
still use AlTCP and the code _syntactically_ does so too (it only turns
off the config option, which creates relevant typedefs/defines to turn
the code into basic TCP manipulation in compile time).

This saves about 3kB of flash and maybe 0.5kB of RAM.

### DIFF
--- a/tests/unit/lib/WUI/nhttp/server_tests.cpp
+++ b/tests/unit/lib/WUI/nhttp/server_tests.cpp
@@ -217,13 +217,10 @@ private:
     friend class MockServer;
     vector<shared_ptr<ConnInfo>> &infos;
     static const constexpr Selector *selectors_array[] = { &validate_request, &fake_index, &fake_api, &unknown_request };
-    altcp_pcb *mock_alloc_inner() {
+    altcp_pcb *new_conn() const {
         MockConn *conn = new MockConn;
         infos.push_back(conn->info);
         return conn;
-    }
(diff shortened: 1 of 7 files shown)


tests/unit/lib/WUI/nhttp/server_tests.cpp
```diff
@@ -217,29 +217,22 @@ class MockServerDefs final : public ServerDefs {
...
-        altcp_allocator_t alloc = {
-            mock_alloc,
-            (void *)this,
-        };
-        return alloc;
-    }
-    virtual uint16_t port() const override { return 80; }
+    virtual altcp_pcb *listener_alloc() const override {
+        auto conn = new_conn();
+        return altcp_listen(conn);
```
espr14, line 234:
```suggestion
        return altcp_listen(new_conn());
```

espr14 (approved):

tests/unit/lib/WUI/nhttp/server_tests.cpp
vorner: I sometimes prefer not to nest the function calls too much, mostly for readability. In here, I split it mostly because I want to hint that creating the connection and listening are two separate actions, not two parts of a single one. Therefore, I'll prefer keeping it this way as it is.

PR #1897, opened 2022-02-25
Rip out alsockets implementation

vorner:
They don't switch anything anymore. Use the sockets emulation layer from
LwIP for that instead.

Saves some of our (broken!) code. It also saves about 8kB of flash.


vladamatena (approved):
Looks good, works

PR #1899, opened 2022-02-25
Provide filament (un)load user abort

espr14:


src/common/marlin_server.hpp
```diff
@@ -25,6 +25,9 @@ void fsm_change(ClientFSM type, T phase, fsm::PhaseData data) {
     _fsm_change(type, fsm::BaseData(GetPhaseIndex(phase), data));
 }
 
+// true if user can stop waiting for heating/cooling by pressing a button
+extern bool can_stop_wait_for_heatup;
```
RadekVana, line 29: don't extern variables
include header containing getter/setter function declaration

btw it seams to be unused, is it dead code?

src/marlin_stubs/pause/M70X.cpp
```diff
@@ -238,22 +239,26 @@ static PreheatStatus::Result M1400_no_parser(uint32_t val) {
     case PreheatMode::None:
         return Filaments::CurrentIndex() == filament_t::NONE ? PreheatStatus::Result::DoneNoFilament : PreheatStatus::Result::DoneHasFilament;
     case PreheatMode::Purge:
+        Pause::Instance().StopReset();
```
RadekVana, line 242: note:
I think table (mask) containing which state should call `Pause::Instance().StopReset();` would be better

RadekVana: or just do use one more switch before this one
```
switch (data.Mode()) {
    case PreheatMode:: some_state :
    ...
    case PreheatMode:: some_state :
      Pause::Instance().StopReset();
    default:
      break;
}
```

src/marlin_stubs/pause/pause.cpp
```diff
@@ -263,12 +263,20 @@ bool Pause::ensureSafeTemperatureNotifyProgress(uint8_t progress_min, uint8_t pr
         return true;
     }
 
-    setPhase(PhasesLoadUnload::WaitingTemp, progress_min);
+    setPhase(can_stop ? PhasesLoadUnload::WaitingTemp_stoppable : PhasesLoadUnload::WaitingTemp_unstoppable, progress_min);
```
RadekVana, line 266: note:
this `can_stop ? ... : ...` is OK now
but we really should add transition table (and unify all state machines in FW)
it is my code, so it is not your fault 😃

RadekVana (changes requested):

src/common/marlin_server.hpp
espr14: A lot of added code with no advantage.

RadekVana (approved):

PR #1908, opened 2022-03-07
Add heater PWM metrics.

mkbel:


src/common/safe_state.cpp
```diff
@@ -33,4 +43,7 @@ void hwio_safe_state(void) {
     yEnable.write(Pin::State::high);
     zEnable.write(Pin::State::high);
     e0Enable.write(Pin::State::high);
+
+    force_record_metric_zero_integer_and_disable(metric_nozzle_PWM);
+    force_record_metric_zero_integer_and_disable(metric_bed_PWM);
```
dragomirecky, line 48: What is the motivation for this here? Why do we need to record zero and disable those metrics in safe state?

src/common/hwio_buddy_2209_02.cpp
```diff
@@ -532,14 +535,22 @@ void digitalWrite(uint32_t marlinPin, uint32_t ulVal) {
 #ifdef SIM_HEATER_BED_ADC
         sim_bed_set_power(ulVal ? 100 : 0);
 #else //SIM_HEATER_BED_ADC
-        _hwio_pwm_analogWrite_set_val(HWIO_PWM_HEATER_BED, ulVal ? _pwm_analogWrite_max[HWIO_PWM_HEATER_BED] : 0);
+    {
+        const uint32_t value = ulVal ? _pwm_analogWrite_max[HWIO_PWM_HEATER_BED] : 0;
+        _hwio_pwm_analogWrite_set_val(HWIO_PWM_HEATER_BED, value);
```
dragomirecky, line 540: As there is more places where the bed/nozzle pwm is being set, wouldn't it be cleaner to record the metric within `_hwio_pwm_analogWrite_set_val`?

src/common/hwio_buddy_2209_02.cpp
```diff
@@ -49,6 +49,9 @@ enum {
 };
 static const int _FAN_CNT = _FAN_ID_MAX - _FAN_ID_MIN + 1;
 
+metric_t metric_nozzle_PWM = METRIC("nozzle_PWM", METRIC_VALUE_INTEGER, 1000, METRIC_HANDLER_DISABLE_ALL);
+metric_t metric_bed_PWM = METRIC("bed_PWM", METRIC_VALUE_INTEGER, 1000, METRIC_HANDLER_DISABLE_ALL);
```
dragomirecky, line 53: /subjective, low priority/ I would slightly prefer `nozzle_pwn` and `bed_pwm` names. The original naming style was snake_case (lowercase only). However, it has been broken with some metrics already :(

dragomirecky: /suggestion/ This seem to record integers between 0-255. Please change to float values (0.0 - 1.0). The 8 bit values are implementation specific and we don't really care about them in metrics.

src/common/safe_state.cpp
mkbel: Because analog output pin here is changed to digital pin and its output value is set to 0. So any subsequent setting of PWM duty cycle done from hwio_....cpp has no effect on output value.

src/common/hwio_buddy_2209_02.cpp
mkbel: > /suggestion/ This seem to record integers between 0-255. Please change to float values (0.0 - 1.0). The 8 bit values are implementation specific and we don't really care about them in metrics.

I don't like that suggestion. As using this raw value already helped me to find a bug - maximum value was 254 and not 255 as intended. Here we are implementing [Arduino analogWrite](https://www.arduino.cc/reference/en/language/functions/analog-io/analogwrite/) which has range 0 to 255 so it is kind a platform independent. Value in float is here not available - this would mean we need to compute it and I don't like that idea.

mkbel: > /subjective, low priority/ I would slightly prefer `nozzle_pwn` and `bed_pwm` names. The original naming style was snake_case (lowercase only). However, it has been broken with some metrics already :(

I can rename it if you wish.

src/common/hwio_buddy_2209_02.cpp
mkbel: Yeah, this makes sense.

src/common/hwio_buddy_2209_02.cpp
mkbel: preferably using nozzle_pwm to nozzle_pwned

src/common/hwio_buddy_2209_02.cpp
mkbel: fixed

src/common/hwio_buddy_2209_02.cpp
mkbel: I put the metric to Arduino layer, maybe it would make sense to put it even deeper and log value put to HW register - at this moment value range is the same 0 - 255 but it may change in future and it may help us to find potential error in abstraction layer.

Downside would be, that repeatedly setting the same value would not produce any metric. I don't know if it is good or bad.
We can even implement not-sending-the-same value right now for free if this is considered beneficial. In such case it would be worth to enable this metric during compile time forever, as it would not send useless data when printer is idling - PWM value is 0 then and would be sent just once.

src/common/safe_state.cpp
dragomirecky: but why do we care? We are in a safe state. We aren't going to record any metrics whatsoever

src/common/hwio_buddy_2209_02.cpp
dragomirecky: > > /suggestion/ This seem to record integers between 0-255. Please change to float values (0.0 - 1.0). The 8 bit values are implementation specific and we don't really care about them in metrics.
> 
> I don't like that suggestion. As using this raw value already helped me to find a bug - maximum value was 254 and not 255 as intended. Here we are implementing [Arduino analogWrite](https://www.arduino.cc/reference/en/language/functions/analog-io/analogwrite/) which has range 0 to 255 so it is kind a platform independent. Value in float is here not available - this would mean we need to compute it and I don't like that idea.

You would find the bug either way (as you wouldn't see the value 1.0, but 0.996). However I am fine with keeping it as a raw value...

dragomirecky: > Downside would be, that repeatedly setting the same value would not produce any metric. I don't know if it is good or bad. We can even implement not-sending-the-same value right now for free if this is considered beneficial. In such case it would be worth to enable this metric during compile time forever, as it would not send useless data when printer is idling - PWM value is 0 then and would be sent just once.

That is a wanted behaviour. We want to periodically send values for metrics even if they did not change.

mkbel: ok. Please mark it as resolved.

src/common/safe_state.cpp
mkbel: removed

dragomirecky (approved):

PR #1914, opened 2022-03-09
gui: Send all FSM messages from marlin_server

espr14:
More messages can be in the queue but only one was sent and the flag was cleared. Now all the messages are sent before the flag is cleared.

### DIFF
--- a/src/common/marlin_server.cpp
+++ b/src/common/marlin_server.cpp
@@ -130,7 +130,7 @@ extern osMessageQId marlin_client_queue[MARLIN_MAX_CLIENTS]; // input queue hand
 
 static void _server_print_loop(void);
 static int _send_notify_to_client(osMessageQId queue, variant8_t msg);
-static int _send_notify_event_to_client(int client_id, osMessageQId queue, MARLIN_EVT_t evt_id, uint32_t usr32, uint16_t usr16);
+static bool _send_notify_event_to_client(int client_id, osMessageQId queue, MARLIN_EVT_t evt_id, uint32_t usr32, uint16_t usr16);
 static uint64_t _send_notify_events_to_client(int client_id, osMessageQId queue, uint64_t evt_msk);
 static uint8_t _send_notify_event(MARLIN_EVT_t evt_id, uint32_t usr32, uint16_t usr16);
 static int _send_notify_change_to_client(osMessageQId queue, uint8_t var_id, variant8_t var);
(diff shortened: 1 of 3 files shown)


src/common/marlin_server.cpp
```diff
@@ -775,19 +775,35 @@ static int _send_notify_to_client(osMessageQId queue, variant8_t msg) {
     return 1;
 }
 
+// send all FSM messages from the FSM queue
+static bool _send_FSM_event_to_client(int client_id, osMessageQId queue, MARLIN_EVT_t evt_id, uint32_t usr32, uint16_t usr16) {
```
RadekVana, line 779: unused parameters `MARLIN_EVT_t evt_id, uint32_t usr32, uint16_t usr16`

src/common/marlin_server.cpp
```diff
@@ -775,19 +775,35 @@ static int _send_notify_to_client(osMessageQId queue, variant8_t msg) {
...
+    if (evt_id != MARLIN_EVT_FSM)
```
RadekVana, line 780: this in unnecessary

src/common/marlin_server.cpp
```diff
@@ -775,19 +775,35 @@ static int _send_notify_to_client(osMessageQId queue, variant8_t msg) {
...
+
+        //erase sent item from queue
+        fsm_event_queues[client_id].Pop();
+    }
+}
+
 // send event notification to client (called from server thread)
-static int _send_notify_event_to_client(int client_id, osMessageQId queue, MARLIN_EVT_t evt_id, uint32_t usr32, uint16_t usr16) {
+static bool _send_notify_event_to_client(int client_id, osMessageQId queue, MARLIN_EVT_t evt_id, uint32_t usr32, uint16_t usr16) {
+    if (evt_id == MARLIN_EVT_FSM)
```
RadekVana, line 799: move this code in following switch
```
case MARLIN_EVT_FSM:
return _send_FSM_event_to_client(client_id, queue);
```

src/guiapi/src/window.cpp
```diff
@@ -32,6 +32,7 @@ void window_t::Validate(Rect16 validation_rect) {
 }
 
 void window_t::Invalidate(Rect16 invalidation_rect) {
+    // FIXME: invalidation if invalidation_rect is empty is dangerous
```
RadekVana, line 35: note:
this should not be part of this PR

RadekVana (changes requested):

RadekVana (approved):

PR #1915, opened 2022-03-09
gui: Improve redraw handling

espr14:
If all messages are received then redraw immediately. If a message is still in queue then skip the redraw to avoid 2 redraws one after another (screen blink). If no message is read then only one redraw is called in gui_loop.
BFW-2443

### DIFF
--- a/src/gui/dialogs/DialogHandler.hpp
+++ b/src/gui/dialogs/DialogHandler.hpp
@@ -4,6 +4,12 @@
 #include "fsm_types.hpp"
 #include "DialogFactory.hpp"
 
+enum class redraw_cmd_t : int {
+    none,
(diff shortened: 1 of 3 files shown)


src/gui/dialogs/DialogHandler.hpp
```diff
@@ -4,6 +4,12 @@
 #include "fsm_types.hpp"
 #include "DialogFactory.hpp"
 
+enum class redraw_cmd_t : int {
```
RadekVana, line 7: note:
`enum class redraw_cmd_t : int {`
int is default so consider just `enum class redraw_cmd_t {`

RadekVana (approved):

PR #1924, opened 2022-03-14
Delete an unused and broken function

vorner:
The media_get_SFN_path wasn't used anywhere. Furthermore, it didn't
work:
* It could infinite loop.
* It probably could overwrite data outside of the provided buffer.


wavexx (approved):
100% code removal diffstat is always good in my book ;)

PR #1927, opened 2022-03-15
Marlin: Provide draining of motion buffer

espr14:


wavexx (approved):

lib/Marlin/Marlin/src/module/probe.cpp
```diff
@@ -555,6 +556,9 @@ static float run_z_probe() {
 
     // Do a first probe at the fast speed
     if (do_probe_move(z_probe_low_point, MMM_TO_MMS(Z_PROBE_SPEED_FAST))) {
+      if (planner.draining())
+        return NAN;
+
```
mkbel, line 561: Ineffective. Wrong branch of the code checked. If do_probe_move() fails we are returning NAN anyway 3 lines of code later.

PR #1928, opened 2022-03-17
Use External Flash for Resources

dragomirecky:
tbd


src/resources/bootstrap.cpp
```diff
@@ -0,0 +1,491 @@
...
+
+    int percent_done() {
+        if (scan_result.has_value() == false) {
+            return false;
+        }
+        const int points_per_file = 100;
+        const int points_per_directory = 100;
+        const int points_per_byte = 1;
+
+        int total_points = scan_result->files_count * points_per_file
```
vorner, line 168: Again, should these be `int`s?

src/resources/bootstrap.cpp
```diff
@@ -0,0 +1,491 @@
...
+    }
+
+    void assign_scan_result(ResourcesScanResult result) {
+        scan_result = result;
+    }
+
+    void reset() {
+        files_copied = directories_copied = bytes_copied = 0;
+        scan_result.reset();
+        last_reported_percent_done = -1;
```
vorner, line 207: :raised_eyebrow: Would something like `optional<size_t>` work better?

src/resources/bootstrap.cpp
```diff
@@ -0,0 +1,491 @@
...
+}
+
+static bool has_bbf_suffix(const char *fname) {
+    char *dot = strrchr(fname, '.');
+
+    if (dot == nullptr) {
+        return 0;
+    }
+
+    return strcmp(dot, ".bbf") == 0 || strcmp(dot, ".BBF") == 0;
```
vorner, line 243: strcasecmp?

src/resources/bootstrap.cpp
```diff
@@ -0,0 +1,491 @@
...
+        bool success = do_bootstrap(progress_hook);
+        if (success) {
+            log(LOG_SEVERITY_INFO, "Bootstrap successful");
+            return true;
+        } else {
+            log(LOG_SEVERITY_INFO, "Bootstrap failed. Retrying in a moment...");
+            osDelay(1000);
+        }
+    }
+    return false;
```
vorner, line 490: Unreacable?

src/resources/bootstrap.cpp
```diff
@@ -0,0 +1,491 @@
...
+
+bool buddy::resources::bootstrap(ProgressHook progress_hook) {
+    while (true) {
+        bool success = do_bootstrap(progress_hook);
+        if (success) {
+            log(LOG_SEVERITY_INFO, "Bootstrap successful");
+            return true;
+        } else {
+            log(LOG_SEVERITY_INFO, "Bootstrap failed. Retrying in a moment...");
+            osDelay(1000);
```
vorner, line 487: So, if I don't have the right file around, will this just wait during boot (without much info on the screen) doing nothing?

vorner (commented):
First, let me check if I'm reading it correctly. If I flash by bbf, restart, it „just works“ after the reboot and everything is great. If I flash by the programmer, then I have to _also_ provide the bbf to the USB drive and make sure it's the right one, and it'll get „stuck“ during boot unless I comply. Right?

I have few minor nitpicks in the code (style, if using `int` is the right type, etc)… nothing serious.

I also have two bigger concerns. Nevertheless, these are more like „leftovers“ than „bugs“, so if you'd want to deal with them later, that's fine.

# Feedback

The infinite loop „give me the right bbf“ _might_ use some feedback to the user. I mean, end-users don't really flash by the programmer, so they are unlikely to hit this problem (well, they could flash and remove the USB before rebooting or something, but, well...)

# Crash recovery

Maybe I just didn't pay attention enough (I've done a review, not a complete audit 😈), but what happens if the USB fails/is removed during the bootstrap of the littlefs? Can it so happen that the new hash file is written before the USB is removed, but some files are then missing? Or is it (through some side effect) made sure that the file is written last?

Is there / will there be a way for the user to manually trigger an „emergency“ re-flash of the BBF?

# Removing files

What if some file is no longer needed?

Specifically, the web resources are often named like `some-hash.css`, where the hash changes between versions. I didn't see code to prune old files, so these could be accumulating over time and eventually fill the flash. What's the planned solution for this?

Is there / will there be a way to completely wipe the xflash, same as there's a way to wipe eeprom?

src/common/bbf.cpp
```diff
@@ -0,0 +1,34 @@
...
+    while (!feof(bbf) && !ferror(bbf)) {
+        TLVType current_type;
+        if (fread(&current_type, 1, sizeof(current_type), bbf) != sizeof(current_type)) {
+            break;
+        }
+        if (fread(&length, 1, sizeof(length), bbf) != sizeof(length) || length == 0) {
+            break;
+        }
+        if (current_type == type) {
+            found = true;
```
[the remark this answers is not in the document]
dragomirecky: 👍 just a leftover from moving the code around

src/common/bbf.cpp
```diff
@@ -0,0 +1,34 @@
+#include "bbf.hpp"
+#include "log.h"
+
+bool buddy::bbf::seek_to_tlv_entry(FILE *bbf, TLVType type, uint32_t &length) {
+    // read firmware size
+    fseek(bbf, 96, SEEK_SET);
```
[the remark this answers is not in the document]
dragomirecky: yep, extracted :)

src/guiapi/include/guitypes.hpp
```diff
@@ -16,6 +17,26 @@ struct point_t {
 using point_i16_t = point_t<int16_t>;
 using point_ui16_t = point_t<uint16_t>;
 
+enum class layout_color : uint8_t { leave_it,
+    black,
+    red };
+
+struct GUIStartupProgress {
+    int percent_done;
```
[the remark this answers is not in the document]
dragomirecky: Hmm, this is intentional. But subjective and most likely I will just change it to unsigned as that is the general „go to“ way.

_I think I am influenced here by Swift. Its design guidelines suggest using int everywhere, where you don't explicitly need unsigned range. Main reason being code interoperability and getting rid of the need to convert often between different types. I somehow really like the simplicity of it and it applies well to C too..._

src/resources/bootstrap.cpp
dragomirecky: see comment above :)

src/resources/bootstrap.cpp
dragomirecky: Yep, rn, it is just gonna be saying „Waiting for bbf“. Until release, we definitely need to improve the UX in such cases. But for now, I think it is enough

dragomirecky:
> First, let me check if I'm reading it correctly. If I flash by bbf, restart, it „just works“ after the reboot and everything is great. If I flash by the programmer, then I have to _also_ provide the bbf to the USB drive and make sure it's the right one, and it'll get „stuck“ during boot unless I comply. Right?

Exactly.

When flashing via bootloader, the bootloader finishes the flashing, starts the firmware and firmware continues in copying the files. The plan is to adjust the UI of the process so user would not notice that part of the update process is done by the firmware. But we can implement this later.

> I have few minor nitpicks in the code (style, if using `int` is the right type, etc)… nothing serious.
> 
> I also have two bigger concerns. Nevertheless, these are more like „leftovers“ than „bugs“, so if you'd want to deal with them later, that's fine.
> 
> # Feedback
> The infinite loop „give me the right bbf“ _might_ use some feedback to the user. I mean, end-users don't really flash by the programmer, so they are unlikely to hit this problem (well, they could flash and remove the USB before rebooting or something, but, well...)

As mentioned above, better UX is definitely needed. For now, the user is stuck with „Waiting for bbf“ on screen which I think is enough for internal testing.

> # Crash recovery
> Maybe I just didn't pay attention enough (I've done a review, not a complete audit 😈), but what happens if the USB fails/is removed during the bootstrap of the littlefs? Can it so happen that the new hash file is written before the USB is removed, but some files are then missing? Or is it (through some side effect) made sure that the file is written last?

It should be resilient to that. You can plug it and unplug it in the process as you want. Unplugging reset the process and next time you start from the beginning. 

It should not happen that a new hash file would be written before the process ends. The hash file is not part of the resources files, but it sits next to it and is handled manually in the code (see `src/resources/revision.cpp`). It is updated after a successful update of the resources directory.

> Is there / will there be a way for the user to manually trigger an „emergency“ re-flash of the BBF?

Emergency reflash of bbf is already supported by the bootloader. But you might mean emergency reflash of the external flash? I think we might need that (as a way of a full factory reset). I have a note that we should look into it at some point.

> # Removing files
> What if some file is no longer needed?
> 
> Specifically, the web resources are often named like `some-hash.css`, where the hash changes between versions. I didn't see code to prune old files, so these could be accumulating over time and eventually fill the flash. What's the planned solution for this?
> 
> Is there / will there be a way to completely wipe the xflash, same as there's a way to wipe eeprom?

All the resources are stored in a `/internal/res/` subdirectory. Changing the revision of resources mean that we 1) remove the directory and 2) create a new one and copy files into it. So we have no problem of the resources accumulating there :)

src/guiapi/include/guitypes.hpp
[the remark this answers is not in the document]
vorner: Hmm, I'm probably on the other side of spectrum, influenced by Rust, where even integral types of different sizes are not (implicitly) compatible with each other and tend to „hint“ the purpose of the variable by the type. Mostly, I tend to ask:

* Can the value naturally contain negative values, or would negative values do evil things if ever found there (sizes, counts, indexes ‒ if I _somehow_ get -5 items in an array, or index of -13, then crazy things happen).
* Do I need _specific_ size (port numbers are 16 bits by definition, etc), or am I OK with the „default“ size? And, if it's something that's supposed to mean number of items that fit somewhere, I tend to use `size_t`.

I don't want to push that too much, but I'm a bit of a fan of avoiding the possibility of „invalid states“. Yes, that's a bit revolutionary stance in this team, I'm afraid 😈

vorner (approved):

PR #1929, opened 2022-03-18
Handle JSON responses bigger than a buffer

vorner:
I've had a leftover, I kind of hoped responses would fit into a packet. It turned out this is not always true (depending on the length of paths and URLs), so this is probably the right time to solve it properly.

This introduces a JSON renderer that can pause in the middle and some framework to help writing the content for that renderer. Ported all the parts to it (not only the ones that I've seen to not fit, but everything just to be on the safe side). It's also probably a little bit more readable code.

The usage of the preprocessor is not a perfect and elegant solution, but it seems good enough (not worth creating something more overengineered).

Note that the `basic_gets.cpp` is for the most part just moved `wui_REST_api.c`, but modified to use this renderer. So all the switches deciding what state the printer is and such are not new code 😇

BFW-2362


lib/WUI/nhttp/file_info.cpp
```diff
@@ -37,153 +143,83 @@ FileInfo::FileInfo(const char *filename, bool can_keep_alive, bool after_upload)
...
+            renderer = DirRenderer(this, dir_attempt);
+            written = write_headers(output, output_size, after_upload ? Status::Created : Status::Ok, ContentType::ApplicationJson, handling);
+        } else if (stat(filename, &finfo) == 0) {
+            renderer = FileRenderer(this, finfo.st_size);
+            written = write_headers(output, output_size, after_upload ? Status::Created : Status::Ok, ContentType::ApplicationJson, handling);
+        } else if (strcmp(filename, "/usb") == 0) {
+            // We are trying to list files in the root and it's not there -> USB is missing.
+            // Special case it, we return empty list of files.
+            renderer = DirRenderer(); // Produces empty file list
+            written = write_headers(output, output_size, after_upload ? Status::Created : Status::Ok, ContentType::ApplicationJson, handling);
```
espr14, line 177: 3x the same is worth deduplication.

lib/WUI/nhttp/file_info.cpp
```diff
@@ -37,153 +143,83 @@ FileInfo::FileInfo(const char *filename, bool can_keep_alive, bool after_upload)
...
-     */
-    if (DIR *dir_attempt = opendir(filename); dir_attempt) {
-        dir.reset(dir_attempt);
-
-        size_t written = write_headers(output, output_size, after_upload ? Status::Created : Status::Ok, ContentType::ApplicationJson, handling);
-
+    // Process both the renderers in uniform way.
+    JsonRenderer *rend = get_if<DirRenderer>(&renderer);
+    if (rend == nullptr) {
+        rend = get_if<FileRenderer>(&renderer);
```
espr14, line 191:
```suggestion
    rend = rend ?: get_if<FileRenderer>(&renderer);
```

lib/WUI/nhttp/file_info.cpp
```diff
@@ -20,6 +24,108 @@ namespace nhttp::printer {
...
+        // Check this is GCODE, not some other thing. Skip the others.
+        if (filename_is_gcode(candidate->d_name)) {
+            ent = candidate;
+            return;
+        }
+    }
+    // Left to the default nullptr when runs out of entities in the directory
+}
+
+JsonRenderer::ContentResult FileInfo::DirEntryRenderer::content(size_t resume_point, Output &output) {
```
espr14, line 42: `using namespace JsonRenderer`?

lib/WUI/link_content/basic_gets.cpp
```diff
@@ -0,0 +1,223 @@
...
+        break;
+    case mpsFinishing_WaitIdle:
+    case mpsFinishing_ParkHead:
+        busy = true;
+        ready = operational = false;
+        break;
+    case mpsAborted:
+    case mpsFinished:
+    case mpsIdle:
+    default:
```
espr14, line 67: This should not be here because compiler won't check if all states are mentioned in the switch. And we want to handle all states including the future ones.

lib/WUI/link_content/basic_gets.cpp
```diff
@@ -0,0 +1,223 @@
...
+     * FIXME: The netdev_get_hostname doesn't properly synchronize. That needs
+     * a fix of its own. But to not make things even worse than they are, we
+     * make sure to copy it out to our side first and make sure it doesn't
+     * change during the JSON stringification which could lead to a different
+     * length of the output and stack smashing.
+     */
+    const char *hostname_unsynchronized = netdev_get_hostname(netdev_get_active_id());
+    const size_t hostname_in_len = strlen(hostname_unsynchronized);
+    char hostname[hostname_in_len + 1];
+    memcpy(hostname, hostname_unsynchronized, hostname_in_len);
```
espr14, line 130: Might include size of `char`.

lib/WUI/link_content/basic_gets.cpp
```diff
@@ -0,0 +1,223 @@
...
+            JSON_FIELD_OBJ("job");
+                JSON_FIELD_INT("estimatedPrintTime", vars->print_duration + vars->time_to_end) JSON_COMMA;
+                JSON_FIELD_OBJ("file")
+                    JSON_FIELD_STR("name", vars->media_LFN) JSON_COMMA;
+                    JSON_FIELD_STR("path", vars->media_SFN_path) JSON_COMMA;
+                    JSON_FIELD_STR("display", vars->media_LFN);
+                JSON_OBJ_END;
+            JSON_OBJ_END JSON_COMMA;
+            JSON_FIELD_OBJ("progress");
+                JSON_FIELD_FFIXED("completion", ((double)vars->sd_percent_done / 100.0)) JSON_COMMA;
```
espr14, line 211: Division should be done in float, the result can be converted. Double is not supported by HW so it is emulated. Also multiplication by 0.01f will be faster. Not sure if compiler will convert it.

lib/WUI/nhttp/file_info.cpp
```diff
@@ -83,26 +72,43 @@ JsonRenderer::ContentResult FileInfo::DirRenderer::content(size_t resume_point,
     JSON_OBJ_START;
         JSON_FIELD_ARR("files");
 
-        JSON_ITERATE(*this);
+        // Note: ent, as the control valiable, needs to be preserved inside the
```
espr14, line 75:
```suggestion
        // Note: ent, as the control variable, needs to be preserved inside the
```

espr14 (changes requested):

dragomirecky (commented):
In general, looks good to me. The comments are minor. 

One non-minor concern I have is regarding consistency of JSON responses. As stated in the documentation for segmented json:
> /// Note that state that needs to survive between resumption either needs to be
 /// stored in the object or re-computed at the start of the content method (but
 /// it gets recomputed on each resume, therefore it should produce the same or at
 /// least compatible results).

This makes sense in respect to the implementation. However, I don't see it being reflected in the implementation of our JSON responses. For example, the `get_printer` API returns JSON with values like `is_printing` and `is_paused`. It seems to me, that if the response gets divided at the right time & in the right place, we might produce invalid responses like `ready = True` and `busy = True`. 

It might be „this never happens“ situation in the current implementation (because of the buffer size), but it looks like something which might eventually create bugs which would be very hard to find. If am not missing something, I would prefer some API design which would make it harder to write such incorrect code.

lib/WUI/nhttp/segmented_json.cpp
```diff
@@ -0,0 +1,110 @@
...
+#include "../json_encode.h"
+
+#include <cassert>
+#include <cstdio>
+#include <cstdarg>
+#include <cinttypes>
+
+namespace nhttp {
+
+JsonRenderer::ContentResult JsonRenderer::Output::output(size_t resume_point, const char *format, ...) {
```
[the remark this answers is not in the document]
vorner: Unfortunately, can't. JsonRenderer is a class, not a namespace.

lib/WUI/nhttp/segmented_json.cpp
```diff
@@ -0,0 +1,110 @@
...
+            assert(written <= buffer_size);
+            buffer += written;
+            buffer_size -= written;
+            written_something = true;
+            this->resume_point = resume_point;
+            return ContentResult::Incomplete;
+        case ContentResult::Abort:
+            return ContentResult::Abort;
+        case ContentResult::BufferTooSmall:
+            if (written_something) {
```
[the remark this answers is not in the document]
vorner: Which version of code is this commented on? I believe the last version no longer contains this.

tests/unit/lib/WUI/nhttp/segmented_json_rendering.cpp
```diff
@@ -0,0 +1,153 @@
...
+class ValueIterator final : public JsonRenderer::Iterator {
+private:
+    static const constexpr int64_t MAX_VALUE = 3;
+    InnerRenderer renderer;
+
+public:
+    ValueIterator()
+        : renderer(true, 0) {}
+    virtual JsonRenderer *get() override {
+        if (renderer.value <= MAX_VALUE) {
```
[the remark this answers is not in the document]
vorner: Again, this is no longer in the code. Iterators got removed.

lib/WUI/nhttp/file_info.cpp
```diff
@@ -20,6 +24,108 @@ namespace nhttp::printer {
...
+            ent = candidate;
+            return;
+        }
+    }
+    // Left to the default nullptr when runs out of entities in the directory
+}
+
+JsonRenderer::ContentResult FileInfo::DirEntryRenderer::content(size_t resume_point, Output &output) {
+    // We don't use the convenience FIELD_STR at many places here, because we
+    // provide the same escaped version multiple times and we we need to
```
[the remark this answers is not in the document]
vorner: Again, no longer present.

lib/WUI/nhttp/file_info.cpp
```diff
@@ -20,6 +24,108 @@ namespace nhttp::printer {
...
+        JSON_ITERATE(*this);
+
+        JSON_ARR_END;
+    JSON_OBJ_END;
+    JSON_END;
+    // clang-format on
+}
+
+JsonRenderer *FileInfo::DirRenderer::get() {
+    if (renderer.ent != nullptr) {
```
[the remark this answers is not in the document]
vorner: Again…

lib/WUI/nhttp/file_info.cpp
vorner: I'm not entirely sure if this is more readable, but :shrug:

lib/WUI/nhttp/file_info.cpp
vorner: Again, not a namespace :-(

lib/WUI/link_content/basic_gets.cpp
vorner: Not a new code in this pull request, it got just moved from the original `wui_REST_api.c`; if you believe this should be solved, let's deal with it in a separate pull request.

Anyway, I think we have already talked about this and the issue is that it issues a warning _even if everything is handled_.

lib/WUI/link_content/basic_gets.cpp
vorner: Again, not a new code. Furthermore, `sizeof char` is by definition `1`. That's how sizeof itself is defined, by number of chars.

lib/WUI/link_content/basic_gets.cpp
vorner: OK with the float, though I believe it's minor (this code's speed is dominated by other factors, including all the sending of packets over network).

The compiler probably replaces the whole division by bitshifts or some such. Doing these things manually can actually throw it off a bit, so it's better to just write the code with the human reader in mind, the compiler will do a better job than us in this low-level stuff.

lib/WUI/nhttp/file_info.cpp
```diff
@@ -20,6 +24,114 @@ namespace nhttp::printer {
 
 using namespace handler;
 
+#if 0
```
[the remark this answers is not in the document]
vorner: Only a leftover, not an intention. When I was refactoring the old code, I kept the original around during iteratively compiling it, to know what's left to port. When it got gutted of all the important things, I forgot to remove this snippet.

lib/WUI/nhttp/segmented_json.cpp
```diff
@@ -0,0 +1,89 @@
...
+JsonRenderer::ContentResult JsonRenderer::Output::output_field_bool(size_t resume_point, const char *name, bool value) {
+    return output(resume_point, "\"%s\":%s", name, jsonify_bool(value));
+}
+
+JsonRenderer::ContentResult JsonRenderer::Output::output_field_int(size_t resume_point, const char *name, int64_t value) {
+    return output(resume_point, "\"%s\":%" PRIi64, name, value);
+}
+
+JsonRenderer::ContentResult JsonRenderer::Output::output_field_float_fixed(size_t resume_point, const char *name, double value) {
+    return output(resume_point, "\"%s\":%.1f", name, value);
```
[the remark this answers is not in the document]
vorner: Fixed number of decimal points. Yes, with the specification it'll be nicer.

vorner:
You may have gotten some comments already (github sends some right away and some stashed :confused:)… pausing for the weekend, will finish later.

lib/WUI/nhttp/file_info.cpp
DRracer: The proposed `?:` is definitely LESS readable than the original `if`. Moreover, the ternary operator was designed to return one of two options. This is not the case - we only want `rend` to be set to something when it comes with a `nullptr`, otherwise don't touch it.

lib/WUI/nhttp/segmented_json.cpp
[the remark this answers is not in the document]
DRracer: Moreover, the best practices of C++ suggest against (ab)use of `using namespace`, especially at the top level of a source file. It gets even trickier if you did this in a header file.

lib/WUI/nhttp/segmented_json.cpp
```diff
@@ -0,0 +1,85 @@
...
+}
+
+JsonRenderer::ContentResult JsonRenderer::Output::output_field_str_format(size_t resume_point, const char *name, const char *format, ...) {
+    va_list params1, params2;
+    va_start(params1, format);
+    va_copy(params2, params1);
+    // First, discover how much space we need for the formatted string.
+    char first_buffer[1];
+    // +1 for \0
+    const size_t needed = vsnprintf(first_buffer, 1, format, params1) + 1;
```
DRracer, line 45: I'm not sure if I understand the intent correctly - `first_buffer` is 1-byte long, so vsnprintf is only used to return the buffer size necessary without actually writing anything into `first_buffer`?

It looks that way since the C standard says:
_vsnprintf returns the number of characters that would have been written if n had been sufficiently large, not counting the terminating null character._

In this case of non-time critical code it is completely ok to call the relatively costly vsnprintf twice.

vorner:
New changes :-).

lib/WUI/nhttp/segmented_json.h
```diff
@@ -0,0 +1,122 @@
...
+/// The user inherits this, adds the state and provides the content function.
+///
+/// This does not validate the produced JSON in any way (and probably could be
+/// used to render something else than JSON, actually).
+///
+/// Few notes for implementers:
+/// * The idea behind the content function is that it tries to call methods on
+///   the provided output until it returns ContentResult::Incomplete. Then the
+///   content function returns and it'll get called in the future with a new buffer
+///   and the resume_point parameter set to whatever was passed to the output that
```
DRracer, line 29: Do I understand correctly, that you render the JSON as many times as many segments are used? I.e. when rendering the second segment, you just skip dumping the first characters until you reach the resume_point offset?

DRracer (commented):
Generally, I like the approach. It looks very elegant and cleverly designed :+1: .
Please address my comments in the code (mostly explanation).

lib/WUI/nhttp/segmented_json.cpp
vorner: Yes.

lib/WUI/nhttp/segmented_json.h
vorner: No. The internal (hidden through the macros) switch is used as kind of a goto and it jumps to the place that didn't fit previously. So that one is retried, but all the old ones are just skipped and not even called.

vorner:
> It might be „this never happens“ situation in the current implementation (because of the buffer size), but it looks like something which might eventually create bugs which would be very hard to find. If am not missing something, I would prefer some API design which would make it harder to write such incorrect code.

Hmm, yes, you're right :-(.

For the API, I'm not sure I know of an idea to make it harder to do :-(.

The answer is consistent at least to the point where it'll always be valid JSON (no missing commas or so), but you're right it isn't entirely correct with regards to the flags. I could rewrite it with saving the marlin vars, or something like that. But that would mean I have to create a "special" HTTP handler, not this generic one, which I've wanted to avoid. And, it seems that the new prusa link API will lead into the direction of not having all these flags (we don't agree on everything, but there's a fully shared hatred for this aspect of the API), so this is probably temporary anyway. What's you're opinion on that?

lib/WUI/link_content/basic_gets.cpp
espr14: I meant `typeof(*hostname)` which is `char` which is `1`. But let's close this.

dragomirecky:
> Hmm, yes, you're right :-(.
> 
> For the API, I'm not sure I know of an idea to make it harder to do :-(.
> 
> The answer is consistent at least to the point where it'll always be valid JSON (no missing commas or so), but you're right it isn't entirely correct with regards to the flags. I could rewrite it with saving the marlin vars, or something like that. But that would mean I have to create a "special" HTTP handler, not this generic one, which I've wanted to avoid. And, it seems that the new prusa link API will lead into the direction of not having all these flags (we don't agree on everything, but there's a fully shared hatred for this aspect of the API), so this is probably temporary anyway. What's you're opinion on that?

I am not worried about the currently-implemented endpoints. I am more worried about the endpoints someone else will implement in future, which will eventually break because that someone else didn't know this.

My thoughts were to design API, which default usage would force you to a) store your state into a separate structure (of your choosing) and b) use this captured state (the structure) to build the response json. 

Of course still leaving you with some „lower level“ API which would work as does the current design, but that API should be harder to find and discouraged.

dragomirecky:
Looks good 👍  See my comment regarding JSON_FIELD_STR. But that is something we can fix later if you want to merge this now

lib/WUI/nhttp/file_info.cpp
```diff
@@ -20,6 +24,73 @@ namespace nhttp::printer {
...
+    JSON_OBJ_END;
+    JSON_END;
+    // clang-format on
+}
+
+JsonResult FileInfo::FileRenderer::renderState(size_t resume_point, JsonOutput &output, FileInfo::FileState &state) const {
+    char *filename = state.owner->filename;
+    JSONIFY_STR(filename);
+    char long_name[FILE_NAME_BUFFER_LEN];
+    get_LFN(long_name, sizeof long_name, filename);
```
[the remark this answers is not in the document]
dragomirecky: My mistake – I didn't notice JSONIFY_STR is used together with JSON_CUSTOM. Not because it would be required for JSON_FIELD_STR.

dragomirecky (approved):

espr14 (approved):

PR #1937, opened 2022-03-22
Fix Windows build.

mkbel:
Change --tlv separator from : to @


dragomirecky:
Fixed in 50335bc85f089e351976bdb5eb1a91d7b03caed1 by properly splitting at most once

PR #1944, opened 2022-03-23
Web in xflash

vorner:
Saves a little bit over 30kB of precious CPU flash.

BFW-2360


dragomirecky (commented):
I haven't done a full review yet, but as mentioned on our call, I think we should not polute `mklittlefs.py` with gzip-specific functions. 

Here I have prepared a commit 2c292469c62a1bde4fb951dcfcdabad0c41fa496 which adds a similar `add_gzip_resource` function to CMake, but in a simpler and more modular way.

Feel free to cherry pick this commit and rebase your work on top of it. I briefly tested it and it should work.

include/lwipopts.h
```diff
@@ -130,7 +130,7 @@ extern "C" {
     #define CHECKSUM_CHECK_ICMP6 1
     /*-----------------------------------------------------------------------------*/
     /* USER CODE BEGIN 1 */
-    #define HTTPD_USE_CUSTOM_FSDATA 1 // uses the web resources from fsdata_custom.c (buddy web pages)
+    #define HTTPD_USE_CUSTOM_FSDATA 0 // uses the web resources from fsdata_custom.c (buddy web pages)
```
DRracer, line 133: the comment is probably not true anymore, is it?

DRracer (approved):

PR #1947, opened 2022-03-24
Fix load homing

espr14:


src/marlin_stubs/pause/pause.cpp
```diff
@@ -625,6 +627,95 @@ bool Pause::wait_or_stop() {
     return false;
 }
 
+// TODO: Implement non-blocking homing, put it in homing.cpp and replace this
```
MistrKernnunos, line 630: I would propose to create wrapper function, which would parse the Gcode and then call function G28_no_parser(arguments) and this function could be called from other parts of the FW, even in other Gcodes. The same way is done M1400 in M70X.cpp file.

src/marlin_stubs/pause/pause.cpp
```diff
@@ -651,18 +742,26 @@ void Pause::park_nozzle_and_notify() {
     //home the X or Y axis if it is not homed and we want to move it
     //homing is after Z move to be clear of all obstacles
     //Should not affect other operations than Load/Unload/Change filament run from home screen without homing. We are homed during print
-    if ((axis_homed & _BV(X_AXIS)) == 0 && current_position.x - park_pos.x != 0) {
+    if (!isnan(park_pos.x) && (axis_homed & axis_known_position & _BV(X_AXIS)) == 0) {
```
MistrKernnunos, line 745: If we change this it will break functionality of M1400 beacuse it sets park_pos to current pos if it does not want to move.

src/common/marlin_server.h
```diff
@@ -50,7 +50,7 @@ extern void marlin_server_do_babystep_Z(float offs);
 extern void marlin_server_move_axis(float pos, float feedrate, size_t axis);
 
 // direct call of 'enqueue_and_echo_command', returns 1 if command enqueued, otherwise 0
```
mkbel, line 52: Documentation not updated regarding to return value.

src/marlin_stubs/pause/pause.cpp
mkbel: I agree.

mkbel:
I don't see anywhere in commit message or PR description of what was the problem and how this is solving that problem.

src/marlin_stubs/pause/pause.cpp
```diff
@@ -709,6 +711,101 @@ bool Pause::wait_or_stop() {
...
+    planner.settings.max_acceleration_mm_per_s2[Y_AXIS] = slow_homing.acceleration.y;
+    #if HAS_CLASSIC_JERK
+    planner.max_jerk = slow_homing.jerk_xy;
+    #endif
+    planner.reset_acceleration_rates();
+#endif
+}
+
+static void home_axis_in_pause(AxisEnum axis) {
+    prepare_for_homing();
```
MistrKernnunos, line 804: I would remove all of this code and rework G28 in its file and create the G28_no_parser.

src/marlin_stubs/pause/pause.cpp
```diff
@@ -739,20 +740,20 @@ void Pause::park_nozzle_and_notify() {
...
-    // TODO: make homeaxis non-blocking to allow quick_stop
-    if ((axis_homed & _BV(X_AXIS)) == 0 && current_position.x - park_pos.x != 0) {
-        homeaxis(X_AXIS);
-    }
-    if ((axis_homed & _BV(Y_AXIS)) == 0 && current_position.y - park_pos.y != 0) {
-        homeaxis(Y_AXIS);
+    LOOP_XY(axis) {
+        // TODO: make homeaxis non-blocking to allow quick_stop
+        if (!isnan(park_pos.pos[axis]) && (axis_homed & axis_known_position & _BV(axis)) == 0) {
+            GcodeSuite::G28_no_parser(false, false, 0, false, axis == X_AXIS, axis == Y_AXIS, false);
```
mkbel, line 746: I think we can not use G28 here, as it resets MBL. And we definitely can not reset MBL while parking nozzle.

PR #1957, opened 2022-03-28
Use bool in enqueueing G code

espr14:


RadekVana (approved):

src/common/marlin_server.cpp
```diff
@@ -380,13 +380,13 @@ extern void marlin_server_move_axis(float pos, float feedrate, size_t axis) {
...
-    return queue.enqueue_one(gcode) ? 1 : 0;
+bool marlin_server_enqueue_gcode(const char *gcode) {
+    return queue.enqueue_one(gcode);
 }
 
-int marlin_server_inject_gcode(const char *gcode) {
+bool marlin_server_inject_gcode(const char *gcode) {
     queue.inject_P(gcode);
-    return 1;
+    return true;
```
mkbel, line 389: Return value is redundant here, as it can not fail.
On the other hand this command can not be legally called from marlin_server_idle() as this might lead to re-entrant call and modifying *gcode pointer earlier than it is fully executed.

PR #1965, opened 2022-03-29
gui: Better radio buttons

RadekVana:
no BFW

### DIFF
--- a/src/common/client_response_texts.cpp
+++ b/src/common/client_response_texts.cpp
@@ -3,45 +3,51 @@
 
 #include "client_response_texts.hpp"
 #include "i18n.h"
-//todo make some automatic checks names vs enum
-//list of all button types
-// order and count must match to enum class Response!
-const std::array<const char *, static_cast<size_t>(Response::_last) + 1> BtnTexts::texts {
-    "",                   // _none
-    N_("ABORT"),          // Abort
-    "ABS",                // ABS filament, do not translate
-    "ASA",                // ASA filament, do not translate
-    N_("BACK"),           // Back
-    N_("CANCEL"),         // Cancel
-    N_("CHANGE"),         // Change
-    N_("CONTINUE"),       // Continue
-    N_("COOLDOWN"),       // Cooldown
-    N_("DISABLE SENSOR"), // Filament_removed
-    "FLEX",               //FLEX filament, do not translate
-    "HIPS",               //HIPS filament, do not translate
-    N_("IGNORE"),         // Ignore
-    N_("LOAD"),           // Load
-    N_("NEXT"),           // Next
-    N_("NO"),             // No
-    N_("OK"),             // Ok
-    "PC",                 // PC filament, do not translate
-    "PETG",               // PETG filament, do not translate
-    "PLA",                // PLA filament, do not translate
-    "PP",                 // PP filament, do not translate
-    N_("PURGE MORE"),     // Purge_more
-    N_("REHEAT"),         // Reheat
-    N_("RETRY"),          // Retry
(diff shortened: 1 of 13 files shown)


src/common/client_response_texts.cpp
```diff
@@ -3,45 +3,51 @@
...
+
+// clang-format off
+/**
+ * @brief list of all button types
+ * order and count must match to enum class Response in general_response.hpp!
+ * TODO make some automatic checks names vs enum
+ */
+const std::array<std::pair<const char *, uint16_t>, static_cast<size_t>(Response::_last) + 1> BtnResponse::texts_and_icons {
+    std::make_pair( "", IDR_PNG_question_48px ),                  // _none
+    std::make_pair( N_("ABORT"), IDR_PNG_disconnect_48px ),       // Abort
```
DRracer, line 16: I'd like to see some verification, that the construction of this whole array gets done at compile time.
`std::make_pair` has a constexpr variant in C++14, so theoretically the compiler should be able to do it.

src/common/filament.cpp
```diff
@@ -31,15 +30,15 @@ static constexpr const char *flex_str =  "FLEX     240/ 50";
 //not a member - static_assert cannot access private members
 //fixme generating long names, takes too long
```
DRracer, line 31: Does this comment still apply?

src/gui/dialogs/DialogRadioButton.cpp
```diff
@@ -138,28 +201,139 @@ void RadioButton::draw_n_btns(const size_t btn_count) {
...
+        display::DrawRect(rc_btn, parent_color);
+        rc_btn += Rect16::Left_t(GuiDefaults::RadioButtonCornerRadius);
+        rc_btn -= Rect16::Width_t(2 * GuiDefaults::RadioButtonCornerRadius);
+    }
     render_text_align(rc_btn, text, pf, button_cl, text_cl, { 0, 0, 0, 0 }, Align_t::Center());
 }
 
 bool RadioButton::IsEnabled() const {
-    return responses ? (*responses)[0] != Response::_none : false; //faster than cnt_responses(responses)!=0
+    return responseFromIndex(0) != Response::_none; //faster than cnt_responses(responses)!=0
```
DRracer, line 224: Does this comment still apply? Moreover, if `cnt_responses` was constexpr (which IMO could be), the whole `IsEnabled` may boil down to something like 3 instructions.

src/common/filament.cpp
RadekVana: good eye, those comments are no longer valid

src/gui/dialogs/DialogRadioButton.cpp
RadekVana: comment is still valid
`cnt_responses `use `std::find` and it is not `constexpr` in C++17
also responses are assigned at runtime, so it would not work anyway

src/common/client_response_texts.cpp
RadekVana: I could move it to header to and make it constexpr, but it would consume more codesize (tested)
I have disassembled the code it seems fine to me as it is

src/gui/dialogs/DialogRadioButton.cpp
DRracer: I have a constexpr find_if somewhere

DRracer: ok

src/common/client_response_texts.cpp
DRracer: ok

PR #1985, opened 2022-04-01
More warning cleanups

wavexx:
Remove a few more warnings spotted with gcc 10, showing an actual implementation bug in _server_set_var

### DIFF
--- a/src/common/marlin_vars.c
+++ b/src/common/marlin_vars.c
@@ -62,7 +62,7 @@ marlin_var_id_t marlin_vars_get_id_by_name(const char *var_name) {
     for (int i = 0; i <= MARLIN_VAR_MAX; i++)
         if (strcmp(var_name, __var_name[i]) == 0)
             return i;
-    return -1;
+    return MARLIN_VAR_MAX + 1;
 }
 
 variant8_t marlin_vars_get_var(marlin_vars_t *vars, marlin_var_id_t var_id) {
--- a/src/common/marlin_server.cpp
+++ b/src/common/marlin_server.cpp
@@ -1395,7 +1395,7 @@ static int _server_set_var(const char *const name_val_str) {
     bool changed = false;
     char *val_str = strchr(name_val_str, ' ');
     *(val_str++) = 0;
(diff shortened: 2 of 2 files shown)


src/common/marlin_vars.c
```diff
@@ -62,7 +62,7 @@ marlin_var_id_t marlin_vars_get_id_by_name(const char *var_name) {
     for (int i = 0; i <= MARLIN_VAR_MAX; i++)
         if (strcmp(var_name, __var_name[i]) == 0)
             return i;
-    return -1;
+    return MARLIN_VAR_MAX + 1;
```
espr14, line 65: Isn't it better to define `typedef enum : int {...} marlin_var_id_t;` instead? Or use std::optional?

src/guiapi/src/ScreenShot.cpp
```diff
@@ -90,7 +90,7 @@ bool TakeAScreenshot() {
     success = write(fd, &bmp_header, header_size) == header_size;
 
     if (success) {
-        for (uint8_t block = ST7789V_ROWS / ST7789V_BUFF_ROWS - 1; block >= 0 && block < ST7789V_ROWS / ST7789V_BUFF_ROWS; block--) {
+        for (uint8_t block = ST7789V_ROWS / ST7789V_BUFF_ROWS - 1; block < ST7789V_ROWS / ST7789V_BUFF_ROWS; block--) {
```
espr14, line 93: Let's redo it completely.

espr14 (changes requested):

src/guiapi/src/ScreenShot.cpp
wavexx: Agreed. Dropped the commit

src/common/marlin_vars.c
wavexx: The enum is used as a mask in several other places. In that case I'm a bit against using a signed value.

I agree it's not nice (especially if you look around, we have many + 1 places around this constant), but right now this is just fixing the behavior, no attempt at improving it.

espr14 (approved):

PR #1986, opened 2022-04-01
build: Fix FileNotFound for missing intermediate directories

dragomirecky:


DRracer (approved):

DRracer:
@dragomirecky this commit solves the issues with shadow builds (QtCreator) -> it is safe to merge this PR. 
However, it looks like we have a similar problem with gzipped web resources, which probably needs a similar change.

PR #1988, opened 2022-04-01
Clean snapshot

espr14:


src/guiapi/src/ScreenShot.cpp
```diff
@@ -2,13 +2,18 @@
...
 #include "ScreenShot.hpp"
-#include "st7789v.hpp"
 #include "display.h"
 #include <inttypes.h>
 #include <dirent.h>
 #include <stdio.h>
 #include <unistd.h>
 
+#ifdef USE_ST7789
+static const uint8_t bytes_per_pixel = 2;
```
wavexx, line 12: With this on lines 21-22 we can remove ST7789V_BYTES_PER_PIXEL.
Similarly on 48-52 we can replace it with bytes_per_pixel which makes it much cleaner.

wavexx: Also in the block of lines 65-90

wavexx (changes requested):

src/guiapi/src/ScreenShot.cpp
```diff
@@ -53,62 +56,57 @@ static const unsigned char bmp_header[] = {
...
-
-    // TODO: BMP headers have to know that we are using 2B pixels. Now it is only in [bits per pixel] and clearly it's not enough.
-
-    for (int row = 0; row < ST7789V_BUFF_ROWS / 2; row++) {
-        for (int col = 0; col < ST7789V_COLS * ST7789V_BYTES_PER_PIXEL; col += ST7789V_BYTES_PER_PIXEL) {
-            for (int chan = 0; chan < ST7789V_BYTES_PER_PIXEL; chan++) {
-                std::swap(buffer[buffer[row * ST7789V_COLS * ST7789V_BYTES_PER_PIXEL + col + chan]], buffer[(ST7789V_BUFF_ROWS - row - 1) * ST7789V_COLS * ST7789V_BYTES_PER_PIXEL + col + chan]);
+    // Y-axis mirror image - because BMP pixel format has base origin in left-bottom corner not in left-top like on displays
+    for (int row = 0; row < buffer_rows / 2; row++) {
+        for (int col = 0; col < display::GetW() * bytes_per_pixel; col += bytes_per_pixel) {
```
wavexx, line 61: I've re-read this three times to understand why it was written like this. Why not do:

```col < display::GetW() * bytes_per_pixel; col++```

and remove the second inner loop? We don't need the weird "+ chan"...

File: src/guiapi/src/ScreenShot.cpp [hunk 3]
```diff
@@ -53,62 +56,57 @@ static const unsigned char bmp_header[] = {
...
+            for (int chan = 0; chan < bytes_per_pixel; chan++) {
+#ifdef USE_ST7789
```
wavexx, line 63: I think this should always work since it's byte-wise, irregardless of display. I'd remove the ifdef

wavexx (changes requested):

src/guiapi/src/ScreenShot.cpp
wavexx: Looking good now

src/guiapi/src/ScreenShot.cpp
wavexx: Fixed

wavexx (approved):

PR #1990, opened 2022-04-02
ESP: Handle reinit after flash

vorner:
* Reset the expected intron after ESP reset (it forgets the set one).
* Inform LwIP that the interface is down and that it needs to be brought
  up properly afterwards.

### DIFF
--- a/lib/WUI/espif.cpp
+++ b/lib/WUI/espif.cpp
@@ -91,6 +91,7 @@ static std::atomic<uint16_t> fw_version;
 static std::atomic<ESPIFOperatingMode> esp_operating_mode = ESPIF_UNINITIALIZED_MODE;
 static std::atomic<bool> associated = false;
 static std::atomic<TaskHandle_t> init_task_handle;
+static std::atomic<netif *> active_esp_netif;
 
 // UART
 static const uint32_t NIC_UART_BAUDRATE = 1000000;
@@ -205,6 +206,7 @@ bool espif_link() {
 }
 
 static void process_link_change(bool link_up, struct netif *netif) {
+    assert(netif != nullptr);
     log_info(ESPIF, "Link status changed: %d", link_up);
     associated = link_up;
     if (link_up) {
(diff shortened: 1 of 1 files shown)


lib/WUI/espif.cpp
```diff
@@ -488,6 +505,10 @@ err_t espif_init(struct netif *netif) {
         return ERR_ALREADY;
     }
 
+    struct netif *previous = active_esp_netif.exchange(netif);
+    assert(previous == nullptr);
+    (void)previous; // Avoid warnings in release
```
espr14, line 510:
```suggestion
    assert(nullptr == active_esp_netif.exchange(netif));
```

lib/WUI/espif.cpp
```diff
@@ -443,7 +444,23 @@ static err_t low_level_output(struct netif *netif, struct pbuf *p) {
     return ERR_OK;
 }
 
+static void force_down() {
+    struct netif *iface = active_esp_netif; // Atomic load
+    assert(iface != nullptr);               // Already initialized
+    process_link_change(false, iface);
```
espr14, line 450: Assert should be (also) in `netif_set_link_up` and `netif_set_link_down`. Then there's no need for `iface` here.

espr14 (changes requested):

lib/WUI/espif.cpp
vorner: Certainly not. That `.exchange` is an important side effect and therefore can't be inside the call to assert. It would be removed in release build.

lib/WUI/espif.cpp
vorner: OK for adding it to the `process_link_change`, I've added it there. But:

* `netif_set_link_*` are 3rd-party ones from LwIP, they are not going to be changed.
* I want it here not only to catch bugs (that's nice), but as a documentation. I'm claiming this function is called only after everything has been initialized.

espr14 (approved):

PR #1995, opened 2022-04-04
Fix M70X parsing

RadekVana:


vorner (approved):
OK, but maybe putting parentheses there might improve readability. Not everyone carries the operator precedence table in their head 😇

PR #2000, opened 2022-04-06
Fix axis test failing after X fails

MistrKernnunos:


src/common/selftest/selftest_MINI.cpp
```diff
@@ -305,6 +309,11 @@ bool CSelftest::phaseAxis(const selftest_axis_config_t &config_axis, CSelftestPa
     switch (config_axis.axis) {
     case X_AXIS:
         eeres.xaxis = (*ppaxis)->GetResult();
+        // if X axis test failed, wee need to skip Z axis test
```
espr14, line 312:
```suggestion
        // if X axis test failed, we need to skip Z axis test
```

src/common/selftest/selftest_MINI.cpp
```diff
@@ -305,6 +309,11 @@ bool CSelftest::phaseAxis(const selftest_axis_config_t &config_axis, CSelftestPa
...
+        if (eeres.xaxis == SelftestResult_Failed) {
+            m_Mask = (SelftestMask_t)(m_Mask & ~SelftestMask_t::stmZAxis);
```
espr14, line 314: `clear bit` macro/function would be better but I haven't found proper one :-(

src/common/selftest/selftest_MINI.cpp
```diff
@@ -254,13 +248,23 @@ bool CSelftest::phaseHome() {
...
+    case SelftestResult_Unknown:
+        return SelftestSubtestState_t::undef;
+    case SelftestResult_Passed:
+        return SelftestSubtestState_t::ok;
+    case SelftestResult_Failed:
+        return SelftestSubtestState_t::not_good;
+    case SelftestResult_Skipped:
+        return SelftestSubtestState_t::undef;
+    }
+    return SelftestSubtestState_t::undef;
```
espr14, line 261:
```suggestion
    case SelftestResult_Passed:
        return SelftestSubtestState_t::ok;
    case SelftestResult_Failed:
        return SelftestSubtestState_t::not_good;
    case SelftestResult_Skipped:
    case SelftestResult_Unknown:
    }
    return SelftestSubtestState_t::undef;
```

espr14 (changes requested):

espr14 (approved):

PR #2005, opened 2022-04-07
Reset wifi if not online for too long

vorner:


lib/WUI/espif.cpp
```diff
@@ -194,11 +213,24 @@ void espif_input_once(struct netif *netif) {
     }
 }
 
+static void generate_intron();
```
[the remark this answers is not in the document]
vorner: I'm not sure. It comes back from Vláďa and Davic, I suspect it is some relevant jargon… something like „Introduction sequence“ stuff. Basically, every message sent on the wire is prefixed with this to mark „Hey, a message starts here“.

And as we wanted to make it harder for some remote attacker to put confusing things into packets, we generate a random one at a start and negotiate it with the other side.

lib/WUI/wui.cpp
```diff
@@ -384,6 +400,26 @@ class NetworkState {
...
+
+                // It's OK if the ESP is turned off on purpose or if it's up and running.
+                const bool esp_ok = (iface_mode(ifaces[NETDEV_ESP_ID]) == Mode::Off || ap.ssid[0] == '\0' || espif_link());
+
+                const uint32_t now = sys_now();
+                if (esp_ok) {
+                    last_esp_ok = now;
+                }
+
+                const uint32_t faulty_for = now - last_esp_ok;
```
DRracer, line 414: How about the 32bit boundary overflow? Does this computation work correctly in such a case?

vorner: I _think_ it does, but I always find these considerations mind-boggling, so if you recheck after me, it would be nice.

lib/WUI/espif.cpp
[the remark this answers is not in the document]
DRracer: ok

lib/WUI/wui.cpp
DRracer: Me too :wink: , but it seems to be ok (considering only 8bits):
`0x01 - 0xfe = 3`

DRracer (approved):

PR #2007, opened 2022-04-08
Fix gzipped resources triggering resources reflash with every build

dragomirecky:


vorner (approved):
Nice catch.

Let's put it in for the fix. But maybe this is getting long enough to become its own python script somewhere in `utils`?

PR #2009, opened 2022-04-11
Homing can fail

mkbel:
Detect HOMING FAILED and kill the printer.

It adds general detection of homing failure.
As two bumps are enabled only for Z axis, only homing Z axis can fail at the moment.
2 homing attempts are set.
Homing of Z axis can fail, if PINDA is disconnected or broken or its cable is broken.

Algorithm compares position of two samples. Homing is OK if two attempts are close enough. Limits are set to +-0.1 mm for Z axis.

Known issues:
Error text and landing page is missing.

### DIFF
--- a/include/marlin/Configuration_MINI_adv.h
+++ b/include/marlin/Configuration_MINI_adv.h
@@ -474,6 +474,16 @@
 
 // @section homing
 
+//kill command after probing fails
+//#define HALT_ON_PROBING_ERROR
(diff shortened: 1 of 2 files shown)


RadekVana (approved):

mkbel:
title: "HOMING ERROR"
  text: "MINDA/SPINDA is probably broken, could not home Z properly."

PR #2030, opened 2022-04-16
WUI: Proper locking during network setup

vorner:
* Lock a mutex whenever we access shared data.
* This doesn't happen on hot-path / packet handling, only during
  "management" (setup, reading info for GUI, ...)
* It is probably a bit more conservative (locking more than necessary).
  It's easier to overlock than prove that it's not needed and as it's
  not on the hot path, better safe than sorry.

### DIFF
--- a/lib/WUI/wui.cpp
+++ b/lib/WUI/wui.cpp
@@ -21,6 +21,8 @@
 #include <lwip/netifapi.h>
 #include <lwip/netif.h>
 #include <lwip/tcpip.h>
+#include <freertos_mutex.hpp>
+#include <mutex>
 #include "http_lifetime.h"
 #include "main.h"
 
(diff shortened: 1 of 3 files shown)


lib/WUI/wui.cpp
```diff
@@ -161,6 +163,7 @@ class NetworkState {
 
     void link_callback(netif &iface) {
         status_callback(iface);
+        xSemaphoreTake(mutex, portMAX_DELAY);
```
DRracer, line 166: please consider a RAII pattern if possible

DRracer (changes requested):
Please consider a RAII pattern approach if possible

vorner:
I'm pushing a version based on the std::lock APIs and the rtos mutex class I've already found lying around. I still need to test this version once I get a calm while, but that's not a reason to keep it to myself.

DRracer (approved):

PR #2033, opened 2022-04-19
ESP update error messages correction

zoltan-l:
ESP Flash Error messages do no correspond to the new flashing from xFlash instead of Flash disk

### DIFF
--- a/src/gui/screen_menu_esp_update.cpp
+++ b/src/gui/screen_menu_esp_update.cpp
@@ -191,7 +191,7 @@ void ScreenMenuESPUpdate::windowEvent(EventLock /*has private ctor*/, window_t *
             esp_loader_connect_args_t config = ESP_LOADER_CONNECT_DEFAULT();
             if (ESP_LOADER_SUCCESS == esp_loader_connect(&config)) {
                 log_info(Network, "ESP boot connect OK");
-                help.SetText(_("Successfully connected to ESP. Do not switch the Printer off nor remove the Flash disk."));
+                help.SetText(_("Successfully connected to ESP. Do not switch the Printer off until the update ends."));
                 progress_state = esp_upload_action::Start_flash;
             } else {
                 log_debug(Network, "ESP boot failedK");
@@ -203,7 +203,7 @@ void ScreenMenuESPUpdate::windowEvent(EventLock /*has private ctor*/, window_t *
         case esp_upload_action::Start_flash:
             if ((opened_file = fopen(current_file->filename, "rb")) == nullptr) {
                 log_error(Network, "ESP flash: Unable to open file %s", current_file->filename);
-                help.SetText(_("Unable to open files. Chek the flash drive and start again"));
+                help.SetText(_("Unable to open files. Try to re-flash the last FW and try again"));
                 progress_state = esp_upload_action::ESP_error;
                 break;
             } else {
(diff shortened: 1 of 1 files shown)


src/gui/screen_menu_esp_update.cpp
```diff
@@ -203,7 +203,7 @@ void ScreenMenuESPUpdate::windowEvent(EventLock /*has private ctor*/, window_t *
         case esp_upload_action::Start_flash:
             if ((opened_file = fopen(current_file->filename, "rb")) == nullptr) {
                 log_error(Network, "ESP flash: Unable to open file %s", current_file->filename);
-                help.SetText(_("Unable to open files. Chek the flash drive and start again"));
+                help.SetText(_("Unable to open files. Try to re-flash the last FW and try again"));
```
vorner, line 206: Reflashing the firmware won't help. I don't think there's currently any user-accessible way to try fix that if that problem persists.

However, trying to flash the ESP again sometimes helps.

src/gui/screen_menu_esp_update.cpp
```diff
@@ -191,7 +191,7 @@ void ScreenMenuESPUpdate::windowEvent(EventLock /*has private ctor*/, window_t *
             esp_loader_connect_args_t config = ESP_LOADER_CONNECT_DEFAULT();
             if (ESP_LOADER_SUCCESS == esp_loader_connect(&config)) {
                 log_info(Network, "ESP boot connect OK");
-                help.SetText(_("Successfully connected to ESP. Do not switch the Printer off nor remove the Flash disk."));
+                help.SetText(_("Successfully connected to ESP. Do not switch the Printer off until the update ends."));
```
vorner, line 194: `Ends` is possible, `finishes` is probably slightly better.

src/gui/screen_menu_esp_update.cpp
```diff
@@ -235,7 +235,7 @@ void ScreenMenuESPUpdate::windowEvent(EventLock /*has private ctor*/, window_t *
             log_debug(Network, "ESP read data %ld", readCount);
             if (ferror(opened_file)) {
                 log_error(Network, "ESP flash: Unable to read file %s", current_file->filename);
-                help.SetText(_("Unable to read files. Chek the flash disk and start again"));
+                help.SetText(_("Unable to read files. Try re-flash the last FW and try again"));
```
vorner, line 238: The same....

vorner (changes requested):

src/gui/screen_menu_esp_update.cpp
vorner: Well, I have seen it to fail somewhat randomly. It's probably some race condition or noise on some internal bus. Or it may be related to me powering the knoblet off a USB. Retrying did help.

Either way, if it fails, it is a bug ‒ either because it allowed it to load wrong file to xflash, or because the above problem. Retrying is a workaround that might help ‒ but it's the only one we can reasonably give. Because reflashing the firmware will _not trigger_ reloading of the files to xflash.

src/gui/screen_menu_esp_update.cpp
zoltan-l: According to conversation the text has been corrected.

vorner (approved):
OK, but can you make it only one commit, please?

PR #2042, opened 2022-04-20
Fix dump

mkbel:
Solved issues:

- During dump RTOS is intentionally blocked so DMA transfers can not work. So dump should never work because of this. Strangely it worked sometimes. Solved by not using DMA transfers at all when dumping. In previous PR https://github.com/prusa3d/Prusa-Firmware-Buddy/pull/2036/commits/e25132db05890d28c3a691438fb5752f0fdda2c8
- When read from xFlash was ongoing it can not react on first request. Solved by deselecting xFlash in init with proper delay https://github.com/prusa3d/Prusa-Firmware-Buddy/pull/2042/commits/a43e9ea67a57320af4eabd87cca18b9d679219c2 and calling init before any attempt to communicate with it https://github.com/prusa3d/Prusa-Firmware-Buddy/pull/2042/commits/bef34e5dde1e45aae64e50538b2f2f855c0d59b4
- Very old error in logic of dump which was first created there https://github.com/prusa3d/Prusa-Firmware-Buddy/pull/995 https://github.com/prusa3d/Prusa-Firmware-Buddy/commit/d93543f2bd42d77b8717a6c64441099264d0383f and its spirit was preserved after many additions to logic. Fixed by https://github.com/prusa3d/Prusa-Firmware-Buddy/pull/2042/commits/bef34e5dde1e45aae64e50538b2f2f855c0d59b4
- When the SPI (DMA) transfer is ongoing, it needs to be aborted before any new communication is started. https://github.com/prusa3d/Prusa-Firmware-Buddy/pull/2042/commits/8ce241f70a36b91c18b4f8c6471e6df27ed5fa75
- Dynamic allocation of RAM was unnecessary for dump to work and so is avoided in https://github.com/prusa3d/Prusa-Firmware-Buddy/pull/2042/commits/e6df9cc27cd506c26082de83d7a81066188686db
- If there is ongoing write or erase operation (re)initialization of xFlash needs to wait until it is finished.

Not solved yet:
- Missing synchronization with xFlash filesystem access and reading dump from xFlash.


src/common/w25x_communication.cpp
```diff
@@ -5,6 +5,8 @@
 #include "cmsis_os.h"
 #include "stm32f4xx_hal.h"
 #include "hwio_pindef.h"
+#define F_CPU SystemCoreClock
+#include "../../lib/Marlin/Marlin/src/HAL/shared/Delay.h"
```
mkbel, line 10: [DRracer](https://github.com/DRracer) [yesterday](https://github.com/prusa3d/Prusa-Firmware-Buddy/pull/2036#discussion_r854176409)

I don't like having dependency on Marlin in dump.c, that's completely unrelated. Don't we have a DELAY_NS in some other form? AFAIK it all boils down to tick_ns() or a similar call, which is our code.

src/common/w25x.h
```diff
@@ -21,8 +21,12 @@ extern "C" {
...
 /// and assigned using w25x_spi_assign.
 ///
-/// Returns true on success, false otherwise.
-extern bool w25x_init();
+/// @param init_DMA Dynamically allocate resources for DMA transfers
+///        DMA and RTOS facilities are not used if false,
+///        it has no effect if it was already allocated by previous call
+/// @retval true on success
+/// @retval false otherwise.
+extern bool w25x_init(bool init_DMA);
```
dragomirecky, line 29: /nitpick/

should be `init_dma`, as it uses snake_case...

src/common/dump.c
```diff
@@ -20,24 +20,23 @@ static inline void dump_regs_SCB(void) {
 }
 
 void dump_to_xflash(void) {
-    _Static_assert(sizeof(dumpinfo_t) == 16, "invalid sizeof(dumpinfo_t)");
-    uint32_t addr;
-    if (!dump_in_xflash_is_empty()) {
-        if (dump_in_xflash_is_saved()) {
-            if (!dump_in_xflash_is_displayed()) {
-                return;
+    if (w25x_init(false)) {
```
dragomirecky, line 23: Can't we just early-exit,  instead of adding one more indentation level to the whole function?

src/common/w25x_communication.cpp
dragomirecky: I agree with @DRracer here. We definitely don't want to have this code dependent on Marlin here.

We should just add `delay_ns` (and other) functions to our `timing.h`.

src/common/w25x_communication.cpp
```diff
@@ -49,16 +52,20 @@ enum {
...
-    current_error = 0;
-
+extern "C" bool w25x_communication_init(bool init_event_group) {
     // check SPI handle has been assigned using w25x_spi_assign
     if (spi_handle == NULL)
         return false;
 
+    HAL_SPI_Abort(spi_handle);
+    w25x_cs_high();
+    DELAY_NS(50);
```
dragomirecky, line 62: Please add a documentation what are we waiting for here and why 50 ns

src/common/w25x.h
dragomirecky: I am not a fan of this API change.

Lets first focus on whether it is actually needed. What is the motivation here.

Is it because `xEventGroupCreate` might not be safe to be called without the FreeRTOS running? We can detect such case using `xTaskGetSchedulerState() != taskSCHEDULER_RUNNING`.

Is it because `xEventGroupCreate` might not be safe to be called from within in ISR (like the hardfault)? We can detect that using `xPortIsInsideInterrupt()`.

Or is it because of something else? If just the two cases above are the reason for this parameter, I would get rid of the parameter and intiialize the DMA based on those checks.

mkbel: I want avoid dynamic RAM allocation in case w25x was not initialized before hardfault happened. w25x_init() is used in dump to xFlash and that is used in hard fault. DMA transfer is not used in hard fault, so I consider redundant to initialize it.

Idea behind this is that it might improve crash dump analysis if there would be no dynamic allocation done by dump itself.

src/common/dump.c
mkbel: Sure we can.

src/common/w25x_communication.cpp
mkbel: fixed

src/common/w25x_communication.cpp
mkbel: documented

src/common/dump.c
mkbel: fixed

src/common/w25x.h
mkbel: init_DMA renamed

src/common/dump.c
```diff
@@ -122,18 +123,14 @@ unsigned int dump_in_xflash_read_regs_GEN(void *pRegsGEN, unsigned int size) {
 }
 
 void dump_in_xflash_reset(void) {
-    if (w25x_init()) {
-        for (uint32_t addr = 0; addr < DUMP_XFLASH_SIZE; addr += 0x10000) {
-            w25x_block64_erase(DUMP_OFFSET + addr);
-        }
+    for (uint32_t addr = 0; addr < DUMP_XFLASH_SIZE; addr += 0x10000) {
```
DRracer, line 126: Is it safe to remove the `w25x_init()` from multiple `dump_*` functions? If they are "private" and not part of the dump API, it is ok. However, if they can be called separately in arbitrary order, relying on having the w25x initialized may be dangerous.

src/hw/mainCheck.cpp
```diff
@@ -0,0 +1,34 @@
...
+    rcc_OscInit.PLL = {
+        .PLLState = RCC_PLL_ON,
+        .PLLSource = RCC_PLLSOURCE_HSE,
+        .PLLM = 6,
+        .PLLN = 168,
+        .PLLP = RCC_PLLP_DIV2,
+        .PLLQ = 7
+    };
+    return rcc_OscInit;
+}();
```
DRracer, line 23: nice :+1:

src/common/dump.c
mkbel: w25x_init() is called in the beginning of main() just after initializing integrated peripherals. I think it is OK to rely on the fact those dump functions will not be called before the basic system is initialized.
Reasoning was already explained in commit messages:

[Remove redundant.](https://github.com/prusa3d/Prusa-Firmware-Buddy/pull/2042/commits/6cc1538f797efaec7fa6a84274bcf13f27981074)
It is asured in main(), that w25x is initilized succesfully
otherwise BSOD is called and we don't reach this point.

[Remove redundant.](https://github.com/prusa3d/Prusa-Firmware-Buddy/pull/2042/commits/4929461b8c85510837fb167eb55660a255ee450e)
There is no need to assure w25x_init() as
dump_save_to_usb() is called from GUI
and it is assured w25x to be successfully initialized
before GUI is started.


[Remove redundant.](https://github.com/prusa3d/Prusa-Firmware-Buddy/pull/2042/commits/d8c6170baa89deec07a75b73a42fbea1b8049e9f)
There is no need to assure w25x_init() as
dump_in_xflash_reset() is called from main only after w25x
was successfully initialized.
It is also called from GUI and it is assured w25x being
initialized before GUI is started.

[Remove redundant.](https://github.com/prusa3d/Prusa-Firmware-Buddy/pull/2042/commits/424464098347d99f5f4273dd40138aebdb3b5e40)
There is no need to assure w25x_init() as
dump_in_xflash_delete() is called from GUI
and it is assured w25x to be successfully initialized
before GUI is started.

src/common/dump.c
```diff
@@ -143,42 +140,40 @@ int dump_save_to_usb(const char *fn) {
...
-                bw_total += bw;
+    fd = fopen(fn, "w");
+    if (fd != NULL) {
+        //save dumped RAM and CCRAM from xflash
+        for (addr = 0; addr < DUMP_XFLASH_SIZE; addr += DUMP_BUFF_SIZE) {
+            memset(buff, 0, DUMP_BUFF_SIZE);
+            w25x_rd_data(addr, buff, DUMP_BUFF_SIZE);
+            bw = fwrite(buff, 1, DUMP_BUFF_SIZE, fd);
+            if (bw <= 0) {
+                break;
```
dragomirecky, line 151: Open file leak

src/common/dump.c
```diff
@@ -143,42 +140,40 @@ int dump_save_to_usb(const char *fn) {
...
-                    break;
-                }
-                bw_total += bw;
+            bw_total += bw;
+        }
+        //save OTP
+        for (addr = 0; addr < DUMP_OTP_SIZE; addr += DUMP_BUFF_SIZE) {
+            bw = fwrite((void *)(DUMP_OTP_ADDR + addr), 1, DUMP_BUFF_SIZE, fd);
+            if (bw <= 0) {
+                break;
```
dragomirecky, line 159: The same problem, open file leak

src/common/timing_precise.h
```diff
@@ -0,0 +1,182 @@
...
+ * Naming convention was changed to be more in line with Buddy
+ * coding standard.
+ * Macro DELAY_MS was changed to force inline delay_us_precise function.
+ * Support for non-constexpr parameter of DELAY_NS_PRECISE was
+ * removed as I considered it broken.
+ *
+ * @file
+ */
+
+#ifdef __cplusplus
```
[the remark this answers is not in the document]
mkbel: `#pragma once` fixed during last rebase

src/common/dump.c
mkbel: I don't see it. Can you elaborate more? note: this is `break` statement and not `return` statement.

src/common/timing_precise.h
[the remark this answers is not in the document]
mkbel: `#ifdef __cplusplus` removed and file renamed to .hpp

src/common/w25x.c
```diff
@@ -47,16 +47,18 @@ typedef enum {
...
 
-bool w25x_init() {
-    w25x_cs_high();
+bool w25x_init(bool init_dma) {
 
-    if (!w25x_communication_init())
+    if (!w25x_communication_init(init_dma))
         return false;
 
+    w25x_wait_busy();
```
mkbel, line 60: note: w25x_wait_busy() don't release CPU, ticket has been created https://dev.prusa3d.com/browse/BFW-2556

CMakeLists.txt
```diff
@@ -748,6 +748,7 @@ target_sources(
           src/marlin_stubs/M300.cpp
           src/marlin_stubs/M330.cpp
           src/marlin_stubs/M505.cpp
+          src/hw/mainCheck.cpp
```
dragomirecky, line 751: Why is it camelCase when all our files are named in snake_case? Why is it in between of MXXX gcode files? 🙈

src/common/dump.c
dragomirecky: my bad, I don't see it either! :)

CMakeLists.txt
mkbel: It is camelCase because all compile time checks are already named somethingCheck.
It is not true that all our files are named in snake_case, see folder src/gui.
As we use PascalCase for classes I consider it strange that file would be named differently than class it contains.

> Why is it in between of MXXX gcode files?

Rebasing put it there.

Renamed to main_check.cpp and moved out of MXXX gcode files.

dragomirecky (approved):
👍

PR #2043, opened 2022-04-21
BFW 2528

espr14:


src/marlin_stubs/include/M70X.hpp
```diff
@@ -46,6 +46,10 @@ void mmu_off();
 
 std::pair<std::optional<PreheatStatus::Result>, filament_t> preheat(PreheatData type);
 std::pair<std::optional<PreheatStatus::Result>, filament_t> preheat_for_change_load();
+
+extern int16_t preheat_nozzle_temp;
```
RadekVana, line 50: please don't use global variables

RadekVana (changes requested):
Add parameter to M1700 (S as store or P as persistent store)
And use it to store filament type
And within first layer calibration you can read temperatures from that

RadekVana (approved):

PR #2048, opened 2022-04-22
nozzle: Provide nozzle cooling by a print fan

espr14:
If nozzle is set to lower temperature (e.g. before G29) fan can be activated to speed up cooling.
BFW-1039


Eldenroot:
How this is handled during printing?

espr14:
1. You should not use this G code parameter during extrusion.
2. Once the temperature is reached, the original fan speed is restored. If the fan speed is changed during the cooling then the new speed is discarded (original is restored).

Eldenroot:
Sounds cool, could save some time! I hope it will get into next version.

PR #2053, opened 2022-04-25
Validate bootstrapped resources against its expected hash

dragomirecky:
- Calculates recursive hash of all resources and stores it into the firmware
- Within the bootstrap procedure, after all files are copied, it calculates the recursive hash in the same way in order to make sure the copied resources match the expected state from compile time

### DIFF
--- a/src/resources/bootstrap.cpp
+++ b/src/resources/bootstrap.cpp
@@ -9,81 +9,16 @@
 #include "log.h"
 #include "timing.h"
 #include "cmsis_os.h"
-#include "filesystem_littlefs_bbf.h"
 
 #include "resources/bootstrap.hpp"
 #include "resources/required_revision.hpp"
+#include "resources/hash.hpp"
+
+#include "fileutils.hpp"
 
 LOG_COMPONENT_DEF(Resources, LOG_SEVERITY_DEBUG);
-#define log(severity, ...) _log_event(severity, log_component_find("Resources"), __VA_ARGS__)
 
 using namespace buddy::resources::revision;
-
-class DIRDeleter {
-public:
-    void operator()(DIR *dir) {
-        closedir(dir);
-    }
-};
-
-class FILEDeleter {
-public:
-    void operator()(FILE *file) {
-        fclose(file);
-    }
-};
-
-class ScopedFileSystemLittlefsBBF {
-private:
-    int device;
-
-public:
-    ScopedFileSystemLittlefsBBF(FILE *bbf)
-        : device(filesystem_littlefs_bbf_init(bbf)) {}
-
-    bool mounted() {
-        return device != -1;
-    }
-
-    ~ScopedFileSystemLittlefsBBF() {
-        if (mounted()) {
-            filesystem_littlefs_bbf_deinit();
-        }
-    }
-};
-
-class Path {
-private:
-    char path[104]; // TODO: Provide proper defines of max path length
-public:
-    Path(const char *path) {
-        set(path);
-    }
-
-    void set(const char *path) {
-        strlcpy(this->path, path, sizeof(this->path));
-    }
-
-    const char *get() const {
-        return path;
-    }
-
-    void push(const char *component) {
-        size_t length = strlen(path);
-        if (length == 0 || path[length - 1] != '/') {
-            strlcat(path, "/", sizeof(path));
-        }
-        strlcat(path, component, sizeof(path));
-    }
-
-    void pop() {
-        char *slash = strrchr(path, '/');
-        if (slash == nullptr) {
-            return;
-        }
-        *slash = 0;
-    }
-};
 
 struct ResourcesScanResult {
     unsigned files_count;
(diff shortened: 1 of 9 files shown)


src/resources/fileutils.hpp
```diff
@@ -0,0 +1,74 @@
...
+#define log(severity, ...) _log_event(severity, log_component_find("Resources"), __VA_ARGS__)
+
+class DIRDeleter {
+public:
+    void operator()(DIR *dir) {
+        closedir(dir);
+    }
+};
+
+class FILEDeleter {
```
DRracer, line 17: /nitpick/ may be find a better name ... like `FILEClose` ... when I first saw it in a unique_ptr it looked like we are actually trying to delete the file instead of opening it :wink:

src/resources/hash.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+    // filedata
+    std::unique_ptr<FILE, FILEDeleter> source(fopen(path.get(), "rb"));
+    if (source.get() == nullptr) {
+        log(LOG_SEVERITY_ERROR, "Failed to open file %s", source.get());
+        return false;
+    }
+    uint8_t buffer[128];
+    bool success;
+    while (true) {
```
DRracer, line 61: /nitpick/ may a `do {} while` cycle would be the appropriate semantic construct here since you have the conditions at the end.

src/resources/hash.cpp
```diff
@@ -0,0 +1,163 @@
...
+    const char *filepath = full_path + root_length;
+    hash_ctx.append_data((uint8_t *)filepath, strlen(filepath));
+
+    // mark
+    hash_ctx.append_mark();
+
+    // filedata
+    std::unique_ptr<FILE, FILEDeleter> source(fopen(path.get(), "rb"));
+    if (source.get() == nullptr) {
+        log(LOG_SEVERITY_ERROR, "Failed to open file %s", source.get());
```
DRracer, line 56: Could we extract this `log(...)` to a common place? Most of the failed `fopen` scenarios need logging of the error, at least the filename. And it would be nice to have all the error texts consistent.

One possible solution would be using some custom class with a bool CheckSuccessfullyOpenned() instead of a std::unique_ptr.

src/resources/fileutils.hpp
```diff
@@ -0,0 +1,74 @@
+#pragma once
+#include <dirent.h>
+#include <stdio.h>
+#include <cstring>
+#include <iterator>
+#include "filesystem_littlefs_bbf.h"
+
+#define log(severity, ...) _log_event(severity, log_component_find("Resources"), __VA_ARGS__)
```
DRracer, line 8: Do we need this `#define log()` anywhere in this header file? Or is it something common for the whole resources module? I'm kind of afraid about spoiling the rest of the code base with such a generic `#define log()`, especially when accidentally used by an inexperienced dev.

I'd rather hide the `#define` inside a cpp file.

src/resources/hash.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+    // mark
+    hash_ctx.append_mark();
+
+    // filepath
+    size_t root_length = strlen(root);
+    const char *full_path = path.get();
+    const char *filepath = full_path + root_length;
+    if (strlen(filepath) == 0) {
+        filepath = "/";
```
DRracer, line 93: Are you sure this pointer to a temporary variable `"/"` will be valid beyond the block where it is used? In reality it probably will, but it is a bad practice to rely on it.
Suggestion:
```
static const char rootS[] = "/";
...
filepath=rootS;
```

src/resources/bootstrap.cpp
```diff
@@ -459,15 +448,31 @@ static bool do_bootstrap(buddy::resources::ProgressHook progress_hook) {
...
+    // calculate the hash of the resources we just installed
+    buddy::resources::Hash current_hash;
+    if (!buddy::resources::calculate_directory_hash(current_hash, "/internal/res")) {
+        log(LOG_SEVERITY_ERROR, "Failed to calculate hash of /internal/res directory");
+        return false;
+    }
+
+    if (buddy::resources::revision::required_revision.hash != current_hash) {
+        log(LOG_SEVERITY_ERROR, "Installed resources but the hash does not match!");
+        return false;
```
DRracer, line 473: I assume dealing with the invalid hash will be a subject of following PR's

src/resources/bootstrap.cpp
```diff
@@ -459,15 +448,31 @@ static bool do_bootstrap(buddy::resources::ProgressHook progress_hook) {
         return false;
     }
 
+    // clear installed resources
+    Path target_path("/internal/res");
+    buddy::resources::InstalledRevision::clear();
+    remove_directory_recursive(target_path);
```
DRracer, line 454: What shall we do in case we are unable the delete the directory recursively? Reformat XLFASH? The WD25 chip can have a problem on its own and the user at least needs to know a board replacement is necessary.

src/resources/bootstrap.cpp
```diff
@@ -300,6 +235,63 @@ static bool copy_file(const Path &source_path, const Path &target_path, Bootstra
     }
 }
 
+static bool remove_directory_recursive(Path &path) {
```
DRracer, line 238: I'd recommend using `[[nodiscard]]` for such an important operation ... and probably at other spots as well

src/resources/fileutils.hpp
dragomirecky: I was following codestyle already used by other parts of our project. Try to search for the word `Deleter` and you find a lot of those... deleters.
Also, I don't find it confusing as the template argument is also called deleter. So as soon as I see a class named `SomethingDeleter` I know, it is meant to be used together with std::unique_ptr and a pointer to `Something`

src/resources/hash.cpp
dragomirecky: I am not sure what you think the result should look like. As the conditions are two and based on which condition is true I need to set a variable, I am not sure how to make it more readable with `do{} while()`

src/resources/hash.cpp
dragomirecky: You mean to wrap `FILE` with our custom wrapper class?

I am not against! It sounds to me as something that needs to be carefully designed and then be available for the whole codebase. I don't really want to introduce classes like this just for the sake of the _resources_ module unless necessary. 

However, when implementing such a class, I am not convinced the right approach is to log all its unsuccessful `open()` calls. First, is unsuccessful call to `open` always a failure to be reported? I don't think so. For example, you might be using it to check whether a file exists. Second, having consistent logs from such module (and not logging it as the caller) might be a downside. Yes, it is a less code. But the logs from the caller (not the callee) give you much more information what went wrong. And the slight differences also enable you trace where exactly it went wrong. 

I am not trying to say this code is the best. I just don't think introducing a whole new class for such problem is necessary better. Maybe just a macro to report the error would be enough in this case? 🤷‍♂️ We can discuss in person ☺️

src/resources/fileutils.hpp
dragomirecky: It is common for the resources module.

It is not spoiling the rest of the code base. This header file is considered private to the module and shall not be included by any other module.

I used here an approach I would maybe like us to use in the future. The public API of the `resources` module is under `include/resources`. There is no extra search path, so you should include it as `#include <resources/bootstrap.hpp>` (or with `""`, but that is for another discussion). The private header files are then inside the module's implementation, under `src/resources/` and included by the module just as a local include `#include "fileutils.hpp`

src/resources/hash.cpp
dragomirecky: Yes, the C++ standard says string literals have type of `const char[]` and have static storage duration.

src/resources/bootstrap.cpp
dragomirecky: Yes! As you correctly pointed out in the comment bellow, we should handle some cases with broken filesystem etc. I plan to add this later.

src/resources/bootstrap.cpp
dragomirecky: Yes, reformat and hope in the best. Maybe report it to the user. I plan to add this later 👍

src/resources/bootstrap.cpp
dragomirecky: Good idea! Added :)

src/resources/fileutils.hpp
DRracer: ok

src/resources/hash.cpp
DRracer: ok

src/resources/fileutils.hpp
DRracer: Ok, if the header is not intended to be included anywhere besides the resources module, I'm fine with it.

src/resources/hash.cpp
DRracer: Oh, I didn't know this part of C++ standard. I must have run into some compiler issue back in the old days of "wooden computers" when this was not working thus being afraid of this construct ever since.

DRracer (approved):
All my comments were resolved and/or answered, approving

PR #2057, opened 2022-04-26
BFW-2542 PVB disappeared from preheat and filament menu

zoltan-l:


vorner (commented):
Mostly looks fine. However, the filaments got reordered. Is that OK?

What I worry about is, if someone upgrades the firmware, the eeprom contains the currently loaded filament as a number ‒ and if they get reordered, the printer might get confused and mismatch them. It's not a very big deal if that happens, so if there's a good reason to reorder them, then this is fine I think.

zoltan-l:
It is not reordering it is reinstallment the original order destroyed by some changes. Now it is coherent with the 4.3.4 filament order

zoltan-l:
Based on Radek review and recomendations changed the identical definitions by macro

RadekVana (approved):

DRracer:
The order of filament is correct (matches the previous state defined in BFW-1661)

PR #2071, opened 2022-05-02
Fix bootstrap with empty xflash

dragomirecky:


src/resources/bootstrap.cpp
```diff
@@ -292,6 +293,29 @@ static bool copy_file(const Path &source_path, const Path &target_path, Bootstra
     return remove(path.get()) == 0;
 }
 
+static bool remove_recursive_if_exists(Path &path) {
+    struct stat sb;
+    int stat_retval = stat(path.get(), &sb);
+
+    // does it exists?
+    if (stat_retval == -1 && errno == ENOENT) {
```
RadekVana, line 301: you could write
```
if (stat_retval == -1) {
    // erno == ENOENT does not exist
    // erno != ENOENT something went wrong
    return errno == ENOENT;
}
```
instead of those 2 ifs

RadekVana (approved):

src/resources/bootstrap.cpp
dragomirecky: I know! But less lines isn't a metric of a better code.

This is subjective, but I believe the separate if statements are easier to read and think about (in other words, the code is extremely simple and it doesn't require almost any mental capacity to read it. Your proposed code is much harder to read and think about :)

PR #2082, opened 2022-05-04
Bootloader Update

dragomirecky:
WIP

### DIFF
--- a/src/bootloader/bootloader.cpp
+++ b/src/bootloader/bootloader.cpp
@@ -43,8 +43,6 @@ constexpr const uint8_t *bootloader_sector_get_address(int sector) {
     return (const uint8_t *)base_address;
 }
 
-Version *const bootloader_version = (Version *)0x0801FFFA;
-
 class FileDeleter {
 public:
     void operator()(FILE *file) {
@@ -238,6 +236,7 @@ static void copy_bootloader_to_flash(FILE *bootloader_bin, ProgressCallback prog
 }
 
 Version buddy::bootloader::get_version() {
+    Version *const bootloader_version = (Version *const)0x0801FFFA;
     return *bootloader_version;
 }
 
--- a/src/bootloader/bootloader.cpp
+++ b/src/bootloader/bootloader.cpp
@@ -280,6 +280,8 @@ void buddy::bootloader::update(ProgressHook progress) {
         fatal_error("bootloader.bin failed to open");
     }
 
+    progress(calc_percent_done(100, 0), buddy::bootloader::UpdateStage::PreparingUpdate);
+
     // update the bootloader in FLASH
     int last_reported_percent_done = -1;
     copy_bootloader_to_flash(bootloader_bin.get(), [&](int percent_done) {
--- a/src/bootloader/bootloader.cpp
+++ b/src/bootloader/bootloader.cpp
@@ -1,6 +1,7 @@
 #include "log.h"
 #include "bsod.h"
 #include <iostream>
+#include <tuple>
 #include <memory>
 #include "cmsis_os.h"
 #include "crc32.h"
@@ -248,9 +249,8 @@ bool buddy::bootloader::needs_update() {
     auto current = get_version();
     auto required = buddy::bootloader::required_version;
 
-    return current.major < required.major
-        || (current.major == required.major && current.minor < required.minor)
-        || (current.major == required.major && current.minor == required.minor && current.patch < required.patch);
+    return std::tie(current.major, current.minor, current.patch)
+        < std::tie(required.major, required.minor, required.patch);
 }
 
 void buddy::bootloader::update(ProgressHook progress) {
--- a/src/bootloader/bootloader.cpp
+++ b/src/bootloader/bootloader.cpp
@@ -187,7 +187,7 @@ static void copy_bootloader_to_flash(FILE *bootloader_bin, ProgressCallback prog
             }
 
             uint32_t current_preboot_crc = crc32_calc(bootloader_sector_get_address(0), bootloader_sector_get_size(0));
-            if (current_preboot_crc == expected_preboot_crc && false) {
+            if (current_preboot_crc == expected_preboot_crc) {
                 log(LOG_SEVERITY_INFO, "No need to update preboot. Skipping sector 0.");
                 continue;
             } else {
--- a/src/bootloader/bootloader.cpp
+++ b/src/bootloader/bootloader.cpp
@@ -1,0 +1,286 @@
+#include "log.h"
+#include "bsod.h"
+#include <iostream>
+#include <memory>
+#include "cmsis_os.h"
+#include "crc32.h"
+#include "bsod.h"
+#include <string.h>
+
+#include "scratch_buffer.hpp"
+#include "resources/bootstrap.hpp"
+#include "resources/revision_bootloader.hpp"
+#include "bootloader/bootloader.hpp"
+#include "bootloader/required_version.hpp"
+
+// FIXME: Those includes are here only for the RNG.
+// We should add support for the stdlib's standard random function
+#include "main.h"
+#include "stm32f4xx_hal.h"
+
+using Version = buddy::bootloader::Version;
+using UpdateStage = buddy::bootloader::UpdateStage;
+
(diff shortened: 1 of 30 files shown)


src/bootloader/bootloader.cpp
```diff
@@ -0,0 +1,286 @@
...
+        }
+
+        // seek at the sector in the file
+        if (fseek(bootloader_bin, bootloader_sector_get_address(sector) - bootloader_sector_get_address(0), SEEK_SET) != 0) {
+            fatal_error("bootloader update: failed to seek sector");
+        }
+
+        // erase the sector
+        HAL_FLASH_Unlock();
+        if (!flash_erase_sector(sector)) {
```
DRracer, line 212: Does flash_erase_sector lock the FLASH internally when done?
When it fails there is probably something seriously wrong with the CPU, so some locking before throwing the fatal_error is probably not necessary.

But does HAL_FLASH_Unlock need to be called again after erasing a sector?

src/bootloader/bootloader.cpp
```diff
@@ -0,0 +1,286 @@
...
+            fatal_error("bootloader update: failed to erase sector");
+        }
+
+        // program the sector
+        HAL_FLASH_Unlock();
+        bool flash_successful = flash_program_sector(sector, bootloader_bin, [&](size_t bytes_written) {
+            if (sector != 0) {
```
DRracer, line 219: Considering the fact, that failing to flash the CPU means something seriously wrong with the CPU itself, it is probably not worth it saving the first 16KB elsewhere and try to flash it back intact in case the preboot cannot be flashed (i.e. restore the previous bootloader)?

src/bootloader/bootloader.cpp
```diff
@@ -0,0 +1,286 @@
...
+
+Version buddy::bootloader::get_version() {
+    return *bootloader_version;
+}
+
+bool buddy::bootloader::needs_update() {
+    auto current = get_version();
+    auto required = buddy::bootloader::required_version;
+
+    return current.major < required.major
```
DRracer, line 246: ideally rewrite this using `std::tie` + `operator<` thereof ... I always make errors in these comparisons of sequences

src/bootloader/bootloader.cpp
```diff
@@ -0,0 +1,286 @@
...
+                progress(calc_percent_done(percent_done, 0), buddy::bootloader::UpdateStage::PreparingUpdate);
+                break;
+            }
+        });
+    }
+
+    std::unique_ptr<FILE, FileDeleter> bootloader_bin(fopen("/internal/res/bootloader.bin", "rb"));
+    if (bootloader_bin.get() == nullptr) {
+        log(LOG_SEVERITY_CRITICAL, "bootloader.bin failed to fopen() after its bootstrap");
+        fatal_error("bootloader.bin failed to open");
```
DRracer, line 275: are any of these fatal_errors recoverable somehow, or do they inevitably lead to a board replacement?

src/common/scratch_buffer.hpp
```diff
@@ -1,6 +1,8 @@
-// scratch_buffer.h
-
+#pragma once
 #include <inttypes.h>
+#include <stddef.h>
+
+namespace buddy::scratch_buffer {
 
 /// Single place for definition of large RAM space for various tasks.
 /// Currently, there's no locking mechanism so before you use it, make sure
```
DRracer, line 8: ... aren't you just adding some locking mechanizm :wink: ? Please update the comment accordingly as well.

src/common/scratch_buffer.hpp
```diff
@@ -13,6 +15,26 @@
...
+        return sizeof(buffer);
+    }
+};
+
+class Ownership {
+private:
+    bool acquired;
+
+public:
+    Ownership();
```
DRracer, line 31: Is there a reason for not having this class a singleton? AFAIK CCMRAM area doesn't depend on any HW and can be inited any time - i.e. an ideal example for a singleton.

src/gui/resource.c
```diff
@@ -204,7 +204,7 @@ RESOURCE_ENTRY_PNG(png_arrow_up_orange_8px)
 RESOURCE_ENTRY_PNG(png_arrow_down_orange_8px)
 
 RESOURCE_ENTRY_PNG(png_back_32px)
-RESOURCE_ENTRY_PNG(png_serial_printing)
+RESOURCE_ENTRY_NUL()
```
DRracer, line 207: ??

dragomirecky: Omg

Thanks for reminding me of the small issue: it won't fit the FLASH 🙈
This was outgoing issue so at some point I just ignored it and did this workaround. Now it is the time to fix it. I am gonna look at our options.

src/common/scratch_buffer.hpp
dragomirecky: There is a singleton – the ScratchBuffer instance (intentionally private in scratch_buffer.cpp).

This class represents the Ownership of the buffer (RAII). Thus it is not a singleton.

By making the singleton private, the API does not allow (without acquiring an ownership first) use of the buffer.

src/common/scratch_buffer.hpp
dragomirecky: Nice catch ☺️

src/bootloader/bootloader.cpp
dragomirecky: This one would not lead to board replacement, because we haven't touch the FLASH yet. 

The one below might (in case the erase failed while changing flash content).

src/bootloader/bootloader.cpp
dragomirecky: Great idea 🙏

src/bootloader/bootloader.cpp
```diff
@@ -0,0 +1,286 @@
...
+
+constexpr const uint8_t *bootloader_sector_get_address(int sector) {
+    uint8_t *base_address = (uint8_t *)0x08000000;
+    for (int i = 0; i < sector; i++) {
+        base_address += bootloader_sector_get_size(i);
+    }
+    return (const uint8_t *)base_address;
+}
+
+Version *const bootloader_version = (Version *)0x0801FFFA;
```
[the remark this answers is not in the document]
dragomirecky: Done 👍

src/bootloader/bootloader.cpp
dragomirecky: I don't think so. At that point it is equally safe to flash the preboot or the original part of the bootloader. However, having the original part would add additional complexity to the code.

We might do some more thorough analysis to see all the possible reasons this could fail and see if we can improve its robustness.

One more thing I had on my mind (and I think would be nice to add eventually) is one more check of the bootloader data consistency (ensuring the buffer we are just about to flash to the CPU holds the data we expect it to hold). Something we can discuss in person.

src/bootloader/bootloader.cpp
dragomirecky: I was not sure it needs to be called. 
But I am sure calling it again is fine.
Therefore I picked the safe side here :)

src/gui/guimain.cpp
```diff
@@ -234,7 +270,7 @@ void gui_run(void) {
     Screens::Access()->Loop();
 
 #ifdef HAS_RESOURCES
```
DRracer, line 272: What is the relation between `#if ENABLED(RESOURCES)` and `#ifdef HAS_RESOURCES`?
As far as I understand, we have an explicit definition of `HAS_RESOURCES` in `CMakeLists.txt`:
```
if(RESOURCES)
  target_compile_definitions(firmware PUBLIC HAS_RESOURCES)
  add_subdirectory(src/resources)
endif()
```
Do we really need the macro `HAS_RESOURCES` from now on?

src/common/scratch_buffer.hpp
DRracer: ok, got it

src/gui/resource.c
DRracer: ok, no problem

DRracer (approved):
all my comments and questions have been either resolved or explained -> approving

PR #2090, opened 2022-05-06
WiFi: Check the liveness of ESP

vorner:
Check that we get an intron from time to time, to see ESP is alive.
Provoke it to send something too.

This deals with possibility of the ESP dying or the communication
breaking without it giving us any info it is not OK ‒ therefore we would
assume the link is still up.

If it's not answering, consider it dead and restart.


DRracer (approved):
look correct to me

PR #2096, opened 2022-05-10
BFW2558 BFW2559 BFW2560

RadekVana:


src/gui/MItem_lan.hpp
```diff
@@ -37,9 +37,9 @@ class MI_WIFI_CREDENTIALS_t : public WI_LABEL_t {
 };
 
 class MI_NET_INTERFACE_t : public WI_SWITCH_t<3> {
-    constexpr static const char *const label = "Network Interface"; //do not translate
+    constexpr static const char *const label = "Interface"; //do not translate
 
-    constexpr static const char *str_off = N_("None");
+    constexpr static const char *str_off = "Off";    //do not translate
```
martin357, line 52: Just a small question: why should this not be translated?

RadekVana: Off is multinational

RadekVana: Or at least every body knows what it means

martin357 (approved):
I don't see a problem.

src/gui/wizard/selftest_frame_esp_progress.cpp
```diff
@@ -0,0 +1,103 @@
...
+    case PhasesSelftest::ESP_progress_failed:
+        txt_top = N_("Updating Wi-Fi...");
+        txt_bot = N_("Firmware flashing\nfailed!");
+        show_progress = true;
+        break;
+    default:
+        break;
+    }
+
+    if (txt_top) {
```
martin357, line 74: This is taken from somewhere else? Can it be inherited?

martin357 (approved):
I don't see a problem.

martin357:
fdsdfgdfg

martin357 (approved):

src/gui/wizard/selftest_frame_esp_progress.cpp
RadekVana: I don't know what do you mean
`SelftestFrameESP_progress` could be inherited from `SelftestFrameESP`
Because they both have a textbox (but with different size, alignment and texts)
But I don't see any benefit in that

src/gui/MItem_lan.hpp
```diff
@@ -64,3 +74,47 @@ class MI_NET_IP_t : public WI_SWITCH_t<2> {
...
+
+class MI_NET_IP_VER_t : public WI_SWITCH_t<2> {
+    constexpr static const char *const label = N_("Protocol");
+
+    constexpr static const char *str_v4 = "IPv4"; //do not translate
+    constexpr static const char *str_v6 = "IPv6"; //do not translate
+
+public:
+    MI_NET_IP_VER_t();
+    //virtual void OnChange(size_t old_index) override; //TODO
```
mkbel, line 86: Commented out code/TODO forgotten?

mkbel (approved):

PR #2099, opened 2022-05-11
Reorganize the startup procedure of the firmware

dragomirecky:
This makes sure minimal amount of work is done before starting FreeRTOS,
therefore moving most of the initialization code into an RTOS task and
context.

Even C(++) constructors of global objects (__libc_init_array) are
now called in FreeRTOS context.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -1,0 +1,140 @@
+#include "main.h"
+#include "cmsis_os.h"
+#include "stm32f4xx_hal.h"
+#include "SEGGER_SYSVIEW.h"
+#include "crc32.h"
+#include "eeprom.h"
+#include "tick_timer_api.h"
+
+static void system_clock_configure(void) {
+    // Configure the main internal regulator output voltage
+    __HAL_RCC_PWR_CLK_ENABLE();
+    __HAL_PWR_VOLTAGESCALING_CONFIG(PWR_REGULATOR_VOLTAGE_SCALE1);
+
+    // Initializes the CPU, AHB and APB busses clocks
+    RCC_OscInitTypeDef osc_init;
+    rcc_osc_get_init(&osc_init);
+    if (HAL_RCC_OscConfig(&osc_init) != HAL_OK) {
+        Error_Handler();
+    }
+
+    // Initializes the CPU, AHB and APB busses clocks
+    RCC_ClkInitTypeDef clk_init;
+    rcc_clk_get_init(&clk_init);
+    if (HAL_RCC_ClockConfig(&clk_init, FLASH_LATENCY_5) != HAL_OK) {
+        Error_Handler();
+    }
+
+    SystemCoreClock = system_core_get_clock();
+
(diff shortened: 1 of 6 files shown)


src/main.c
```diff
@@ -0,0 +1,151 @@
...
+}
+
+/// The entrypoint of our firmware
+///
+/// Do not do anything here that isn't essential to starting the RTOS
+/// That is our one and only priority.
+///
+/// WARNING
+/// The C++ runtime hasn't been initialized yet (together with C's constructors).
+/// So make sure you don't do anything that is dependent on it.
```
DRracer, line 127: :+1:

src/main.c
```diff
@@ -0,0 +1,151 @@
...
+    enable_backup_domain();
+    enable_segger_sysview();
+    enable_dfu_entry();
+
+    // define the startup task
+    osThreadDef(startup, startup_task, osPriorityHigh, 0, 1024);
+    osThreadCreate(osThread(startup), NULL);
+
+    // start the RTOS with the single startup task
+    osKernelStart();
```
DRracer, line 150: /nitpick/ return something :wink:

src/main.c
```diff
@@ -0,0 +1,151 @@
...
+static void startup_task(void const *argument) {
+    // init crc32 module. We need crc in eeprom_init
+    crc32_init();
+
+    // init communication with eeprom
+    eeprom_init_i2c();
+
+    // init eeprom module itself
+    taskENTER_CRITICAL();
+    eeprom_init();
```
DRracer, line 106: :thinking: this is walking on thin ice - eeprom_init() is defined inside a C++ source file and may actually call something which is depending on the C++ runtime/globals/constructors. And if not now, someone could add something like that in the future, there is nothing preventing from doing so.

src/main.c
```diff
@@ -0,0 +1,151 @@
...
+/// The C++ runtime isn't initialized at the beginning of this function
+/// and initializing it is the main priority here.
+/// So first, we have to get the EEPROM ready, then we call libc_init_array
+/// and that is the time everything is ready for us to switch to C++ context.
+static void startup_task(void const *argument) {
+    // init crc32 module. We need crc in eeprom_init
+    crc32_init();
+
+    // init communication with eeprom
+    eeprom_init_i2c();
```
DRracer, line 102: :thinking:  this calls ErrorHandler in case of an error, which calls BSOD (which is again a C++ function dependent on GUI)

DRracer (changes requested):
Generally, I like the approach. Parts of the initialization sequence need to be double-checked if they do not rely on the inited C++ runtime. This may require some more refactoring in order to prevent less skilled + unaware devs from accidentally putting some C++-related stuff into it. My primary concern is the EEPROM init, that doesn't feel right (and didn't feel right before anyway).

DRracer (dismissed):
more work will follow, the discussed issues are present also in the current code base

DRracer (approved):

PR #2100, opened 2022-05-11
espif: Check if the link really changed

vorner:
Since we use the link status message for a periodic "ping"...


dragomirecky (approved):
👍

PR #2113, opened 2022-05-13
Enable start print

vorner:
Once this is merged, we probably should close BFW-2300.


lib/WUI/wui_api.cpp
```diff
@@ -347,30 +350,29 @@ uint32_t wui_gcodes_uploaded() {
...
+    // race condition. Doing it properly would be kind of hard and the risk is
+    // we maybe start the print and don't get the print screen or something
+    // like that ‒ annoying, but not entirely dangerous.
+    //
+    // We assume marlin does another check when asked to print and won't start
+    // a print inside a print or such.
+    //
+    // Also, this introduces another code dependency in the „wrong direction“.
+    // GUI should be a neighbor of WUI and should not depend on each other.
+    // But, well, …
```
DRracer, line 364: I agree with your complaints, but the infrastructure is not ready for the task of synchronizing multiple clients among each other.

DRracer (approved):
nice :+1:

PR #2121, opened 2022-05-16
Fix and unify critical sections.

mkbel:
Fix of st7789v_reset() might fix issue of wrongly detected rotary encoder model.

CRITICAL_SECTION_START:
make primask const to catch more usage faults.

CRITICAL_SECTION_END:
__set_PRIMASK(primask) might be more efficient
than if (!primask) __enable_irq()
due to the fact that "cpsie i" ASM instruction is not allowed in ASM condition

Stepper::babystep() might enable interrupts unintentionally before the fix.
We call it from marlin server so better to be safe than sorry.

sys_reset() remove redundant code:
1. Return value of __get_PRIMASK doesn't need to be masked.
2. irq variable is not needed as interrupt masking state is not restored.
3. irq type doesn't match __get_PRIMASK() return value.
4. "cpsid i" ASM instruction is not allowed in condition so who knows what kind of code was generated.

sys_pll_enable(), sys_pll_disable(), sys_spi_set_prescaler() and st7789v_reset() fix:
1. Before the fix, logic was inverted and totally wrong.
__get_PRIMASK() returns 1 if interrupts are masked, 0 if interrupts are enabled
So before the fix interrupts were disabled in the beginning only when already disabled (masked).
And interrupts were enabled in the end if those were disabled before calling this function.
2. Return value of __get_PRIMASK doesn't need to be masked.
3. irq type didn't match __get_PRIMASK() return value.
4. "cpsie i" ASM instruction is not allowed in condition
so who knows what kind of code was generated when __enable_irq() is put
in if condition.
5. irq variable renamed to primask, as irq is misleading as it has opposite meaning.
6. irq was made const.

sys_sscg_disable(void) and sys_sscg_enable(void) might enable interrupts
unintentionally before the fix.

sys_sscg_disable(void) might return with interrupts still disabled as one
of its return point didn't handle interrupts before the fix.

EepromSystemInit():
1. Before the fix, logic was inverted and totally wrong.
__get_PRIMASK() returns 1 if interrupts are masked, 0 if interrupts are enabled
So interrupts were disabled before returning of this function
if those were enabled before calling this function.
2. Return value of __get_PRIMASK doesn't need to be masked.
3. irq type didn't match __get_PRIMASK() return value.
4. "cpsid i" ASM instruction is not allowed in condition
so who knows what kind of code was generated when __disable_irq() is put
in if condition.
5. irq variable renamed to primask, as irq is misleading as it has opposite meaning.
6. irq was made const.

### DIFF
--- a/src/common/disable_interrupts.h
+++ b/src/common/disable_interrupts.h
@@ -1,0 +1,91 @@
+/**
+ * @file
+ *
+ */
+
+#pragma once
+
+#include <stdint.h>
+#include <cmsis_gcc.h>
+
+namespace buddy {
+
+/**
+ * @brief Disable all maskable interrupts
+ *
+ * Disabling all interrupts may be needed in case of dealing with stepper data
+ * or in case timing with nanosecond precision is needed.
+ *
+ * In other cases use CriticalSection instead.
+ *
+ * Interrupts are disabled when object is constructed by its default constructor
+ * and resumed to original state once destroyed.
+ *
+ * When the motors are moving disabling all interrupts for more than
+ *
+ * - 500 ns frequently may produce motor noise
+ * (Induce stepping error bigger than 1% at 200 mm/s 200 step motor 16 microsteps or 400 step motor 8 microsteps)
+ *
+ * - 50 us produce multiple microsteps done at once (noise, may cause motor over-current in no-current-feedback-mode)
+ * - 400 us may cause motor stall (400 step motors)
+ * - 800 us -//- (200 step motors)
+ *
+ * Example usage:
+ * @code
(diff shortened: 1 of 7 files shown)


src/common/disable_interrupts.h
```diff
@@ -0,0 +1,91 @@
...
+ *
+ *     interrupts.disable();
+ *     // critical code
+ *     interrupts.resume();
+ *
+ *     // non-critical code
+ *
+ *     interrupts.disable();
+ *     // critical code
+ * }
```
dragomirecky, line 65: Nice & Clean API 👏

dragomirecky (approved):

PR #2122, opened 2022-05-17
Net menu fixes

RadekVana:


src/common/selftest/selftest_esp_update.cpp
```diff
@@ -297,11 +297,14 @@ bool EspCredentials::make_file() {
         return false;
     }
 
-    fputs(file_str, file.get());
+    bool failed = fputs(file_str, file.get()) == EOF;
+    if (failed) {
```
DRracer, line 301: I'd recommend a more general error check ... something like
`if( fputs(...) < 0 ) ... log error`

RadekVana: https://www.cplusplus.com/reference/cstdio/fputs/

Return Value
On success, a non-negative value is returned.
On error, the function returns EOF.

src/common/selftest/selftest_esp_update.cpp
```diff
@@ -297,11 +297,14 @@ bool EspCredentials::make_file() {
...
+        log_error(Network, "ESP credentials: Unable to write into file %s", file_name);
```
DRracer, line 302: you can return false in this if() branch, the file will get closed correctly

martin357 (approved):
Looks good, maybe some variables could be more descriptive: i.e. file_str, but that is not part of this PR.

PR #2141, opened 2022-05-24
M600 flickering

RadekVana:
Fix unnecessary invalidation of menu items while enabling/disabling.
Make M600 menu item enabling less frequent.


martin357 (approved):
:+1:

PR #2149, opened 2022-05-25
gui: Save RAM in window_dlg_strong_warning_t

RadekVana:


dragomirecky (approved):
Checked and we managed to save 6 KB of RAM 💪

PR #2152, opened 2022-05-26
lwip: Save RAM by shortening supported domain names

dragomirecky:
Saves 640 bytes of RAM


vorner (approved):
That's quite a bit, that's a surprise. I assume we are OK not accepting some technically valid DNS names.

PR #2158, opened 2022-05-30
Force filament selection in load operation

MistrKernnunos:
This also fixes filament change in first layer calibration
BFW-2578

### DIFF
--- a/src/common/filament_sensor_api_no_mmu.cpp
+++ b/src/common/filament_sensor_api_no_mmu.cpp
@@ -24,7 +24,7 @@ void FilamentSensors::evaluate_sensors() {
     std::unique_lock lock_printer(mutex_printer, std::defer_lock);
 
     // lock both unique_locks without deadlock
-    std::lock(lock_mmu, lock_printer);
+    buddy::lock(lock_mmu, lock_printer);
 
     has_mmu = false;
     state_of_mmu_sensor = fsensor_t::Disabled;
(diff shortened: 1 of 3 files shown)


src/marlin_stubs/pause/M701_2.cpp
```diff
@@ -52,8 +52,8 @@ bool filament_gcodes::load_unload(LoadUnloadMode type, filament_gcodes::Func f_l
 void filament_gcodes::M701_no_parser(filament_t filament_to_be_loaded, float fast_load_length, float z_min_pos, std::optional<RetAndCool_t> op_preheat, uint8_t target_extruder, int8_t mmu_slot) {
     Filaments::SetToBeLoaded(filament_to_be_loaded);
     if (op_preheat) {
-        PreheatData data(fast_load_length > 0.F ? PreheatMode::Load : PreheatMode::Purge, *op_preheat);
-        auto preheat_ret = preheat(data);
+        PreheatData data(isnan(fast_load_length) || fast_load_length > 0.F ? PreheatMode::Load : PreheatMode::Purge, *op_preheat);
```
DRracer, line 55: I really don't like the hack of using NaN instead of some std::optional. Anybody else unaware of this hack will run straight into it and spends a lot of time fixing functionality dependent on `fast_load_length`

MistrKernnunos: I don't like it too, but this is just quick fix for a bug. If I change this to std::optional I would like to change all optional arguments to std::optional, should I do it?

PR #2162, opened 2022-06-01
Closing „error“ HTTP

vorner:
When there's an error on HTTP (usually, early in HTTP upload), we send
the response ASAP, close our side of connection and then continue
accepting the data until the other side is satisfied (not closing their
side forcefully).

This is to deal with browsers that don't deal well with the connection
being closed on them. Now, we have to accept the whole transfer (which
takes time), but gives the correct error description.


zoltan-l (approved):
Did not find anything suspicious

zoltan-l (approved):

PR #2172, opened 2022-06-08
Increase wait time in try_lock for FreeRTOS_mutex

MistrKernnunos:
This increase should cause priority inheritance. We need priority inheritance to avoid priority induced deadlock. 
(thread with higher priority us waiting for thread with lower priority to release mutex, but the thread with lower priority does not get any CPU time without the priority inheritance)
FreeRTOS does priority inheritance when the block time is larger than  zero


wavexx:
This breaks the semantics of try_lock though. Doing this might be ok for the specific case that causes priority inversion.

Can you point out where it happens?

MistrKernnunos:
This can happen during filament operatio in state when we are reading filament sensor state. Here the dafult task uses std::lock to lock two mutexes, but one of the mutexes can be locked by measurement task, which have lower priority.

mkbel (changes requested):
I think this is not good solution. try_lock is expected to be non-blocking. I would propose instead implement our buddy::lock which is not broken.
Implement it for just two arguments.
Lock first argument and try_lock second, if ti fails release first lock and lock second argument and try_lock first, if it fails release second lock and loop again.

Create our mutex file, in that file just include system mutex file and GCC poison std::lock. So that broken std::lock is not used anymore.

I checked gcc-arm-none-eabi 11.2.1 and that bug is still present there. So there is no option for us just to update compiler.

Please also report this bug to gcc - that std::lock deadlocks if second or higher argument is locked mutex and std::lock is called from higher priority task than task holding second or higher argument mutex.

mkbel:
It looks like this bug of gcc libstdc++ is not yet reported. I have not found anything close to this.

src/common/freertos_mutex.hpp
```diff
@@ -29,3 +29,38 @@ class FreeRTOS_Mutex {
...
+            lock_first = 1;
+            break;
+        case 1:
+            if (lock_helper(l2, l1)) {
+                return;
+            }
+            lock_first = 0;
+            break;
+        }
+    }
```
mkbel, line 64: What is the point of switch statement?
Wouldn't be:
```
while (true) {
    if (lock_helper(l1, l2)) {
        return;
    }
    if (lock_helper(l2, l1)) {
        return;
    }
}
```
Sufficient?

vorner:
Honestly, the solution proposed by @mkbel seems suspicious at least. The fact one tries to repeatedly lock and unlock in alternating order in fact can cause a livelock instead!

I don't actually see any reason why a `lock` should use `try_lock` internally. A lock should only use the fully blocking ones that give up CPU time until the lock is acquired completely.

The way deadlock prevention is done with locking in a guaranteed-to-work way is by using a consistent locking order. All mutexes involved are ordered, always according to the same rule and always locked from the smallest to largest. The rule that's commonly used is the address at which the mutex resides. But then, if `std::lock` is broken for us, we would need to use our own `lock` at all places where this is relevant.

Nevertheless, it would be even better if we didn't have to lock two things at the same time.

mkbel:
@vorner I think livelock is avoided by the fact we don't have (hopefully) any tasks sharing the same priority. So the locks are inevitably acquired by the highest priority task.

vorner:
That's a very fragile way, we shift with priorities from time to time and we may move code from one task to another. Especially if there exist a solution that doesn't need such special case.

mkbel:
And of course you are right that maybe we even don't need such over-complicated code at all.

MistrKernnunos:
@mkbel pragma poison can be used only for lock not for std::lock.

mkbel (approved):
I am approving this. But I must admit I am in conflict of interests as I am approving solution I proposed. @vorner is right that this solution might lead to livelock in case we happen to use the same priority for more task. I don't see it as big deal as I am anyway opponent of using the same priority for more tasks.  Ordering the locks proposed by @vorner is even better solution and probably it can be implemented easily for arbitrary number of mutexes in template.

MistrKernnunos:
I change the algorith to locking the mutexes in one order. Currently only for two mutexes. 
I think that we should remove the mutexes and use communication between marlin client and server instead of directly accessing the FSInstance.

src/common/freertos_mutex.hpp
```diff
@@ -29,3 +29,19 @@ class FreeRTOS_Mutex {
...
     char xMutexBuffer[buffer_size]; // this size is checked in cpp file
 };
+
+// we need our own lock, because GCC uses try and back off which internally uses try_lock and because FreeRTOS does not use priority inheritance when trying to lock mutex without blocking
+// this results in deadlock (inverse priority problem), because thread with lower priority does not get CPU time while still holding mutex.
+namespace buddy {
+
+template <typename lockable1, typename lockable2>
+void lock(lockable1 &l1, lockable2 &l2) {
+    if (&l1 < &l2) {
```
mkbel, line 39: This probably might break if caller once wrap pointer to mutex in some class with `lock` method. And in one place it locks wrapper class and in different place bare mutex. Maybe it could be solved if this is not template but type of object it operates on is enforced.

mkbel (approved):

mkbel (approved):

PR #2203, opened 2022-06-30
Fix silent error (bootloop) due to flash chip size & MFRID

### LINKED ISSUE #2188: Bootloader not getting updated
I don't want to open a bug report because it may not officially be a bug. However, I am unable to update the bootloader to anything above 1.0.0. Are there any official .bbf's that update to 1.1.1 or 1.0.1? 

I've tried so many combinations and custom builds but no luck getting the bootloader up to the 2.0.2 that I need to use esp's.

Thank you!

CarlosGS:
Currently there is a check for specific MFRID which badly freezes when using memory chips from any different brand.

It causes a white screen at startup, with no information to the user, which would cause soft bricks for printers using other memory chips when upgrading to the official 4.4.0+ firmware.

This was present since two years ago, and the problem simply has appeared now that the flash chip is actually used (with the new & awesome offloading of resources into the flash memory).
The fix is to remove the MFRID check, because checking for DEVID is enough to determine that the correct chip is present. With this change everything works perfectly :tada: 

Fixes [ Bootloader not getting updated #2188 ](https://github.com/prusa3d/Prusa-Firmware-Buddy/issues/2188)

**Update:** Also fixes a similar problem with 16MB flash chips, see https://github.com/prusa3d/Prusa-Firmware-Buddy/pull/2203#issuecomment-1201824784


MinnesotaMean (commented):
This fixed my issue for 4.4.0 updating. Awesome work!

MinnesotaMean (approved):

MinnesotaMean (dismissed):

DRracer:
@CarlosGS do you see a binary at `.dependencies/clang-format-9.0.0-noext/clang-format`? Does it have the executable flag set?
Normally, the build environment is set up using `utils/bootstrap.py`, which shall be called also during normal build via `utils/build.py`

CarlosGS:
@DRracer Thank you, gave it a fresh try and it works! :)
Strict compilation was failing and had to add `__attribute__((unused))` to the variable.

CarlosGS:
Found a similar problem with boards that have a 16MB flash chip, the firmware was halting as it expected either 8MB or 64MB (fixed DEVID and sector count). This was tricky to identify as it occurs so early at boot. Thanks JetSet in Discord for helping troubleshoot!

This last commit completes the support in the flash driver, allowing a normal boot in these cases. It would still be possible to require specific memory sizes in the future, but at least the firmware won't freeze silently, allowing to eventually notify these users.

PeterDHabermehl (approved):
Trying to update my Mini I ran into the bootloop issue.
I did some research on the internet and found @CarlosGS contribution.
I reviewed the changes made, considered it plausible and built a firmware from the recent master branch plus this commit. 
Build process and flashing of the fw worked flawless, so I approve this commit.

PeterDHabermehl (approved):
This commit properly addresses issues with different flash chip sizes.
Reviewed the code and built a fw to test - passed successfully, no issues, was able to flash two printers that failed before.

PeterDHabermehl (approved):
srry, forgot to mark the file as reviewed before.
I can confirm that the fix is working, reviewed the code and found it straightforward. No objections.

bmangodt:
This NEEDS to be in the official release. I still don't get why this isn't merged. This fixes an issue for LOTS of enthusiasts. The changes, as mentioned numerous times, are straight forward and non-intrusive.

thisiskeithb:
@CarlosGS: Would you be able to rebase/update this PR and resolve merge conflicts? I have a third-party Buddy board and would be keen to see if this fix still works.

PR #2205, opened 2022-06-30
Proper parsing of JSON structures (for connect & WUI)

vorner:
+ Implement a sample command to be handled.


src/Connect/command.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+        parse_result = jsmn_parse(&parser, body.data(), body.size(), tokens, sizeof tokens / sizeof *tokens);
+    } // Free the parser
+
+    CommandType command_type = CommandType::Unknown;
+
+    // Error from jsmn_parse will lead to -1 -> converted to 0, refused by json::search as Broken.
+    const bool success = json::search(body.data(), tokens, std::max(parse_result, 0), [&](const Event &event) {
+        if (event.depth != 1 || event.type != Type::String) {
+            return;
```
mkbel, line 47: Is this legal?
From C++11 draft:
6.6.3 The return statement [stmt.return]
1 A function returns to its caller by the return statement.
2 A return statement with neither an expression nor a braced-init-list can be used only in functions that do not
return a value, that is, a function with the return type cv void, a constructor (12.1), or a destructor (12.4).

mkbel:
Is first commit message relevant for the whole pull request?
(
Connect: Preliminary parsing and handling of JSON
Not complete, not 100% correct, but does something sometimes.
)

src/Connect/command.cpp
vorner: Yes, this is legal. This goes under the second point, function returning `void`. Well, it's lambda returning `void`, but that's the same kind of thing. Don't get yourself confused by the `json::search` itself (which returns `bool`), the lambda it takes as a parameter has no `-> ReturnType` specifier, which makes it return `void`.

If passing functions/lambdas into other functions isn't a commonly used technique in our team, I guess we can put it into the language window in the next meeting.

vorner:
> Is first commit message relevant for the whole pull request?

What do you mean, relevant for the whole pull request?

The first one took the `search` from WUI, extracted it to a shared place and used it. But the `search`s functionality was limited to the use cases in WUI and didn't cover all the use cases we'll have around Connect.

So the next two commits generalize it to deal with everything.

mkbel:
> The first one took the search from WUI, extracted it to a shared place and used it. But the searchs functionality was limited to the use cases in WUI and didn't cover all the use cases we'll have around Connect.

Maybe this information should be added to the first commit message. Sounds more useful than "Not complete, not 100% correct, but does something sometimes."

mkbel (approved):

PR #2210, opened 2022-07-06
Add support for loading PHA filament

dhiltonp:
colorFabb has released allPHA, with 190-200C printing temp and a cooled heat bed: https://colorfabb.com/allpha-natural

The code compiles, but this code is not tested; I don't have a mini. I've put an equivalent PR for the i3 so I figured I may as well add PHA support here, too.

### DIFF
--- a/src/common/client_response.cpp
+++ b/src/common/client_response.cpp
@@ -38,7 +38,8 @@ const PhaseResponses ClientResponses::LoadUnloadResponses[CountPhases<PhasesLoad
 const PhaseResponses ClientResponses::PreheatResponses[CountPhases<PhasesPreheat>()] = {
     {}, //_first
     { Response::Abort, Response::Cooldown, Response::PLA, Response::PETG,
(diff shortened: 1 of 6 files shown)


pbozeman:
Is there any desire from prusa to take this pull request?  I've been almost exclusively using PHA because its better biodegradability and lack of microplastics compared to PLA and friends.   

Please accept this, or do your own PHA support.   Thanks!!  :)

(Note: I made my own custom build with a variant of this patch and it's kinda lame that I had to void my warranty by breaking my buddy board appendix to allow flashing unsigned fw.  I totally get not allowing unsigned fw, I'm referring more to needing to void my warranty to use PHA without all the "mismatched filament" warnings and proper load/unload support.)

danopernis:
@dhiltonp @pbozeman Thanks for your contribution and sorry for taking so much time to respond.

There is a better alternative proposed by @bkerler in #3657 which we are planning to eventually merge.

PR #2216, opened 2022-07-09
Connect: fix compilation after gcc upgrade

vorner:


vorner:
Already merged as part of #2217

PR #2217, opened 2022-07-09
PL: Less remove calls

vorner:


laloch (approved):
Looks and works just fine. Thanks!

PR #2218, opened 2022-07-09
PrusaLink: Avoid unnecessary calls to remove

vorner:
Previously, every instance in a move-chain of GcodeUpload attempted to
remove() the temporary file. That failed, because it was held open and
alive by the newer (moved-to) instance, but it was still bothering the
USB storage, trying to delete the (still open) file.

Track if we are the "living" instance or just an empty shell we moved
from and try to close or delete things only in the last instance that
really dies, not moves.


laloch (approved):
Looks and works just fine. Thanks!

PR #2223, opened 2022-07-11
ESP: Show wifi link status on dev

vorner:
Show some more info about the internal / link state of the wifi in the
menu. Only accessible in dev builds („green“ menu).


vorner:
For the record: I've merged it as it is for now (dev only), will talk with relevant „UI“ people if they want it in non-dev. In such case, I'd add translation and make it normal „white“ item.

PR #2256, opened 2022-07-18
Fix Footer item axis value range

MistrKernnunos:


vorner (commented):
I don't see anything particularly wrong with the new code (except that it is somewhat more involved and wordy than would seem necessary). But I don't see what is being fixed in the old code, what was wrong there and why it needs what looks like a bigger rewrite. Can you elaborate (optimally, inside the commit message) and link the relevant Jira issue?

PR #2257, opened 2022-07-18
Workaround for corrupted transmitted packets

vorner:
Sometimes, if we try to send small-ish packet and the ESP is up and
running, the packet gets mangled (repeatedly; no retransmit is accepted
because it has wrong checksum). Our buffer is, however, intact and
correct, so it happens somewhere else. This happens on ethernet too (but
only if ESP is present).

This masks the problem temporarily, further investigation needed.

BFW-2664.


laloch (approved):
This is creepy.

PR #2258, opened 2022-07-18
PrusaLink: Don't hard-cache index.html

vorner:
If the printer is re-flashed, we want the browser to load the new page,
not the old one. The old one could still refer to no longer existing css
and JS files ‒ these return 404 and make the page unusable.

Note that we still _do_ use the etag/if-none-match.

BFW-2666


laloch (approved):
:+1: Thanks @vorner.

PR #2265, opened 2022-07-20
Alternative faster text rendering

michalrudolf:
BFW-2212
Text is no longer sent to display char by char, it loads as many chars as possible to buffer and then it sends them all at once. This is reducing SPI overhead and making text rendering a little bit faster.

When there are too many characters for one package, it divides them to as big doses as possible (display buffer has capacity for about 16-17 chars)

Multiline rendering is taken care of by textWrapper, which seperates lines and makes it a single line problem.


RadekVana (commented):
Fix those unit tests and I'll give you approve

RadekVana (approved):

PR #2271, opened 2022-07-21
nhttp: Reset timers after a step

vorner:
Because the step can take a long time (due to USB write) and could
trigger false "inactivity" timeout.


laloch (approved):
Perfect! Thank's @voner.

PR #2273, opened 2022-07-21
Connect render json

vorner:
Refactor the sending code to use json renderer instead of a bunch of `snprint` calls. Use that to make buffers smaller and to allow conditionals inside the rendering.

Use that two enrich the data being sent.


src/Connect/core_interface.cpp
```diff
@@ -121,16 +121,16 @@ device_params_t core_interface::get_data() {
         switch (marlin_vars->print_state) {
```
TadeasPilar, line 121: Maybe split this function into two? Separate the marlin->connect state mapping from copying of variables.

PR #2278, opened 2022-07-24
Prusalink

RadekVana:
BFW-2700


src/gui/screen_print_preview.hpp
```diff
@@ -8,26 +8,42 @@
...
-    window_icon_button_t back_button;
-    window_text_t back_label;
-    WindowPreviewThumbnail thumbnail;
-    GCodeInfo &gcode;
 
+    GCodeInfo &gcode;
     GCodeInfoWithDescription gcode_description; //cannot be first
+    WindowPreviewThumbnail thumbnail;
+
+    RadioButton radio;
```
vorner, line 29: nitpick: The name doesn't really tell much about what it is used for :-(

vorner (commented):
Some minor points inline.

Overall, I don't have any serious concerns, but it's not code I'd be very familiar with, so something might be hiding there.

I wonder about one thing, though. The questions still seem to be part of the PrintPreview state. From the point of view of PrusaLink, it would be great if:
* When we are showing the actual preview, the printer would still not have a printing state and a new print can be started.
* When we are asking about wrong kind of filament, the PrusaLink probably should already show as „printing“ and _not_ allow starting a new print.

I'm not sure I paid close enough attention, so I don't know how exactly it behaves in that state. Can you elaborate if it does this, or something different?

vorner (approved):
As agreed OOB, let's merge it as it is. The distinction between preview with just a picture and „preview“ asking if the filament is correct is not easy to do, so let's merge it as it is (with the printer showing as „idle“ the whole time) and then discuss if we want to improve it further at this time or sometime later.

PR #2281, opened 2022-07-24
Enable Link Time Optimization

DRracer:
Proof of concept - code size drops by ~31KB

Placement of `set_target_properties(firmware PROPERTIES INTERPROCEDURAL_OPTIMIZATION True)` in CMakeLists.txt is a matter of further discussion.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -303,7 +303,6 @@ add_subdirectory(lib)
 #
 
 add_executable(firmware)
-
 add_subdirectory(src)
 
 if(WUI)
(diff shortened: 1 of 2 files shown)


DRracer:
rebased onto current master

CMakeLists.txt
```diff
@@ -303,7 +303,7 @@ add_subdirectory(lib)
 #
 
 add_executable(firmware)
-
+set_target_properties(firmware PROPERTIES INTERPROCEDURAL_OPTIMIZATION True)
```
dragomirecky, line 306: /very minor/
Can we move this a bit lower in the CMakeLists? I would put it under `# generate linker map file`

dragomirecky (approved):

dragomirecky (commented):
And I guess LTO should be enabled in Release only, right?

CMakeLists.txt
DRracer: I'd put that line under `add_compile_options(-Os)` - that only for the release firmware builds. Does that sound right?

vorner (approved):

PR #2206, opened 2022-06-30
USB reliability

laloch:
Improves USB storage reliability in case of concurrent access and/or slow media.

PR #2287, opened 2022-07-26
ESP: fix broken FW version

laloch:
Fixes the previous commit, where I somehow managed to commit some ESP firmware binary of unknown origin and unknown quality. Sorry folks.


DRracer (commented):
:thinking: how am I supposed to verify the binary file? Do we need this for #2206?

laloch:
@DRracer, yes you are completely right. We need to automate the process and make sure that this faux pas will never happen again.

vorner (approved):
I don't think we are able to verify it right now or come up with proper automation. On the other hand, we need it to reflect the reality. Let's rely on testing, but we should start talking about the automation soon-ish.

Especially because this is not the only stuff with this issue, the PrusaLink content is similar ‒ manually built and commited :-(.

laloch:
This has been already merged in scope of #2206.

PR #2308, opened 2022-07-28
Connect status

vorner:


src/Connect/connect.cpp
```diff
@@ -39,6 +40,30 @@ namespace con {
 
 namespace {
 
+    std::atomic<OnlineStatus> last_known_status = OnlineStatus::Unknown;
+
+    optional<OnlineStatus> err_to_status(const Error error) {
```
RadekVana, line 45: why does it have `optional `return value and return `OnlineStatus::Unknown;`

vorner: You're correct. That must have been some leftover from figuring out the exact stuff as I went 😇. Fixed.

RadekVana (approved):

PR #2339, opened 2022-08-01
WUI: Enable NTP over WiFi

laloch:


vorner (approved):
Looks good.

Possible improvements, if you want:
* Store them as bools, eg `bool eth = netdev_get_status(...) == NETDEV_NETIF_UP` and use them in the conditions this way.
* Considering the primary interface instead of both (but this might work better in practice ‒ I believe the printer will use the other interface anyway if the primary is not active).

Anyway, feel free to merge in this way too.

laloch:
Thanks @vorner! Sorry for not reacting to your review comment. I hit the merge button faster than I read.

PR #2352, opened 2022-08-03
Fix end of print and reprint

RadekVana:
BFW-2730
BFW-2731

### DIFF
--- a/src/common/media.cpp
+++ b/src/common/media.cpp
@@ -412,11 +412,11 @@ void media_loop(void) {
         GCodeFilter::State state;
         char *gcode = gcode_filter.nextGcode(&state);
 
-        if (state == GCodeFilter::State::Timeout) {
+        switch (state) {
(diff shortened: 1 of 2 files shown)


src/common/media.cpp
```diff
@@ -426,42 +426,41 @@ void media_loop(void) {
...
+        case GCodeFilter::State::Eof:
+            // Stop print on EOF
+            // TODO: this is incorrect. We need to wait until the queue is drained before we can stop
+            media_print_stop();
+            return;
+        case GCodeFilter::State::Ok:
+            if (gcode == NULL || gcode[0] == '\0') {
+                // Nothing to process, continue to the next G-Code
+                break;
+            }
```
laloch, line 439: I don't think this can ever happen.

laloch (approved):
Looks fine to me. I only have the one comment, but it's not really that important.

PR #2353, opened 2022-08-04
Connect: Proper fingerprint computation

vorner:
* Generalize the printer code to be able to produce other similar
  hashes.
* Compute the long fingerprint (that doesn't contain appendix and
  signature bits).
* Send the long/short variant in INFO and header.

NOTE: We are not 100% sure this is the final version of the fingerprint
computation, we just need something to have in hand for further
discussion.

BFW-2657.

### DIFF
--- a/src/Connect/core_interface.cpp
+++ b/src/Connect/core_interface.cpp
@@ -229,13 +229,8 @@ printer_info_t core_interface::get_printer_info() {
         synthetic_serial(info.serial_number);
     }
 
-    // Prusa connect requires 16 long fingerprint. Printer code is 8 characters.
-    // Copy printer code behind itself to make it 16 characters long.
-    //FIXME! This is just a temporary solution!
-    printerCode(info.fingerprint); //Compute hash(8 bytes) from uid, serial and mac
-    assert(strlen(info.fingerprint) == 8);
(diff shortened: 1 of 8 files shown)


src/Connect/core_interface.cpp
```diff
@@ -229,13 +229,8 @@ printer_info_t core_interface::get_printer_info() {
...
     }
 
-    // Prusa connect requires 16 long fingerprint. Printer code is 8 characters.
-    // Copy printer code behind itself to make it 16 characters long.
-    //FIXME! This is just a temporary solution!
-    printerCode(info.fingerprint); //Compute hash(8 bytes) from uid, serial and mac
-    assert(strlen(info.fingerprint) == 8);
-    memcpy(info.fingerprint + FINGERPRINT_SIZE / 2, info.fingerprint, FINGERPRINT_SIZE / 2);
-    info.fingerprint[FINGERPRINT_SIZE] = 0;
+    printerHash(info.fingerprint, sizeof info.fingerprint - 1, false);
```
MistrKernnunos, line 232: Would be nicer to use  `sizeof(info.fingerprint) - 1` , to keep the code uniform.

vorner: Yes, sure, thanks for pointing it out :-)

MistrKernnunos (approved):
👌🏻

PR #2360, opened 2022-08-05
connect: Survive server consistently refusing messages

vorner:
Throw away messages the server doesn't "want". Select only few statuses
that'll be retried ‒ generally, timeout-related and rate-limit ones.
These are expected to go away eventually and deliver the message.

Network failures (eg. no status from server received at all) are still
retried, this is only about server providing some unusual error status
code.

Prevents infinite loop of trying to ram a message into the server's
throat while it consistently refuses to accept it, which effectively
blocked the communication until restart.

BFW-2678.


MistrKernnunos (approved):
👌🏻

PR #2363, opened 2022-08-05
Gcode upload: Shorten returned paths

vorner:
After uploading (and starting the print), ensure that the returned paths
are the short ones.

* We want to make sure we fit all the buffers once someone sends them
  back to us.
* We want to get rid of all "weird" characters in the paths (like
  spaces), because we don't do proper percent-decoding yet.

Fix some buffer lengths on the way.


laloch (approved):
Looks good. We should, however, implement proper URL decoding.

vorner:
> We should, however, implement proper URL decoding.

Fully agreed. Especially once we switch to uploading by PUT, it'll be needed. There's a ticket and I hope to get to it soon-ish.

PR #2369, opened 2022-08-08
Add footer setting migration

MistrKernnunos:
BFW-2718

### DIFF
--- a/src/common/eeprom_current.hpp
+++ b/src/common/eeprom_current.hpp
@@ -7,6 +7,7 @@
  */
 
 #include "eeprom_v10.hpp"
+#include "footer_eeprom.hpp"
 
 namespace eeprom::current {
 
@@ -15,8 +16,9 @@ namespace eeprom::current {
 #pragma pack(1)
 
 /**
- * @brief body of eeprom v10
+ * @brief body of eeprom v11
  * without head, padding and crc
+ * eeprom setting storage has changed to using 5 bits pre item and last 3 bits are ones to force default footer setting on FW < 3.3.4
  */
 struct vars_body_t : public eeprom::v10::vars_body_t {
     uint8_t EEVAR_ACTIVE_NETDEV;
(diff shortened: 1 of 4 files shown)


src/common/eeprom_current.hpp
```diff
@@ -60,6 +61,10 @@ inline vars_body_t convert(const eeprom::v10::vars_body_t &src) {
 
     // copy entire v10 struct
     memcpy(&ret, &src, sizeof(eeprom::v10::vars_body_t));
+    // count of eeprom items in last eeprom version is 6
```
RadekVana, line 64: why is not this in footer_eeprom like in master branch?

src/common/footer_eeprom.cpp
```diff
@@ -112,3 +112,25 @@ draw_zero_t footer::eeprom::GetItemDrawZero() {
...
+    for (size_t i = 0; i < count; ++i) {
+        uint32_t decoded = encoded & mask;
+        if (decoded > size_t(items::count_))
+            return footer::DefaultItems; // data corrupted, return default setting
+        ret[i] = items(decoded);
+        encoded >>= min_bit_size;
+    }
+    return ret;
+}
+/**
```
RadekVana, line 128: remove duplicit doxygen

src/common/footer_eeprom.hpp
```diff
@@ -112,24 +112,13 @@ constexpr uint32_t Encode(record rec) {
     return ret;
 }
 
+record DecodeWithSize(uint32_t encoded, size_t min_bit_size);
 /**
```
RadekVana, line 116: * @param encoded - encoded value stored in flash
 * @return record

RadekVana (changes requested):

RadekVana (changes requested):
src\common\eeprom_current.hpp
has bug in comment (not your fault)
`* @brief body of eeprom v10`
and it should be
`* @brief body of eeprom v11`
please add to this comment description how was storing of footer changed

RadekVana (approved):

PR #2380, opened 2022-08-09
First layer calibration

RadekVana:
BFW-2732


lib/Marlin/Marlin/src/gcode/bedlevel/abl/G29.cpp
```diff
@@ -990,4 +990,619 @@ G29_TYPE GcodeSuite::G29() {
   G29_RETURN(isnan(measured_z));
 }
 
+/**
+ * @brief copy of G29 without parser 
+ * 
+ * @return G29_TYPE 
+ */
+void GcodeSuite::G29_parameterless() {
```
mkbel, line 998: It would be impossible to maintain and update copied function. Suggest inject "G29" instead.

PR #2399, opened 2022-08-15
Fix fingerprint computation

vorner:
* Use actual serial number, as intended in comments, and not some random
  uninitialized bytes.
* Include the appendix bit and the signature bit in the produced output.

BFW-2657

### DIFF
--- a/src/common/support_utils.cpp
+++ b/src/common/support_utils.cpp
@@ -20,6 +20,7 @@
 static constexpr char INFO_URL_LONG_PREFIX[] = "HTTPS://HELP.PRUSA3D.COM";
 static constexpr char ERROR_URL_LONG_PREFIX[] = "HTTPS://HELP.PRUSA3D.COM";
 static constexpr char ERROR_URL_SHORT_PREFIX[] = "help.prusa3d.com";
+static constexpr char SERIAL_PREFIX[] = "CZPX";
 
 /// FIXME same code in support_utils_lib
 /// but linker cannot find it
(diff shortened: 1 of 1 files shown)


src/common/support_utils.cpp
```diff
@@ -38,11 +38,11 @@ void printerHash(char *str, size_t size, bool state_prefix) {
     uint8_t toHash[bufferSize];
     /// CPU ID
     memcpy(toHash, (char *)OTP_STM32_UUID_ADDR, OTP_STM32_UUID_SIZE);
-    //snprintf((char *)toHash, buffer, "/%08lX%08lX%08lX", *(uint32_t *)(OTP_STM32_UUID_ADDR), *(uint32_t *)(OTP_STM32_UUID_ADDR + sizeof(uint32_t)), *(uint32_t *)(OTP_STM32_UUID_ADDR + 2 * sizeof(uint32_t)));
     /// MAC
     memcpy(&toHash[OTP_STM32_UUID_SIZE], (char *)OTP_MAC_ADDRESS_ADDR, OTP_MAC_ADDRESS_SIZE);
     /// SN
-    memcpy(&toHash[OTP_STM32_UUID_SIZE + OTP_MAC_ADDRESS_SIZE], (char *)OTP_SERIAL_NUMBER_ADDR, SNSize);
+    memcpy(&toHash[OTP_STM32_UUID_SIZE + OTP_MAC_ADDRESS_SIZE], "CZPX", 4);
+    memcpy(&toHash[OTP_STM32_UUID_SIZE + OTP_MAC_ADDRESS_SIZE + 4], (char *)OTP_SERIAL_NUMBER_ADDR, SNSize - 4);
```
MistrKernnunos, line 45: The CZPX wont change in future, but it would be nicer to use constant for its length.

vorner:
How about now?

MistrKernnunos (approved):

PR #2410, opened 2022-08-17
Gcode upload: Enable upload by PUT

HelclM:
We can handle PUT requests with gcode and store it to USB.

### DIFF
--- a/lib/WUI/link_content/prusa_link_api.cpp
+++ b/lib/WUI/link_content/prusa_link_api.cpp
@@ -64,6 +64,19 @@ optional<ConnectionState> PrusaLinkApi::accept(const RequestParser &parser) cons
         return get_only(StatusPage(Status::NoContent, parser.status_page_handling(), parser.accepts_json));
     } else if (suffix == "settings") {
         return get_only(SendStaticMemory("{\"printer\": {}}", http::ContentType::ApplicationJson, parser.can_keep_alive()));
+    } else if (remove_prefix(suffix, "v1/files").has_value()){
+        static const auto prefix = "/api/v1/files/usb";
+        static const size_t prefix_len = strlen(prefix);
+        char fname[FILE_PATH_BUFFER_LEN + prefix_len];
+        if (!parser.uri_filename(fname, sizeof fname))
+            return StatusPage(Status::NotFound, parser.status_page_handling(), parser.accepts_json);
+
+        static const auto prefix_storage = "/api/v1/files";
+        static const size_t prefix_storage_len = strlen(prefix_storage);
+        char* storage_path = &fname[0] + prefix_storage_len;
+        auto upload = GcodeUpload::start(parser, wui_uploaded_gcode, parser.accepts_json, storage_path, parser.print_after_upload);
+        return std::visit([](auto upload) -> ConnectionState { return std::move(upload); }, std::move(upload));
+
     } else if (remove_prefix(suffix, "files").has_value()) {
         // Note: The check for boundary is a bit of a hack. We probably should
         // be more thorough in the parser and extract the actual content type.
--- a/lib/WUI/link_content/prusa_link_api.cpp
+++ b/lib/WUI/link_content/prusa_link_api.cpp
@@ -86,11 +86,10 @@ optional<ConnectionState> PrusaLinkApi::accept(const RequestParser &parser) cons
                 return StatusPage(Status::Forbidden, parser.status_page_handling(), parser.accepts_json);
             }
 
-            GcodeUpload::UploadParams uploadParams = GcodeUpload::PutParams();
-            GcodeUpload::PutParams &putParams = std::get<0>(uploadParams);
-            std::get<0>(uploadParams).print_after_upload = parser.print_after_upload;
+            GcodeUpload::PutParams putParams;
+            putParams.print_after_upload = parser.print_after_upload;
             strlcpy(putParams.filepath.data(), storage_path, sizeof(putParams.filepath));
-            auto upload = GcodeUpload::start(parser, wui_uploaded_gcode, parser.accepts_json, move(uploadParams));
+            auto upload = GcodeUpload::start(parser, wui_uploaded_gcode, parser.accepts_json, std::move(putParams));
             return std::visit([](auto upload) -> ConnectionState { return std::move(upload); }, std::move(upload));
         } else {
             return StatusPage(Status::MethodNotAllowed, StatusPage::CloseHandling::ErrorClose, parser.accepts_json);
@@ -107,8 +106,8 @@ optional<ConnectionState> PrusaLinkApi::accept(const RequestParser &parser) cons
             char boundary_cstr[boundary.size() + 1];
             memcpy(boundary_cstr, boundary.begin(), boundary.size());
             boundary_cstr[boundary.size()] = '\0';
-            GcodeUpload::UploadParams uploadState = printer::UploadState(boundary_cstr);
-            auto upload = GcodeUpload::start(parser, wui_uploaded_gcode, parser.accepts_json, move(uploadState));
+            printer::UploadState uploadState(boundary_cstr);
+            auto upload = GcodeUpload::start(parser, wui_uploaded_gcode, parser.accepts_json, std::move(uploadState));
             /*
              * So, we have a "smaller" variant (eg. variant<A, B, C>) and
              * want a "bigger" variant<A, B, C, D, E>. C++ templates can't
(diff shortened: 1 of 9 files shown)


lib/WUI/link_content/prusa_link_api.cpp
```diff
@@ -64,6 +64,19 @@ optional<ConnectionState> PrusaLinkApi::accept(const RequestParser &parser) cons
...
+        static const size_t prefix_len = strlen(prefix);
+        char fname[FILE_PATH_BUFFER_LEN + prefix_len];
+        if (!parser.uri_filename(fname, sizeof fname))
+            return StatusPage(Status::NotFound, parser.status_page_handling(), parser.accepts_json);
+
+        static const auto prefix_storage = "/api/v1/files";
+        static const size_t prefix_storage_len = strlen(prefix_storage);
+        char* storage_path = &fname[0] + prefix_storage_len;
+        auto upload = GcodeUpload::start(parser, wui_uploaded_gcode, parser.accepts_json, storage_path, parser.print_after_upload);
+        return std::visit([](auto upload) -> ConnectionState { return std::move(upload); }, std::move(upload));
```
vorner, line 78: I've noticed few things around here, mostly coming from assumption that the input would be correct / benign. This comes over the network, so we can't assume any such thing 😇

* I suspect this'll allow uploading to other places on the printer than just `/usb`.
* The `&fname[0] + ...` may point outside of the file name (not out of the buffer). There isn't much guarantee of what's behind the `\0` in there, so anyone calling a `strlen` or anything might _then_ run out of the buffer.
* That construct is probably more complex than necessary, that's the same as `fname + prefix_storage_len`.
* Conceptually, this is the right place to check for the `PUT` method. The selectors decide _what_ gets done, the `GcodeUpload` is responsible for doing it, not for choosing it. Once we want to support the `HEAD`, it'll be decided here too.

lib/WUI/nhttp/gcode_upload.cpp
```diff
@@ -144,41 +155,60 @@ GcodeUpload::UploadResult GcodeUpload::start(const RequestParser &parser, Upload
...
+        boundary_cstr[boundary.size()] = '\0';
+        UploadState uploader(boundary_cstr);
 
-    return GcodeUpload(move(uploader), json_errors, *parser.content_length, file_idx, move(file), uploaded);
+        if (const auto err = uploader.get_error(); get<0>(err) != Status::Ok) {
+            return StatusPage(get<0>(err), StatusPage::CloseHandling::ErrorClose, json_errors, get<1>(err));
+        }
+
+        return GcodeUpload(move(uploader), json_errors, *parser.content_length, file_idx, move(file), uploaded);
+    } else /*(parser.method == http::Method::Put)*/ {
```
vorner, line 175: Is it? Will I get here even if I send another method? I didn't see anything checking for PUT anywhere on the path.

lib/WUI/nhttp/gcode_upload.cpp
```diff
@@ -189,6 +219,39 @@ Step GcodeUpload::step(string_view input, bool terminated_by_client, uint8_t *,
     }
 }
 
+Step GcodeUpload::stepPUT(string_view input) {
+    const size_t read = std::min(input.size(), size_rest);
+    // remove the "/usb/" prefix
+    const char* filename = filepath + USB_MOUNT_POINT_LENGTH;
+
+    // check only once
+    static bool filename_checked = false;
```
vorner, line 228: So, check only on the first ever file that gets uploaded to the printer during its runtime? :-O

What exactly are you trying to solve here? The fact that the step might be called multiple times? Don't you want to put the check at the beginning, somewhere in the `start` static method, which is full of other one-time checks?

If at least possible, avoid any kind of "globals" unless they are conceptually needed.

lib/WUI/nhttp/gcode_upload.cpp
```diff
@@ -189,6 +219,39 @@ Step GcodeUpload::step(string_view input, bool terminated_by_client, uint8_t *,
...
+    if (not filename_checked){
```
vorner, line 229: I don't think I've ever seen `not` in our code. Does it actually _work_?

lib/WUI/nhttp/gcode_upload.cpp
```diff
@@ -189,6 +219,39 @@ Step GcodeUpload::step(string_view input, bool terminated_by_client, uint8_t *,
...
+        filename_checked = true;
+    }
+
+    auto error = data(input.substr(0, read));
+    if (std::get<0>(error) != Status::Ok)
+        return { read, 0, StatusPage(std::get<0>(error), StatusPage::CloseHandling::ErrorClose, json_errors, std::get<1>(error)) };
+
+    size_rest -= read;
+    if (size_rest == 0){
+        filename_checked = false;
```
vorner, line 242: You (currently wrongly) assume there's only one upload running at a time.

lib/WUI/nhttp/gcode_upload.cpp
```diff
@@ -189,6 +219,39 @@ Step GcodeUpload::step(string_view input, bool terminated_by_client, uint8_t *,
...
+        // remove the "/usb/" prefix
```
vorner, line 243: What does this comment refer to?

lib/WUI/nhttp/gcode_upload.h
```diff
@@ -46,14 +46,22 @@ class GcodeUpload final : private UploadHooks {
     virtual Result finish(const char *final_filename, bool start_print) override;
     virtual Result check_filename(const char *filename) const override;
 
+    // filepath in case of PUT (we know it from the start)
+    char filepath [FILE_PATH_BUFFER_LEN];
```
vorner, line 50: The filepath is somewhat big thing. As we have either the uploader or the filepath, we could put these into a std::variant to save valuable space.

Then we'll also be able to dispatch between them by `std::visit` and an overloaded method instead of `stepPOST` and `stepPUT`.

lib/WUI/nhttp/gcode_upload.cpp
```diff
@@ -47,16 +47,29 @@ GcodeUpload::GcodeUpload(UploadState uploader, bool json_errors, size_t length,
...
 
+GcodeUpload::GcodeUpload(bool json_errors, size_t length, size_t upload_idx, unique_file_ptr file, UploadedNotify *uploaded, const char* filepath_other, bool print_after_upload)
+    : uploaded_notify(uploaded)
+    , size_rest(length)
+    , json_errors(json_errors)
+    , cleanup_temp_file(true)
+    , tmp_upload_file(move(file))
+    , file_idx(upload_idx)
+    , print_after_upload(print_after_upload){
+    strcpy(this->filepath, filepath_other);
```
vorner, line 58: According to the header file, the `filepath_other` may be `nullptr`. I don't think that's valid.

lib/WUI/nhttp/gcode_upload.h
```diff
@@ -46,14 +46,22 @@ class GcodeUpload final : private UploadHooks {
...
     GcodeUpload(UploadState uploader, bool json_errors, size_t length, size_t upload_idx, unique_file_ptr file, UploadedNotify *uploaded);
+    GcodeUpload(bool json_errors, size_t length, size_t upload_idx, unique_file_ptr file, UploadedNotify *uploaded, const char* filepath, bool print_after_upload);
 
 public:
     bool want_read() const { return size_rest > 0; }
     bool want_write() const { return false; }
     handler::Step step(std::string_view input, bool terminated_by_client, uint8_t *output, size_t output_size);
     using UploadResult = std::variant<handler::StatusPage, GcodeUpload>;
-    static UploadResult start(const handler::RequestParser &parser, UploadedNotify *uploaded, bool json_errors);
+    static UploadResult start(const handler::RequestParser &parser, UploadedNotify *uploaded, bool json_errors, const char* filename = nullptr, bool print_after_upload = false);
```
vorner, line 64: Do I understand correctly that in case of `POST`, you (implicitly) set `print_after_upload` to `false`, but still may print after upload? That sounds like something a bit confusing. The situation there isn't „We don't want to print after upload“, but „We don't know if we want to print after upload or not“.

Furthermore, considering we want to add more such flags (eg. `overwrite`), I'm also not entirely sure adding a parameter is the right way or if we should just pull it out of the parser that's already passed in… similarly to pulling out the boundary.

vorner (changes requested):
The big picture seems to be in line with what we talked about. Nevertheless, there are some (more or less) important details that still need some attention.

Also, the CI doesn't pass. It complains about formatting ‒ which suggests you didn't install the git hook as recommended by the README.

vorner (commented):
I have only few minor nitpicks.

Also, before merging, this would be great if we didn't have such „prettier upload…“… can you squash these together?

vorner (approved):
Cool, let's get it in :-)

PR #2426, opened 2022-08-22
log: Unify cpp logging with c logging

Derevin:
Logging in c++ always suffered from having to `log_component_find("component")`. This PR adjusts logging so that instead of `log_component_find` logging in c++ has the 'normal' interface (`log_info(component, "text")`, ...), with the addition that the component needs to be referenced at the top of the file with `LOG_COMPONENT_REF(Component)`

### DIFF
--- a/src/common/selftest/selftest_axis.cpp
+++ b/src/common/selftest/selftest_axis.cpp
@@ -13,6 +13,7 @@
 #include "algorithm_scale.hpp"
 
 using namespace selftest;
+LOG_COMPONENT_REF(Selftest);
 
 static const char AxisLetter[] = { 'X', 'Y', 'Z', 'E' };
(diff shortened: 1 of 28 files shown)


src/common/selftest/selftest_axis.cpp
```diff
@@ -110,7 +111,7 @@ void CSelftestPart_Axis::sg_sample_cb(uint8_t axis, uint16_t sg) {
 }
 
 void CSelftestPart_Axis::sg_sample(uint16_t sg) {
-    int32_t pos = stepper.position((AxisEnum)config.axis);
+    [[maybe_unused]] int32_t pos = stepper.position((AxisEnum)config.axis);
```
LukasLendvorsky, line 114: Neni to spíš na ostranění?

LukasLendvorsky (approved):

LukasLendvorsky (approved):

PR #2429, opened 2022-08-23
SEND_FILE_INFO support for connect

vorner:
There's some amount of cleanup and refactoring needed, mostly to make sure we can share the same path buffer for various purposes (will be also used in future for more things).

The amount of data sent from there is somewhat limited, but it prepares the place where we'll put all the other information once we take the time to extract it.


src/Connect/run.hpp
```diff
@@ -0,0 +1,7 @@
+#pragma once
+
+namespace con {
+
+void run() __attribute__((noreturn));
```
TadeasPilar, line 5: Run what? I think this deserves better name, or at least comment.

src/Connect/buffer.hpp
```diff
@@ -0,0 +1,98 @@
...
+    public:
+        Borrow(const Borrow &) = delete;
+        Borrow &operator=(const Borrow &) = delete;
+        Borrow(Borrow &&other)
+            : buff(other.buff) {
+            other.buff = nullptr;
+        }
+        Borrow &operator=(Borrow &&other) {
+            if (buff != nullptr) {
+                assert(buff->borrowed);
```
TadeasPilar, line 37: Where are all these aserts implemented? Are we sure this resoults in some form of error screen?

src/Connect/printer.hpp
```diff
@@ -0,0 +1,101 @@
...
+        char fingerprint[FINGERPRINT_BUFF_LEN];
+    };
+
+    enum class DeviceState {
+        Unknown,
+        Idle,
+        Printing,
+        Paused,
+        Finished,
+        Prepared,
```
TadeasPilar, line 31: State "Prepared" was renamed to "Ready" in connect. Maybe we should do the same in FW to avoid confusion.

src/Connect/buffer.hpp
vorner: I'm 100% sure that asserts:

* Don't result in any kind of error screen, the thread (unfortunately, only that one thread) just stops. That's somewhat less useful than what it could do, but it does make the developer go and investigate and the debugger points very clearly at the culprit (unlike any kind of `bsod`, which just resets the printer and loses all the valuable stack trace, so `bsod` is even less useful).
* Don't get into release builds. They are only part of the debug builds to help catch bugs while implementing things.

My primary intention with using asserts is, however, documentation of the intention. It's more useful (and often more precise in the information) than a comment. Given there's nothing better right now, I use at least asserts that are sub-optimal then just not have anything at all. @dragomirecky mentioned he would like to make asserts more useful in the future in some way, but I'm not sure if there was any progress or any concrete plan to do so.

src/Connect/run.hpp
vorner: It is in the `namespace con`, so it's `con::run()`. Is that still unclear? (and yes, the `con` should really be eventually renamed to `connect`). What name would you suggest? I'm not happy with java-style whole-sentence names and I don't prefer repeating a namespace or other context inside the name. So, suggestions could be:

* `con::main()`
* `con::main_loop()`

src/Connect/printer.hpp
vorner: I'm looking into the docs… and it can contain both `IDLE`, `READY` and `PREPARED`. I'm going to inquire about what each one of them mean and what the current status is before doing that and then somehow unify it.

vorner:
I'm also adding a tiny (but important) fix in there; I'll auto-squash these into the right commits once review is done.

src/Connect/printer.hpp
vorner: Done…

src/Connect/buffer.hpp
TadeasPilar: If that is the intention, it might be useful to invoke SW breakpoint in the assert, but that's beyond the scope of this PR.
Just a thought for future improvements.

src/Connect/run.hpp
TadeasPilar: Adding comment would be enough for me. I would consider  `main_loop()` easiest to understand, but that's a matter of taste.

TadeasPilar: But all other task main functions are named `..._run`. It's probably better to keep the convention. Adding comment is fine.

TadeasPilar (approved):

PR #2430, opened 2022-08-23
fix: cmake 20 property, Marlin _OX macros, LWIP socket write

Derevin:
There is a mistake in a previous PR https://github.com/prusa3d/Prusa-Firmware-Buddy/pull/2403  bumping cpp to 20. There is a property in `CMakeLists.txt` that was forgotten and needed to be set. This PR fixes that.

Fixing the property created compiler errors which can be categorized into these two categories:

1. @mkbel Error related to Marlin's `macro.h` due to `#define _O2 ...` macro, which created collisions with `ranges_algo.h` (std lib, included from `algorithms`). The fix of this error is in accordance to a commit fixing this in Marlin's repository https://github.com/MarlinFirmware/Marlin/commit/659b4172aa49d82e54a08b5ed674b3ba4ad51fb0
2. @vorner Error related to LWIP's `socket.h` due to `#define write...` macro, which created collisions with `ostream.tcc` (std lib, included from `memory`). This error is 'fixed' by commenting the problem macro out, and changing corresponding calls from `write` to `lwip_write`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,12 +39,12 @@ endif()
 # eclipse sets those variables, so lets just use them so we don't get a warning about unused
 # variables
 set(unused "${CMAKE_VERBOSE_MAKEFILE} ${CMAKE_RULE_MESSAGES}")
+set(CMAKE_CXX_STANDARD 20)
 
 # enable all warnings (well, not all, but some)
 add_compile_options(-Wall -Wsign-compare)
 add_compile_options(
-  $<$<COMPILE_LANGUAGE:CXX>:-Wno-register> $<$<COMPILE_LANGUAGE:CXX>:-std=c++20>
-  $<$<COMPILE_LANGUAGE:CXX>:-Wno-volatile>
+  $<$<COMPILE_LANGUAGE:CXX>:-Wno-register> $<$<COMPILE_LANGUAGE:CXX>:-Wno-volatile>
   )
 
 # append custom C/C++ flags
@@ -306,8 +306,6 @@ add_subdirectory(lib)
 add_executable(firmware)
 
 add_subdirectory(src)
-
-set_target_properties(firmware PROPERTIES CXX_STANDARD 17)
 
 if(WUI)
   target_compile_definitions(firmware PRIVATE BUDDY_ENABLE_WUI)
--- a/include/buddy/lwipopts.h
+++ b/include/buddy/lwipopts.h
@@ -69,7 +69,7 @@ extern "C" {
 #define LWIP_SINGLE_NETIF            0
 #define LWIP_NETIF_HOSTNAME          1
 #define LWIP_HTTPD_SUPPORT_POST      0
-#define LWIP_COMPAT_SOCKETS          1
+#define LWIP_COMPAT_SOCKETS          2
 #define LWIP_ALTCP                   0
 #define LWIP_HTTPD_DYNAMIC_FILE_READ 0
 #define LWIP_TIMERS                  1
--- a/lib/Marlin/Marlin/src/core/macros.h
+++ b/lib/Marlin/Marlin/src/core/macros.h
@@ -39,11 +39,11 @@
 #define _FORCE_INLINE_ __attribute__((__always_inline__)) __inline__
 #define  FORCE_INLINE  __attribute__((always_inline)) inline
 #define _UNUSED      __attribute__((unused))
-#define _O0          __attribute__((optimize("O0")))
-#define _Os          __attribute__((optimize("Os")))
-#define _O1          __attribute__((optimize("O1")))
-#define _O2          __attribute__((optimize("O2")))
-#define _O3          __attribute__((optimize("O3")))
+#define __O0         __attribute__((optimize("O0")))
+#define __Os         __attribute__((optimize("Os")))
+#define __O1         __attribute__((optimize("O1")))
+#define __O2         __attribute__((optimize("O2")))
+#define __O3         __attribute__((optimize("O3")))
 
 #ifndef UNUSED
   #if defined(ARDUINO_ARCH_STM32) && !defined(STM32GENERIC)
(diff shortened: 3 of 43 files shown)


mkbel:
Does c++ standard need to be set in two places?
`$<$<COMPILE_LANGUAGE:CXX>:-std=c++20`
`set_target_properties(firmware PROPERTIES CXX_STANDARD 20)`

vorner (changes requested):
I'm afraid to comment out the `write` macro. That header is a third-party library and editing it now brings a considerable risk we'll forget/lose the change once we upgrade the library to some future version. Is there a problem with the collision itself (specifically, considering we don't want to use C++ streams as being a bit expensive) or with the use of said macro? Would the changes in `src/Connect` be enough? And, is it the only use of write throughout our code?

Alternatively, we could just disable the whole section of compat socket macros and use `lwip_...` everywhere.

Derevin:
@mkbel After some testing, both lines need to be there because:
1. Removing the flag causes issues in a scenario where we include a header with 'new stuff' in a lib without it
2. Removing the property doesn't properly enable c++20 in firmware target because it defaults to gnu++17 and overrides the flag

@vorner Yes, currently it is the only place where `lwip/sockets.h` is included and its `write` is used

To elaborate on the collision it is because of:
- `src/Connect/socket.cpp` includes `"lwip/socket.h"` and later includes `<memory>` (causing the macros to be valid during the compilation of the lib)
- `memory` includes `unique_ptr` which includes `ostream` since c++20 because of `operator<<` (which was added in c++20)
- `ostream` tries to declare `__ostream_type& write(const char_type* __s, streamsize __n);` but preprocessor tries to replace the `write` with the macro, causing compiler errors.

There are two ways to solve this problem (that i thought of):
1. 'removing' the problem macro (as done in the first commit)
2. reordering it so that `<memory>` is included before `"lwip/socket.h"`

I am strongly against the second option, which left the first option to be the only one proper one. Thank you for raising the concern about updating the library. In that case I am leaning towards disabling the whole section of macros as well.

vorner:
> I am strongly against the second option

Can you elaborate on why? I mean, yes, it's kind of subtle and easy to break, but if it's just in this one file and we put a comment near the includes, it could be the least disruptive solution.

If you still prefer the disabling of the macro section, will you do it, or do you want some help from me in there?

Derevin:
> Can you elaborate on why? I mean, yes, it's kind of subtle and easy to break, but if it's just in this one file and we put a comment near the includes, it could be the least disruptive solution.

Right now it would be a 'quick and dirty' fix but having 'include order' matter is something that would be very unintuitive when trying to solve compile errors and it feels like (and most likely is) a very bad practice.

> If you still prefer the disabling of the macro section, will you do it, or do you want some help from me in there?

I can do the dirty refactoring, however I would appreciate if you could point me to the right direction of where to put `#undef LWIP_COMPAT_SOCKETS` (so far the two options i have found are "sockets.h" right before the `#if LWIP_COMPAT_SOCKETS`, or in `opt.h`, right below the `#if !defined LWIP_COMPAT_SOCKETS` check)

> Also, what's the effect on code size after the change? Because recently we had some kind of streams included somewhere and removing them freed several kBs of flash and even RAM. I think @laloch knows more about it and if this could have the reverse effect.

I have not tested changes on RAM, but these are the code sizes based on different profiles and standards:
debug noboot cpp20: overflow by 96416
debug noboot cpp17: overflow by 96640

release boot cpp20: .text 851888
release boot cpp17: .text 851712

vorner:
>  matter is something that would be very unintuitive when trying to solve compile errors and it feels like (and most likely is) a very bad practice.

Agreed it is a bad practice in general. I just feel like one may break good practice principles when there's a good enough reason to do so and the reason is well documented somewhere close by (pragmatic approach with some effort to minimize the downsides). But I'm not trying to persuade this is one of the places where it's actually worth it.

> however I would appreciate if you could point me to the right direction of where to put #undef LWIP_COMPAT_SOCKETS (so far the two options i have found are "sockets.h" right before the #if LWIP_COMPAT_SOCKETS, or in opt.h, right below the #if !defined LWIP_COMPAT_SOCKETS check)

I think we should not undef it (that would completely disable the functionality). We want to disable only the renaming and the right place is probably setting this one to 2 instead of 1:

https://github.com/prusa3d/Prusa-Firmware-Buddy/blob/master/include/buddy/lwipopts.h#L72

> I have not tested changes on RAM, but these are the code sizes based

That seems OK :+1:

Derevin:
> setting this one to 2 instead of 1:

Thank you! That is a much better solution to this problem, changing the option to 2 instead of 1 made it so that there's no more need for changes in either lib/* or src/* in regards to the whole `<memory>` problem.

I have rebased in the 'proper' fix

vorner (dismissed):
I've been confused for a while, because I expected to there'll be a need to change all the calls to lwip_write and such and wondered how this could work, but according to the docs, the lwip will now provide "real functions". I haven't tracked where they come from, but :shrug:.

So, OK from the networking part. I haven't checked the marlin parts, please wait for the other reviewer for that one 😇

Derevin:
@mkbel I have done the changes as we have discussed - set cxx20 only once but globally in CMakeLists and make required changes in Marlin.

@vorner As we have discussed, since the last review there has been a commit into master which now blocks this PR due to the introduction of the `namespace connect` which conflicts with the `lwip_connect` function. I do not know the best way forward myself, but i thought of these three options (so far):

1. Renaming the `namespace connect`
2. Renaming the conflicting function `connect`
3. Changing the option in `lwipopts.h` to 0 (essentially undefining it) and then rewriting all the uses of `lwip_` functions to use the `lwip_` prefix.

Personally the third option with renaming the functions to use the prefix feels the cleanest. What do you think?

mkbel (dismissed):

vorner:
Thinking about it, I'll go through the options:

* Renaming the `connect` function is not really an option. It's inside the LwIP library and we definitely should not mess inside it (we don't dare poke in that code too much, and we might want to update it to a newer version eventually).
* Changing all uses to `lw_*` functions. That is an option, in theory. But the use of networking is a bit widespread and the idea is we have are compatible with BSD sockets, so we can just take code in the wild that does networking. So, this option has the downside we would break this compatibility. It's not a hard blocker, but a big disadvantage.

So I think the renaming of the namespace is the least bad option. I'd suggest something like `connect_client`. Should I do a PR to prepare it, or will you include it here?

Derevin:
> So I think the renaming of the namespace is the least bad option. I'd suggest something like `connect_client`. Should I do a PR to prepare it, or will you include it here?

I think it makes sense that renaming the namespace would be a part of this PR, in a separate commit. If you're okay with the name `connect_client`, I'll go and do the renaming (so that you don't have to).

vorner (approved):

PR #2436, opened 2022-08-24
Fix print preview after file was uploaded via plink

RadekVana:
BFW-2711

### DIFF
--- a/lib/WUI/wui_api.cpp
+++ b/lib/WUI/wui_api.cpp
@@ -353,6 +353,14 @@ uint32_t wui_gcodes_uploaded() {
     return uploaded_gcodes;
 }
 
+static void set_current_file(char *filename) {
+    strlcpy(marlin_vars()->media_LFN, basename(filename), FILE_NAME_BUFFER_LEN);
+    // Turn it into the short name, to improve buffer length, avoid strange
+    // chars like spaces in it, etc.
+    get_SFN_path(filename);
+    marlin_set_current_file(filename);
+}
+
 bool wui_start_print(char *filename) {
     // Note: By checking now and starting it later, we are introducing a short
     // race condition. Doing it properly would be kind of hard and the risk is
@@ -377,6 +385,8 @@ bool wui_start_print(char *filename) {
         // chars like spaces in it, etc.
         get_SFN_path(filename);
         print_begin(filename);
+    } else {
+        set_current_file(filename);
     }
 
     return can_start_print;
@@ -388,6 +398,7 @@ bool wui_uploaded_gcode(char *filename, bool start_print) {
     if (start_print) {
         return wui_start_print(filename);
     } else {
+        set_current_file(filename);
         return true;
     }
 }
--- a/src/common/marlin_client.c
+++ b/src/common/marlin_client.c
@@ -569,6 +569,16 @@ void marlin_print_start(const char *filename) {
     _send_request_to_server_and_wait(request);
 }
 
+void marlin_set_current_file(const char *filename) {
+    char request[MARLIN_MAX_REQUEST];
+    const int len = snprintf(request, sizeof(request), "!%c%s", MARLIN_MSG_SET_CURRENT_FILE, filename);
+    if (len < 0)
+        bsod("Error formatting request.");
+    if ((size_t)len >= sizeof(request))
+        bsod("Request too long.");
+    _send_request_to_server_and_wait(request);
+}
+
 void marlin_print_abort(void) {
     _send_request_id_to_server_and_wait(MARLIN_MSG_PRINT_ABORT);
 }
--- a/src/common/marlin_server.cpp
+++ b/src/common/marlin_server.cpp
@@ -535,6 +535,16 @@ void marlin_server_print_start(const char *filename) {
     }
 }
 
+void marlin_server_set_current_file(const char *filename) {
+    if (filename == nullptr)
+        return;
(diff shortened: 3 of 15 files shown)


lib/WUI/wui_api.cpp
```diff
@@ -353,6 +353,14 @@ uint32_t wui_gcodes_uploaded() {
     return uploaded_gcodes;
 }
 
+static void set_current_file(char *filename) {
+    strlcpy(marlin_vars()->media_LFN, basename(filename), FILE_NAME_BUFFER_LEN);
```
vorner, line 357: What is the purpose here? I mean, I'm setting this only for my copy of the buffer, right?

I'd like to point out that the filename isn't guaranteed to be LFN. Actually, most of the time, it _isn't_, most of the time it's a SFN path. There's a rare occasion (starting print just after an upload) where it might be LFN and that's why it is explicitly always "shortened" in the original code.

If we want this to be LFN, we should convert it to LFN (there's a function for that direction too).

src/common/marlin_client.c
```diff
@@ -569,6 +569,16 @@ void marlin_print_start(const char *filename) {
     _send_request_to_server_and_wait(request);
 }
 
+void marlin_set_current_file(const char *filename) {
```
vorner, line 572: What happens here if this happens while printing?

Because it is possible (and desirable) to be able to upload while printing. Then the file is ready & waiting there once the job is finished. You can then just return to the home screen after cleaning the bed and it'll pop it up as the next one, avoiding several ways between the printer and the computer.

But I assume marlin already uses the "current file" while printing and changing it at that time is not the best course of action?

src/common/marlin_server.cpp
```diff
@@ -535,6 +535,16 @@ void marlin_server_print_start(const char *filename) {
     }
 }
 
+void marlin_server_set_current_file(const char *filename) {
+    if (filename == nullptr)
+        return;
+    if ((marlin_server.print_state == mpsIdle) || (marlin_server.print_state == mpsFinished) || (marlin_server.print_state == mpsAborted)) {
+        media_preselect_file(filename);
+        _set_notify_change(MARLIN_VAR_FILEPATH);
+        _set_notify_change(MARLIN_VAR_FILENAME);
```
vorner, line 544: I see. Soo, if I upload while printing, what file does "pop up" after finishing the current print? :thinking:

vorner (commented):
I like the general idea of passing the file name instead of trying to find it later by the date/time, which might be wrong (though I'd still be interested to know why that approach was glitchy and didn't work reliably). It completely sidesteps the problem and looks more "correct".

On the other hand, (ab)using the marlin's buffer to pass it through seems somehow wrong. Aside from not being elegant (the buffer is supposed to mean the gcode we are currently printing, I assume), it also _probably_ (I haven't tried) do a sane thing when the upload happens during a print; that scenario is however quite useful and I personally use that regularly, I'd assume other users would too.

If we do want to go that way (with the marlin's buffer), we probably at least should make sure to not pop up anything when returning to the home screen after a print. That way the user needs to go to the print menu instead of getting the last uploaded file, but that at least seems to be a consistent and sane behaviour, while now I suspect it would offer the file I've just finished printing and went away from the screen offering a „reprint“ button.

src/common/marlin_server.cpp
RadekVana: It would require additional buffers.

RadekVana:
> I like the general idea of passing the file name instead of trying to find it later by the date/time, which might be wrong (though I'd still be interested to know why that approach was glitchy and didn't work reliably). It completely sidesteps the problem and looks more "correct".
> 
> On the other hand, (ab)using the marlin's buffer to pass it through seems somehow wrong. Aside from not being elegant (the buffer is supposed to mean the gcode we are currently printing, I assume), it also _probably_ (I haven't tried) do a sane thing when the upload happens during a print; that scenario is however quite useful and I personally use that regularly, I'd assume other users would too.
> 
> If we do want to go that way (with the marlin's buffer), we probably at least should make sure to not pop up anything when returning to the home screen after a print. That way the user needs to go to the print menu instead of getting the last uploaded file, but that at least seems to be a consistent and sane behaviour, while now I suspect it would offer the file I've just finished printing and went away from the screen offering a „reprint“ button.

There was a bug. Picking newest file does not use the same comparation algorithm as sorting.
I have already fixed that in master (fix was not requested in release). Also slicer generates weird time stamps (modified older than created). I have some files with created timestamp in 2095 this messes up the sorting a lot ..

Those buffers are not from marlin, this is our implementation, You can insert that gcode to those media buffers only when printer is not printing. Which will try to start the print.

Sequenece
wui -> marlin client -> marlin server -> gui (print preview) -> marlin server (print)
can be much simplified in master to
wui -> marlin client -> marlin server (print preview + print)

and since master considers print preview as printing, it will be cleaner

src/gui/screen_home.cpp
```diff
@@ -232,7 +244,8 @@ void screen_home_data_t::printBtnDis() {
     w_labels[0].SetText(_(labels[labelNoUSBId]));
 }
 
-bool screen_home_data_t::moreGcodesUploaded() {
+bool screen_home_data_t::MoreGcodesUploaded() {
```
vorner, line 247: This capitalization change goes against the style guideline we agreed on. Is there a good reason for it?

vorner (approved):

PR #2438, opened 2022-08-24
Add basic bed leveling screen to Calibration menu

martin357:
- Use screen_t as ancestor and get the last marlin message localy from the screen
 - Block MenuTimeout when Mesh Bed Leveling is in progress
 - Add progressBar

BFW-2586

### DIFF
--- a/src/gui/screen_bed_leveling.cpp
+++ b/src/gui/screen_bed_leveling.cpp
@@ -1,0 +1,85 @@
+#include "screen_bed_leveling.hpp"
+#include "log.h"
+#include "ScreenHandler.hpp"
+#include "gui.hpp"
+
+ScreenBedLeveling::ScreenBedLeveling()
+    : AddSuperWindow<screen_t>()
+    , m_statusMsg(this, { 10, 128, 220, 80 }, is_multiline::no)
+    , m_progressBar(this, Rect16(10, 70, GuiDefaults::RectScreen.Width() - 2 * 10, 16))
+    , header(this)
+    , footer(this) {
+    header.SetIcon(IDR_PNG_heatbed_16px);
+    header.SetText(_("BED LEVELING"));
+    Screens::Access()->DisableMenuTimeout();
+    const uint8_t *recastPointer = reinterpret_cast<const uint8_t *>((const char *)this->m_statusBuffer);
+    this->m_statusMsg.SetText(string_view_utf8::MakeRAM(recastPointer));
+}
+
+void ScreenBedLeveling::windowEvent(EventLock /*has private ctor*/, window_t *sender, GUI_event_t event, void *param) {
+    if (event == GUI_event_t::LOOP) {
+        switch (this->state) {
+        case state_t::idle: // Start homing
+            // Start by clearing events and homing
+            marlin_error_clr(MARLIN_ERR_ProbingFailed);
+            marlin_event_clr(MARLIN_EVT_CommandBegin);
+            marlin_event_clr(MARLIN_EVT_CommandEnd);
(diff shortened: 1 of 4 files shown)


src/gui/screen_bed_leveling.cpp
```diff
@@ -0,0 +1,85 @@
...
+}
+
+void screen_bed_leveling_t::updateStatusMsg() {
+    // Update status message
+    const char *msg = MsgCircleBuffer().GetLastIfAble();
+    if (msg) {
+        const uint8_t *data = reinterpret_cast<const uint8_t *>(msg);
+        auto newText = string_view_utf8::MakeRAM(data);
+        if (newText != this->m_statusMsg.GetText()) {
+            this->m_statusMsg.SetText(newText);
```
MistrKernnunos, line 81: The data could change under our hands between the call and invalidate. 
MakeRAM does not allocate any memory, it uses the memory where the pointer points.

src/gui/screen_bed_leveling.hpp
```diff
@@ -0,0 +1,53 @@
...
+        homing,
+        leveling,
+        done,
+        error
+    };
+    state_t state { state_t::idle };
+
+    /** While an instance exists, MenuTimeout will be disabled.
+    **/
+    class MenuTimeoutBlocker {
```
MistrKernnunos, line 24: Just call `DisableMenuTimeout()` in constructor. and it will disable timeout for this screen. This object is not needed.

With `Screens::Access()` you get pinter to current screen

src/gui/screen_bed_leveling.hpp
```diff
@@ -0,0 +1,53 @@
...
+#include "gui.hpp"
+#include "status_footer.hpp"
+#include "window_text.hpp"
+#include "window_icon.hpp"
+#include "window_term.hpp"
+#include "window_progress.hpp"
+#include "screen.hpp"
+#include "ScreenHandler.hpp"
+
+class screen_bed_leveling_t : public AddSuperWindow<screen_t> {
```
MistrKernnunos, line 12: Use CamelCase for the name of the screen.

src/gui/screen_bed_leveling.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+screen_bed_leveling_t::screen_bed_leveling_t()
+    : AddSuperWindow<screen_t>()
+    , m_statusMsg(this, { 10, 128, 220, 80 }, is_multiline::no)
+    , m_progressBar(this, Rect16(10, 70, GuiDefaults::RectScreen.Width() - 2 * 10, 16))
+    , header(this)
+    , footer(this) {
+    header.SetIcon(IDR_PNG_heatbed_16px);
+    header.SetText(_("BED LEVELING"));
+    this->m_statusMsg.SetText(_(""));
```
MistrKernnunos, line 14: Text will be initialized as empty, but if you want use directly stringview insted of translating empty string.

MistrKernnunos (changes requested):

src/gui/screen_bed_leveling.cpp
martin357: This should be fixed by keeping an internal copy.

src/gui/screen_bed_leveling.hpp
martin357: Removed, switched to DisableMenuTimeout()

src/gui/screen_bed_leveling.hpp
martin357: Renamed

PR #2440, opened 2022-08-25
Fix template syntax error

vorner:
No idea why it is accepted in some circumstances, but local test
compilation fails on it.


Derevin:
I can confirm that when I was attempting upgrade to arm-gcc 11, that this same syntax error was causing compilation faults during normal builds as well.

wavexx (approved):

PR #2441, opened 2022-08-25
Skip from filament load to calibration

BlinCT:
BFW-2761


src/common/selftest/selftest_firstlayer.cpp
```diff
@@ -162,7 +162,9 @@ LoopResult CSelftestPart_FirstLayer::stateFilamentLoadWaitFinished() {
         // IPartHandler::SetFsmPhase(PhasesSelftest::);
         return LoopResult::RunNext;
     } else {
-        return LoopResult::GoToMark;
+        state_selected_by_user = StateSelectedByUser::Calib;
```
RadekVana, line 165: In case of abort we  want to return  LoopResult::GoToMark.
You should use switch.

RadekVana (changes requested):

RadekVana (approved):

PR #2443, opened 2022-08-26
Get head v1 request

HelclM:
Implementing GET and HEAD on `/api/v1/files` url according to  https://dev.prusa3d.com/browse/BFW-2787

### DIFF
--- a/lib/WUI/link_content/prusa_link_api.cpp
+++ b/lib/WUI/link_content/prusa_link_api.cpp
@@ -25,6 +25,11 @@ using nhttp::printer::FileInfo;
 using nhttp::printer::GcodeUpload;
 using nhttp::printer::JobCommand;
 
+enum class RemapPolicy {
+    Octoprint,
+    NoRemap
+};
+
 namespace {
     optional<string_view> remove_prefix(string_view input, string_view prefix) {
         if (input.size() < prefix.size() || input.substr(0, prefix.size()) != prefix) {
@@ -34,7 +39,7 @@ namespace {
         return input.substr(prefix.size());
     }
 
-    optional<ConnectionState> parse_file_url(const RequestParser &parser, const size_t prefix_len, char *filename, const size_t filename_len) {
+    optional<ConnectionState> parse_file_url(const RequestParser &parser, const size_t prefix_len, char *filename, const size_t filename_len, RemapPolicy remapPolicy) {
         /*
          * We do *not* use the uri with prefix removed. We need the safe
          * transformation into a file name (removal of query params,
@@ -47,33 +52,34 @@ namespace {
          *
          * The FileInfo handler doesn't know recursive yet either.
          */
-        char fname[FILE_PATH_BUFFER_LEN + prefix_len];
-        const char *fname_real = fname + prefix_len;
-        if (parser.uri_filename(fname, sizeof fname)) {
-            size_t len = strlen(fname);
+        const char *fname_real = filename + prefix_len;
+        if (parser.uri_filename(filename, filename_len)) {
+            size_t len = strlen(filename);
             if (prefix_len > len) {
                 return StatusPage(Status::NotFound, parser.status_page_handling(), parser.accepts_json);
             }
 
-            /*
-             * The octoprint API gives special meaning to /local and
-             * /sdcard. For us, everything lives in the USB (/usb). We
-             * remap these. Nevertheless, we never _generate_ these /local
-             * or such URIs, so we don't remap anything else but the "root".
-             */
-            static const char *const roots[] = {
-                "",
-                "/",
-                "/local",
-                "/local/",
-                "/sdcard",
-                "/sdcard/",
-            };
+            if (remapPolicy == RemapPolicy::Octoprint) {
+                /*
+                 * The octoprint API gives special meaning to /local and
+                 * /sdcard. For us, everything lives in the USB (/usb). We
+                 * remap these. Nevertheless, we never _generate_ these /local
+                 * or such URIs, so we don't remap anything else but the "root".
+                 */
+                static const char *const roots[] = {
+                    "",
+                    "/",
+                    "/local",
+                    "/local/",
+                    "/sdcard",
+                    "/sdcard/",
+                };
 
-            for (size_t i = 0; i < sizeof roots / sizeof roots[0]; i++) {
-                if (strcmp(fname_real, roots[i]) == 0) {
-                    fname_real = "/usb/";
-                    break;
+                for (size_t i = 0; i < sizeof roots / sizeof roots[0]; i++) {
+                    if (strcmp(fname_real, roots[i]) == 0) {
+                        fname_real = "/usb/";
+                        break;
+                    }
                 }
             }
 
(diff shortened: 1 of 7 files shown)


lib/WUI/link_content/prusa_link_api.cpp
```diff
@@ -33,6 +33,64 @@ namespace {
...
+             * or such URIs, so we don't remap anything else but the "root".
+             */
+            static const char *const roots[] = {
+                "",
+                "/",
+                "/local",
+                "/local/",
+                "/sdcard",
+                "/sdcard/",
+            };
```
vorner, line 71: Oops. I know it was my suggestion to re-use the code in the `/api/files`. But looking at it here, we probably don't want this particular re-mapping hack in the `/api/v1/files` ‒ one reason for creating the new API was to get rid of hardcoded `sdcard` and `local` and all the hacks related to that.

So I'd say we either want to somehow pass the list of mappings in as a parameter, or disable this part by some parameter.

lib/WUI/link_content/prusa_link_api.cpp
```diff
@@ -66,32 +124,28 @@ optional<ConnectionState> PrusaLinkApi::accept(const RequestParser &parser) cons
...
+            strlcpy(putParams.filepath.data(), filename, sizeof(putParams.filepath));
             auto upload = GcodeUpload::start(parser, wui_uploaded_gcode, parser.accepts_json, std::move(putParams));
             return std::visit([](auto upload) -> ConnectionState { return std::move(upload); }, std::move(upload));
-        } else {
+        }
+        case Method::Get: {
+            return FileInfo(filename, parser.can_keep_alive(), parser.accepts_json, false);
+        }
+        case Method::Head: {
+            return FileInfo(filename, parser.can_keep_alive(), parser.accepts_json, false, true);
```
vorner, line 146: I'd prefer to not put defaults in the parameters here; the FileInfo is called only from these two places.

Also, I'm thinking if passing the actual method (Head/Get) here would be more readable then true/false.

lib/WUI/link_content/prusa_link_api.cpp
```diff
@@ -120,88 +174,41 @@ optional<ConnectionState> PrusaLinkApi::accept(const RequestParser &parser) cons
...
-                }
-
-                /*
-                 * Now, we make 100% sure the user won't get a file that's not
-                 * on the USB drive (eg. our xflash).
-                 */
-                if (strncmp(fname_real, "/usb/", 5) != 0) {
-                    return StatusPage(Status::Forbidden, parser.status_page_handling(), parser.accepts_json);
-                }
+            char filename[FILE_PATH_BUFFER_LEN + prefix_len];
```
vorner, line 179: Am I right the parse_file_url outputs the filename here? Then it probably doesn't need the extra `prefix_len` any more?

I'm wondering about two things here, though.

* The buffer is not exactly small. And you now use two on the stack at the same time (the other one is inside the function). Would it make sense to declare that this parameter is not only output, but also the workspace?
* Alternatively, would it be cleaner to return some kind of `variant<array<char, ...>, error>`? I'm not sure if C++ guarantees some kind of return-placement optimization then or what, though, so maybe that could _also_ use two different buffers.

lib/WUI/nhttp/file_info.cpp
```diff
@@ -123,7 +125,7 @@ FileInfo::FileInfo(const char *filename, bool can_keep_alive, bool json_errors,
 Step FileInfo::step(std::string_view, bool, uint8_t *output, size_t output_size) {
     const size_t last_chunk_len = strlen(LAST_CHUNK);
     size_t written = 0;
-    const ConnectionHandling handling = can_keep_alive ? ConnectionHandling::ChunkedKeep : ConnectionHandling::Close;
+    const ConnectionHandling handling = can_keep_alive && !header_only ? ConnectionHandling::ChunkedKeep : ConnectionHandling::Close;
```
vorner, line 128: Why do you disable keep-alive on the header_only? That probably isn't desirable.

lib/WUI/nhttp/file_info.cpp
```diff
@@ -147,7 +149,22 @@ Step FileInfo::step(std::string_view, bool, uint8_t *output, size_t output_size)
...
+        static const char *extra_hdrs_no_printing[] = {
+            "Read-Only: false\r\n",
+            "Currently-Printing: false\r\n",
+            nullptr
+        };
+        static const char *extra_hdrs_printing[] = {
+            "Read-Only: false\r\n",
+            "Currently-Printing: true\r\n",
+            nullptr
+        };
```
vorner, line 162: Hmm, this doesn't look great. I think lifetime of the parameter is OK to be just during the call to `write_headers`, so you could use one static string, another buffer on stack to render the header in and have the array also on stack.

lib/WUI/wui_api.cpp
```diff
@@ -17,6 +17,7 @@
 #include "stm32f4xx_hal.h"
 #include "print_utils.hpp"
 #include "marlin_client.hpp"
+#include "marlin_client.h"
```
vorner, line 20: Wait a moment. Do we have two marlin_client headers? That seems super-weird.

lib/WUI/wui_api.cpp
```diff
@@ -404,3 +405,15 @@ bool wui_uploaded_gcode(char *filename, bool start_print) {
...
     }
 }
+
+bool wui_is_file_being_printed(const char *filename) {
+    marlin_client_loop();
+    if (!marlin_is_printing()) {
+        return false;
+    }
+
+    char SFN[FILE_PATH_BUFFER_LEN];
```
vorner, line 415: Style nitpick: `SFN` capitalization would look like a constant, not a variable.

lib/WUI/wui_api.cpp
```diff
@@ -404,3 +405,15 @@ bool wui_uploaded_gcode(char *filename, bool start_print) {
...
+    strlcpy(SFN, filename, sizeof(SFN));
+    get_SFN_path(SFN);
+    return strcmp(SFN, marlin_vars()->media_SFN_path) == 0;
```
vorner, line 418: `strcasecmp`?

lib/WUI/wui_api.cpp
HelclM: Apparently we do :man_shrugging: , but the `.h`  is included in `.hpp`, so the include here is redundant.

lib/WUI/link_content/prusa_link_api.cpp
```diff
@@ -120,88 +183,41 @@ optional<ConnectionState> PrusaLinkApi::accept(const RequestParser &parser) cons
...
+            auto error = parse_file_url(parser, prefix_len, filename, FILE_PATH_BUFFER_LEN + prefix_len, RemapPolicy::Octoprint);
```
vorner, line 189: Maybe use `sizeof(filename)` as the less error-prone alternative?

lib/WUI/nhttp/file_info.cpp
vorner: Ping? Is this a leftover?

vorner (commented):
There are few more nitpicks.

vorner (approved):
:+1:

PR #2451, opened 2022-08-28
Make error loud

RadekVana:
BFW-2793


mkbel (approved):

mkbel:
Do we really want loud sound even in cases like failed to home? I would suggest force loud sound only in cases something can be damaged by non-action - like thermal errors. In most minimalist case I would suggest force loud sound only in case of MAXTEMP.

PR #2462, opened 2022-08-30
Provide the time function

vorner:
Instead of using our own custom function to get time based on NTP,
provide it as the standard time function.

Besides being a bit cleaner as the solution, this makes mbedTLS work
without too much gymnastics and makes certificate expiration work.

Enabling the expiration checks.

BFW-2767.

### DIFF
--- a/src/buddy/fatfs.c
+++ b/src/buddy/fatfs.c
@@ -1,5 +1,6 @@
 #include "fatfs.h"
-#include <wui_api.h>
+
+#include <time.h>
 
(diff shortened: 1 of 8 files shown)


src/buddy/fatfs.c
```diff
@@ -1,5 +1,6 @@
 #include "fatfs.h"
-#include <wui_api.h>
+
+#include <time.h>
```
DRracer, line 3: :pray: much cleaner this way

DRracer (approved):
nice :+1:

PR #2463, opened 2022-08-30
static_file: Make the extra headers nicer

vorner:
Don't have two long constants around there (it wouldn't scale), make
something that would allow multiple such conditionals.

### DIFF
--- a/lib/WUI/link_content/static_file.cpp
+++ b/lib/WUI/link_content/static_file.cpp
@@ -45,16 +45,13 @@ optional<ConnectionState> StaticFile::accept(const RequestParser &parser) const 
 
     FILE *f = fopen(fname, "rb");
     if (f) {
-        static const char *extra_hdrs_cache[] = {
+        const char *extra_hdrs[] = {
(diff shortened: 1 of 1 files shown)


lib/WUI/link_content/static_file.cpp
```diff
@@ -45,16 +45,13 @@ optional<ConnectionState> StaticFile::accept(const RequestParser &parser) const
 
     FILE *f = fopen(fname, "rb");
     if (f) {
-        static const char *extra_hdrs_cache[] = {
+        const char *extra_hdrs[] = {
```
HelclM, line 48: Maybe I am missing something, but I think the static was there for a good reason, because SendFile stores only the pointer, does not do a copy, so it goes out of scope and is invalid at the time it is passed to write_headers in SendFile::step.

vorner: Ah, crap, you're right. That was the reason here.

So, I'm going to close this one, because it's broken. Nevertheless, I think this problem doesn't apply to your use case, you could still use something like this.

PR #2465, opened 2022-08-31
Support endpoints for PUT

HelclM:
Support endpoints needed for uploading gcodes by PUT request, specifically adding /api/v1/storage endpoint and enriching /api/version endpoint with upload-by-put boolean.

### DIFF
--- a/lib/WUI/link_content/basic_gets.cpp
+++ b/lib/WUI/link_content/basic_gets.cpp
@@ -6,6 +6,7 @@
 
 #include <segmented_json_macros.h>
 
+#include <dirent.h>
 #include <cstring>
 #include <cstdio>
 
@@ -149,7 +150,10 @@ JsonResult get_version(size_t resume_point, JsonOutput &output) {
         // PrusaLink API to replace this?
         JSON_FIELD_STR("server", LWIP_VERSION_STRING) JSON_COMMA;
         JSON_FIELD_STR("text", "PrusaLink MINI") JSON_COMMA;
-        JSON_FIELD_STR("hostname", hostname);
+        JSON_FIELD_STR("hostname", hostname) JSON_COMMA;
+        JSON_FIELD_OBJ("capabilities");
+            JSON_FIELD_BOOL("upload-by-put", true);
+        JSON_OBJ_END;
     JSON_OBJ_END;
     JSON_END;
     // clang-format on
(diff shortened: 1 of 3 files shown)


lib/WUI/link_content/basic_gets.cpp
```diff
@@ -249,4 +253,17 @@ JsonResult get_job(size_t resume_point, JsonOutput &output) {
...
 }
 
+JsonResult get_storage(size_t resume_point, JsonOutput &output) {
+    // Keep the indentation of the JSON in here!
+    // clang-format off
+    JSON_START;
+        JSON_OBJ_START
+            JSON_FIELD_STR("path", "/usb") JSON_COMMA;
+            JSON_FIELD_STR("type", "USB") JSON_COMMA;
+            JSON_FIELD_BOOL("available", opendir("/usb"));
```
vorner, line 263: The `opendir` is missing its closedir (+style-wise, this should probably be `opendir(..) != nullptr`, to avoid the implicit pointer->bool conversion).

I'd suggest pulling this out into a helper function (otherwise the macros around the JSON will get unhappy about a local-scoped variable) in an anonymous namespace.

As another point. Would something more lightweight work? I think `stat` for example won't allocate dynamically, unlike `opendir`, but I'm not sure if it does what we need.

vorner (changes requested):

lib/WUI/link_content/basic_gets.cpp
```diff
@@ -249,4 +253,29 @@ JsonResult get_job(size_t resume_point, JsonOutput &output) {
...
+    // but fatfs doesn't support calling it on root and from it's
+    // perspective /usb is root
+    if (DIR *dir = opendir("/usb"); dir != nullptr) {
+        available = true;
+        closedir(dir);
+    }
+    return available;
+}
+
+JsonResult get_storage(size_t resume_point, JsonOutput &output) {
```
vorner, line 268: I'm sorry to still bother with minor details. But the function is used only here, locally. Can you put it into an anonymous namespace, so it doesn't become linkable and can't collide with some other `get_storage` function somewhere else?

vorner (changes requested):

lib/WUI/link_content/basic_gets.cpp
HelclM: I don't get this, `get_storage` is used in `prusa_link_api.cpp` don't you mean the function `usb_available` above?

vorner: Yes, of course. I've attached it to the wrong function.

vorner (approved):
:+1:

PR #2469, opened 2022-08-31
Marlin client

RadekVana:
BFW-2816


DRracer:
@RadekVana what is the key benefit of this PR besides breaking all other PRs? I don't see any code safety / compilation checks improvements, but plain renaming of variables, types and enums.

src/common/marlin_server.hpp
```diff
@@ -1,16 +1,182 @@
...
+
 #pragma once
 
+#include "../../lib/Marlin/Marlin/src/inc/MarlinConfig.h"
+#include "marlin_events.h"
+#include "marlin_vars.hpp"
+#include "marlin_errors.h"
+#include "client_fsm_types.h"
+
+#include <stddef.h>
```
vorner, line 14: C++ should be using `cstddef` instead.

src/common/marlin_server.hpp
```diff
@@ -1,16 +1,182 @@
...
+static constexpr uint16_t MARLIN_SFLG_STARTED = 0x0001; // server started (set in marlin_server_init)
+static constexpr uint16_t MARLIN_SFLG_PROCESS = 0x0002; // loop processing in main thread is enabled
+static constexpr uint16_t MARLIN_SFLG_BUSY = 0x0004;    // loop is busy
+static constexpr uint16_t MARLIN_SFLG_PENDREQ = 0x0008; // pending request
+static constexpr uint16_t MARLIN_SFLG_EXCMODE = 0x0010; // exclusive mode enabled (currently used for selftest/wizard)
+
+// server variable update interval [ms]
+static constexpr uint8_t MARLIN_UPDATE_PERIOD = 100;
+
+typedef void(marlin_server_idle_t)(void);
```
vorner, line 33: The `void` in params may be omitted now?

src/common/marlin_vars.hpp
```diff
@@ -99,48 +102,48 @@ enum {
...
-    mpsPowerPanic_AwaitingResume,
-} marlin_print_state_t;
+static constexpr uint8_t MARLIN_VAR_MOTION_MSK_X = 1 << MARLIN_VAR_INDEX_X;
+static constexpr uint8_t MARLIN_VAR_MOTION_MSK_Y = 1 << MARLIN_VAR_INDEX_Y;
+static constexpr uint8_t MARLIN_VAR_MOTION_MSK_Z = 1 << MARLIN_VAR_INDEX_Z;
+static constexpr uint8_t MARLIN_VAR_MOTION_MSK_E = 1 << MARLIN_VAR_INDEX_E;
+
+static constexpr uint32_t TIME_TO_END_INVALID = (uint32_t)-1;
+
+enum class marlin_print_state_t : uint8_t {
```
vorner, line 112: When we are touching the type and moving everything to C++, I might as well poke into the naming conventions.

We kind of agreed types should be `CamelCase`. I know specifically the marlin and around it, this is historically different, so I'm not pushing too strongly for that, but when changing it to enum class and effectively enforcing everything to prefix it with the type, it sounds like a good time to fix the name.

Similarly, the prefix and suffix makes it really long in many places. The `_t` is a `C` relic and isn't usually used in C++. And, C++ can prefer to use a namespace instead of prefix ‒ for example `marlin::PrintState`.

src/common/marlin_vars.hpp
```diff
@@ -176,24 +179,23 @@ typedef enum {
 
 // variables structure - used in server and client
 // deliberately ordered from longest data types to shortest to avoid alignment issues
-typedef struct _marlin_vars_t {
+struct marlin_vars_t {
```
vorner, line 182: Again, the naming…

vorner (commented):
:question: Do we use `h` or `hpp` for C++ headers? I don't think we are unified on that.

It's not harmful, but I still have few notes.

The use of the namespace is weird. Usually, one wraps the whole C++ file into the namespace same as the header file, not this way where you `using namespace ...` and then prefix all the functions ‒ is that even universally supported? :-O

And I'm not sure about the name of the namespace. Where does `core` come from? Wouldn't something like `marlin::client` be better (yes, I'm suggesting nesting the namespaces into a tree).

src/common/marlin_vars.hpp
RadekVana: I wanted to rename this one in next PR, we already have too many changes in this one.

src/common/marlin_server.cpp
```diff
@@ -66,7 +67,7 @@ static_assert(MARLIN_VAR_MAX < 64, "MarlinAPI: Too many variables");
     #include "power_panic.hpp"
 #endif
 
-static_assert(MARLIN_VAR_MAX < 64, "MarlinAPI: Too many variables");
+static_assert(MARLIN_VAR_COUNT <= 64, "MarlinAPI: Too many variables");
```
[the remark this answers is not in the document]
RadekVana: But i definitely want to rewrite this enum too
But in some other PR, because it will be huge amount of changes

src/common/marlin_events.c
```diff
@@ -1,4 +1,4 @@
-// marlin_events.c
+// core_client::events.c
```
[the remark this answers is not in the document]
RadekVana: thx, missed that one
was tired when I was checking all changes

src/common/marlin_client.hpp
```diff
@@ -1,20 +1,233 @@
...
+int motion();
+
+// synchronously wait for motion short timeout
+int wait_motion(uint32_t timeout);
+
+// enqueue gcode - thread-safe version  (request '!g xxx')
+void gcode(const char *gcode);
+
+// enqueue gcode from ethernet command (json parsed)
+void json_gcode(const char *gcode);
```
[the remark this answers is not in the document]
RadekVana: removed

src/common/marlin_events.c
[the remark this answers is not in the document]
RadekVana: fixed

src/common/marlin_events.h
```diff
@@ -1,4 +1,4 @@
-// marlin_events.h
+// core_client::events.h
```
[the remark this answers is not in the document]
RadekVana: fixed

src/common/marlin_server.hpp
RadekVana: fixed

src/common/marlin_server.hpp
RadekVana: I wanted to use alias, but I have forgot about it
Thx

src/common/marlin_vars.hpp
RadekVana: it should be print processor independent (GUI should not know about marlin ... mostly)
So I will use `PrintState`

vorner (commented):
Few style nitpicks, nothing blocking…

src/common/marlin_client.cpp
```diff
@@ -1,24 +1,26 @@
...
 #include <string.h>
 #include "config.h"
 #include "bsod.h"
 #include "ffconf.h"
 #include "timing.h"
 #include "log.h"
-#include "marlin_vars.h"
+#include "marlin_vars.hpp"
+
+using namespace print_client;
```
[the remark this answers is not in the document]
RadekVana: I have it wrapped in namespace in header. What is wrong with `using namespace print_client;` in source file?

PR #2473, opened 2022-09-01
Add translations for 4.4.0x

DRracer:
Pointless in case of translations, but some more interesting stuff comes along, so why not 

BFW-2708

### DIFF
--- a/src/lang/translation_provider_CPUFLASH.hpp
+++ b/src/lang/translation_provider_CPUFLASH.hpp
@@ -53,7 +53,7 @@ public:
     }
 
     // bucket_count is being computed at compile time (lang.py is searching for the lowest possible number of buckets where collisions do not occur)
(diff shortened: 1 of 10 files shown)


src/lang/translation_provider_CPUFLASH.hpp
```diff
@@ -54,7 +54,7 @@ class CPUFLASHTranslationProviderBase : public ITranslationProvider {
 
     // bucket_count is being computed at compile time (lang.py is searching for the lowest possible number of buckets where collisions do not occur)
     // 310 is the maximum total number of strings the translator array can hold. To be increased in the future as new strings come into the FW
```
MistrKernnunos, line 56: Just small thing, change the comment to current number of strings in fw

MistrKernnunos (approved):

PR #2476, opened 2022-09-02
Support for Overwrite header parameter in PUT.

HelclM:


lib/WUI/nhttp/gcode_upload.cpp
```diff
@@ -193,7 +194,14 @@ Step GcodeUpload::step(string_view input, PutParams &putParams) {
     // it can be used inside UploadState as a function of UploadHooks.
     if (!filename_checked) {
         auto filename_error = check_filename(filename);
-        if (std::get<0>(filename_error) != Status::Ok)
+        if (std::get<0>(filename_error) == Status::Conflict and putParams.overwrite) {
+            if (wui_is_file_being_printed(putParams.filepath.data())) {
```
vorner, line 198: This seems like a bit of an abuse. I mean, it should probably be the job of check_filename to handle the situation, not guessing by what HTTP error code it returned and then trying to „fix“ the situation.

lib/WUI/nhttp/gcode_upload.cpp
```diff
@@ -245,24 +253,35 @@ namespace {
             strlcat(fn, dest_fname, FILE_PATH_BUFFER_LEN - USB_MOUNT_POINT_LENGTH);
         }
 
+        FILE *attempt = fopen(fn, "r");
+        if (attempt) {
+            fclose(attempt);
+            if (overwrite) {
+                int result = remove(fn);
+                if (result == -1) {
+                    switch (errno) {
```
vorner, line 262: Do I read correctly that this branch leaves the temp file in place?

lib/WUI/nhttp/gcode_upload.cpp
```diff
@@ -245,24 +253,35 @@ namespace {
...
+                    case EBUSY:
+                        return make_tuple(Status::Conflict, "File is busy");
+                    default:
+                        return make_tuple(Status::NotFound, "Unknown error");
```
vorner, line 266: If it's unknown error, then we probably shouldn't return 404 Not Found, which is _client_ error. This should go to some 5?? error.

lib/WUI/nhttp/gcode_upload.cpp
```diff
@@ -292,7 +310,9 @@ UploadHooks::Result GcodeUpload::finish(const char *final_filename, bool start_p
     // Close the file first, otherwise it can't be moved
     ftruncate(fileno(tmp_upload_file.get()), ftell(tmp_upload_file.get()));
     tmp_upload_file.reset();
-    return try_rename(fname, final_filename, [&](char *filename) -> UploadHooks::Result {
+    auto putParams = std::get_if<PutParams>(&upload);
+    bool overwrite = putParams and putParams->overwrite;
```
vorner, line 314: Style nitpick:

* Using of `and` and `or` in C++ is some new kind of decadence (or, actually, they seem to be ancient and never have reached any kind of widespread use :-O … anyway, they seem unusual). When trying to find out why, it seems they tend to be a bit confusing, because `and` is equivalent to `&&`, while `and_eq` is equivalent to `&=` which acts differently.
* We prefer to _explicitly_ compare pointers to `nullptr` in contrast to treating them as bools implicitly.

vorner (approved):

PR #2477, opened 2022-09-02
Add new eeprom impplementation

MistrKernnunos:
# Configuration store
Configuration store is replacement of current EEPROM implementation. This implementation should solve these problems:
- need to add things to multiple files in FW
- unaven wearing 
- removes the need to use variant_8

It is split into two layers. Configuration store and Eeprom access. Configuration store is structure which holds the data in runtime and enables access to items. Eeprom access enables access to the eeprom itself. 
## Configuration store
Configuration store is structure which holds items as its members and each item represents data in eeprom.
Items are initialized with default values and then eeprom access is initialized and loads changed values to the items.
Item is stored in eeprom only when its value is changed. So it wont take up space if the value is left as default value. 
The items are generated from JSON with python script. This python script support creating:
- Simple one value items
- Arrays of values
- Struct items
- Strings

## Eeprom access
This layer stores the data in eeprom in journal like style. Each write is written after the previous one. When end of bank is reached, cleanup routine is called.
It only supports write operation during runtime. Reading data is possible only during initialization. To be able to read data during runtime we would need some cache for storing position of each item to prevent O(n) complexity for read. 
### Banks
The EEPROM is split in two same sized banks and one bank is in use. Eeprom access stores which bank is used in byte before the first bank. End of bank is marked with 0xFF.
### Item
| size | data | crc32  |
|------|------|--------|

Each item consists of size, data and crc32. And data is serialized as msgpack. 
### Cleanup routine
Cleanup routine copies only newest values for each key to the new bank. The old bank is left valid in case new bank gets corupted.


MistrKernnunos:
Cleanup method is not yet implemented, because I am not still 100% how I want to implement it. 
1. Store all non default items to begging of eeprom.
 - danger of loosing large chunk off eeprom if the printer is reset during this operation
3. Iterate throuhg eeprom and update first items with the latest value.
- danger of corrupting some item in middle of eeprom in case of reset  during write of one item and rendering all items behind this one inaccesible. 
- This one has smaller windows but more windows when reset can cause problems.

tests/unit/common/eeprom/eeprom_acces.h
```diff
@@ -0,0 +1,131 @@
...
+    // find address for hash in index
+    std::optional<uint16_t> get_address(uint32_t hash);
+
+    // stores item to eeprom and calculates the crc of the data
+    // creates header of the data
+    // the header  element ext 8 from msgpack spec and the value for type is EepromAccess::CUSTOM_ITEM_ID
+    // +--------+--------+--------+========+
+    // |  0xc7  |XXXXXXXX|  type  |  data  |
+    // +--------+--------+--------+========+
+    // where XXXXXXXX is a 8-bit unsigned integer which represents the length of the data including crc
```
vorner, line 65: What's the 0xc7 thing? And, is it possible to combine the type and this tag together? :thinking: I mean, the eeprom is not that big, so maybe being able not to waste the space would be nice.

tests/unit/common/eeprom/eeprom_acces.h
```diff
@@ -0,0 +1,131 @@
...
+
+    // Saves data item after last item
+    // if it hits max number of items it will remove duplicate items
+    template <class T>
+    void Set(const char *key, const T &data);
+
+    // looks in the index if we have the key, if yes it load s the data from eeprom if not returns nullopt
+    // should be called only after init before any call to set, because set can invalidate the index
+    template <class T>
+    std::optional<T> Get(const char *key);
```
vorner, line 87: * It would be really nice if the type was checked at compile time or at least by some kind of assert. I mean, the code should be 100% sure what type it is. Or do I get it wrong when it can give me `nullopt`?
* I don't like the stuff about invalidation. The eeprom is being used in a way where someone is setting it and someone (potentially someone _else_) is reading it without the knowledge of the set.

tests/unit/common/eeprom/eeprom_access.cpp
```diff
@@ -0,0 +1,96 @@
+#include "eeprom_acces.h"
+#include "st25dv64k.h"
+#include <algorithm>
+
+void EepromAccess::store_item(const std::vector<uint8_t> &data) {
+    // we have only one byte to store length of the data
+    if (data.size() >= (std::numeric_limits<uint8_t>::max() - 1)) {
+        // general_error("data too large);
+        // FIXME remove before commit
+        throw std::out_of_range("data too large");
```
vorner, line 10: I don't think we support exceptions in the firmware code. We do have bsod.

tests/unit/common/eeprom/eeprom_access.cpp
```diff
@@ -0,0 +1,96 @@
...
+        auto data = item.value();
+        uint32_t key = msgpack::unpack<Key>(data).key;
+        // push back and sort
+        index.emplace_back(key, address);
+        std::push_heap(index.begin(), index.end());
+        address += item.value().size() + HEADER_SIZE + CRC_SIZE;
+    }
+    std::sort_heap(index.begin(), index.end());
+    // reversing it because if we used backward iterator we would have the unwanted indexes in front of the ones we want
+    std::reverse(index.begin(), index.end());
```
vorner, line 79: Can you explain what this whole thing does and what's the index even for? Didn't we kind of want to mirror the values inside the RAM instead? Why sorting? Isn't it overcomplicating things?

tests/unit/common/eeprom/msgpack.h
```diff
@@ -0,0 +1,1020 @@
+//
+// Created by Mike Loomis on 6/22/2019.
```
vorner, line 2: Should this go somewhere into lib/third_party/...? Should it be a git subproject, to allow future easy updates?

vorner (commented):
I understand this is a draft. I also assume this is why it completely lives in unittests and not somewhere inside the source code, where it would eventually move. So some high-level comments:

* Methods starting with upper case letter aren't really the capitalization style we kind of agreed on (I hope it was agreed on _generally_, but at least it's written down somewhere).
* As for the interface:
  - I'd prefer if `get` couldn't fail at runtime, or at least not by returning a `nullopt`. The only reason why it could fail is IMO because the type is wrong. That's not a runtime error, that's programmer error and it would be great to have some kind of hard check on that.
  - Should the keys be some kind of enum, preferably?
  - Can we make the definition of one field a bit less verbose? If I'm looking at the definition of the version, it's a bit… big.
* Internals:
  - We've already talked about allocation. It would be better without them.
  - I don't understand why you need the index (I didn't do a really deep dive), but I've assumed that we would still have in-RAM cache of the _values_. That is, we wouldn't need the index, we could just load or store from the right place in RAM.
  - I think we could get rid of some markers inside the format. Have only length-block and the stop could be 0-sized block. Saves quite a bit.
* :question: What _should_ happen if we hit an incorrect CRC? In current implementation, can we get an inconsistent eeprom content?

tests/unit/common/eeprom/eeprom_acces.h
MistrKernnunos: This is ext 8 type from msgpack spec, I could remove the type and just use the 0xc7, but then we will not be compatible with msgpack.

tests/unit/common/eeprom/eeprom_acces.h
MistrKernnunos: - This should give you nullopt when the data are not in eeprom or are corrupted. 
- You could get bad data if you use different type than the one which is in eeprom. 
- Eeprom access does not know which data type is stored in eeprom at compile time. 
- Eeprom access shoul be used just through Eeprom, which knows how to use it.

tests/unit/common/eeprom/Eeprom.h
```diff
@@ -0,0 +1,91 @@
+#pragma once
+#include "eeprom_acces.h"
+using EepromKey = const char *;
+
+enum class EepromVersion : uint16_t {
+    v1,
+    v2,
+    v3,
+    v4,
+    count = v4
```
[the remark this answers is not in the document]
MistrKernnunos: I am using the count for bounds checking.

tests/unit/common/eeprom/msgpack.h
MistrKernnunos: Maybe,I am thinking about ditching it and using something ours to cut down usage of eeprom and get rid of dynamic allocation

tests/unit/common/eeprom/Eeprom.h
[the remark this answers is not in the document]
vorner: Hmm, I can't find anything in the C++ library. So, OK…

tests/unit/common/eeprom/Eeprom.h
```diff
@@ -0,0 +1,91 @@
...
+    T max;
+    T def;
+    bool validate(const T &data) {
+        return (data >= min && data <= max);
+    }
+};
+
+template <class T, class CovertTo = T>
+struct MemConfigItem {
+    EepromAccess &eeprom;
```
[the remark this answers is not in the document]
vorner: That's what I'm challenging.

Conceptually, you have independent config items that can update themselves inside the eeprom. You can just hold the item and you're OK. Besides that feeling somewhat strange (an item just living in the ether?), this brings the disadvantage that each single config item will be at least 4 bytes bigger because of the reference. And if I understand it correctly (which I'm not sure of), these are long-living somewhere, or even are the „definitions“ for the values.

What I would think about is eeprom „has“ bunch of items. Or the config store has. So you'd call something like `store.set(...)` and the eeprom would be available in the set.

tests/unit/common/eeprom/eeprom_acces.h
vorner: Well, we can get _inspired_ by msgpack, or reuse parts of it. If adhering to it means having some 25% ineffeciency ratio, then it's a bit expensive.

So we can have length-prefixed chunk (very simple encoding) and msgpack inside, it would still give us most of the benefits.

tests/unit/common/eeprom/msgpack.h
vorner: That's also a possibility. We could have something simple/simplish. Another option is looking at how protobuf does it (I mean only the lower-level thing without the generators, etc, but they somehow combine the „name“ tag with the „type“ tag). But on the other hand, they have some variable-length thingie, so that's maybe a complication.

tests/unit/common/eeprom/Eeprom.h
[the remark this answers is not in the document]
MistrKernnunos: Yes this items are definitions for the items and are also the ram mirror of the eeprom. 

I have changed the reference to the eeprom_access to pointer to the store. 

T think this is good tradeoff between the implementation and the posibility to access the eeprom items as member variables.

tests/unit/common/eeprom/eeprom_acces.h
MistrKernnunos: I aggree, I willl change it to just lenght prefix.

utils/configuration_store_generator/configuration_store_structure_generator.py
```diff
@@ -0,0 +1,251 @@
...
+
+    def __create_array_item(self,
+                            name: str,
+                            data_type: str,
+                            length: int,
+                            item: dict,
+                            default=None) -> Item:
+        member = self.__parse_type("item", item)
+
+        return ArrayItem(name, length, member)
```
dragomirecky, line 148: It seems to me that you can't create an array of N elements and set the default to N distinctive values. Is that intentional? 🤔

tests/unit/common/eeprom/eeprom_test.cpp
```diff
@@ -0,0 +1,126 @@
...
+    REQUIRE(unpacked.key == crc32_calc(reinterpret_cast<const uint8_t *>(key), strlen(key)));
+}
+
+TEST_CASE("Test item storing and loading") {
+    std::vector<uint8_t> data { 0x11, 0x22, 0x33, 0x44 };
+    eeprom_chip.clear();
+    EepromAccess eeprom;
+    ConfigurationStore<> store;
+    auto updater = ItemUpdater(store);
+    eeprom.Init(updater);
```
dragomirecky, line 23: All members & methods should be snake_case

src/common/configuration_store/mem_config_item.hpp
```diff
@@ -0,0 +1,108 @@
...
+    void Set(std::array<char, SIZE> new_data);
+
+    const std::array<char, SIZE> &Get() {
+        return data;
+    }
+
+    constexpr MemConfigItem(EepromKey key, const char *def_val)
+        : key(key)
+        , def_val(def_val) {
+        factory_reset();
```
dragomirecky, line 106: /nitpick/
Can we please rename this to something like "set_to_default"?
Everytime I read "factory_reset" I though "OMG why does he do factory reset here?! 😅"

src/common/configuration_store/configuration_store.hpp
```diff
@@ -0,0 +1,39 @@
...
+ * This class is configuration store which provides easy access to its values.
+ * It initializes the values from eeprom and persists them into it on change.
+ * Data is written to eeprom only if the value is non-default or if it is overwriting non-default data with default data
+ *
+ * To add new item to configuration store, you need to add it to json from which it is generated
+ *
+ * To access the values get singleton instance of the store and access the items like members of any other class. To change the items use their Set and Get methods.
+ */
+template <class CONFIG>
+class ConfigurationStore : public CONFIG {
```
dragomirecky, line 19: /suggestion/ The configuration store is implemented across several files and its public API consists of more classes etc. What about placing all of it into some namespace?

dragomirecky (commented):
idea: filter board & printers

utils/configuration_store_generator/configuration_store_structure_generator.py
```diff
@@ -0,0 +1,251 @@
...
+@dataclass
+class Item(ABC):
+    name: str
+
+    def generate_item(self):
+        pass
+
+    def get_item_declaration(self) -> str:
+        # TODO create bounds set which will reduce the number of bounds
+        return f"MemConfigItem<{self.get_type_name()}> {self.get_name()} {{ \"{self.name}\" , {str(self.default)} }};"
```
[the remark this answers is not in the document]
MistrKernnunos: They should end up in ram, because they hold the current value at runtime.

utils/configuration_store_generator/configuration_store_structure_generator.py
MistrKernnunos: Yes, I havent found reason to do have different defaults. But if tyou have some I will add the posibility to add different defaults. 

Strings have their own type.

utils/configuration_store_generator/configuration_store_structure_generator.py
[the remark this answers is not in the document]
dragomirecky: Of course, my bad 🙈😅

src/common/st25dv64k.h
```diff
@@ -7,16 +7,15 @@
...
+extern void st25dv64k_user_write_bytes(uint16_t address, void const *pdata, uint16_t size);
 
 extern uint8_t st25dv64k_user_read(uint16_t address);
 
 extern void st25dv64k_user_write(uint16_t address, uint8_t data);
 
-extern void st25dv64k_user_read_bytes(uint16_t address, void *pdata, uint16_t size);
-
-extern void st25dv64k_user_write_bytes(uint16_t address, void const *pdata, uint16_t size);
+extern void st25dv64k_init(void);
```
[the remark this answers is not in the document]
MistrKernnunos: No, just left over from development

utils/configuration_store_generator/configuration_store_structure_generator.py
```diff
@@ -0,0 +1,251 @@
+import argparse
```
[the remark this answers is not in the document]
MistrKernnunos: Agree, I will shorten it.

utils/configuration_store_generator/configuration_store_structure_generator.py
```diff
@@ -0,0 +1,257 @@
...
+class DataType(Enum):
+    i8 = "int8_t"
+    i16 = "int16_t"
+    i32 = "int32_t"
+    u8 = "uint8_t"
+    u16 = "uint16_t"
+    u32 = "uint32_t"
+    flt = "float"
+    bol = "bool"
+    cha = "char"
```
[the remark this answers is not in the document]
MistrKernnunos: Yes

mkbel:
Make sure new implementation doesn't mess with EEPROM flag read by bootloader:
```
#define FW_UPDATE_FLAG_ADDRESS 0x40B
uint8_t fw_update_flag = eeprom_read_byte(FW_UPDATE_FLAG_ADDRESS);
```

vorner:
In an attempt to revive this, I've went over the code again, ignoring all the above discussion and such. I'm not doing a line-by-line review right now, nevertheless there's bunch of topics than would probably need addressing:

**Minor nitpicks**

* You're consistently using the "wrong" include style ‒ `"..."` is for use of relative paths, if it's a search-path-include, it _should_ be `<...>`. The fact it works with `"..."` is more of an accident than a feature.
* In a lot of places, there were changes, but surrounding comments were not updated (like, changing a return value to `void`, but not removing commend describing the result).
* Is the `config_store_c_api` generated or written manually? It is in git, but seems very repetitive.

**Possibly unnecessary wasting?**

* I still feel that using a type prefix and CRC for every single item is a very high overhead and that we don't have that much luxury.
* Do we really need a hashmap? What is it used for, only getting rid of duplicates? Would it be acceptable to just use some kind of insert-sort or the keys, or just put everything into an array, sort and eliminate in a linear pass? The amount of hash-map code seems a lot.
* I'm no longer convinced that messagepacks gives us any particular benefit in here… but I might be wrong.
* Is the „new“ storage placed in a different place of eeprom, the old place being completely unused from now on and wasted?

**Packing and recovery**

* The recovery code seems to be somewhat chaotic. If there's a problem, do we know into what state are we returning? Is it for sure the newest state we are able to recover? How often do the banks switch, is it possible we are returning very long time into the past, or is a cleanup done on each startup?
* Is it possible we could have a configuration that doesn't fit into a bank even after a cleanup? What happens then?

**Lack of naming consistency**

In general, the way items are accessed or set seems to be arbitrary and seems to be done different ways across places? I'm still not sure if there's a pattern, but it seems there's usually some `get()` involved, but it is often somewhere in the middle of method chain and member access. But I've also seen things like `set_some_item()` instead of `item.set()` (while there _was_ an `item.get()` nearby), I've seen free-standing `get_some_item()`…

Maybe I'm just not getting it.

**Thread safety**

Or the lack of? Again, I don't see a clearly-correct pattern in the way it should be ensured. Things I've noticed that are at least suspicious:

* The eeprom access locks a mutex at init, but nowhere else. Do we ensure two threads are not writing an update or performing a cleanup at the same time?
* Disable-Interrupts has a race between reading the mask and disabling it, it could have changed in meantime. And I think if two threads at once try to use this, we could end up with eternally disabled interrupts (first disables, the second one reads the already-disabled state and once the second one restores, it restores to the disabled state).
* The move constructor of FreeRTOS mutex a) talks about eeprom, while it is something used in many other places too, b) seems at least suspicious ‒ creating a copy of the mutex somewhere else, possibly already locked (does it work that way?)? What happens to the old one, does it get unlocked? Or are there now two, both locked?
* Reading strings ‒ in WUI and Connect ‒ is not thread safe (it wasn't before either, but I think the new API shall not have this kind of broken-by-design thing). It is returning a pointer, but that memory might be overwritten in another thread during the later `strncpy` ‒ it doesn't happen under any lock.
* Are items locked during reading, or are they atomic?

**Way forward**

I'd say we need answers to these questions/issues first.

Then someone probably should go and deal with a line-by-line review, fixes, etc.

Also, the migration of old eeprom seems to be missing?

PR #2480, opened 2022-09-05
Podpora delete requestu do /api/v1/files.

HelclM:


lib/WUI/link_content/prusa_link_api.cpp
```diff
@@ -158,6 +158,19 @@ optional<ConnectionState> PrusaLinkApi::accept(const RequestParser &parser) cons
...
+                    switch (errno) {
+                    case EBUSY:
+                        return StatusPage(Status::Conflict, parser.status_page_handling(), parser.accepts_json, "File is busy");
+                    default:
+                        return StatusPage(Status::NotFound, parser.status_page_handling(), parser.accepts_json);
+                    }
+                } else {
+                    return StatusPage(Status::NoContent, parser.status_page_handling(), parser.accepts_json);
+                }
+            }
```
vorner, line 173: Can you extract the code instead of copy-pasting it?

vorner (changes requested):

vorner (approved):
:+1:

PR #2486, opened 2022-09-06
marlin_client_loop -> marlin_update_vars

HelclM:
Changed marlin_client_loop to marlin_update_vars in wui thread, to make
the update actually work.

### DIFF
--- a/lib/WUI/link_content/basic_gets.cpp
+++ b/lib/WUI/link_content/basic_gets.cpp
@@ -29,7 +29,7 @@ JsonResult get_printer(size_t resume_point, JsonOutput &output) {
     bool ready = true;
     bool busy = false;
 
-    marlin_client_loop();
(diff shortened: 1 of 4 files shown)


lib/WUI/link_content/basic_gets.cpp
```diff
@@ -29,7 +29,7 @@ JsonResult get_printer(size_t resume_point, JsonOutput &output) {
     bool ready = true;
     bool busy = false;
 
-    marlin_client_loop();
+    marlin_update_vars(MARLIN_VAR_MSK_TEMP_ALL | MARLIN_VAR_PRNSPEED | MARLIN_VAR_POS_Z | MARLIN_VAR_PRNSPEED | MARLIN_VAR_PRNSTATE);
```
RadekVana, line 32: you cannot use those enum values directly
use MARLIN_VAR_MSK(MARLIN_VAR_PRNSPEED)

RadekVana (changes requested):

lib/WUI/link_content/basic_gets.cpp
RadekVana: it applies to all your changes

RadekVana (approved):

PR #2492, opened 2022-09-09
URL percent decoding of file names to upload.

HelclM:


tests/unit/common/url_decode_tests.cpp
```diff
@@ -0,0 +1,48 @@
...
+        char decoded_url[0];
+        CHECK(url_decode(common_url, decoded_url, sizeof(decoded_url)) == false);
+    }
+
+    SECTION("empty url") {
+        std::string url = "";
+        char decoded_url[100];
+        CHECK(url_decode(url, decoded_url, sizeof(decoded_url)) == true);
+        CHECK(strcmp(decoded_url, "") == 0);
+    }
```
vorner, line 47: I'd propose few more tests:

* Something with an incomplete escape sequence at the end, eg `/x/a%2`.
* Some integration with the Parser structure ‒ combination with real `?` that should end the path vs. encoded `?` that should be part of the resulting `?`. Similarly, we probably need to make sure one can't overcome the restrictions not to contain `..` in the path by escaping it. That is, we still don't want to allow someone to escape from `/usb` by encoding the dots or slashes into percents.

src/common/http/url_decode.h
```diff
@@ -0,0 +1,37 @@
...
+                decoded_url[out_index] = ' ';
+            else
+                decoded_url[out_index] = url[i];
+        } else {
+            int ascii_value;
+            auto curr_iter = url.begin() + i;
+            auto result = std::from_chars(curr_iter + 1, curr_iter + 3, ascii_value, 16);
+            // don't go over 7-bit ascii chars
+            if (result.ec != std::errc {} || ascii_value > 127)
+                return false;
```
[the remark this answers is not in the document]
HelclM: Seems like it isn't, but it will produce weird question mark characters on the screen, but maybe if we get garbage we should return garbage?

lib/WUI/nhttp/req_parser.cpp
```diff
@@ -210,28 +212,31 @@ bool RequestParser::uri_filename(char *buffer, size_t buffer_size) const {
         len = amp - url.begin();
     }
 
-    // Make sure the user is not able to "escape" from directory.
+    if (buffer_size < len) {
+        return false;
+    }
```
vorner, line 217: :thinking: I wonder… it's probably not worth fixing/supporting right now, but maybe a note would not go amiss. If we have an URL with a bunch of `%??` in it, the result is shorter. So this is a bit stricter than necessary.

src/common/http/url_decode.h
```diff
@@ -0,0 +1,41 @@
+#pragma once
+
+#include <charconv>
+#include <string_view>
+
+namespace http {
+bool url_decode(std::string_view url, char *decoded_url, size_t decoded_url_len) {
```
vorner, line 7: Looking at this. There doesn't seem much reason why this needs to be an inline function in a header file. Can you put it into a proper .cpp file, please?

src/common/http/url_decode.h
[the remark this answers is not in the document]
vorner: I'd say we should accept as much as possible / allowed here, unless there's a reason not to. After all, users are allowed to put weirdly-named files on the USB stick on the computer too and we should be able to work with these.

vorner (commented):
I've noticed just few more minor nitpicks.

vorner (approved):

PR #2493, opened 2022-09-09
Enable Crash Detection & Recovery.

mkbel:
Enable and configure Unified Bed Leveling.
Enable and configure Precise Homing.
Enable and configure Crash Detection & Recovery.
Enable and configure Power Panic.

Known issues:
 - Unified Bed Leveling
   - redundant first move - PINDA goes outside of print area
 - Precise Homing
   - none
 - Crash Detection & Recovery
   - Disabling Crash Detection doesn't work well - very old bug - part of the move is skipped when stall is detected (stall detection should be disabled when crash detection is disabled) (I have not tested, if the bug is still there. I just assume it as I did nothing to fix it.)
 - Power Panic
   - does not work at all (It is not triggered by power outage.)


mkbel:
Power panic was ditched for the moment. As it is not possible to edit originating branch I am closing it and opening it again as https://github.com/prusa3d/Prusa-Firmware-Buddy/pull/2514

PR #2494, opened 2022-09-09
Fix release debugging

vladamatena:
- Avoid reading bootloader vars in noboot scenario.
- Avoid breaking debug when checking appendix presence.

### DIFF
--- a/src/common/support_utils.cpp
+++ b/src/common/support_utils.cpp
@@ -132,28 +132,37 @@ void error_url_short(char *str, const uint32_t str_size, const int error_code) {
 }
 
 bool appendix_exist() {
+    // With recent bootloader this can be done the easy way
+#ifdef BOOTLOADER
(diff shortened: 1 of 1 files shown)


src/common/support_utils.cpp
```diff
@@ -132,28 +132,37 @@ void error_url_short(char *str, const uint32_t str_size, const int error_code) {
...
-#else //In debug version the appendix status has to be tested by bootloader version greater or equal than 1.1
-        pinState = ram_data_exchange.model_specific_flags & APPENDIX_FLAG_MASK
-            ? GPIO_PIN_SET
-            : GPIO_PIN_RESET;
+    }
 #endif
-        return pinState == GPIO_PIN_RESET;
+
+    // If debugging session is active there is no appendix
+    if (DBGMCU->CR) {
```
dragomirecky, line 144: 😍
we should add this to the bootloader too

dragomirecky (approved):

PR #2507, opened 2022-09-15
Connect preview

vorner:


lib/WUI/nhttp/gcode_preview.cpp
```diff
@@ -21,7 +21,7 @@ using std::string_view;
 
 GCodePreview::GCodePreview(FILE *f, const char *path, bool can_keep_alive, bool json_errors, uint16_t width, uint16_t height, uint32_t if_none_match)
     : gcode(f)
-    , decoder(f, width, height)
+    , decoder(f, width, height, true)
```
HelclM, line 24: Minor nitpick: wouldn't this maybe be more readable as a some kind of enum? Right now I have to look inside the constructor to find out what does it mean.

HelclM (approved):

lib/WUI/nhttp/gcode_preview.cpp
vorner: I know these bools get a bit unreadable, especially if there's a bunch of them.

On the other hand, defining a whole new enum and write something like `Base64::Decode` and `Base64::DontDecode` adds a lot of code and tend to clutter the headers, the code that uses it (because then one needs to compare it using `== Base64::Whatever`).

Considering this one is used at 3 places only, I'd say the benefit of defining an enum doesn't outweight the downsides here.

PR #2511, opened 2022-09-16
Fix junction deviation

mkbel:
When doing update of the Marlin, I noticed some problem workaround feature called JD_SMALL_SEGMENT_HANDLING. I decided to backport this trivial fix to our firmware.

This doesn't fix our problem. That upset me so much I fixed the problem myself.

The problem was in junction angle computation - before the fix junction angle was computed after position was rounded up to whole machine steps. For short segments it resulted in +- 6° angle error. After some debugging and realizing that the problem is in segment angle I tried lot of things to fix it. E.g. use double instead float for computation - this has not fixed the problem. I checked the gcode, if the problem is not already presented there - gcode is not problematic, there are 3 decimal points for position, which is sufficient for angle computation.

And then the final fix - compute the angle from float position from gcode - before it is rounded to machine steps coordinates. - that solved the problem. First video JD enabled - before the fix. Second video JD enabled - after the fix.
[test gcode](https://github.com/prusa3d/Prusa-Firmware-Buddy/files/10270973/valec_JD_0.2mm_PLA_MINI_25m.zip)

[JD_gcode_analysis.ods](https://github.com/prusa3d/Prusa-Firmware-Buddy/files/9586791/JD_gcode_analysis.ods)

https://user-images.githubusercontent.com/35807926/190718132-6eb1c4ab-5534-4db1-8564-9e566f15dd6b.mp4

https://user-images.githubusercontent.com/35807926/190718160-32382076-175f-4f0e-94ca-688930cfb98c.mp4

### DIFF
--- a/lib/Marlin/Marlin/src/module/planner.cpp
+++ b/lib/Marlin/Marlin/src/module/planner.cpp
@@ -2353,22 +2353,27 @@ bool Planner::_populate_block(block_t * const block, bool split_move,
     static xyze_float_t prev_unit_vec;
 
     xyze_float_t unit_vec =
-      #if IS_KINEMATIC && DISABLED(CLASSIC_JERK)
+      #if IS_KINEMATIC
         delta_mm_cart
       #else
-        { delta_mm.x, delta_mm.y, delta_mm.z, delta_mm.e }
+        #if HAS_POSITION_FLOAT
+          { target_float.x - position_float.x, target_float.y - position_float.y, target_float.z - position_float.z,  target_float.e - position_float.e}
+        #else
+          { delta_mm.x, delta_mm.y, delta_mm.z, delta_mm.e }
+        #endif
       #endif
     ;
-    unit_vec *= inverse_millimeters;
 
-    #if IS_CORE && DISABLED(CLASSIC_JERK)
-      /**
-       * On CoreXY the length of the vector [A,B] is SQRT(2) times the length of the head movement vector [X,Y].
-       * So taking Z and E into account, we cannot scale to a unit vector with "inverse_millimeters".
-       * => normalize the complete junction vector
-       */
+    /**
+     * On CoreXY the length of the vector [A,B] is SQRT(2) times the length of the head movement vector [X,Y].
+     * So taking Z and E into account, we cannot scale to a unit vector with "inverse_millimeters".
+     * => normalize the complete junction vector
+     * Also always normalize when float position is not available and there is E component.
+     */
+    if (IS_CORE || (!HAS_POSITION_FLOAT && (esteps > 0)))
       normalize_junction_vector(unit_vec);
-    #endif
+    else
+      unit_vec *= inverse_millimeters;
 
     // Skip first block or when previous_nominal_speed is used as a flag for homing and offset cycles.
     if (moves_queued && !UNEAR_ZERO(previous_nominal_speed_sqr)) {
@@ -2376,6 +2381,9 @@ bool Planner::_populate_block(block_t * const block, bool split_move,
       // NOTE: Max junction velocity is computed without sin() or acos() by trig half angle identity.
       float junction_cos_theta = (-prev_unit_vec.x * unit_vec.x) + (-prev_unit_vec.y * unit_vec.y)
                                + (-prev_unit_vec.z * unit_vec.z) + (-prev_unit_vec.e * unit_vec.e);
+      #if ENABLED(JD_DEBUG_OUTPUT)
+        SERIAL_ECHO_F(junction_cos_theta, 7);
+      #endif
 
       // NOTE: Computed without any expensive trig, sin() or acos(), by trig half angle identity of cos(theta).
       if (junction_cos_theta > 0.999999f) {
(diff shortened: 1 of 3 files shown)


thinkyhead:
Does Marlin (2.1.1) continue to exhibit the same issue repaired here?

mkbel:
@thinkyhead 

> Does Marlin (2.1.1) continue to exhibit the same issue repaired here?

I don't know, I can't test yet. Update to Marlin to 2.1.x is still far from to be build-able https://github.com/prusa3d/Prusa-Firmware-Buddy/pull/2488

The problem in JD is caused by wild fluctuation of angle between segments. That angle fluctuates wildly, because it is not computed from original float coordinates from gcode (which doesn't fluctuate so much) but from position after it was rounded to machine micro steps. 

E.g. from our slicer coordinates are in thousandths of milimeters

```
G1 X20.937 Y209.063 E9.96284
G1 X22.437 Y209.063 E.05077
```
this is sufficient to compute stable angle between segments.

But our machines uses 100 micro steps per 1 mm, so coordinates becomes rounded to hundredths of milimeter and that rounding produces big jitter in angle between segments so JD limited junction speed jerk leads to wild speed oscillations when printing e.g. big finely sliced cylinder. (And for sure you need to slice big cylinder finely to allow high speed.)

You can check yourself upstream what is the source of coordinates to `unit_vec`.

thinkyhead:
This PR remains in my peripheral vision. Hopefully there will be a more concerted cross-pollination between Prusa Firmware and Marlin Firmware going forward. As we have our first implementation of Input Shaping, that has the potential to boost the productivity of the Prusa Print Farm considerably, and I would love to see a reaffirmation of our mutual RepRap love!

mkbel:
@thinkyhead I removed all the unnecessary changes from this PR and it still does what is intended. It removes speed fluctuating when printing detailed circles in high speed.

mkbel:
> Does Marlin (2.1.1) continue to exhibit the same issue repaired here?

@thinkyhead I can also confirm this bug was already fixed in upstream Marlin. I tested it on five months old branch 2.1.x on following commit:

Author:			tombrazier <68918209+tombrazier@users.noreply.github.com>
Author date:	5 months ago (7/23/2022 6:32:28 AM)
Committer:		Scott Lahteine <thinkyhead@users.noreply.github.com>
Commit date:	5 months ago (7/23/2022 6:33:30 AM)
Commit hash:	733ca940c0ed582fcd0e0a39b10b1fe8986b574c
Parent:			c880c7ed
�🐛 Fix 2d mesh print (#24536)

The problem is not manifesting itself anymore visibly during print. It is also not visible on my debug output. https://github.com/mkbel/Prusa-Firmware-Buddy/tree/test_JD_Marlin_2.1.x

mkbel:
@thinkyhead But I am absolutely unable to tell what upstream change fixed the problem. As upstream `xyze_float_t unit_vec` is still computed from `steps_dist_mm` which is computed from `(target - position) * mm_per_step` where `target` and `position` are `abce_long_t` and `xyze_long_t` (integer position in steps). So the upstream computation of `junction_cos_theta` should still suffer from position being rounded to steps.

mkbel:
@thinkyhead I managed to identify which upstream change fixed the problem. It is this commit:
https://github.com/MarlinFirmware/Marlin/pull/17938/commits/b45d97ae608b8e626e92f118b45e5659295deff8
After that change `if (esteps > 0)` `normalize_junction_vector(unit_vec)` is always called. That normalization somehow mitigates `junction_cos_theta` fluctuations on PRUSA MINI printer. It is probably because E axis is most coarse during the print on MINI so rounding to whole E steps has most detrimental effect on `junction_cos_theta` precision. The whole concept of computing junction angle in four dimensional space is very dubious for me. In my first proof of concept fix to the problem I completely excluded E axis from angle computation.

My solution to the problem is probably more efficient as it avoids costly `normalize_junction_vector` computation. And more logical as it uses more accurate source of position instead of dubiously fixing E step quantization error. On the other hand my fix relies on `HAS_POSITION_FLOAT` and is not possible if float position is not available.

lib/Marlin/Marlin/src/module/planner.cpp
```diff
@@ -2356,7 +2356,7 @@ bool Planner::_populate_block(block_t * const block, bool split_move,
       #if IS_KINEMATIC && DISABLED(CLASSIC_JERK)
         delta_mm_cart
       #else
-        { delta_mm.x, delta_mm.y, delta_mm.z, delta_mm.e }
```
wavexx, line 2359: Shouldn't we also shield this with JD_SMALL_SEGMENT_HANDLING?

mkbel: No, there is no connection with JD_SMALL_SEGMENT_HANDLING. In reality there is no connection between JD_SMALL_SEGMENT_HANDLING and fix of junction deviation. The only reason why I pulled JD_SMALL_SEGMENT_HANDLING from upstream is that I thought it might fix the problem observed. It fixed nothing so I analyzed the problem and found out, that the problem is caused by imprecise junction angle computation. The problem is there regardless JD_SMALL_SEGMENT_HANDLING  is enabled or disabled. I improved junction angle precision by using float source of position from gcode instead of position rounded to steps.

I also checked if the problem is still present in upstream. I have noticed the problem doesn't manifest itself upstream. Even though imprecise position is still used upstream. In upstream Marlin, the problem is somehow mitigated by always normalizing unit vector if there is any E move.

It might have sense to use float position if HAS_POSITION_FLOAT is enabled and fall back to original code if HAS_POSITION_FLOAT is disabled.

wavexx: Ok, I would check for HAS_POSITION_FLOAT then. Although we have several places that use/require POSITION_FLOAT, my point was just to not to require it outright.

mkbel: I have gated it by HAS_POSITION_FLOAT. And also melded in upstream workaround for case HAS_POSITION_FLOAT is unavailable.

mkbel: I have also removed redundant `&& DISABLED(CLASSIC_JERK)` as everything is inside DISABLED(CLASSIC_JERK) block.

wavexx (approved):

PR #2520, opened 2022-09-23
Wizard map

RadekVana:
BFW-2796

### DIFF
--- a/src/gui/test/screen_test_selftest.cpp
+++ b/src/gui/test/screen_test_selftest.cpp
@@ -49,6 +49,45 @@ static void LoadUnloadTest() {
         WaitAndShot(get_selftest_state_name(i));
     }
 
+    //alternative stated for Result (default state is unknown == some tests did not pass)
+    SelftestResult_t result_data;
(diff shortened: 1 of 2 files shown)


src/gui/test/screen_test_selftest.cpp
```diff
@@ -8,19 +8,20 @@
...
+    while ((gui::GetTick_IgnoreTickLoop() - switch_time) < switch_period) {
         gui::TickLoop();
         gui_loop();
         DialogHandler::Access().Loop();
     }
 }
 
-static void WaitShotWait(const char *file_name, uint32_t wait = 512) {
+static void WaitAndShot(const char *file_name, uint32_t wait = 512) {
     char buff[128];
```
DRracer, line 25: Do we really need 128 bytes? What is the longest filename we'll process? Think about stack usage :wink:

RadekVana: I have plenty of stack. It is debug only feature. Printer does not do anything else.
Ok will limit it to 64 to make you happy

DRracer (approved):

PR #2523, opened 2022-09-24
Connect dir info

vorner:


src/common/segmented_json.h
```diff
@@ -85,6 +86,28 @@ class ChunkRenderer {
...
+};
+
+template <class... V>
+class VariantRenderer : public ChunkRenderer {
+private:
+    std::variant<V...> data;
+
+public:
+    VariantRenderer() = default;
+    VariantRenderer(std::variant<V...> d)
```
DRracer, line 102: :thinking: why do you pass the parameter by value and then move it into `data`?
Usually the parameter is an r-value reference in similar cases.

src/connect/planner.cpp
```diff
@@ -59,7 +59,8 @@ namespace {
 
     bool path_allowed(const char *path) {
         constexpr const char *const usb = "/usb/";
-        const bool is_on_usb = strncmp(path, usb, strlen(usb)) == 0;
+        // Note: allow even "bare" /usb
+        const bool is_on_usb = strncmp(path, usb, strlen(usb)) == 0 || strcmp(path, "/usb") == 0;
```
DRracer, line 63: :thinking: I don't like the 2 separate calles to strcmp but at this moment I don't see a simple way of elliminating one or the other

DRracer (approved):
looks almost ok to me, just please think about my comments.

src/common/segmented_json.h
vorner: Fixed. Probably a leftover from fights with different kinds of C++ references, templates (that were involved more in some intermediate version), failures with `std::forward`… sometimes C++ goes off the rail and this had the benefit of actually being able to compile 😈

src/connect/planner.cpp
vorner: I don't see an elegant one either. Writing some kind of search automaton is obviously an overkill and the compiler is likely doing that under the hood (I wouldn't be surprised if it does one 4-byte int compare and then checking for either `/` or `\0` or something like that).

src/common/segmented_json.h
Derevin: I'd like to point out that the previous version was also fine, pass-by-value and then std::move is generally a preferred way of doing setters, rather than having to specify multiple overloads (`const T &` and `T &&`). Performance wise it seems to be about the same, per https://stackoverflow.com/a/10706636

DRracer (approved):

PR #2526, opened 2022-09-26
Move shared files from connect to common/http

Derevin:
Some files from connect can be used to implement general purpose networking code.


src/common/http/httpc.hpp
```diff
@@ -1,16 +1,12 @@
...
 
-#include "connection.hpp"
 #include <cstdint>
 #include <optional>
 #include <variant>
-#include <http/types.h>
-
-namespace connect_client {
+#include "connection.hpp"
+#include "http/types.h"
```
vorner, line 7: This should become "types.h", as it now lives in the same directory.

src/common/http/httpc.hpp
```diff
@@ -1,16 +1,12 @@
...
 
-enum class Method {
-    Post
-};
```
vorner, line 13: The fact we can now reuse the same `http::Method` for both is good.

But I suspect there's some kind of translation of Method -> string somewhere and it now contains only the `Post`, while `Method` now has more variants. 

https://github.com/prusa3d/Prusa-Firmware-Buddy/blob/master/src/connect/httpc.cpp#L29

Also, that now probably belongs to types.cpp.

src/connect/connect.cpp
```diff
@@ -1,9 +1,9 @@
 #include "connect.hpp"
-#include "httpc.hpp"
+#include "http/httpc.hpp"
```
vorner, line 2: These should (and the below one too) should become `#include <http/httpc.hpp>`, as it is path include now, not a relative one. Also, it's worth moving to the relevant group with includes from other our directories below.

src/connect/connect.cpp
```diff
@@ -16,7 +16,16 @@
...
+using http::Connection;
+using http::ConnectionFactory;
 using http::ContentType;
+using http::Error;
+using http::HeaderOut;
+using http::HttpClient;
+using http::Method;
+using http::Request;
+using http::Response;
+using http::socket_con;
```
vorner, line 28: In general, I'm a bit reluctant about using `using namespace`, but this might be a place where it's the lesser evil…

src/connect/connect.hpp
```diff
@@ -1,7 +1,7 @@
 #pragma once
 
 #include "buffer.hpp"
-#include "httpc.hpp"
+#include "http/httpc.hpp"
```
vorner, line 4: Again, path include.

src/connect/tls/tls.hpp
```diff
@@ -1,7 +1,7 @@
 #pragma once
 
-#include "connection.hpp"
-#include "connect_error.h"
+#include "http/connection.hpp"
+#include "http/connect_error.h"
```
vorner, line 4: Again…

tests/unit/connect/http_client.cpp
```diff
@@ -1,4 +1,4 @@
-#include <httpc.hpp>
+#include "http/httpc.hpp"
```
vorner, line 1: Path include.

Also, this whole test file seems like it's also a good candidate to move, because it's testing the moved file.

vorner (commented):
There's just some nitpicking… but might be worth addressing first.

tests/unit/common/http/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
+add_executable(
+  common_http_tests
+  ${CMAKE_CURRENT_BINARY_DIR}/http_resp_automaton.cpp
+  ${CMAKE_SOURCE_DIR}/src/common/automata/core.cpp
+  ${CMAKE_SOURCE_DIR}/src/common/http/httpc.cpp
+  ${CMAKE_SOURCE_DIR}/src/common/http/resp_parser.cpp
+  ${CMAKE_SOURCE_DIR}/src/common/http/os_porting.cpp
+  http_client.cpp
+  # ${CMAKE_SOURCE_DIR}/tests/stubs/jsmn_impl.c
+  # ${CMAKE_SOURCE_DIR}/tests/stubs/strlcpy.c
```
vorner, line 10: Remove the commented-out things?

src/common/http/types.h
```diff
@@ -14,6 +15,41 @@ enum Method {
...
+    case Post:
+        return "POST";
+    case Put:
+        return "PUT";
+    case Delete:
+        return "DELETE";
+    default:
+        assert(0);
+    case UnknownMethod:
+        return "UNKNOWNMETHOD";
```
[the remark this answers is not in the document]
Derevin: The reason I chose to not `assert` the `UnknownMethod` the `to_str` one is because the general use case for this method I have in my head is logging. So when a user decides to output log an unknown state (let's say during an initialization of some object somewhere), the user might be confused why a simple 'log' would crash their program. However in the second case, calling it with  `UnknownMethod` feels like 'always a bug'.

vorner (approved):

PR #2527, opened 2022-09-26
Marlin 2.1.x incremental update

wavexx:
This updates a few core headers to Mariln 2.1.x.

- src/core/macros.h: pristine copy
- src/core/types.h: all local changes moved at the end for future removal (they're kept for current source compatibility only)
- src/core/debug_*: as above

A few extra changes were pulled as necessary (notably Conditionals_LCD/post introduce the new axis macros and enumerations which are used in types.h) without breaking current source compatibility.

Code contained inside unused conditionals (for example, DELTA) was pulled as-is to minimize any future conflict. We assume that to be intentionally broken until the migration is complete.


lib/Marlin/Marlin/src/gcode/bedlevel/M420.cpp
```diff
@@ -87,12 +87,12 @@ void GcodeSuite::M420() {
     }
   #endif
 
+  const xyz_pos_t oldpos = current_position;
+
   // If disabling leveling do it right away
   // (Don't disable for just M420 or M420 V)
   if (seen_S && !to_enable) set_bed_leveling_enabled(false);
 
-  xyz_pos_t oldpos = current_position;
```
mkbel, line 94: Does have this change any effect?

wavexx: In al the unrelated changes I took the upstream resolution. In this case, oldpos was moved up and should have no effect. This was something I did as I manually introduced const to fix overload resolution, but I'm going to fix this by fixing types.h instead shortly.

wavexx: Removed this and another similar change in M420 which is no longer required.

mkbel (approved):

PR #2528, opened 2022-09-27
Dump error message to xflash

michalrudolf:
BFW-2818

Add dumping and aquiring of error message and error title through xflash dump.

### DIFF
--- a/src/common/bsod_gui.cpp
+++ b/src/common/bsod_gui.cpp
@@ -149,16 +149,13 @@ char nth_char(const char str[], uint16_t nth) {
 //! Fatal error that causes Redscreen
 void fatal_error(const char *error, const char *module) {
     uint16_t *perror_code_short = (uint16_t *)(DUMP_INFO_ADDR + 1);
+    bool dump_error_message = false;
 
     /// Decision tree to define error code
     using namespace Language_en;
     /// TODO share these strings (saves ~100 B of binary size)
-    if (strcmp(MSG_INVALID_EXTRUDER_NUM, error) == 0) {
-        *perror_code_short = 0;
-    } else if (strcmp("Emergency stop (M112)", error) == 0) {
+    if (strcmp("Emergency stop (M112)", error) == 0) {
         *perror_code_short = 510;
-    } else if (strcmp("Inactive time kill", error) == 0) {
-        *perror_code_short = 0;
     } else if (strcmp(MSG_ERR_HOMING, error) == 0) {
         *perror_code_short = 301;
     } else if (strcmp(MSG_HEATING_FAILED_LCD_BED, error) == 0) {
@@ -178,11 +175,14 @@ void fatal_error(const char *error, const char *module) {
     } else if (strcmp(MSG_ERR_MINTEMP, error) == 0) {
         *perror_code_short = 208;
     } else {
-        *perror_code_short = 0;
+        *perror_code_short = 0; // Unknown error code = we don't have help.prusa3d site support for this error
+        dump_error_message = true;
     }
 
     DUMP_FATALERROR_TO_CCRAM();
     dump_to_xflash();
+    if (dump_error_message)
+        dump_err_to_xflash(error, module);
     sys_reset();
 }
 
--- a/src/common/crash_dump/dump.cpp
+++ b/src/common/crash_dump/dump.cpp
@@ -11,7 +11,8 @@ static constexpr uint32_t dump_offset = w25x_dump_start_address;
 static constexpr uint16_t dump_buff_size = 0x100;
 static constexpr uint32_t dump_xflash_size = DUMP_RAM_SIZE + DUMP_CCRAM_SIZE;
 
-static_assert(dump_xflash_size <= w25x_pp_start_address, "Dump overflows reserved space.");
+static_assert(dump_xflash_size <= w25x_error_start_adress, "Dump overflows reserved space.");
+static_assert(sizeof(dumpmessage_t) <= (w25x_pp_start_address - w25x_error_start_adress), "Error message overflows reserved space.");
 
 #define _STR(arg)  #arg
 #define __STR(arg) _STR(arg)
@@ -229,4 +230,50 @@ int dump_hardfault_test_1(void) {
     volatile int b = 0;
     volatile int c = 1 / b;
     return c;
+}
+
+// Dumping error message
+
+void dump_err_to_xflash(const char *error, const char *title) {
+    w25x_sector_erase(w25x_error_start_adress);
+    w25x_fetch_error();
+
+    dumpmessage_t dump_message;
+    dump_message.invalid = 0;
+    strlcpy(dump_message.title, title, sizeof(dump_message.title));
+    strlcpy(dump_message.msg, error, sizeof(dump_message.msg));
+    w25x_program(w25x_error_start_adress + sizeof(dumpmessage_t::not_displayed), ((uint8_t *)(&dump_message)) + sizeof(dumpmessage_t::not_displayed), sizeof(dumpmessage_t) - sizeof(dumpmessage_t::not_displayed)); // not_displayed have to stay untouched
+    w25x_fetch_error();
+}
+
+int dump_err_in_xflash_is_valid() {
+    uint8_t invalid;
+    w25x_rd_data(w25x_error_start_adress + sizeof(dumpmessage_t::not_displayed), &invalid, sizeof(dumpmessage_t::invalid)); // reading second byte of error space (invalid flag)
+    if (w25x_fetch_error())
+        return 0; // Behave as invalid message
+    return invalid ? 0 : 1;
+}
+
+int dump_err_in_xflash_is_displayed() {
+    uint8_t not_displayed;
+    w25x_rd_data(w25x_error_start_adress, &not_displayed, sizeof(dumpmessage_t::not_displayed)); // reading first byte of error space (not_displayed flag)
+    if (w25x_fetch_error())
+        return 0;
+    return not_displayed == 0 ? 1 : 0;
+}
+
+void dump_err_in_xflash_set_displayed(void) {
+    uint8_t not_displayed = 0;
+    w25x_program(w25x_error_start_adress, &not_displayed, sizeof(dumpmessage_t::not_displayed)); // writing to second byte of error space (not_displayed flag)
+    w25x_fetch_error();
+}
+
+int dump_err_in_xflash_get_message(char *msg_dst, uint16_t msg_dst_size, char *tit_dst, uint16_t tit_dst_size) {
+    dumpmessage_t dump_message;
+    w25x_rd_data(w25x_error_start_adress, (uint8_t *)(&dump_message), sizeof(dumpmessage_t));
+    if (w25x_fetch_error())
+        return 0;
+    strlcpy(tit_dst, dump_message.title, sizeof(dump_message.title) > tit_dst_size ? tit_dst_size : sizeof(dump_message.title));
+    strlcpy(msg_dst, dump_message.msg, sizeof(dump_message.msg) > msg_dst_size ? msg_dst_size : sizeof(dump_message.msg));
+    return 1;
 }
--- a/src/common/bsod.h
+++ b/src/common/bsod.h
@@ -29,6 +29,10 @@ extern "C" {
 
 __attribute__((noreturn)) void _bsod(const char *fmt, const char *fine_name, int line_number, ...); //with file name and line number
 
+/** Fatal error that causes redscreen
+ * @param error - error message, that will be displayed as error description (MAX length 107 chars)
+ * @param module - module affected by error will be displayed as error title (MAX length 20 chars)
+*/
 void fatal_error(const char *error, const char *module);
 
 void ScreenHardFault(void);
--- a/src/common/crash_dump/dump.h
+++ b/src/common/crash_dump/dump.h
@@ -30,6 +30,9 @@ enum {
     DUMP_FLASH_SIZE = 0x00100000,
 };
 
+// DUMP constants for error message
+#define DUMP_MSG_TITLE_MAX_LEN 20
+#define DUMP_MSG_MAX_LEN       107
 // general registers stored to ccram
 // r0-r12, sp, lr, pc - 64 bytes
 // xpsr, fpcsr, PRIMASK, BASEPRI, FAULTMASK, CONTROL, MSP, PSP - 32 bytes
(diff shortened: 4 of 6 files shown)


src/common/bsod_gui.cpp
```diff
@@ -153,12 +153,8 @@ void fatal_error(const char *error, const char *module) {
     /// Decision tree to define error code
     using namespace Language_en;
     /// TODO share these strings (saves ~100 B of binary size)
-    if (strcmp(MSG_INVALID_EXTRUDER_NUM, error) == 0) {
-        *perror_code_short = 0;
-    } else if (strcmp("Emergency stop (M112)", error) == 0) {
+    if (strcmp("Emergency stop (M112)", error) == 0) {
         *perror_code_short = 510;
-    } else if (strcmp("Inactive time kill", error) == 0) {
-        *perror_code_short = 0;
```
mkbel, line 161: Why MSG_INVALID_EXTRUDER_NUM and Inactive time kill cases are removed?

mkbel: I know that it does nothing at the moment, but I would consider leaving it there in case perror_code_short becomes defined for those error states.

michalrudolf: It seems to me, that leaving it there with unassigned error code is not very clean. Once new error codes are assigned, then new cases will be created...

src/common/dump.h
```diff
@@ -30,6 +30,11 @@ enum {
     DUMP_FLASH_SIZE = 0x00100000,
 };
 
+// DUMP constants for error message
+#define DUMP_MSG_TITLE_MAX_LEN 20
+#define DUMP_MSG_MAX_LEN       107
+static const uint32_t DUMP_MSG_ADDR = 0x1000fe80;
```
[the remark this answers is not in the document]
michalrudolf: ```e2e5a09```

src/common/dump.cpp
```diff
@@ -44,6 +44,44 @@ void dump_to_xflash(void) {
...
+    strlcpy(msg_dst, dump_message.msg_content, sizeof(dump_message.msg_content));
+    return 1;
+}
+
+void dump_in_xflash_clear_message_flag(void) {
+    uint8_t dumpmessage_flag;
+    w25x_rd_data(w25x_error_start_adress, &dumpmessage_flag, 1);
+    if (!w25x_fetch_error() && dumpmessage_flag == 0) {
+        dumpmessage_flag = 0xFF;
+        w25x_program(w25x_error_start_adress, &dumpmessage_flag, 1);
```
mkbel, line 72: You can not write to the same flash location twice without erasing whole sector. You already written there 0 in dump_err_to_xflash().

src/common/dump.cpp
```diff
@@ -44,6 +44,44 @@ void dump_to_xflash(void) {
...
+    if (w25x_fetch_error())
+        return 0;
+    strlcpy(title_dst, dump_message.msg_title, sizeof(dump_message.msg_title));
+    strlcpy(msg_dst, dump_message.msg_content, sizeof(dump_message.msg_content));
+    return 1;
+}
+
+void dump_in_xflash_clear_message_flag(void) {
+    uint8_t dumpmessage_flag;
+    w25x_rd_data(w25x_error_start_adress, &dumpmessage_flag, 1);
```
mkbel, line 69: This is suspitious. As dumpmessage_t.flag is defined as unsigned short earlier.

src/common/dump.cpp
mkbel: I would suggest change
```
typedef struct _dumpmessage_t {
    unsigned short flag; // flags if message is valid (== 0x00) or not valid (!= 0x00)
    char msg_title[DUMP_MSG_TITLE_MAX_LEN];
    char msg_content[DUMP_MSG_MAX_LEN];
} dumpmessage_t;
```
to
```
typedef struct _dumpmessage_t {
    uint8_t invalid; // flags if message is valid (== 0x00) or invalid (!= 0x00)
    uint8_t not_displayed; // flags if message was not displayed (== 0xff) or displayed (== 0x00)
    char msg_title[DUMP_MSG_TITLE_MAX_LEN];
    char msg_content[DUMP_MSG_MAX_LEN];
} dumpmessage_t;
```
And then in `dump_err_to_xflash()` just write `.invalid` to 0 and don't touch `.not_displayed`. That way `.not_displayed` would be kept erased ==0xff meaning not displayed. You can switch it later in dump_in_xflash_clear_message_flag to 0, so `.not_displayed` would be turned to displayed.

src/common/dump.cpp
```diff
@@ -44,6 +44,44 @@ void dump_to_xflash(void) {
...
+    w25x_sector_erase(w25x_error_start_adress);
+    dump_message.flag = 0;
+    strlcpy(dump_message.msg_title, title, sizeof(dump_message.msg_title));
+    strlcpy(dump_message.msg_content, error, sizeof(dump_message.msg_content));
+    w25x_program(w25x_error_start_adress, (uint8_t *)(&dump_message), sizeof(dumpmessage_t));
+    w25x_fetch_error();
+}
+
+int dump_in_xflash_get_message(char *msg_dst, char *title_dst) {
+    dumpmessage_t dump_message;
```
mkbel, line 58: I would suggest this function to take `dumpmessage_t dump_message` by reference instead of taking pointers to char arrays. That way you avoid passing too small array by `char *msg_dst` or `char *title_dst` which may overflow, as the size is not passed nor checked.

Alternatively if you don't want to expose message flags, you can pass reference to fixed size arrays `char (&msg_dst)[DUMP_MSG_MAX_LEN]` and `char title_dst [DUMP_MSG_TITLE_MAX_LEN]` https://www.nextptr.com/question/a6212599/passing-cplusplus-arrays-to-function-by-reference

Alternatively you can use c-style and pass length of each array after each `char *msg_dst` and `char *title_dst` parameter.

src/common/dump.h
```diff
@@ -30,6 +30,10 @@ enum {
     DUMP_FLASH_SIZE = 0x00100000,
 };
 
+// DUMP constants for error message
+#define DUMP_MSG_TITLE_MAX_LEN 20
+#define DUMP_MSG_MAX_LEN       107
+static const uint32_t DUMP_MSG_SIZE = DUMP_MSG_TITLE_MAX_LEN + DUMP_MSG_TITLE_MAX_LEN + 1;
```
mkbel, line 36: note: DUMP_MSG_SIZE is never used anywhere and is wrong anyway as sizeof(unsigned short flag) is probably 2.

src/common/dump.cpp
mkbel: Or instead of flag `invalid` you can call it `version` and you can start with `version == 0` (with all the logic being same as if it would be `invalid`). That way if someone decides to alter `dumpmessage_t` he can increment `version`, so if there would be something stored with older version after firmware updatate, instead of displaying untangible nonsense later firmware would ignore data stored with older version format.

src/common/dump.cpp
michalrudolf: This feels utterly unnecessary, since strlcpy copies up to `sizeof(c-string ptr, not decayed to char ptr)` characters and it ensures, that it will be NULL-terminated.

There is no need of passed string size, because only size of buffer is important here. We want developers to write whatever msg they want. If it is longer - it will be truncated - if it is shorter (most of the time) - it will fill up only part of the buffered space and strlcpy ensures that.

src/common/dump.h
michalrudolf: That was a mistake, the flag should be uint8_t. Fixed in `7c21391`

src/common/dump.cpp
michalrudolf: Yes, sizes are passed in `7c21391`.

src/common/dump.cpp
michalrudolf: `7c21391`

src/common/dump.cpp
michalrudolf: Mistake, flag is now defined as uint8_t in `7c21391`

src/common/bsod_gui.cpp
```diff
@@ -178,11 +175,16 @@ void fatal_error(const char *error, const char *module) {
     } else if (strcmp(MSG_ERR_MINTEMP, error) == 0) {
         *perror_code_short = 208;
     } else {
-        *perror_code_short = 0;
+        // write error message to CCRAM - will be dumped to xflash
```
mkbel, line 178: This comment is not relevant anymore.

src/common/dump.cpp
```diff
@@ -44,6 +44,39 @@ void dump_to_xflash(void) {
...
+    w25x_rd_data(w25x_error_start_adress, (uint8_t *)(&dump_message), sizeof(dumpmessage_t));
+    if (w25x_fetch_error())
+        return 0;
+    strlcpy(tit_dst, dump_message.msg_title, sizeof(dump_message.msg_title));
+    strlcpy(msg_dst, dump_message.msg_content, sizeof(dump_message.msg_content));
+    return 1;
+}
+
+void dump_in_xflash_clear_message(void) {
+    w25x_sector_erase(w25x_error_start_adress);
```
mkbel, line 68: It looks like you are calling `w25x_sector_erase(w25x_error_start_adress)`
twice per error screen saved and displayed. This is probably lowering xFlash durability to half of what is achievable.

There are two possible solutions to that problem. 
1. Either you can read the whole sector, or at least `sizeof(dumpmessage_t)` in `dump_err_to_xflash()` and erase the sector only if it is already not erased.

or

2. You can implement what I already proposed earlier. Put two flags (`invalid`, `not_diplayed`) instead of just one (`not_diplayed`) into `dumpmessage_t` and switch them one by one. Make the message valid and not displayed in `dump_err_to_xflash()`. And make it displayed in `dump_in_xflash_clear_message()` without `w25x_sector_erase` call.

src/common/dump.cpp
```diff
@@ -44,6 +44,39 @@ void dump_to_xflash(void) {
...
+    dumpmessage_t dump_message;
+    w25x_sector_erase(w25x_error_start_adress);
+    dump_message.not_displayed = 0; // validate message
+    strlcpy(dump_message.msg_title, title, sizeof(dump_message.msg_title));
+    strlcpy(dump_message.msg_content, error, sizeof(dump_message.msg_content));
+    w25x_program(w25x_error_start_adress, (uint8_t *)(&dump_message), sizeof(dumpmessage_t));
+    w25x_fetch_error();
+}
+
+int dump_in_xflash_get_message(char *msg_dst, uint16_t msg_dst_size, char *tit_dst, uint16_t tit_dst_size) {
```
mkbel, line 57: It looks like parameters `msg_dst_size` and `tit_dst_size` are unused. It is big mystery for me why are you not getting compiler warning for that.

src/common/bsod_gui.cpp
michalrudolf: `e611ffc`

src/common/dump.cpp
michalrudolf: Oops, forgot that one. fixed in `e611ffc`

src/common/dump.cpp
```diff
@@ -230,3 +230,54 @@ int dump_hardfault_test_1(void) {
...
+
+void dump_err_to_xflash(const char *error, const char *title) {
+    // This is called after w25x initialization (in dumpinfo_t dumping), so no additional w25x_init() is needed
+    uint8_t valid_flag = dump_err_in_xflash_get_valid_flag();
+    if (valid_flag != 0xFF) {                                       // ==> message space is not empty
+        if (valid_flag != 0 || dump_err_in_xflash_is_displayed()) { // ==> erase sector if flag is corrupted (!= 0x00) or is valid and is displayed
+            w25x_sector_erase(w25x_error_start_adress);
+            w25x_fetch_error();
+        }
+    }
```
mkbel, line 244: I would suggest to remove all those conditions and simply always call w25x_sector_erase(w25x_error_start_adress).

Alternatively if you really want to avoid that one in printer lifetime redundant w25x_sector_erase call which happens only when newly manufactured printer wants to show error message for the first time. You would need to do this properly - read whole to be written area and check if the whole area is erased (contains 0xff).

src/common/dump.cpp
```diff
@@ -230,3 +230,54 @@ int dump_hardfault_test_1(void) {
...
+    dumpmessage_t dump_message;
```
mkbel, line 245: This is not initialized, later you are initializing fields `invalid`, `title` and `msg`. Field `not_displayed` is never initialized so it might contain any random data what was on stack.

src/common/dump.cpp
```diff
@@ -230,3 +230,54 @@ int dump_hardfault_test_1(void) {
...
+    dump_message.not_displayed = dump_message.invalid = 0;
+    strlcpy(dump_message.title, title, sizeof(dump_message.title));
+    strlcpy(dump_message.msg, error, sizeof(dump_message.msg));
+    w25x_program(w25x_error_start_adress, (uint8_t *)(&dump_message), sizeof(dumpmessage_t));
```
mkbel, line 249: Here you write whole dump_message. So any random data contained in `not_displayed` are written also. Even if you would initialize `not_displayed` field into `0xff` still requirement not to write the same memory location would be violated.

I would suggest to split w25x_program() into two or more calls. Write invalid flag, write messages, do not write anything else.

src/common/dump.cpp
michalrudolf: Yep, that will be a lot easier. It means, that printer will show the latest redscreen, not the first occured, but I'm fine with that. `9d28bd0`

src/common/dump.cpp
michalrudolf: It was initialized to 0, which is wrong of course, here is a fix `9d28bd0`

src/common/dump.cpp
michalrudolf: `9d28bd0`

src/common/crash_dump/dump.cpp
```diff
@@ -11,7 +11,7 @@ static constexpr uint32_t dump_offset = w25x_dump_start_address;
 static constexpr uint16_t dump_buff_size = 0x100;
 static constexpr uint32_t dump_xflash_size = DUMP_RAM_SIZE + DUMP_CCRAM_SIZE;
 
-static_assert(dump_xflash_size <= w25x_pp_start_address, "Dump overflows reserved space.");
+static_assert(dump_xflash_size <= w25x_error_start_adress, "Dump overflows reserved space.");
```
mkbel, line 14: Suggestion:
Add check that error is not overflowing reserved space:
e.g. `static_assert(sizeof(dumpmessage_t) <= (w25x_pp_start_address - w25x_error_start_adress), "Error message overflows reserved space.");`

mkbel (dismissed):

src/common/crash_dump/dump.cpp
michalrudolf: Good point, added in squashed commit

mkbel (approved):

PR #2529, opened 2022-09-27
Disable the corrupt-packet workarounds

vorner:
The whole need for pauses in transmission is now gone due to changes in the ESP firmware.

* Disable the copy-packet trick. That one apparently worked only because the packet got assembled into a bigger block instead of being composed from parts.
* Disable the pauses.

BFW-2664.


laloch (approved):
OK, thanks. ESP update incoming.

PR #2531, opened 2022-09-27
network: Add PostFile request and SocketConnectionFactory

Derevin:
Implements PostFile request and SocketConnectionFactory for the use case of generic sending a file via HTTP POST.

### DIFF
--- a/src/common/http/post_file_request.cpp
+++ b/src/common/http/post_file_request.cpp
@@ -1,0 +1,36 @@
+#include "post_file_request.hpp"
+
+namespace http {
+PostFile::PostFile(const char *file_path, const char *url_string, const char *file_size)
+    : hdrs {
+        { "Content-Length", file_size, std::nullopt },
+        { nullptr, nullptr, std::nullopt },
+    }
+    , url_string(url_string)
+    , file(fopen(file_path, "rb")) {
+}
+const char *PostFile::url() const {
+    return url_string;
+}
+ContentType PostFile::content_type() const {
+    return ContentType::ApplicationOctetStream;
+}
+Method PostFile::method() const {
+    return Method::Post;
+}
+const HeaderOut *PostFile::extra_headers() const {
+    return hdrs;
+}
+std::variant<size_t, Error> PostFile::write_body_chunk(char *data, size_t size) {
+    if (!file || !size) {
+        return Error::InternalError;
+    }
+
+    if (auto written = fread(data, 1, size, file.get()); ferror(file.get())) {
+        return Error::InternalError;
+    } else {
+        return written;
+    }
+}
+
+};
--- a/src/common/http/post_file_request.hpp
+++ b/src/common/http/post_file_request.hpp
@@ -1,0 +1,21 @@
+#pragma once
+#include "httpc.hpp"
+#include "unique_file_ptr.hpp"
+
+namespace http {
+class PostFile : public Request {
+public:
+    PostFile(const char *file_path, const char *url_string, const char *file_size);
+    virtual ~PostFile() = default;
+    const char *url() const override;
+    ContentType content_type() const override;
+    Method method() const override;
+    const HeaderOut *extra_headers() const override;
+    std::variant<size_t, Error> write_body_chunk(char *data, size_t size) override;
+
+private:
+    HeaderOut hdrs[2];
+    const char *url_string;
+    unique_file_ptr file;
+};
+};
--- a/src/common/http/socket_connection_factory.hpp
+++ b/src/common/http/socket_connection_factory.hpp
@@ -1,0 +1,21 @@
+#pragma once
+#include "httpc.hpp"
+#include "socket.hpp"
+
+namespace http {
+class SocketConnectionFactory : public ConnectionFactory {
+public:
+    SocketConnectionFactory(const char *host, uint16_t port);
+    virtual ~SocketConnectionFactory() = default;
+
+protected:
+    std::variant<Connection *, Error> connection() override;
+    const char *host() override;
+    void invalidate() override;
+
+private:
+    const char *hostname;
+    uint16_t port;
+    socket_con con;
+};
+};
--- a/src/common/http/CMakeLists.txt
+++ b/src/common/http/CMakeLists.txt
@@ -1,4 +1,10 @@
 target_sources(
-  firmware PRIVATE url_decode.cpp httpc.cpp socket.cpp resp_parser.cpp os_porting.cpp
-                   socket_connection_factory.cpp
+  firmware
+  PRIVATE url_decode.cpp
+          httpc.cpp
+          socket.cpp
+          resp_parser.cpp
+          os_porting.cpp
+          socket_connection_factory.cpp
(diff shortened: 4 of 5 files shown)


src/common/http/post_file_request.hpp
```diff
@@ -0,0 +1,21 @@
...
+public:
+    PostFile(const char *file_path, const char *url_string, const char *file_size);
+    virtual ~PostFile() = default;
+    const char *url() const override;
+    ContentType content_type() const override;
+    Method method() const override;
+    const HeaderOut *extra_headers() const override;
+    std::variant<size_t, Error> write_body_chunk(char *data, size_t size) override;
+
+protected:
```
vorner, line 16: Why not private?

src/common/http/post_file_request.hpp
```diff
@@ -0,0 +1,21 @@
+#pragma once
+#include "httpc.hpp"
+#include "unique_file_ptr.hpp"
+
+namespace http {
+class PostFile : public Request {
+public:
+    PostFile(const char *file_path, const char *url_string, const char *file_size);
```
vorner, line 8: Passing the file size as string is… interesting. I probably see why you do that, but feels like this needs a solution.

I won't insist on solving it inside this PR.

src/common/http/post_file_request.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+    size_t written { 0 };
+    while (written < size) {
+        char ch = fgetc(file.get());
+        if (feof(file.get())) {
+            break;
+        }
+        data[written] = ch;
+        ++written;
+    }
```
vorner, line 37: Huh? Why are you doing it by one-character reads? That's probably _very_ ineficient.

src/common/http/socket_connection_factory.hpp
```diff
@@ -0,0 +1,20 @@
...
+namespace http {
+class SocketConnectionFactory : public ConnectionFactory {
+public:
+    SocketConnectionFactory(const char *host, uint16_t port);
+    virtual ~SocketConnectionFactory() = default;
+
+protected:
+    std::variant<Connection *, Error> connection() override;
+    const char *host() override;
+    void invalidate() override;
```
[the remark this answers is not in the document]
Derevin: Yes, clang-tidy in default settings even warns about using `virtual` when `override` is the only thing needed

src/common/http/socket_connection_factory.hpp
```diff
@@ -0,0 +1,20 @@
+#pragma once
+#include "httpc.hpp"
+#include "socket.hpp"
+
+namespace http {
+class SocketConnectionFactory : public ConnectionFactory {
+public:
+    SocketConnectionFactory(const char *host, uint16_t port);
+    virtual ~SocketConnectionFactory() = default;
```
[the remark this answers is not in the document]
Derevin: Not technically, it adds some clarity to all who might derive from this class.. It's a toin coss for me, I don't have an opinion on this

src/common/http/socket_connection_factory.hpp
```diff
@@ -0,0 +1,20 @@
...
+
+protected:
+    std::variant<Connection *, Error> connection() override;
+    const char *host() override;
+    void invalidate() override;
+
+    const char *hostname;
+    uint16_t port;
+    socket_con con;
```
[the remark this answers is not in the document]
Derevin: Personally, my practice is to pretty much 'almost forget the keyword private exists'. It allows me ease of programming while my classes get derived. Or when I eventually decide "Hey this class actually makes sense to not be final". No strong opinion on this and I could go back to 'including private into my dictionary' :smile:

src/common/http/post_file_request.cpp
Derevin: You're right. I should rethink this.

src/common/http/post_file_request.hpp
Derevin: I completely get that it doesn't look nor feel ideal, and it was a coin toss for me again. Either it can be done this way, which allows the class to be 'clean', it could be accepted as an int and parsed to string "just in time" or parsed to a class-level char buffer.
Neither feels nice. Maybe some different way to handle headers, so that the second argument accepts whatever and parses it to strings during an actual transaction? Just a random thought.

src/common/http/socket_connection_factory.hpp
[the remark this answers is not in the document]
vorner: Does it? Doesn't seem to be doing that for us 😇. Ok, learning something new.

src/common/http/socket_connection_factory.hpp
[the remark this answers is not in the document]
vorner: So long as you're doing it on purpose, I'm fine with it :shrug:

src/common/http/post_file_request.hpp
vorner: > parses it to strings

To clarify… you mean „render“?

I was thinking about putting some `std::variant<const char *, size_t>` into `HeaderOut`. I'm not sure how much work it would be to do that ‒ it would be cleaner, but I if it takes another day of work, then it's not worth it.

src/common/http/post_file_request.cpp
vorner: `fread` should do what you want out of the box 😇.

src/common/http/post_file_request.hpp
Derevin: > To clarify… you mean „render“?

The verb i use for doing something to/from strings is 'parse'. So... maybe?

I did the `std::variant` you proposed, it feels clean and it was easy to implement, so I did it.

src/common/http/httpc.cpp
```diff
@@ -187,8 +187,17 @@ optional<Error> HttpClient::send_request(const char *host, Connection *conn, Req
...
     for (const HeaderOut *extra_hdrs = request.extra_headers() ?: &term; extra_hdrs->name; extra_hdrs++) {
-        CHECKED(buffer.header(extra_hdrs->name, extra_hdrs->value, extra_hdrs->size_limit));
+        if (std::holds_alternative<const char *>(extra_hdrs->value)) {
+            CHECKED(buffer.header(extra_hdrs->name, std::get<const char *>(extra_hdrs->value), extra_hdrs->size_limit));
+        } else if (std::holds_alternative<size_t>(extra_hdrs->value)) {
+            snprintf(buff, buff_size, "%u", std::get<size_t>(extra_hdrs->value));
+            CHECKED(buffer.header(extra_hdrs->name, buff, extra_hdrs->size_limit));
+        } else {
+            return Error::InternalError;
+        }
```
vorner, line 200: Can we use some kind of `std::visit` that guarantees all variants are handled in constant time?

vorner (approved):

src/common/http/httpc.cpp
Derevin: Continued in https://github.com/prusa3d/Prusa-Firmware-Buddy/pull/2556

PR #2533, opened 2022-09-27
Marlin 2.1.x G28 incremental update

wavexx:
Update G28 from Marlin 2.1.x:

* Marlin/src/gcode/calibrate/G28: preserves changes for broken cable detection, alternate feedrate calculations and other rather minor code changes.
* Conditionals_LCD|post: update for the few new predicates required
* src/module/motion: stub a few functions to keep G28 clean

As before, unused defines are pulled as-is to reduce changes from Marlin2 and are assumed to be non-working until the transition is complete.

The resulting differences in G28 from 2.1.x are now easily diffable.


lib/Marlin/Marlin/src/gcode/calibrate/G28.cpp
```diff
@@ -293,23 +387,34 @@ void GcodeSuite::G28_no_parser(bool always_home_all, bool O, float R, bool S, bo
...
 
   #if ENABLED(DELTA)
 
+    constexpr bool doZ = true; // for NANODLP_Z_SYNC if your DLP is on a DELTA
+
     home_delta();
-    UNUSED(always_home_all);
 
-  #else // NOT DELTA
+    TERN_(IMPROVE_HOMING_RELIABILITY, end_slow_homing(saved_motion_state));
```
mkbel, line 423: This is supitious, as code above:
```
  #if ENABLED(IMPROVE_HOMING_RELIABILITY)
    slow_homing_t slow_homing{0};
    slow_homing.acceleration.set(planner.settings.max_acceleration_mm_per_s2[X_AXIS],
                                 planner.settings.max_acceleration_mm_per_s2[Y_AXIS]);
    planner.settings.max_acceleration_mm_per_s2[X_AXIS] = XY_HOMING_ACCELERATION;
    planner.settings.max_acceleration_mm_per_s2[Y_AXIS] = XY_HOMING_ACCELERATION;
    #if HAS_CLASSIC_JERK
      slow_homing.jerk_xy = planner.max_jerk;
      planner.max_jerk.set(XY_HOMING_JERK, XY_HOMING_JERK);
    #endif
    planner.reset_acceleration_rates();
  #endif
```
was not altered by this commit. I am not sure this change is compatible with the mentioned code above.

lib/Marlin/Marlin/src/gcode/calibrate/G28.cpp
```diff
@@ -448,36 +567,28 @@ void GcodeSuite::G28_no_parser(bool always_home_all, bool O, float R, bool S, bo
...
+      TERN_(IMPROVE_HOMING_RELIABILITY, saved_motion_state = begin_slow_homing());
 
       // Always home the 2nd (right) extruder first
       active_extruder = 1;
-      homeaxis(X_AXIS
-        #if ENABLED(PRECISE_HOMING)
-          , 0.0f, false, !parser.seen('D')
-        #endif
-      );
+      homeaxis(X_AXIS);
```
mkbel, line 592: Looks like PRECISE_HOMING 'D' - do not calibrate feature is dropped here.

lib/Marlin/Marlin/src/gcode/calibrate/G28.cpp
```diff
@@ -448,36 +567,28 @@ void GcodeSuite::G28_no_parser(bool always_home_all, bool O, float R, bool S, bo
...
+      inactive_extruder_x = current_position.x;
 
       // Home the 1st (left) extruder
       active_extruder = 0;
-      homeaxis(X_AXIS
-        #if ENABLED(PRECISE_HOMING)
-          , 0.0f, false, !parser.seen('D')
-        #endif
-      );
+      homeaxis(X_AXIS);
```
mkbel, line 599: Looks like PRECISE_HOMING 'D' - do not calibrate feature is dropped here.

lib/Marlin/Marlin/src/gcode/calibrate/G28.cpp
wavexx: Intentional. We have no idea or ever tested this on any non-cartesian setup. At the same time, this just creates potential for conflicts with future merges. For these reasons I rather ignore these changes or flatly fail to compile if one attempts to enable it.

I would love to change the prototype of homeaxis to something that doesn't require these changes as an exception, but I will do so in a separate PR.

lib/Marlin/Marlin/src/gcode/calibrate/G28.cpp
wavexx: IMPROVE_HOMING_RELIABILITY is currently disabled, and it breaks with this PR as I'm keeping this just about G28 to reduce the amount of changes to review. I preserved the changes to it in this case.

I already completed IMPROVE_HOMING_RELIABILITY but this spans planner/motion, will be submitted in another PR.

lib/Marlin/Marlin/src/gcode/calibrate/G28.cpp
wavexx: As above. I'm not against the change itself, but I don't like how we specialize each call. Fixing this now would be a mistake though, so I'm waiting until all of 2.1.x is merged before we do cosmetic fixes of this sort.

lib/Marlin/Marlin/src/gcode/calibrate/G28.cpp
mkbel: IMPROVE_HOMING_RELIABILITY is enabled on mk4. It was enabled just in mk4 branch but today when we started new mk4 release we noticed this unmerged and is merged now.

wavexx: I wasn't aware. In this case, I added an additional commit that completes IMPROVE_HOMING_RELIABILITY as a whole.

lib/Marlin/Marlin/src/gcode/calibrate/G28.cpp
```diff
@@ -58,8 +63,16 @@ static inline void MINDA_BROKEN_CABLE_DETECTION__END() {}
 
 #include "../../lcd/ultralcd.h"
 
-#if HAS_DRIVER(L6470)                         // set L6470 absolute position registers to counts
-  #include "../../libs/L6470/L6470_Marlin.h"
+#if ENABLED(EXTENSIBLE_UI)
+  //#include "../../lcd/extui/ui_api.h"
```
mkbel, line 67: Why this is commented out?

In case this file doesn't exist I would rather the whole part of the patch not applied. That way it becomes updated later also with that file present.

wavexx: Unfortunately EXTENSIBLE_UI exists both in old and new and the API has diverged quite a bit. I kept the ifdef as it does apply cleanly in more cases, but I had to comment 3 places where ExtUI is used.

lib/Marlin/Marlin/src/module/endstops.cpp
```diff
@@ -87,6 +87,9 @@ Endstops::esbits_t Endstops::live_state = 0;
 #if ENABLED(SPI_ENDSTOPS)
   Endstops::tmc_spi_homing_t Endstops::tmc_spi_homing; // = 0
 #endif
+#if ENABLED(IMPROVE_HOMING_RELIABILITY)
+  millis_t sg_guard_period; // = 0
+#endif
```
mkbel, line 92: I intentionally removed it in https://github.com/prusa3d/Prusa-Firmware-Buddy/pull/2276/files

wavexx: I introduced an extra define HOMING_SG_GUARD_DURATION (which we don't use) to elide the code instead.

lib/Marlin/Marlin/src/module/planner.cpp
```diff
@@ -1372,6 +1372,34 @@ void Planner::check_axes_activity() {
...
+    else {
+      settings.max_acceleration_mm_per_s2[X_AXIS] = saved_motion_state.acceleration.x;
+      settings.max_acceleration_mm_per_s2[Y_AXIS] = saved_motion_state.acceleration.y;
+      TERN_(DELTA, settings.max_acceleration_mm_per_s2[Z_AXIS] = saved_motion_state.acceleration.z);
+      TERN_(HAS_CLASSIC_JERK, max_jerk = saved_motion_state.jerk_state);
+    }
+    refresh_acceleration_rates();
+  }
+
+#endif
```
mkbel, line 1401: note:
Looks partially duplicate to:

```
/**
 * Class provides storage of speed and acceleration parameters of the motion.
 */
class Motion_Parameters {
  public:
    // save motion parameters
    void save();
    // save motion parameters and reset them
    void save_reset();
    // reset motion parameters
    static void reset();
    // load motion parameters back
    void load();
  private:
    motion_parameters_t mp;
};
/**
 * Class provides temporary storage of speed and acceleration parameters of the motion.
 * Constructor saves parameters and resets them, destructor retrieves them back.
 */
class Temporary_Reset_Motion_Parameters {
  public:
    Temporary_Reset_Motion_Parameters() {
      mp.save();
      mp.reset();
    }
    ~Temporary_Reset_Motion_Parameters() {
      mp.load();
    }
  private:
    Motion_Parameters mp;
};
```
and its usage in 
lib/Marlin/Marlin/src/feature/prusa/homing.cpp:
Motion_Parameters reset_acceleration_if(bool condition) {

mkbel (approved):

lib/Marlin/Marlin/src/module/planner.cpp
wavexx: Yes, I noticed. I'm not touching motion unless necessary in this PR.

PR #2540, opened 2022-09-29
homing: Set configured homing sensitivity on X motor

wavexx:
Looks like that when CRASH_RECOVERY is enabled, homing_sensitivity wasn't set for stepperX.

### DIFF
--- a/lib/Marlin/Marlin/src/module/motion.cpp
+++ b/lib/Marlin/Marlin/src/module/motion.cpp
@@ -1153,6 +1153,9 @@ feedRate_t get_homing_bump_feedrate(const AxisEnum axis) {
           #endif
 
           stealth_states.x = tmc_enable_stallguard(stepperX);
+          #if ENABLED(CRASH_RECOVERY)
+            stepperX.stall_sensitivity(crash_s.home_sensitivity[0]);
+          #endif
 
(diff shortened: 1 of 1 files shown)


lib/Marlin/Marlin/src/module/motion.cpp
```diff
@@ -1153,6 +1153,9 @@ feedRate_t get_homing_bump_feedrate(const AxisEnum axis) {
           #endif
 
           stealth_states.x = tmc_enable_stallguard(stepperX);
+          #if ENABLED(CRASH_RECOVERY)
+            stepperX.stall_sensitivity(crash_s.home_sensitivity[0]);
+          #endif
```
mkbel, line 1158: This code is not supposed to be there. I recently replaced it with `crash_s.start_sensorless_homing_per_axis(axis);`
This also applies for Y axis.

I forgot to remove 

```
          #if ENABLED(CRASH_RECOVERY)
            stepperY.stall_sensitivity(crash_s.home_sensitivity[1]);
          #endif
```

wavexx: Do you want me to make a PR for this?

wavexx:
Replaced by #2547

PR #2568, opened 2022-10-05
Connect: Reasons in Rejected events

vorner:
Provide the (debug) reason for rejected events when complaining to the server about something.

BFW-2828.

### DIFF
--- a/src/connect/planner.cpp
+++ b/src/connect/planner.cpp
@@ -178,19 +178,19 @@ void Planner::action_done(ActionResult result) {
 }
 
 void Planner::command(const Command &command, const UnknownCommand &) {
-    planned_event = Event { EventType::Rejected, command.id };
+    planned_event = Event { EventType::Rejected, command.id, nullopt, nullopt, "Unknown command" };
 }
 
-void Planner::command(const Command &command, const BrokenCommand &) {
-    planned_event = Event { EventType::Rejected, command.id };
+void Planner::command(const Command &command, const BrokenCommand &c) {
+    planned_event = Event { EventType::Rejected, command.id, nullopt, nullopt, c.reason };
 }
 
 void Planner::command(const Command &command, const GcodeTooLarge &) {
-    planned_event = Event { EventType::Rejected, command.id };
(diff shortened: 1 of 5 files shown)


tests/unit/connect/render.cpp
```diff
@@ -16,8 +16,9 @@ using namespace json;
 namespace {
 
 constexpr const char *const print_file = "/usb/box.gco";
-constexpr const char *const rejected_event_printing = "{\"state\":\"PRINTING\",\"command_id\":11,\"event\":\"REJECTED\"}";
+constexpr const char *const rejected_event_printing = "{\"reason\":\"Job ID doesn't match\",\"state\":\"PRINTING\",\"command_id\":11,\"event\":\"REJECTED\"}";
```
Derevin, line 19: The last `const` is already implied by `constexpr`, to my knowledge `constexpr const char *` is exactly the same as `constexpr const char *const` except shorter

src/connect/planner.cpp
```diff
@@ -205,25 +205,35 @@ void Planner::command(const Command &command, const Gcode &gcode) {
...
+    const char *reason = nullptr;
+    if (!path_allowed(path)) {
+        reason = "Forbidden path";
+    } else if (!path_exists(path)) {
+        reason = "File not found";
+    } else if (!printer.start_print(path)) {
+        reason = "Can't print now";
+    }
+
+    if (reason == nullptr) {
```
Derevin, line 233: Generally I think checking for `nullptr` directly with `if (reason == nullptr)` => `if (!reason)` and vice versa is cleaner. Just like one doesn't write `if (boolean_flag == false)` but `if (!boolean_flag)`.

tests/unit/connect/render.cpp
Derevin, line 19: There are `raw string literals` in c++ which allow cleaner writing of const JSON strings, so that you don't have to escape every `"` in the string
`R"(string that doesn't need to  be escaped and i can put as many " as i want in it)"`

Derevin (approved):
Just some nitpicks. Already giving it an approve, up to you whether you want to incorporate the suggested changes :)

PR #2574, opened 2022-10-05
Marlin 2.1.x: incremental endstops update

wavexx:
- Stub SERIAL_ECHOF/SERIAL_ECHONLF to avoid updating serial.*
- Similarly, avoid using SERIAL_ECHOPAIR for the moment.
- Update M119 along the way


lib/Marlin/Marlin/src/module/endstops.cpp
```diff
@@ -949,11 +1592,117 @@ void Endstops::update() {
...
+      #if HAS_W_MIN
+        ES_REPORT_CHANGE(W_MIN);
+      #endif
+      #if HAS_W_MAX
+        ES_REPORT_CHANGE(W_MAX);
+      #endif
+
       SERIAL_ECHOLNPGM("\n");
-      analogWrite(pin_t(LED_PIN), local_LED_status);
+      hal.set_pwm_duty(pin_t(LED_PIN), local_LED_status);
```
mkbel, line 1639: Do we have now hal.set_pwm_duty() ?

wavexx: This is part of PINS_DEBUGGING, so it's dead code for us as of now. It's going to be there when we complete the update.

mkbel: Ok but I suggest let this not updated with this partial update PR, this will be updated when complete marlin update will be merged.

wavexx:
Since we're bringing this all at once, closing these incremental merges.

PR #2585, opened 2022-10-07
Marlin 2.1.x incremental motion endstops update

wavexx:
This updates HAS_SOFTWARE_ENDSTOPS/HAS_ENDSTOPS code in motion from Marlin 2.1.x.

- Stacked on top of PR #2574 
- Includes updated axis homed/trusted bits and stubs to keep surrounding code working
- Preserves WORKSPACE_HOME changes in set_axis_is_at_home


lib/Marlin/Marlin/src/module/motion.cpp
```diff
@@ -1166,44 +1166,31 @@ void prepare_move_to_destination() {
...
-      );
+      char msg[30];
+      #define _AXIS_CHAR(N) TEST(axis_bits, _AXIS(N)) ? STR_##N : ""
+      sprintf_P(msg, home_first, MAPLIST(_AXIS_CHAR, MAIN_AXIS_NAMES));
       SERIAL_ECHO_START();
       SERIAL_ECHOLN(msg);
-      #if HAS_DISPLAY
-        ui.set_status(msg);
-      #endif
+      ui.set_status(msg);
```
mkbel, line 1193: Isn't there any side effect with this change? Is ui.set_status() always available?

wavexx: In marlin 2.1.x the display logic is more decoupled from the LCD. 2.1.x now always calls the "ui" abstraction, which then dispatches the call to the real display. As such, the UI calls are generally always done, so that code that is completely headless (say, remote display) can still catch the UI and redirect it as needed.

For us there's no functional change, since HAS_DISPLAY was always already set.

wavexx:
Since we're bringing this all at once, closing these incremental merges.

PR #2594, opened 2022-10-10
network: Add PostLambda request

Derevin:
This PR adds a PostLambda request, which allows "write_body_chunk" to be given by user via a stateful lambda.

### DIFF
--- a/src/common/http/post_lambda_request.cpp
+++ b/src/common/http/post_lambda_request.cpp
@@ -1,0 +1,27 @@
+#include "post_lambda_request.hpp"
+
+namespace http {
+PostLambda::PostLambda(LambdaT lam, const char *url_string)
+    : url_string(url_string)
+    , lamb(lam) {
+}
+const char *PostLambda::url() const {
+    return url_string;
+}
+ContentType PostLambda::content_type() const {
+    return ContentType::ApplicationOctetStream;
+}
+Method PostLambda::method() const {
+    return Method::Post;
+}
+
+std::variant<size_t, Error> PostLambda::write_body_chunk(char *data, size_t size) {
+    assert(size && lamb);
+    if (!size || !lamb) {
+        return Error::InternalError;
+    }
+
+    return lamb(data, size);
+}
+
+};
--- a/src/common/http/post_lambda_request.hpp
+++ b/src/common/http/post_lambda_request.hpp
@@ -1,0 +1,25 @@
+#pragma once
+#include "httpc.hpp"
+#include <functional>
+
+namespace http {
+/**
+* @brief Post Request with 'write_body_chunk' given by user via a (potentially stateful) lambda
+*
+*/
+class PostLambda : public Request {
+public:
+    using LambdaT = std::function<std::variant<size_t, Error>(char *, size_t)>;
+
+    PostLambda(LambdaT lambda, const char *url_string);
+    virtual ~PostLambda() = default;
+    const char *url() const override;
+    ContentType content_type() const override;
+    Method method() const override;
+    std::variant<size_t, Error> write_body_chunk(char *data, size_t size) override;
+
+private:
+    const char *url_string;
+    LambdaT lamb;
+};
+};
--- a/src/common/http/CMakeLists.txt
+++ b/src/common/http/CMakeLists.txt
@@ -3,6 +3,7 @@ target_sources(
   PRIVATE httpc.cpp
           os_porting.cpp
           post_file_request.cpp
+          post_lambda_request.cpp
           resp_parser.cpp
           socket.cpp
           socket_connection_factory.cpp


src/common/http/post_lambda_request.cpp
```diff
@@ -0,0 +1,31 @@
+#include "post_lambda_request.hpp"
+
+namespace http {
+PostLambda::PostLambda(LambdaT lam, const char *url_string)
+    : hdrs {
+        { nullptr, nullptr, std::nullopt },
+    }
```
vorner, line 7: This is not needed. You're allowed to return `nullptr` from `extra_headers` and it'll do what you expect (even then you wouldn't have to keep it in the instance). In addition, it's what the default implementation does, so even that one is not needed.

src/common/http/post_lambda_request.cpp
```diff
@@ -0,0 +1,31 @@
...
+Method PostLambda::method() const {
+    return Method::Post;
+}
+const HeaderOut *PostLambda::extra_headers() const {
+    return hdrs;
+}
+std::variant<size_t, Error> PostLambda::write_body_chunk(char *data, size_t size) {
+    if (!size || !lamb) {
+        return Error::InternalError;
+    }
```
vorner, line 26: I think you'd prefer to assert on these things, because they are likely bugs, not errors to expect and somehow deal with at runtime.

src/common/http/post_lambda_request.hpp
```diff
@@ -0,0 +1,27 @@
...
+#include <functional>
+
+namespace http {
+/**
+* @brief Post Request with 'write_body_chunk' given by user via a lambda
+*
+*/
+class PostLambda : public Request {
+public:
+    using LambdaT = std::function<std::variant<size_t, Error>(char *, size_t)>;
```
vorner, line 12: Is that thing using dynamic allocation internally? If so, is it better to have the class as a template?

vorner (commented):
Apart from the suggestions to trim it down. I have a general worry, even more so after reading the code. This seems like a layer that doesn't solve any particular problem and tends to make the code overall harder to read (having to dig through the layers that do nothing) ‒ the so called lasagna code and some kind of code bloat.

I won't block the PR on that regard, but I'll ask you to think again about it, if you really find this helps you in making the code more readable as opposed to less, and if it serves an actual purpose.

src/common/http/post_lambda_request.hpp
Derevin: I want to allow the user to pass in a stateful lambda, so it needs  to be std::function, and std::function cannot be a template parameter

src/common/http/post_lambda_request.cpp
Derevin: I agree partially, I will add the assert but I think the check should still be there (better to 'not do anything' than 'crash when invoking a nullptr')

Derevin:
> I won't block the PR on that regard, but I'll ask you to think again about it, if you really find this helps you in making the code more readable as opposed to less, and if it serves an actual purpose.

Personally I think this will allow the user to 'keep relevant code in one place' rather than having to create a one-purpose requests for anything and everything

Here's a use case I'm trying to solve with this (work in progress):
I am trying to send a dump to a server. This is a one-purpose code request (there's 0 chance of the code being reused elsewhere). 

So I have two options:
- Create the request hpp and cpp, add it to cmakelist, implement it... 
- Implement how the request will be parsed right in the place that actually creates the request

In my opinion, I think it will be less "digging through layers" if the code is right there where it's relevant. I can tag you in the PR that uses this (once it exists) and we can reconsider based on the real example :)

src/common/http/post_lambda_request.hpp
vorner: Huh? Why couldn't you use that stateful lambda without first wrapping it into a `std::function`?

Ah, I see. It could be a template parameter, but one can't _name_ the parameter. So, you couldn't do:

```cpp
PostLambda<???> post([&](){ ... });
```

Because that `???` type can't be named. So you'd need a helper function and would have to do something like

```cpp
auto post(make_post([&](){ ... });
```

Which starts getting a bit annoying. I do believe you can have `std::function<whatever>` as a template parameter, but then, there's no actual point, because you'd still have the `std::function`.

vorner (approved):
Well, I've always considered that single-purpose request class to be some kind of „lambda with additional tags on it“ without the special syntax of lambda.

For that, I'd assume it would _not_ go into a separate cpp file, cmake, whatever, but would be local to where it is used, either in an anonymous namespace or even inside the local function. This is not Java, it's not like every class needs to have its own file to live in 😈. Yes, doing it the „class“ way is a bit less comfortable than just doing a lambda. On the other hand, I think this class will be used _exactly once_, because you hardcode no extra headers, that it is Post, that it's `application/octet-stream`…

PR #2601, opened 2022-10-11
Revert "network: Add PostLambda request"

Derevin:
Reverts prusa3d/Prusa-Firmware-Buddy#2594

In the previous PR I argued heavily for the introduction of PostLambda.

After having it merged and doing tests with the whole thing and thought about it a little, I found out that in the scenario where instead of doing a `PostLambda` you create a one-purpose `Request` right there in the source code, not only is the code slightly more readable (due to better encapsulation), you also get more freedom in handling the other parameters (headers, method....). 

That would be okay and there still would be a use case for the `PostLambda` (i think). However I found out that there is also a significant code-size difference (260 bytes in 50ish line write_body_chunk)), plus the fact that std::function doesn't have to be used (and all that comes with that).

All in all i think my decision to push for PostLambda was done in error and I think the previous PR should be reverted as to have us rid of this 'ineffective' abomination. :smile: 

Better late than too late :innocent:


vorner (approved):
Don't stress about it :-). Sometimes, the best way to explore is to try things out.

PR #2623, opened 2022-10-19
Connection timeout

Derevin:
Socket timeout is now made as a constructor parameter for `http::Connection`. This allows it to be specified per use case rather than once globally. 

The dump server has gotten it's response changed to 204. This makes the transaction 'clean'.

Also subsequently increase the timeout of crash dump sending to 6s.

### DIFF
--- a/src/common/http/connection.cpp
+++ b/src/common/http/connection.cpp
@@ -1,0 +1,10 @@
+#include "connection.hpp"
+
+namespace http {
+
+Connection::Connection(uint8_t timeout_s)
+    : timeout_s(timeout_s) {}
+
+uint8_t Connection::get_timeout_s() const { return timeout_s; }
+
+}
--- a/utils/dumpserver/server.py
+++ b/utils/dumpserver/server.py
@@ -28,4 +28,4 @@ def post_dump():
     with open(metadata_filepath, 'w') as f:
         f.write(json.dumps(metadata))
 
-    return 'ok'
+    return '', 204
--- a/src/common/crash_dump/crash_dump_distribute.cpp
+++ b/src/common/crash_dump/crash_dump_distribute.cpp
@@ -11,6 +11,8 @@ LOG_COMPONENT_DEF(CrashDump, LOG_SEVERITY_INFO);
 
 namespace crash_dump {
 
+inline constexpr uint8_t socket_timeout_s { 6 };
+
 bool escape_url_string(std::span<char> escaped_url_string, const std::array<char, url_buff_size> &url_buff) {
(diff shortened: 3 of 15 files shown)


utils/dumpserver/server.py
```diff
@@ -28,4 +28,4 @@ def post_dump():
     with open(metadata_filepath, 'w') as f:
         f.write(json.dumps(metadata))
 
-    return 'ok'
+    return '', 204
```
vorner, line 31: Not sure if this should be `''` (empty string/empty body) or something like `None` (no body at all). Maybe it doesn't matter in this particular python library or maybe it doesn't make the distinction, but I could imagine there being a conceptual difference 😇.

src/common/http/connection.cpp
```diff
@@ -0,0 +1,10 @@
+#include "connection.hpp"
+
+namespace http {
+
+Connection::Connection(uint8_t timeout_s)
+    : timeout_s(timeout_s) {}
+
+uint8_t Connection::get_timeout_s() const { return timeout_s; }
```
vorner, line 8: Considering how trivial both these are, does it make sense to have them inline in the class?

src/common/crash_dump/crash_dump_distribute.hpp
```diff
@@ -13,6 +13,8 @@ inline constexpr uint16_t port { 8888 };                  // temporary port
 
 inline constexpr size_t url_buff_size { 128 };
 
+inline constexpr uint8_t socket_timeout_s { 6 };
```
vorner, line 16: Shall these constants be publicly available in the header file (both of them), or are they more of an implementation fine-tuning parameters the caller is not interested in?

vorner (commented):
I have some style nitpicks there. None of them are blocking issues, so if you prefer to keep it as it is, I'm fine with that, just pointing them out for your attention.

utils/dumpserver/server.py
Derevin: using `''` seems to be the general 'way' of returning 204 for flask... or at least that's quick google / stack overflow thinks :smile:

src/common/crash_dump/crash_dump_distribute.hpp
Derevin: The reason I've put them in the header is so that 'they're easy to find'.. but I see your point and I will change it accordingly, it makes sense

src/common/http/connection.cpp
vorner: I don't know if we have a consensus or a rule. I use it based on how I feel at the time and nobody seemed to try beating mi with an existing rule or anything, so :shrug:

PR #2632, opened 2022-10-21
Initialize translations to a valid default even if empty

DRracer:
A safety precaution - initialize the translations array with valid pointers even if the cells are empty - prevents accessing invalid pointers in some edge cases.

### DIFF
--- a/src/lang/translator.cpp
+++ b/src/lang/translator.cpp
@@ -15,7 +15,7 @@ ProviderRegistrator prEN("en", EmptyTranslationProvider::Instance());
 Translations::Translations()
     : currentProvider(EmptyTranslationProvider::Instance()) {
     // this is a safety precaution - initialize the array with valid pointers even if the cells are empty
-    std::fill(translations.begin(), translations.end(), TranRec({ currentProvider, 0U }));
+    translations.fill(TranRec({ currentProvider, 0U }));
 }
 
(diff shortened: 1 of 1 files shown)


src/lang/translator.cpp
```diff
@@ -14,6 +14,8 @@ ProviderRegistrator prEN("en", EmptyTranslationProvider::Instance());
 // even if there were no registered languages at all
 Translations::Translations()
     : currentProvider(EmptyTranslationProvider::Instance()) {
+    // this is a safety precaution - initialize the array with valid pointers even if the cells are empty
+    std::fill(translations.begin(), translations.end(), TranRec({ currentProvider, 0U }));
```
Derevin, line 18: I'd suggest using `std::ranges::fill(translations, TranRec({ currentProvider, 0U }));`
However this change has an interesting effect on code size; release gets smaller by 16B (yay) debug size increases by 16B (ouch). Though I'd still argue that 16B is a good price to pay for readability

Derevin (approved):

src/lang/translator.cpp
DRracer: That's exactly why I gave it to you - to find the balance between code readability and size, thanks :+1: . I'm not used to ranges (yet).

Derevin (approved):

PR #2643, opened 2022-10-24
Merge release

MistrKernnunos:


lib/WUI/CMakeLists.txt
```diff
@@ -31,10 +31,13 @@ target_sources(
             nhttp/status_page.cpp
             nhttp/upload_state.cpp
             random.cpp
+            random.cpp
+            sntp/sntp.c
             sntp/sntp.c
             sntp/sntp.c
             sntp/sntp_client.c
             sntp/sntp_client.c
+            sntp/sntp_client.c
```
vorner, line 40: These are kind of plaguing the file, aren't they? How came we have `random` twice, and the `sntp` things _three times_? :-O

vorner (changes requested):

lib/WUI/CMakeLists.txt
MistrKernnunos: Should be fixed now

vorner (approved):

PR #2645, opened 2022-10-24
homing: Preserve correct SG threhold

wavexx:
Call crash_s.start_sensorless_homing_per_axis after tmc_enable_stallguard, as the latter resets the SG threshold.


mkbel:
I don't see any reason why we should change the order of setting sensitivity and enabling the stallguard. So I recommend to close this PR.

wavexx:
Seconded

PR #2647, opened 2022-10-24
Disable translations temporarily

vorner:
To free some flash space, before we have some more actual savings ready.


Eldenroot:
Like some other 3D printer producers - they offer one firmware file for one language (with English as a fallback). This seems to be a good idea, reflashing is fast and user friendly.

PR #2671, opened 2022-11-01
Connect & gui

vorner:
Make the status line production menu (non-green), translated.

Provide a menu item to load connect settings (only).

BFW-2801.


michalrudolf (approved):
Looks ok

PR #2673, opened 2022-11-02
Fix unit tests on windows (except connect, lazyfilelist, wui)

RadekVana:


src/common/selftest_eeprom.hpp
```diff
@@ -28,7 +28,8 @@ union SelftestResultEEprom_t {
         uint8_t reserved_1 : 2;   // bit 16-17
         uint8_t eth : 2;          // bit 17-19
         uint8_t wifi : 2;         // bit 20-21
-        uint16_t reserved : 10;   // bit 22-31
+        uint8_t reserved1 : 2;    // bit 22-23
+        uint8_t reserved0;        // bit 24-31
```
vorner, line 32: Why is this needed?

tests/stubs/malloc.h
```diff
@@ -1,5 +1,5 @@
 #pragma once
 
-#include <cstddef>
+#include <stddef.h>
```
vorner, line 3: Again, why is this needed?

vorner (approved):
For the record, I'd like the os_porting and debug „modules“ to be gone, they are somewhat weird things that were left from previous authors and don't really make much sense. But this is probably not the right time.

I don't have _objections_ to the other notes, so there's an approval of the changes, but I'm still curious to know why they are needed.

PR #2676, opened 2022-11-02
Fix possible bsod within log_destination_find

dragomirecky:


src/logging/log.c
```diff
@@ -68,11 +68,12 @@ log_component_t *log_component_find(const char *name) {
 }
 
 log_destination_t *log_destination_find(const char *name) {
-    for (log_destination_t **destination_pp = &destinations_head; *destination_pp != NULL; destination_pp++) {
+    for (log_destination_t **destination_pp = &destinations_head; *destination_pp != NULL;) {
         log_destination_t *destination = *destination_pp;
         if (strcmp(destination->name, name) == 0) {
             return destination;
         }
+        destination_pp = &((*destination_pp)->next);
```
mkbel, line 76: Is this equivalent to? 
log.c:49
```
for (log_destination_t **destination_pp = &destinations_head; *destination_pp != NULL; destination_pp = &(*destination_pp)->next) {
```

mkbel: The same construct which is fixed here is also used there without being fixed, is this right?
```
void log_destination_unregister(log_destination_t *destination) {
    for (log_destination_t **destination_pp = &destinations_head; *destination_pp != NULL; destination_pp++)

log_component_t *log_component_find(const char *name) {
    for (log_component_t *component = __start_log_components; component < __end_log_components; component++) {
```

mkbel (commented):
.

src/logging/log.c
dragomirecky: nice catch 👍 fixed the other occurrence & unified it

mkbel (dismissed):

mkbel (approved):

PR #2683, opened 2022-11-02
Back preview

RadekVana:
BFW-2927


src/gui/guimain.cpp
```diff
@@ -359,8 +359,11 @@ void gui_run(void) {
         // one click print is closed automatically from main thread, because it is opened for wrong gcode
         if ((marlin_update_vars(MARLIN_VAR_MSK(MARLIN_VAR_PRNSTATE))->print_state == mpsWaitGui)) {
             if ((!DialogHandler::Access().IsAnyOpen()) && can_start_print_at_current_screen) {
+                bool have_file_browser = Screens::Access()->IsScreenOnStack<screen_filebrowser_data_t>();
```
DRracer, line 362: this is a brutal hack ... but afterall it might work

DRracer (approved):

PR #2695, opened 2022-11-07
connect: React to new config in snappier way

vorner:
If connect can't connect, it wants to sleep for a long time. But interrupt the sleep if a new config comes earlier, as that might make it possible to connect now and don't the user let wait for a whole minute or so.

BFW-2812.

### DIFF
--- a/src/connect/connect.cpp
+++ b/src/connect/connect.cpp
@@ -146,8 +146,8 @@ namespace {
     // for that.
     const constexpr size_t MAX_RESP_SIZE = 256;
 
-    // Wait one second between config retries and similar.
-    const constexpr uint32_t IDLE_WAIT = 1000;
+    // Wait half a second between config retries and similar.
+    const constexpr uint32_t IDLE_WAIT = 500;
 
     using Cache = variant<monostate, tls, socket_con, Error>;
 }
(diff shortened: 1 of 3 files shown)


src/connect/connect.cpp
```diff
@@ -147,7 +147,7 @@ namespace {
     const constexpr size_t MAX_RESP_SIZE = 256;
 
     // Wait one second between config retries and similar.
-    const constexpr uint32_t IDLE_WAIT = 1000;
+    const constexpr uint32_t IDLE_WAIT = 500;
```
HelclM, line 150: The comment seems to be a bit off now...

src/connect/connect.cpp
```diff
@@ -275,7 +275,16 @@ optional<OnlineStatus> connect::communicate(CachedFactory &conn_factory) {
...
+        for (size_t i = 0; i < s->milliseconds / IDLE_WAIT; i++) {
+            // In case there's a change in situation during a long sleep, we
+            // want to retry sooner (new config might lead to being able to
+            // connect or something).
+            if (std::get<1>(printer.config(false))) {
+                return nullopt;
+            }
+        }
+
+        osDelay(s->milliseconds % IDLE_WAIT);
```
HelclM, line 287: Isn't this suppose to be in the loop and the amount slept be `IDLE_WAIT`, so we wait `s->miliseconds / IDLE_WAIT` times for `IDLE_WAIT` giving us total amount of `s->milliseconds` of sleeping? Seems to me, that we just check a bunch of times in a row and then wait for smaller amount of time. :thinking:

HelclM (changes requested):

src/connect/connect.cpp
vorner: Good catch!

HelclM (approved):

PR #2698, opened 2022-11-08
Suspend other tasks while firmware version is printed onto serial

DRracer:
Based on comments from @dragomirecky and @mkbel - we must explicitly prevent Marlin's task from taking over while printing the important messages.

BFW-2913


dragomirecky (commented):
I must be missing something, but where is the Marlin task actually suspended? I only see it being resumed 🧐

Btw just for simplicity & getting rid of dirty externs, what about taskENTER_CRITICAL/taskEXIT_CRITICAL?

DRracer:
@dragomirecky ok I'll rework it into taskENTER/EXIT

dragomirecky (approved):

PR #2702, opened 2022-11-08
Inicialized marlin_logging before wait_for_dependecies

Hagridpon3:
Fix bug in USBSerial. Set lineBufferHook in constructor to nullptr and check if correctly set hook in flush

### DIFF
--- a/lib/Arduino_Core_Buddy/cores/arduino/USBSerial.cpp
+++ b/lib/Arduino_Core_Buddy/cores/arduino/USBSerial.cpp
@@ -47,10 +47,12 @@ void USBSerial::flush(void) {
     if (enabled)
         tud_cdc_write_flush();
 
-    if (lineBufferUsed) {
-        lineBufferHook(&lineBuffer[0], lineBufferUsed);
-        lineBufferUsed = 0;
+    if (!lineBufferHook || !lineBufferUsed) {
+        return;
     }
+
+    lineBufferHook(&lineBuffer[0], lineBufferUsed);
+    lineBufferUsed = 0;
 }
 
 void USBSerial::LineBufferAppend(char character) {
--- a/lib/Arduino_Core_Buddy/cores/arduino/USBSerial.h
+++ b/lib/Arduino_Core_Buddy/cores/arduino/USBSerial.h
@@ -33,7 +33,7 @@ public:
     virtual size_t write(const uint8_t *buffer, size_t size);
     operator bool(void);
 
-    void (*lineBufferHook)(const uint8_t *buf, int len);
+    void (*lineBufferHook)(const uint8_t *buf, int len) { nullptr };
 };
 
 extern USBSerial SerialUSB;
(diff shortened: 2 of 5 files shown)


lib/Arduino_Core_Buddy/cores/arduino/USBSerial.h
```diff
@@ -18,7 +18,8 @@ class USBSerial : public Stream {
         : enabled(false)
         , isWriteOnly(false)
         , lineBuffer()
-        , lineBufferUsed(0) {}
+        , lineBufferUsed(0)
+        , lineBufferHook(nullptr) {}
```
Derevin, line 22: Rather than adding default values to constructor, it's better to add the default values to declarations (like so in this case: 
`void (*lineBufferHook)(const uint8_t *buf, int len) { nullptr };` )

lib/Arduino_Core_Buddy/cores/arduino/USBSerial.cpp
```diff
@@ -47,7 +47,7 @@ void USBSerial::flush(void) {
     if (enabled)
         tud_cdc_write_flush();
 
-    if (lineBufferUsed) {
+    if (lineBufferUsed && lineBufferHook) {
```
Derevin, line 50: generally I prefer to 'escape early' (guarding) rather than 'do if valid' (current state), ie in this case that would mean:
```
if (!lineBufferHook || !lineBufferUsed){
   return;
}

lineBufferHook(&lineBuffer[0], lineBufferUsed);
lineBufferUsed = 0;
```

However It's up to you, both options are valid.

src/buddy/main.cpp
```diff
@@ -32,6 +32,7 @@
 #include <option/filament_sensor.h>
 #include <option/has_gui.h>
 #include "tasks.h"
+#include "appmain.hpp"
```
Derevin, line 35: should be in `<>`

src/common/appmain.cpp
```diff
@@ -26,6 +26,8 @@
 #include <Arduino.h>
 #include "trinamic.h"
 #include "../Marlin/src/module/configuration_store.h"
+#include "tasks.h"
```
Derevin, line 29: should be in `<>`

src/common/appmain.cpp
```diff
@@ -108,8 +119,6 @@ void app_idle(void) {
 void app_setup_marlin_logging();
```
Derevin, line 119: remove this line

src/common/appmain.cpp
```diff
@@ -85,6 +87,15 @@ extern uartslave_t uart6slave; // PUT slave
 extern osThreadId webServerTaskHandle; // Webserver thread(used for fast boot mode)
 #endif                                 //BUDDY_ENABLE_ETHERNET
 
+void app_setup_marlin_logging();
```
Derevin, line 90: It would be better to move the function as a whole here, rather than create a forward declaration

Derevin (changes requested):
Overall just small nitpicky stuff

Derevin (approved):

PR #2709, opened 2022-11-09
Differential telemetry

vorner:


vorner:
I'll retract this one for a while. I'll add few more things in here, because they are related and will share a bit of code.

PR #2714, opened 2022-11-10
In code renaming of api_key to password.

HelclM:
Internally renamed api_key to password, so that it corresponds with what we externally expose to users.


lib/WUI/wui_api.h
```diff
@@ -145,22 +145,22 @@ void add_time_to_timestamp(int32_t secs_to_add, struct tm *timestamp);
 /// @brief Authorization key for PrusaLink
 ///
 /// @return Return an x-api-key
```
DRracer, line 147: /nitpick/: documentation might be updated here as well (`Return an x-api-key`)

DRracer (changes requested):
Just check the doxygen doc as well, at some spots the x-api-key is still present.
Otherwise the PR looks ok.

DRracer (approved):

Issue #2558, opened 2022-10-03
[BUG] No unique serial number in 4.4.0-RC1


paukstelis:
**Printer type** -
Mini

**Printer firmware version** - 
4.4.0-RC1

**Original or Custom firmware** - [Original/Custom]
Original

**Optional upgrades** - [e.g. Filament Runout Sensor]
None

**USB drive or USB/Octoprint**
Both

**Describe the bug**
 Updating bootloader and firmware to 4.4.0-RC1 results in disruption of the unique serial number. Serial number is now reported as: SerialNumber: 00000000001A

**How to reproduce**
Install bootloader and firmware.

**Expected behavior**
 Unique serial numbers should be retained. This is the most useful way for identifying individual printers using udev rules for host control systems (OctoPrint).

Related but not identical to issue raised here: https://github.com/prusa3d/Prusa-Firmware-Buddy/issues/2542

Prusa-Support:
The issue should be fixed in the final [FW 4.4.0](https://github.com/prusa3d/Prusa-Firmware-Buddy/releases/tag/v4.4.0) release now. Please check it out.

Michele Moramarco
Prusa Research

Issue #2542, opened 2022-09-29
Prusa Mini expose not only the usbserial, but a usb storage device too


czo:
Please, before you create a new bug report, please make sure you searched in open and closed issues and couldn't find anything that matches.

**Printer type** - Prusa Mini

**Printer firmware version** - 4.4.0-RC1

**Original or Custom firmware** - Original

**Optional upgrades** - Filament Runout Sensor

**Describe the bug**
The printer seems to me, has exporting an unknown "usb-storage" device, and the kernel try to spin up the disk forever.

kernel log:
Sep 29 19:45:08 dwarf.czo.hu kernel: usb 1-1: new full-speed USB device number 5 using orion-ehci
Sep 29 19:45:08 dwarf.czo.hu kernel: usb 1-1: New USB device found, idVendor=2c99, idProduct=000c, bcdDevice= 1.00
Sep 29 19:45:08 dwarf.czo.hu kernel: usb 1-1: New USB device strings: Mfr=1, Product=2, SerialNumber=3
Sep 29 19:45:08 dwarf.czo.hu kernel: usb 1-1: Product: Original Prusa PRINTER
Sep 29 19:45:08 dwarf.czo.hu kernel: usb 1-1: Manufacturer: Prusa Research (prusa3d.com)
Sep 29 19:45:08 dwarf.czo.hu kernel: usb 1-1: SerialNumber: 00000000001A
Sep 29 19:45:08 dwarf.czo.hu kernel: cdc_acm 1-1:1.0: ttyACM0: USB ACM device
Sep 29 19:45:08 dwarf.czo.hu kernel: usb-storage 1-1:1.2: USB Mass Storage device detected
Sep 29 19:45:08 dwarf.czo.hu kernel: scsi host2: usb-storage 1-1:1.2
Sep 29 19:45:09 dwarf.czo.hu kernel: scsi host2: scsi scan: INQUIRY result too short (5), using 36
Sep 29 19:45:09 dwarf.czo.hu kernel: scsi 2:0:0:0: Direct-Access     Prusa    MINI             1.0  PQ: 0 ANSI: 2
Sep 29 19:45:09 dwarf.czo.hu kernel: sd 2:0:0:0: Attached scsi generic sg1 type 0
Sep 29 19:45:09 dwarf.czo.hu kernel: sd 2:0:0:0: [sdb] Spinning up disk...

root@dwarf:~# cat /proc/scsi/scsi 
Attached devices:
[...]

PR #2724, opened 2022-11-15
Use serial number from OTP and disable USB MSC

MistrKernnunos:
solves #2558 and #2542
BFW-2945

### DIFF
--- a/src/common/otp.c
+++ b/src/common/otp.c
@@ -13,16 +13,3 @@ const char *otp_get_mac_address_str() {
     }
     return mac_address_str;
 }
-
-const char *otp_get_serial_number() {
-    static char serial_number_str[OTP_SERIAL_NUMBER_SIZE] = { 0 };
-
-    // if the string is empty, read the serial number
-    if (serial_number_str[0] == 0) {
-        // we need to do this to avoid UB when casting volatile variable to non-volatile
-        for (uint8_t i = 0; i < OTP_SERIAL_NUMBER_SIZE; ++i) {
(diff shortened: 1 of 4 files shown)


src/common/otp.c
```diff
@@ -13,3 +13,7 @@ const char *otp_get_mac_address_str() {
     }
     return mac_address_str;
 }
+
+const char *otp_get_serial_number() {
+    return (const char *)(OTP_SERIAL_NUMBER_ADDR);
```
vorner, line 18: I'm not sure this is valid. Isn't this a cast from volatile to non-volatile?

vorner (commented):
I have a concern about a possible UB there. Can we get away without the unvolatilement of the pointer?

src/common/otp.c
MistrKernnunos: fixed

src/common/otp.c
```diff
@@ -13,3 +13,16 @@ const char *otp_get_mac_address_str() {
     }
     return mac_address_str;
 }
+
+const char *otp_get_serial_number() {
+    static char serial_number_str[16] = { 0 };
```
lukash, line 18: Shouldn't you make the size be `OTP_SERIAL_NUMBER_SIZE`? You have a mismatch in the size of the array and the upper bound of the `i` when you write to it below. I assume they're equal and chances of `OTP_SERIAL_NUMBER_SIZE` changing are low, but anyway.

vorner (approved):

PR #2731, opened 2022-11-18
Connect info sending

vorner:
(In the same PR, as this shares some similar/same code)

*Differential telemetry*

Depending on how much changed in telemetry, send the full version or smaller set, to save the amount of changes transferred & processed by the server.

*Notifications about changes to USB/network/files*

Proactively sending the relevant INFO message when something changes.

### DIFF
--- a/src/connect/connect.cpp
+++ b/src/connect/connect.cpp
@@ -149,6 +149,9 @@ namespace {
     // Wait half a second between config retries and similar.
     const constexpr uint32_t IDLE_WAIT = 500;
 
+    // Send a full telemetry every 5 minutes.
+    const constexpr uint32_t FULL_TELEMETRY_EVERY = 5 * 60 * 1000;
+
     using Cache = variant<monostate, tls, socket_con, Error>;
 }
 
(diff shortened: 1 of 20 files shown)


src/connect/printer.cpp
```diff
@@ -7,16 +7,73 @@
...
+    }
+
+    return crc
+        .add(int(pos[Printer::Z_AXIS_POS]))
+        .add(print_speed)
+        .add(flow_factor)
+        .add(job_id)
+        // The RPM values are in thousands and fluctuating a bit, we don't want
+        // that to trigger the send too often, only when it actually really
+        // changes.
```
HelclM, line 57: I don't see any mechanism by which you would achieve this, am I missing something, or did you forget to divide the rpm by some amount as you do with the `filament_used`?

src/connect/connect.cpp
```diff
@@ -312,11 +319,21 @@ optional<OnlineStatus> connect::communicate(CachedFactory &conn_factory) {
...
+    // Underflow should naturally work
+    if (now - last_full_telemetry >= FULL_TELEMETRY_EVERY) {
+        // The server wants to get a full telemetry from time to time, despite
+        // it not being changed. Some caching reasons/recovery/whatever?
+        //
+        // If we didn't send a new telemetry for too long, reset the
+        // fingerprint, which'll trigger the resend.
+        telemetry_changes.mark_dirty();
+    }
+    const bool is_full_telemetry = holds_alternative<SendTelemetry>(action) && !get<SendTelemetry>(action).empty;
```
HelclM, line 334: minor nitpick: Would maybe be more readable if moved closer to where it is actually used.

src/connect/printer.cpp
vorner: That's when I'm testing with a device that has no fans connected :facepalm: Of course.

src/connect/connect.cpp
vorner: I moved it a bit, but it's only a few lines.

HelclM (approved):

PR #2736, opened 2022-11-21
Fix invalidation of the connect status menu

vorner:
In addition, set the status at start too.

Broken in f10bb8b77d1a4385529f4d2da1bd056817ffd3d0.

BFW-3070.


src/guiapi/include/WindowMenuInfo.hpp
```diff
@@ -55,7 +55,10 @@ class WiInfo : public AddSuper<IWiInfo> {
     }
 
     void ChangeInformation(string_view_utf8 str) {
+        // There doesn't seem to be an easy way to compare string_view_ut8 with
```
RadekVana, line 58: string_view_utf8 has operator ==

RadekVana (changes requested):

src/guiapi/include/WindowMenuInfo.hpp
vorner: It has an `operator ==` to compare with another `string_view_utf8`. But `WiInfo` contains a `char *` string. Comparing to _that_ string is complicated, as it seems, the easiest way is to create _another_ string buffer, copy the `string_view_utf8` into that and compare _them_. At that point, I'm not sure it's worth it.

But maybe the redrawing is significantly more effort than even such comparison. In that case, I'm OK implementing it.

RadekVana (approved):

PR #2739, opened 2022-11-21
Generate png data files at buildtime

MistrKernnunos:


RadekVana (changes requested):
Delete
src\resources\png\resources.pngs

RadekVana (dismissed):

Derevin (approved):
debug compiles (but doesn't link due to flash limitations which is OK)

PR #2753, opened 2022-11-24
Ignore hidden directories

tomasjakubik:
Please review another hidden files pull.
I forgot to ignore folders starting with a dot as well.

### DIFF
--- a/src/gui/file_raii.hpp
+++ b/src/gui/file_raii.hpp
@@ -41,12 +41,14 @@ struct F_DIR_RAII_Iterator {
     }
 
     bool EntryAccepted() const {
+        if (fno->lfn[0] == '.') { // ignore hidden files/directories
+            return false;
+        }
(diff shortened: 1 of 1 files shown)


Derevin (commented):
I'd suggest doing some kind of a guard like
```
if (fno->lfn[0] == '.' ) { // ignore hidden files/directories
return false; 
}
... original function
```
rather than checking for the same condition in multiple places

Derevin (changes requested):
Code looks good.

We're also attempting to have nice and clean commit history, which in this case at the very least means to rebase and fixup the last commit (cleanup) into the previous one (and once the PR is approved, merge with the 'rebase' option rather than create a merge commit)

tomasjakubik:
Rebased and squshed, please review once more.

Derevin (approved):

PR #2759, opened 2022-11-24
eeprom: Fix footer encoding

RadekVana:
BFW-3098

### DIFF
--- a/src/common/footer_eeprom.hpp
+++ b/src/common/footer_eeprom.hpp
@@ -106,7 +106,7 @@ uint8_t GetCenterNAndFewer();
  * @return constexpr uint32_t
  */
 constexpr uint32_t Encode(record rec) {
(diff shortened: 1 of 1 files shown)


src/common/footer_eeprom.hpp
```diff
@@ -106,7 +106,7 @@ uint8_t GetCenterNAndFewer();
  * @return constexpr uint32_t
  */
 constexpr uint32_t Encode(record rec) {
-    uint32_t ret = uint32_t(rec[0]);
+    uint32_t ret = uint32_t(rec[0]) << count_of_trailing_ones;
     for (size_t i = 1; i < count; ++i) {
```
Derevin, line 110: You can change it so that it's `uint32_t ret{0};` and then init the loop index at 0, which would result in a slightly cleaner code

Hagridpon3 (approved):

PR #2779, opened 2022-11-29
Remove unnecesary initializations in redscreen or BSOD

MistrKernnunos:
We want to init just bare  minimum to display red screen or BSOD. For it we need to init SPI for xFlash and dislay, some timers for hwio, gui and buzzer and init eeprom. We also init dma for SPI comunication

BFW-2924

### DIFF
--- a/src/buddy/main.cpp
+++ b/src/buddy/main.cpp
@@ -33,6 +33,7 @@
 #include <option/has_gui.h>
 #include "tasks.h"
 #include <appmain.hpp>
+#include "safe_state.h"
 
 #if ENABLED(POWER_PANIC)
     #include "power_panic.hpp"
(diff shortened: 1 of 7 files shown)


src/buddy/main.cpp
```diff
@@ -330,6 +314,19 @@ void spi_set_prescaler(SPI_HandleTypeDef *hspi, int prescaler_num) {
...
 }
 
+void init_redscreen() {
+    // init lcd spi and timer for buzzer
+#if HAS_GUI()
+    SPI_INIT(lcd);
+    hw_tim2_init(); // TIM2 is used to generate buzzer PWM. Not needed without display.
+#endif
+
+    if (option::has_gui) {
```
michalrudolf, line 324: Use one condition type (Not `HAS_GUI()` and `if (option::has_gui)` next to it) and encapsule calling of the function

src/buddy/main.cpp
```diff
@@ -330,6 +314,19 @@ void spi_set_prescaler(SPI_HandleTypeDef *hspi, int prescaler_num) {
     HAL_SPI_Init(hspi);
 }
 
+void init_redscreen() {
```
michalrudolf, line 317: Rename function to something like `init_error_screen`. Redscreen is not the only relevant screen here.

michalrudolf (changes requested):

michalrudolf (dismissed):

michalrudolf (approved):

PR #2783, opened 2022-11-30
Merge 4.4 -> 4.5

HelclM:


lib/WUI/wui_api.cpp
```diff
@@ -307,8 +307,9 @@ void get_MAC_address(mac_address_t *dest, uint32_t netdev_id) {
 
 void sntp_set_system_time(uint32_t sec) {
 
-    RTC_TimeTypeDef currTime;
-    RTC_DateTypeDef currDate;
+    // RTC_TimeTypeDef has attributes like TimeFormat (AM/PM) and DayLightSaving, which we don't use
+    RTC_TimeTypeDef currTime = { 0 };
+    RTC_DateTypeDef currDate = { 0 };
```
vorner, line 312: I guess this should have been at the very original MR, but why is there the `0`? Just using `{}` should initialize all to zero-equivalents for all the fields (all the fields not listed there, which would be all of them).

@lukash :question:

vorner (approved):

lib/WUI/wui_api.cpp
lukash: AFAIU this is how it'd be done in C, according to https://en.cppreference.com/w/c/language/initialization the struct can be initialized with an empty initializer (`{}`) since C23. This is actually C++ here though (just the struct and function is from C, that's how I got confounded I guess), so you are correct, `{}` should have worked just as well here.

PR #2793, opened 2022-12-02
Experiment, do not merge: TX power experiments

vorner:


vorner:
The production version got merged in #2817, #2818.

PR #2797, opened 2022-12-04
Home fix

RadekVana:


src/gui/screen_home.cpp
```diff
@@ -56,7 +62,7 @@ uint32_t screen_home_data_t::lastUploadCount = 0;
 screen_home_data_t::screen_home_data_t()
     : AddSuperWindow<screen_t>()
     , usbInserted(marlin_vars()->media_inserted)
-    , esp_flash_being_openned(false)
+    , event_in_progress(false)
```
Derevin, line 65: I'd suggest setting this variable right during declaration rather than initialize it in a constructor

Derevin (approved):
Just a minor nitpick

PR #2826, opened 2022-12-13
Unload filament fix unhomed z moves

libchap:
BFW-3137


michalrudolf (approved):

Derevin (approved):
If you move in Z, let's say to 39, and then do an action (unload), then the action moves it by 40 (to 79), however the Move Z menu still thinks it's only at 40. Would be nice if it move only by 1 or the menu reflected the 79.

However, not a blocking issue in my opinion and this fix is 'acceptable' for now.

PR #2831, opened 2022-12-15
gui: Use proper variables in liveadjust_z

tomasjakubik:
Use proper variables in liveadjust_z instead of hardcoded values.

### DIFF
--- a/src/gui/dialogs/liveadjust_z.cpp
+++ b/src/gui/dialogs/liveadjust_z.cpp
@@ -21,10 +21,9 @@
 
 WindowScale::WindowScale(window_t *parent, point_i16_t pt)
     : AddSuperWindow<window_frame_t>(parent, Rect16(pt, 10, 100))
-    , scaleNum0(parent, getNumRect(pt), 0)
-    , scaleNum1(parent, getNumRect(pt), -1)
-    , scaleNum2(parent, getNumRect(pt), -2) {
-    scaleNum0.SetFormat("% .0f");
+    , scaleNum0(parent, getNumRect(pt), z_offset_max, "% f")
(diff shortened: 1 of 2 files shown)


src/gui/dialogs/liveadjust_z.cpp
```diff
@@ -192,13 +191,13 @@ LiveAdjustZ::LiveAdjustZ()
 
 void LiveAdjustZ::moveNozzle() {
     uint16_t old_top = nozzle_icon.Top();
-    Rect16 moved_rect = nozzleRect;                     // starting position - 0%
-    float percent = adjuster.GetValue() / z_offset_min; // z_offset value in percent
+    Rect16 moved_rect = nozzleRect;                                                        // starting position - 0%
+    float relative = (z_offset_max - adjuster.GetValue()) / (z_offset_max - z_offset_min); // relative z_offset value
```
Derevin, line 195: A `static_assert` making sure that `min != max` to ensure no 0 div would be nice, but it's just a minor nitpick.

Derevin (approved):

src/gui/dialogs/liveadjust_z.cpp
tomasjakubik: Then this whole screen wouldn't make sense. I think in that case HardFault will be fine.

Derevin: What can be caught at compile time should be caught at compile time, `static_assert` is essentially free and exactly for making sure that stuff 'makes sense'. However in this case it would require constexpr enabling the constants, which might be effort that's not worth it (however small).

PR #2843, opened 2022-12-20
Transfer monitor

vorner:
BFW-3179.


src/transfers/monitor.hpp
```diff
@@ -0,0 +1,259 @@
...
+    Type type;
+    Timestamp start;
+    size_t expected;
+    size_t transferred;
+    char destination_path[FILE_PATH_BUFFER_LEN];
+
+    // History related
+    static constexpr size_t HISTORY_MAX_LEN = 2;
+    size_t history_len = 0;
+    TransferId history_end;
```
HelclM, line 212: This is probably a matter of personal taste, but `history_end` sounds like the oldest thing not the latest to me (also kind of like an end iterator for some array), IMHO something like `previous_id` or `latest_id` would be more clear, but I leave that up to you, if you want to change it or not, I'm fine with it if you think it's better this way.

HelclM (dismissed):

vorner:
Updated (the only difference is the name).

HelclM (approved):

PR #2855, opened 2022-12-22
Transfer monitor might not known destination of the file being uploaded.

HelclM:
This happens in case of POST from PrusaLink.

### DIFF
--- a/src/connect/render.cpp
+++ b/src/connect/render.cpp
@@ -367,7 +367,11 @@ namespace {
                     JSON_FIELD_INT_G(transfer_status.has_value(), "size", transfer_status->expected) JSON_COMMA;
                     JSON_FIELD_INT_G(transfer_status.has_value(), "transferred", transfer_status->transferred) JSON_COMMA;
                     JSON_FIELD_INT_G(transfer_status.has_value(), "time_transferring", (transfer_status->start - ticks_ms()) / 1000) JSON_COMMA;
-                    JSON_FIELD_STR_G(transfer_status.has_value(), "path", transfer_status->destination) JSON_COMMA;
+                    // Note: This works, because destination cannot go from non null to null
+                    // (if one transfer ends and another starts mid report, we bail out)
+                    if (transfer_status->destination) {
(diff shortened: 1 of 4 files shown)


src/connect/render.cpp
```diff
@@ -348,7 +348,9 @@ namespace {
                     JSON_FIELD_INT_G(transfer_status.has_value(), "size", transfer_status->expected) JSON_COMMA;
                     JSON_FIELD_INT_G(transfer_status.has_value(), "transferred", transfer_status->transferred) JSON_COMMA;
                     JSON_FIELD_INT_G(transfer_status.has_value(), "time_transferring", (transfer_status->start - ticks_ms()) / 1000) JSON_COMMA;
-                    JSON_FIELD_STR_G(transfer_status.has_value(), "path", transfer_status->destination) JSON_COMMA;
+                    if (transfer_status->destination) {
+                        JSON_FIELD_STR_G(transfer_status.has_value(), "path", transfer_status->destination) JSON_COMMA;
+                    }
```
vorner, line 353: I agree this'll actually work in practice, but is a bit fragile and non-trivial to prove (and is based on the fact destination never changes from non-null to null).

I think the reason _why_ this works needs a comment.

src/transfers/monitor.hpp
```diff
@@ -208,6 +212,8 @@ class Monitor {
     Timestamp start;
     size_t expected;
     size_t transferred;
+
+    bool path_known = false;
```
vorner, line 216: Is it worth the effort, or are we OK conflating empty path and null path here?

I guess `optional<char [FILE_PATH_BUFFER_LEN]>` would be too much of a mental exercise? 😈

vorner (commented):
Minor nitpicks, probably worth addressing. But it is something you can safely build other things on without fear it'll change, in case you happen to return to work before me 😇

src/transfers/monitor.hpp
HelclM: I agree we could just omit the bool, but AFAIK you can't put an array type into a optional. (taken from [cppreference](https://en.cppreference.com/w/cpp/utility/optional))
image

vorner (approved):

PR #2869, opened 2023-01-02
Integrating transfer monitor to PrusaLink.

HelclM:


vorner (commented):
You've asked for an idea about too many calls to `done`. Currently, there's the `Dropped` status that's auto-assigned if the slot is dropped without setting the result. That's kind of only for handling inappropriate uses.

But now I see the code, it seems that the most common call to `done` is an `Error` and, at the same time, the slot is going to be dropped (because of the error). I think it would make sense to just remove the `Dropped` status and auto-assign an `Error`.

I also think there's no `.done` called in case the GcodeUpload object is destroyed prematurely (because of eg. timeout in the server)… which _would_ get handled correctly with the above change.

lib/WUI/nhttp/gcode_upload.h
```diff
@@ -38,6 +39,9 @@ class GcodeUpload final : private UploadHooks {
 
 private:
     UploadParams upload;
+    //Note: mutable, so it can be changed in check_filename
+    // which is const.
+    mutable transfers::Monitor::Slot monitor_slot;
```
vorner, line 44: Is the better way to change check_filename to non-const? 😇

tests/unit/lib/WUI/nhttp/time_mock.cpp
```diff
@@ -15,4 +15,8 @@ extern "C" {
 uint32_t ticks_s() {
     return mock_time;
 }
+
+uint32_t ticks_ms() {
+    return mock_time;
+}
```
vorner, line 21: It's kind-of suspicious that the `ticks_ms` and `ticks_s` would return the same thing. At least multiply 1000?

lib/WUI/nhttp/gcode_upload.cpp
```diff
@@ -113,6 +117,14 @@ GcodeUpload::UploadResult GcodeUpload::start(const RequestParser &parser, Upload
         return StatusPage(Status::LengthRequired, StatusPage::CloseHandling::ErrorClose, json_errors);
     }
 
+    const char *path = holds_alternative<PutParams>(uploadParams) ? get<PutParams>(uploadParams).filepath.data() : nullptr;
+    auto slot = Monitor::instance.allocate(Monitor::Type::Link, path, *parser.content_length);
```
vorner, line 121: Just to confirm (I think it is correct), but does this content-length correspond to the correct values in case of the Post, where this is the size of the form, not the file? That is, the `progress` is being called with the undecoded data?

vorner: I know it is the length of the whole form. I'm just making sure this size corresponds to the values that are being reported by `progress` below ‒ that is, that `progress` is being called _before_ decoding the form and will not terminate at 95% or so, because the latter included only the „real“ data.

vorner (approved):

PR #2877, opened 2023-01-04
wifi: remove wpa/wep knob which does nothing...

libchap:
...only distinguish non/empty passwd

https://dev.prusa3d.com/browse/BFW-2634

### DIFF
--- a/doc/ESP/README.md
+++ b/doc/ESP/README.md
@@ -26,8 +26,6 @@ Be aware please, that the Prusa-Firmware-Buddy/Master repository is a **developm
 ssid= enter here the SSID of your wireless network
 
 psk= enter here the valid password
-
-key_mgmt=WPA
 ```
 
 - Store the `prusa_printer_settings.ini` file in the root folder of the USB flash drive and plug in to the printer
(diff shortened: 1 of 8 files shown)


vorner (commented):
The code changes look good, but I have two things I'd like to check first:

* I'm not sure if we want to remove these flags from the eeprom or _set_ them according to what we find in the ini file. They may be a bit useless now, but I expect to see some AP-mode (eg. we are an AP, we let a mobile phone connect to us and select which real AP to connect to or something like that). Maybe, possibly… But I won't argue in either direction, only adding the consideration to the table.
* Is it now possible to _remove_ a password? The non-existence of the field keeps the old one. Does changing to „empty“ one actually work?

libchap:
> * I'm not sure if we want to remove these flags from the eeprom or _set_ them according to what we find in the ini file. They may be a bit useless now, but I expect to see some AP-mode (eg. we are an AP, we let a mobile phone connect to us and select which real AP to connect to or something like that). Maybe, possibly… But I won't argue in either direction, only adding the consideration to the table.

So we will add them back in the future. It would be weird to keep random bits in EEPROM and re-use them later for different use-case (AP). I've been taught to remove things that "might be used in far future", rather than keeping unnecessary stuff complicating the code.

> * Is it now possible to _remove_ a password? The non-existence of the field keeps the old one. Does changing to „empty“ one actually work?

I thried over-configuring from WPA2 PSK to no-security by changing the password to "" and it worked for me.

vorner (approved):

vorner (approved):

PR #2885, opened 2023-01-05
Make mergeable

RadekVana:
BFW-3259


src/gui/include_MINI/gui_config_printer.hpp
```diff
@@ -1,6 +1,12 @@
-// gui_config_mini.h
+/**
+ * @file gui_config_printer.hpp
+ * @brief printer specific gui config
+ */
```
vorner, line 4: Is this new part? I think we've agreed not to have these headers / not have the `@file` part (because it doesn't particularly do anything in doxygen anyway).

Goes for other files too...

src/gui/screen_menu_lan_settings.hpp
```diff
@@ -12,7 +12,7 @@
 // Container for this base class contains all MI from both ETH and WIFI screen
 // There can be MI, that will not be used in derived class (MI_WIFI_... won't be used in ETH Screen)
 // This is a solution to the problem that base class container have to define what MIs will be used, but derived classes will have slightly different ones.
-using ScreenMenuConnectionBase__ = ScreenMenu<GuiDefaults::MenuFooter, MI_RETURN, MI_WIFI_STATUS_t, MI_WIFI_INIT_t, MI_WIFI_CREDENTIALS_INI_FILE_t, MI_WIFI_CREDENTIALS_t, MI_NET_IP_VER_t, MI_NET_IP_t, MI_IP4_ADDR, MI_IP4_NMSK, MI_IP4_GWAY, MI_MAC_ADDR>;
+using ScreenMenuConnectionBase__ = ScreenMenu<EFooter::Off, MI_RETURN, MI_WIFI_STATUS_t, MI_WIFI_INIT_t, MI_WIFI_CREDENTIALS_INI_FILE_t, MI_WIFI_CREDENTIALS_t, MI_NET_IP_VER_t, MI_NET_IP_t, MI_IP4_ADDR, MI_IP4_NMSK, MI_IP4_GWAY, MI_MAC_ADDR>;
```
vorner, line 15: Seems like at some places, it changes from `GuiDefaults::MenuFooter` to something, some other places have it the other way around.

Can you explain why? Would it have been better to make the change in a separate PR to make it easier to review?

src/gui/screen_menu_info.cpp
```diff
@@ -18,5 +18,7 @@ void ScreenMenuInfo::windowEvent(EventLock /*has private ctor*/, window_t *sende
 ScreenMenuInfo::ScreenMenuInfo()
     : ScreenMenuInfo__(_(label)) {
     EnableLongHoldScreenAction();
+#if (PRINTER_TYPE != PRINTER_PRUSA_MINI)
     header.SetIcon(&png::info_16x16);
+#endif //PRINTER_PRUSA_MINI
```
vorner, line 23: This code doesn't seem to be the same thing as before, this effectively disables the line here. Is that on purpose?

src/gui/screen_menu_move.cpp
```diff
@@ -147,6 +149,9 @@ bool ScreenMenuMove::IsTempOk() {
 
 ScreenMenuMove::ScreenMenuMove()
     : ScreenMenuMove__(_(label)) {
+#if PRINTER_TYPE != PRINTER_PRUSA_MINI
+    header.SetIcon(&png::move_16x16);
+#endif
```
vorner, line 154: Similar as above.

vorner (commented):
Honestly, I'm not sure what to say about this PR. I mean, the better mergeability is nice and I'm all for it.

But it's kind of impossible to give it an effective review, if something breaks. I could have taken your word that it's just restructuring the code and moving bits around… but that doesn't seem to be true, this _seems_ to be bringing changes of functionality to the code. Is it possible to split it up so we have the changes and the restructuring in separate PRs?

src/gui/screen_menu_lan_settings.hpp
RadekVana: You have it in separate commit.
GuiDefaults::MenuFooter is default setting (of for MK4 on for MINI)
EFooter::Off is of for all printers

It was wrong in public and GuiDefaults::MenuFooter was set to off for NINI too

src/gui/screen_menu_info.cpp
RadekVana: yes
all other menus in MINI does not use header icon, this one was an exception

I think it is stupid and we should have same icons in MK4/MINI but it is for another discussion

src/gui/screen_menu_move.cpp
RadekVana: MK4 has this icon, MINI das not
it does not change anything, just makes the same code public/private

src/gui/include_MINI/gui_config_printer.hpp
RadekVana: I don't know anything about agreement not having this headers.

vorner (approved):

PR #2890, opened 2023-01-06
Menu sz2

RadekVana:


src/gui/menu_items_open.cpp
```diff
@@ -0,0 +1,86 @@
...
+#include "screen_menu_eeprom.hpp"
+#include "screen_menu_footer_settings.hpp"
+#include "screen_prusa_link.hpp"
+#include "screen_menu_connect.hpp"
+#include "screen_menu_experimental_settings.hpp"
+#include "screen_menu_network.hpp"
+#include "screen_menu_eeprom_diagnostics.hpp"
+#include "gui/test/screen_menu_test.hpp"
+
+#define GENERATE_OPENER_ITEM_DEFINITION__(NAME, SCREEN, ICON, VISIBILITY) \
```
Derevin, line 31: I am very strongly against creating obfuscation define macros likes these. It's not really adding to readability (i have to go and find out what the macro actually does), it's bloating global namespace and it's nigh impossible to debug easily. 
Not to mention extendability goes also completely out of the picture.
And as a side note, you're doing in headers what could be done in sources.

Derevin (changes requested):
I haven't checked the rest of the PR but this is personally a major issue that needs to be addressed.

danopernis:
This is not mergeable at this point.

PR #2895, opened 2023-01-06
Do not preheat bed on filament change

tomasjakubik:
BFW-3246

### DIFF
--- a/src/marlin_stubs/pause/M70X_preheat.cpp
+++ b/src/marlin_stubs/pause/M70X_preheat.cpp
@@ -82,7 +82,6 @@ void filament_gcodes::preheat_to(filament_t filament) {
     if (thermalManager.degTargetHotend(0) < fil_cnf.nozzle) {
         thermalManager.setTargetHotend(fil_cnf.nozzle, 0);
         marlin_server_set_temp_to_display(fil_cnf.nozzle);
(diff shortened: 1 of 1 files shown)


davidnovacek (commented):
This is confusing
The change seems to be correct - do not heat up bed when you want to load or unload filament. Gcode is for load and unload, but name of pull request suggest that we don't want to preheat bed on filament change - and that is not true. 
Just for sure we are at the same note. Is it change for only loading and unloading before you printing? And for filament change (during printing) it will still preheat bed?

tomasjakubik:
> This is confusing The change seems to be correct - do not heat up bed when you want to load or unload filament. Gcode is for load and unload, but name of pull request suggest that we don't want to preheat bed on filament change - and that is not true. Just for sure we are at the same note. Is it change for only loading and unloading before you printing? And for filament change (during printing) it will still preheat bed?

As far as I can tell, this change is relevant only for the filament menu, where there are load, unload, change and purge. Purge works only with filament sensor, so I didn't test that, but it should be the same.

Filament change during print won't allow you to change filament type or temperature. Even if you could have one material at the bottom, you wouldn't want to change bed temperature when you are printing something else on top.

davidnovacek (approved):

PR #2903, opened 2023-01-09
Fix serial number displaying in version info

Hagridpon3:
Now is show full serial number with last digit


danopernis:
Looks like this is already fixed, we can even fit BOMID, so closing.

PR #2925, opened 2023-01-12
crash detection: fix re-enabling during print

libchap:
BFW-3284

### DIFF
--- a/lib/Marlin/Marlin/src/feature/prusa/crash_recovery.cpp
+++ b/lib/Marlin/Marlin/src/feature/prusa/crash_recovery.cpp
@@ -217,6 +217,7 @@ void Crash_s::update_machine() {
     if(!m_axis_is_homing[0])
     {
         if (enabled) {
+            tmc_enable_stallguard(stepperX);
             stepperX.stall_max_period(0);
             #if AXIS_DRIVER_TYPE_X(TMC2130)
                 stepperX.sfilt(filter);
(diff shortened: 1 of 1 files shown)


lib/Marlin/Marlin/src/feature/prusa/crash_recovery.cpp
```diff
@@ -223,6 +223,7 @@ void Crash_s::update_machine() {
             #endif
             stepperX.stall_sensitivity(crash_s.sensitivity.x);
             stepperX.stall_max_period(crash_s.max_period.x);
+            tmc_enable_stallguard(stepperX);
```
mkbel, line 226: This call will undesirably clear stepperX.sfilt(filter) settings and maybe even something else set there earlier.

libchap: Re-pushed fixed version with correct `sfilt` setting.

lib/Marlin/Marlin/src/feature/prusa/crash_recovery.cpp
```diff
@@ -217,6 +217,7 @@ void Crash_s::update_machine() {
     if(!m_axis_is_homing[0])
     {
         if (enabled) {
+            tmc_enable_stallguard(stepperX);
```
mkbel, line 220: Stallguard might trigger falsely at this point, as it is enabled earlier than sensitivity is set, so original sensitivity (homing sensitivity) might be higher than intended for crash detection.
I think there is no possible place to put this function call in this if{} block. E.g. if you put it right after `stepperX.stall_sensitivity(crash_s.sensitivity.x);` it would be OK from the point of sensitivity, but it might trigger falsely because it might set longer `stall_max_period` (TCOOLTHRS) so it might be active even stall_max_period for crash detection prevent its activity at current feedrate.

libchap:
Solved differently.

PR #2927, opened 2023-01-12
Adding /api/transfer endpoint to link.

HelclM:
[BFW-3184](https://dev.prusa3d.com/browse/BFW-3184)


lib/WUI/link_content/prusa_link_api.cpp
```diff
@@ -255,7 +254,12 @@ optional<ConnectionState> PrusaLinkApi::accept(const RequestParser &parser) cons
             return StatusPage(Status::MethodNotAllowed, StatusPage::CloseHandling::ErrorClose, parser.accepts_json);
         }
     } else if (suffix == "printer") {
-        return get_only(StatelessJson(get_printer, parser.can_keep_alive()));
+        return get_only(SendJson(EmptyRenderer(get_printer), parser.can_keep_alive()));
+    } else if (suffix == "download" || suffix == "transfer") {
+        if (transfers::Monitor::instance.status().has_value())
+            return get_only(SendJson(TransferRenderer(), parser.can_keep_alive()));
```
vorner, line 260: I _think_ this is a little bit racy. Is it possible status has value now, but will disappear before entering the transfer renderer?

Let me have a look at that one 😇

lib/WUI/nhttp/send_json.cpp
```diff
@@ -0,0 +1,124 @@
...
+    // Keep the indentation of the JSON in here!
+    // clang-format off
+    JSON_START;
+        JSON_OBJ_START
+            if (transfer_status.has_value()) {
+                if (resume_point != 0 && transfer_status->id != state.transfer_id)
+                    // if transfer changes mid report, bail out
+                    transfer_status.reset();
+                else
+                    state.transfer_id = transfer_status->id;
```
vorner, line 27: I'm not sure we (the team) explicitly agreed on this one, but… I tend to use the block braces `{ }` even with single-line blocks around if-else.

The reason is, it is somewhat easy to accidentally add another (indented) line into the „block“, but it would not be conditional. This is somewhat alleviated by the fact that the automatic formatter would reformat it one level up again, so the reviewer would have less chance to get confused, but I think the precaution is worth having two more characters in the source code 😇

I think I might point this precaution out in the next retrospective

lib/WUI/nhttp/send_json.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+                JSON_FIELD_STR("target", "usb") JSON_COMMA;
+                JSON_FIELD_STR_FORMAT_G(transfer_status.has_value(), "type", "%s", to_str(transfer_status->type)) JSON_COMMA;
+                JSON_FIELD_STR_FORMAT_G(transfer_status.has_value(), "size", "%d", transfer_status->expected) JSON_COMMA;
+                //FIXME: Right now should be seconds from epoch start, would be much nicer
+                // for us to make it seconds from print start and make the client compute the rest,
+                // also it would solve some potential problems with NTP and time zones etc.
+                JSON_FIELD_STR_FORMAT_G(transfer_status.has_value(), "start_time", "%d", time(nullptr) - (ticks_s() - transfer_status->start)) JSON_COMMA;
+                // FIXME: Does this make sense in case of connect? Probably should be added
+                // once we actually have a working connect transfer implementation.
```
vorner, line 37: No, connect doesn't have any notion of „Transfer & print imediatelly“. The server keeps track of these things and would send a separate command to start printing once the transfer is finished.

Feel free to add such flags into the monitor as needed.

lib/WUI/nhttp/send_json.cpp
```diff
@@ -0,0 +1,124 @@
...
+                JSON_FIELD_BOOL_G(transfer_status.has_value(), "to_print", false) JSON_COMMA;
+                // Note: This works, because destination cannot go from non null to null
+                // (if one transfer ends and another starts mid report, we bail out)
+                if (transfer_status->destination) {
+                    JSON_FIELD_STR_G(transfer_status.has_value(), "destination", transfer_status->destination) JSON_COMMA;
+                }
+                JSON_FIELD_STR_FORMAT_G(transfer_status.has_value(), "remaining_time", "%.2f", static_cast<double>(transfer_status->expected - transfer_status->transferred) / (static_cast<double>(transfer_status->transferred) / (ticks_s() - transfer_status->start))) JSON_COMMA;
```
vorner, line 44: Is there a risk of division-by-zero here?

lib/WUI/nhttp/send_json.cpp
```diff
@@ -0,0 +1,124 @@
...
+                JSON_FIELD_STR_FORMAT_G(transfer_status.has_value(), "progress", "%.2f", static_cast<double>(transfer_status->transferred) / transfer_status->expected);
```
vorner, line 45: Similar?

lib/WUI/nhttp/send_json.h
```diff
@@ -0,0 +1,78 @@
...
+private:
+    Content generator;
+
+public:
+    EmptyRenderer(Content content)
+        : JsonRenderer(Empty {})
+        , generator(content) {}
+    virtual json::JsonResult renderState(size_t resume_point, json::JsonOutput &output, Empty &state) const override {
+        return generator(resume_point, output);
+    }
```
vorner, line 44: I wonder if the combination of `virtual` and (implicit) `inline` is a good idea. I mean, the compiler will make _something_ of it, but I wonder what the implications for code size, etc, are.

There's likely no chance for the compiler to opmitize the dynamic dispatch out here anyway, so maybe this could better go into the cpp file?

lib/WUI/nhttp/send_json.h
```diff
@@ -0,0 +1,78 @@
...
+};
+
+/// Renderer of generic JSON.
+///
+/// This'll generate a JSON response, provided a class derived from JsonRenderer
+/// that may or may not have a state.
+template <class Renderer>
+class SendJson {
+public:
+    //typedef json::JsonResult (*Content)(size_t resume_point, json::JsonOutput &output);
```
vorner, line 54: Leftover?

vorner (commented):
There are few minor nitpicks, almost ready to go in.

On some level, I would have preferred the refactoring of stateless_json->send_json to be a separate commit. It's probably not worth splitting it now, though.

lib/WUI/link_content/prusa_link_api.cpp
HelclM: It is, maybe we could allow stale inside `TransferRenderer` every time, not depending on resume_point not being 0 and then we would be guaranteed to get a status.

vorner: And maybe pass the expected transfer_id into the constructor, which could also eliminate a bit of repeated logic inside the render step.

lib/WUI/nhttp/send_json.cpp
```diff
@@ -0,0 +1,128 @@
...
+namespace nhttp::handler {
+
+json::JsonResult EmptyRenderer::renderState(size_t resume_point, json::JsonOutput &output, Empty &state) const {
+    return generator(resume_point, output);
+}
+
+JsonResult TransferRenderer::renderState(size_t resume_point, json::JsonOutput &output, TransferState &state) const {
+    // Note: We are guaranteed to get a status, because we checked if it is present
+    // before calling this and we set allow_stale to true
+    auto transfer_status = Monitor::instance.status(true);
```
vorner, line 23: I'd say we may either change the comment above to assert (to verify that our expectations are true) or, as you also do the `_G` stuff and `.reset()` below, simply guard the `transfer_status->id` with a check.

lib/WUI/nhttp/send_json.cpp
```diff
@@ -0,0 +1,128 @@
...
+            JSON_FIELD_STR_FORMAT_G(transfer_status.has_value(), "type", "%s", to_str(transfer_status->type)) JSON_COMMA;
+            JSON_FIELD_STR_FORMAT_G(transfer_status.has_value(), "size", "%d", transfer_status->expected) JSON_COMMA;
+            //FIXME: Right now should be seconds from epoch start, would be much nicer
+            // for us to make it seconds from print start and make the client compute the rest,
+            // also it would solve some potential problems with NTP and time zones etc.
+            JSON_FIELD_STR_FORMAT_G(transfer_status.has_value(), "start_time", "%d", time(nullptr) - (ticks_s() - transfer_status->start)) JSON_COMMA;
+            JSON_FIELD_BOOL_G(transfer_status.has_value(), "to_print", transfer_status->print_after_upload) JSON_COMMA;
+            // Note: This works, because destination cannot go from non null to null
+            // (if one transfer ends and another starts mid report, we bail out)
+            if (transfer_status->destination) {
```
vorner, line 43: Not really trivial to see that we don't need `if (transfer_status.has_value() && transfer_status->destination != nullptr)`… (though it is true).

lib/WUI/nhttp/send_json.cpp
```diff
@@ -0,0 +1,128 @@
...
+                JSON_FIELD_STR_G(transfer_status.has_value(), "destination", transfer_status->destination) JSON_COMMA;
+            }
+            // To prevent division by zero
+            if (transfer_status->transferred !=0 && ticks_s() > transfer_status->start) {
```
vorner, line 47: I'm not sure this is 100% correct, for several reasons:
* The field may be retried with a different value.
* The time may (and does) overflow (or doesn't it, in case of seconds?)

I'd actually go with something like `... / std::min(1, transfer_status->expected)`

vorner (commented):
I should learn not to nitpick that much 😇

lib/WUI/nhttp/send_json.cpp
HelclM: `ticks_s` overflows every 136 years, so we are fine there, so the only time we are in danger of division by zero is in the very beginning, where no data was transferred or the very first second of the transfer. 

IMHO this is secure, because once those checks passed once, they can never stop being true. Neither transferred nor ticks_s can ever decrease, nevertheless your solution may be more explicitly correct.

lib/WUI/nhttp/send_json.cpp
```diff
@@ -0,0 +1,132 @@
...
+    auto transfer_status = Monitor::instance.status(true);
+    // Keep the indentation of the JSON in here!
+    // clang-format off
+    JSON_START;
+        JSON_OBJ_START
+            if (transfer_status.has_value()) {
+                if (transfer_status->id != state.transfer_id) {
+                    // if transfer changes mid report, bail out
+                    transfer_status.reset();
+                }
```
vorner, line 32: Damn. I've noticed this issue only now, should have pointed it out earlier :-(.

The `JSON_START` is a switch statement that jumps to the place where it left previously. That is, if you place this guard _inside_ the `JSON_START`, it'll be skipped over. It needs to go outside / before the JSON_START.

I wonder, considering how much JSON rendering we have here if we should somehow explore the C++ coroutines, here we have quite many ugly cornercases and traps :-|.

vorner (changes requested):
Just that one last issue and we shall be ready to go.

vorner (approved):

PR #2937, opened 2023-01-17
Remove unnecessary "extern C"

HelclM:
Probably an artefact from the past, not needed anymore.


Derevin:
We're using a convention that .h files can be included in a C context. If you're converting the header to C++, then the header should be renamed to `.hpp` to reflect the fact

HelclM:
Ok, nobody told me and I found many files that do not reflect this convention in just a very quick scan, but I can do that, it seems like a good idea.

Derevin (approved):

PR #2939, opened 2023-01-18
Avoid global python package installs in bootstrap

vladamatena:
Add --user to install commandline to avoid breaking system packages.

### DIFF
--- a/utils/bootstrap.py
+++ b/utils/bootstrap.py
@@ -214,7 +214,7 @@ def main() -> int:
             continue
         print('Installing Python package %s' % package)
         run(sys.executable, '-m', 'pip', 'install', package,
-            '--disable-pip-version-check')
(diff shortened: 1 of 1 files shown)


vorner:
Will this work OK with my setup of using virtualenv setup? I use that one instead, I don't want it to break my _user_ packages either 😇

vladamatena:
> Will this work OK with my setup of using virtualenv setup? I use that one instead, I don't want it to break my _user_ packages either innocent
Good point, it will refuse to install for virtualenv. I will look for better solution.

dragomirecky:
I think the best option would be to create custom virtualenv automatically (and run everything within that virtualenv). 
I have this as a todo for some time in my head but didn't get to it yet

danopernis:
Since we are using `venv` now, this can be closed.

PR #2949, opened 2023-01-22
Selftest result

RadekVana:
BFW-2844

### DIFF
--- a/src/guiapi/src/display_helper.cpp
+++ b/src/guiapi/src/display_helper.cpp
@@ -266,6 +266,109 @@ void render_icon_align(Rect16 rc, const png::Resource *res, color_t clr_back, ic
     }
 }
 
+/**
+ * @brief calculate size (in number of characters) of required rectangle, independent on font
+ * TODO use common algorithm with draw and merge those functions in one
+ *
+ * @param str                           text to be measured
+ * @param max_chars_per_line            max returned number of characters per line
+ * @param ret_numOfUTF8Chars            optional return number of characters (use nullptr if not needed)
+ * @return std::optional<size_ui16_t>   size of needed rectangle, it might be narrower than max_width
+ */
+std::optional<size_ui16_t> characters_meas_text(string_view_utf8 &str, uint16_t max_chars_per_line, uint16_t *numOfUTF8Chars) {
+    if (max_chars_per_line == 0)
+        return std::nullopt;
+
+    std::optional<int> last_space_index = std::nullopt;
+    unichar c = 0;
+
+    int chars_this_line = 0;
+    int chars_longest_line = 0;
+    int chars_tot = 0;
+    int row_no = 0;
+
+    while ((c = str.getUtf8Char()) != 0) {
+        ++chars_tot;
+        switch (c) {
+        case '\n': // new line
+            if (chars_this_line >= chars_longest_line)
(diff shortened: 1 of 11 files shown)


src/guiapi/src/display_helper.cpp
```diff
@@ -266,6 +266,108 @@ void render_icon_align(Rect16 rc, const png::Resource *res, color_t clr_back, ic
...
+
+    while ((c = str.getUtf8Char()) != 0) {
+        ++chars_tot;
+        switch (c) {
+        case '\n': // new line
+            if (chars_this_line >= chars_longest_line)
+                chars_longest_line = chars_this_line;
+            ++row_no;
+            chars_this_line = 0;
+            break;
```
libchap, line 298: Here should be re-initialized `last_space_index` to 0.

src/guiapi/src/display_helper.cpp
```diff
@@ -266,6 +266,108 @@ void render_icon_align(Rect16 rc, const png::Resource *res, color_t clr_back, ic
...
+                // chars_longest_line can become last_space_index
+                chars_longest_line = std::max(chars_longest_line, *last_space_index);
+
+                // new line
+                ++row_no;
+
+                // count chars in next line
+                chars_this_line -= ((*last_space_index) + 1);
+                chars_this_line = std::max(0, chars_this_line);
+
```
libchap, line 337: Here should be re-initialized `last_space_index` to 0 as well.

src/guiapi/src/display_helper.cpp
```diff
@@ -266,6 +266,108 @@ void render_icon_align(Rect16 rc, const png::Resource *res, color_t clr_back, ic
...
+                // and set chars_this_line 13 - 5 - 1 (1 = space)
+
+                // chars_longest_line can become last_space_index
+                chars_longest_line = std::max(chars_longest_line, *last_space_index);
+
+                // new line
+                ++row_no;
+
+                // count chars in next line
+                chars_this_line -= ((*last_space_index) + 1);
```
libchap, line 335: I guess correct would be just `chars_this_line -= *last_space_index;`. You are correctly decrementing by 1 for the space being omitted, however, you should also increment by 1 for the character being processed right now (because the statement `++chars_this_line;` is skipped later). In your example "Hellow world!!!", you `set chars_this_line 13 - 5 - 1 (1 = space)` which results in 7, but "world!!!" is 8.

src/guiapi/src/display_helper.cpp
```diff
@@ -266,6 +266,108 @@ void render_icon_align(Rect16 rc, const png::Resource *res, color_t clr_back, ic
...
+                chars_this_line = std::max(0, chars_this_line);
```
libchap, line 336: I'd expect that if this algorithm is correct, it will be always >=0 at this point. I would rather see an assert for this.

libchap (changes requested):

src/guiapi/src/display_helper.cpp
RadekVana: done

src/guiapi/src/display_helper.cpp
RadekVana: done

PR #2952, opened 2023-01-24
gui: Improve drawing of preheat

RadekVana:
BFW-3336

### DIFF
--- a/src/common/filament.cpp
+++ b/src/common/filament.cpp
@@ -22,29 +22,17 @@ const char *get_selected_filament_name() {
 #endif                // USE_<display>
 
 // clang-format off
-// to keep the texts aligned for easier checking of their alignment on the LCD
-static constexpr const char *pla_str =      "PLA      " SPACES "215/ 60";
-static constexpr const char *pet_g_str =    "PETG     " SPACES "230/ 85";
-static constexpr const char *pet_g_nbh_str ="PETG_NH  " SPACES "230/  0";
-static constexpr const char *asa_str =      "ASA      " SPACES "260/100";
-static constexpr const char *pc_str =       "PC       " SPACES "275/100";
-static constexpr const char *pvb_str =      "PVB      " SPACES "215/ 75";
-static constexpr const char *abs_str =      "ABS      " SPACES "255/100";
-static constexpr const char *hips_str =     "HIPS     " SPACES "220/100";
-static constexpr const char *pp_str =       "PP       " SPACES "240/100";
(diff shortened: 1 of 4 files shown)


src/common/filament.cpp
```diff
@@ -22,29 +22,17 @@ const char *get_selected_filament_name() {
...
-    { BtnResponse::GetText(Response::PLA), pla_str,     215, 170,  60, Response::PLA },
-    { BtnResponse::GetText(Response::PETG), pet_g_str,  230, 170,  85, Response::PETG },
-    { BtnResponse::GetText(Response::ASA), asa_str,     260, 170, 100, Response::ASA },
-    { BtnResponse::GetText(Response::PC), pc_str,       275, 170, 100, Response::PC },
-    { BtnResponse::GetText(Response::PVB), pvb_str,     215, 170,  75, Response::PVB },
-    { BtnResponse::GetText(Response::ABS), abs_str,     255, 170, 100, Response::ABS },
-    { BtnResponse::GetText(Response::HIPS), hips_str,   220, 170, 100, Response::HIPS },
-    { BtnResponse::GetText(Response::PP), pp_str,       240, 170, 100, Response::PP },
-    { BtnResponse::GetText(Response::FLEX), flex_str,   240, 170,  50, Response::FLEX },
+    { "---",                                  0,   0,   0, Response::Cooldown }, // Cooldown sets long text instead short, not a bug
```
lukash, line 26: `long_name` was removed, the comment is no longer relevant.

src/gui/dialogs/window_dlg_preheat.cpp
```diff
@@ -14,6 +14,13 @@
 
 /*****************************************************************************/
 //NsPreheat::I_MI_Filament
+NsPreheat::I_MI_Filament::I_MI_Filament(string_view_utf8 name, unsigned t_noz, unsigned t_bed)
+    : WI_INFO_t(name, nullptr, is_enabled_t::yes, is_hidden_t::no) {
+    char buff[sizeof("999/999 ")];
+    snprintf(buff, sizeof(buff), "%3u/%3u ", t_noz, t_bed); // extra space at the end is intended
```
lukash, line 20: Could you also mention the reason for the extra space? I assume it's to shift the numbers to the left a bit, but why is it needed in this particular case?

src/common/filament.cpp
RadekVana: I'll fix that, thx

lukash (approved):

PR #2954, opened 2023-01-25
Transfer body of the download from Connect

vorner:
BFW-3188.

### DIFF
--- a/src/connect/connect.cpp
+++ b/src/connect/connect.cpp
@@ -1,11 +1,9 @@
 #include "connect.hpp"
 #include <http/httpc.hpp>
-#include <http/os_porting.hpp>
 #include "tls/tls.hpp"
 #include "render.hpp"
 #include <http/socket.hpp>
 
-#include <cmsis_os.h>
 #include <log.h>
 
 #include <atomic>
@@ -24,7 +22,6 @@ using std::get;
 using std::get_if;
 using std::holds_alternative;
 using std::is_same_v;
-using std::min;
 using std::monostate;
 using std::move;
 using std::nullopt;
@@ -148,9 +145,6 @@ namespace {
     // for that.
     const constexpr size_t MAX_RESP_SIZE = 256;
 
-    // Wait half a second between config retries and similar.
-    const constexpr uint32_t IDLE_WAIT = 500;
-
     // Send a full telemetry every 5 minutes.
     const constexpr uint32_t FULL_TELEMETRY_EVERY = 5 * 60 * 1000;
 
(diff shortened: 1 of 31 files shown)


src/connect/printer.hpp
```diff
@@ -1,13 +1,15 @@
...
+
 #include <cstdint>
 #include <cstddef>
 #include <tuple>
 #include <optional>
 
 namespace connect_client {
 
-class Printer {
+class Printer : public transfers::NotifyFilechange {
```
HelclM, line 12: Do we expect `NotifyFilechange` to be used soon somewhere else too? Adding another inheritance layer only for this one use seems maybe a bit unnecessary  to me :thinking:

HelclM (changes requested):

src/connect/connect.cpp
```diff
@@ -438,9 +420,6 @@ optional<OnlineStatus> connect::communicate(CachedFactory &conn_factory) {
 
 void connect::run() {
     log_debug(connect, "%s", "Connect client starts\n");
-    // waits for file-system and network interface to be ready
-    //FIXME! some mechanisms to know that file-system and network are ready.
-    osDelay(IDLE_WAIT);
```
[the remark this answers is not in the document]
vorner: Let me answer this in a bit roundabout way.

I've migrated all sleeps in this file to the Sleep class. That allows me to remove the direct interfacing of the low-level calls (and remove the header). On the other hand, the Sleep needs a real planner prepared and working, etc, which is not yet available at this point.

So I've thought about why the sleep is here in the first place (this one was there from the very beginning from the previous author 😇) and decided it doesn't really solve any actual problem. Now, connect will start early and likely try to connect before networking is fully initiated ‒ IP addresses, etc, are not ready yet. So it'll fail to connect and retry later (it needs that mechanism of retrying anyway).

But that sleep was arbitrary. It sometimes did sleep long enough to have IP address, sometimes it failed anyway and sometimes it slept for too long.

As a result of that, I've decided to just drop it.

src/connect/planner.cpp
```diff
@@ -156,9 +162,9 @@ Action Planner::next_action() {
     }
 
     if (auto current_transfer = Monitor::instance.id(); observed_transfer != current_transfer) {
-        optional<Monitor::Outcome> outcome = observed_transfer.has_value() ? Monitor::instance.outcome(*observed_transfer) : nullopt;
-
         auto terminated_transfer = *observed_transfer;
+        optional<Monitor::Outcome> outcome = observed_transfer.has_value() ? Monitor::instance.outcome(terminated_transfer) : nullopt;
```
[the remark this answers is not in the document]
vorner: Good catch. Result of moving the code around carelessly during development…

src/connect/printer.hpp
vorner: I'm not sure.

Eventually, we'll want to be able to deliver specific notifications about changes ‒ eg. not only „Something has changed“ but „File of this name was deleted“. This could be a good place to put it and in such case, we would want to use the same mechanism from both Connect and Link.

Nevertheless, the reason I've added this now is that I've needed a „callback“ from the Download:
* I didn't want to use a simple function pointer. Such function would have to call `wui_gcode_modified` directly and it would introduce such dependency into the wrong file ‒ into the planner. The planner is, however, supplied with a printer implementation independent of marlin in tests through the Printer interface. I needed a „this“ pointer / functor object / something like that and passing already existing Printer instance pointer seemed best / most lightweight solution.
* On the other hand, depending on connect::Printer from transfers::Download would create a dependency in the wrong direction ‒ these two directories would then circularly depend on each other. That wouldn't directly break anything, but it „smells“.

So in a sense, this is probably a somewhat underutilizing the potential of the approach a bit, but I got to that in a kind of elimination of other possibilities.

That is, if you have a more elegant solution that doesn't have any of the above unwanted properties, I'm happy to adopt it :-).

HelclM: Ok, you convinced me :D

HelclM (approved):

PR #2960, opened 2023-01-27
Send free space to Connect

vorner:
BFW-3355.

### DIFF
--- a/src/connect/marlin_printer.cpp
+++ b/src/connect/marlin_printer.cpp
@@ -17,6 +17,7 @@
 #include <cstring>
 #include <cctype>
 #include <mbedtls/sha256.h>
+#include <sys/statvfs.h>
 
 using std::nullopt;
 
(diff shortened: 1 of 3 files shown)


src/connect/marlin_printer.cpp
```diff
@@ -257,6 +258,21 @@ Printer::Params MarlinPrinter::params() const {
     params.filament_used = Odometer_s::instance().get(Odometer_s::axis_t::E);
     params.has_usb = marlin_vars->media_inserted;
 
+    struct statvfs fsbuf = {};
+    if (params.has_usb && statvfs("/usb/", &fsbuf) == 0) {
+        // Contrary to the "unix" documentation for statvfs, our FAT implementation stores:
+        // * Number of free *clusters*, not blocks in bfree.
+        // * Number of blocks per cluster in frsize.
```
DRracer, line 265: I have a strange feeling, that blocks vs. clusters don't have a well defined meaning. And it gets worse :wink: :
- "traditional" ext2,3,4 filesystems use the word "blocks" for an allocation unit
- FAT filesystems on the other hand use the word "clusters" for the same thing, but a block size is also often referenced here - usually it describes the size of an allocation unit on the physical device (a cluster contains is made of 1 or more physical blocks)

So IMO the `statvfs` documentation is on par with our implementation, it is "just" the traditional confusion between clusters and blocks on FAT filesystems.

DRracer (approved):

PR #2964, opened 2023-01-31
Link support folders

HelclM:
BFW-3148


lib/WUI/nhttp/file_info.cpp
```diff
@@ -33,6 +35,63 @@ namespace nhttp::printer {
 using namespace handler;
 
 JsonResult FileInfo::DirRenderer::renderState(size_t resume_point, JsonOutput &output, FileInfo::DirState &state) const {
+    if (api == APIVersion::Octoprint)
+        return renderStateOctoprint(resume_point, output, state);
+    else
+        return renderStateV1(resume_point, output, state);
+}
```
vorner, line 42: Nitpick: I'd prefer to always use the block brackets, even for 1-line branches, eg:

```cpp
if (...) {
    return stuff;
} else {
    return other_stuff;
}
```

That's a bit more resilient against future adding of lines.

I've already wanted to take this up during a retrospective, but the last retrospective kind of didn't really happen 😇

lib/WUI/nhttp/file_info.cpp
```diff
@@ -33,6 +35,63 @@ namespace nhttp::printer {
...
+
+JsonResult FileInfo::DirRenderer::renderStateV1(size_t resume_point, JsonOutput &output, FileInfo::DirState &state) const {
+    struct stat st;
```
vorner, line 45: Just for extra care, can we do `struct stat st = {};`, so it's all properly zeroed out?

lib/WUI/nhttp/file_info.cpp
```diff
@@ -33,6 +35,63 @@ namespace nhttp::printer {
...
+    // Keep the indentation of the JSON in here!
+    // clang-format off
+    JSON_START;
+    JSON_OBJ_START;
+
+        JSON_CONTROL("\"type\":\"FOLDER\"") JSON_COMMA;
+        JSON_FIELD_BOOL("ro", false) JSON_COMMA;
+        // Note: we can only get a timestamp for folders that are not
+        // /usb/ because that is the root of FATFS and stat() don't work on it
+        if (stat(state.filepath, &st) == 0) {
```
vorner, line 56: This call might be skipped in case the `m_timestamp` doesn't fit and it is resumed before that again.

lib/WUI/nhttp/file_info.cpp
```diff
@@ -33,6 +35,63 @@ namespace nhttp::printer {
...
+            JSON_FIELD_INT("m_timestamp", st.st_mtime) JSON_COMMA;
+        }
+        JSON_FIELD_STR("name", state.filename) JSON_COMMA;
+        JSON_FIELD_ARR("children");
+        while (state.dir.get() && (state.ent = readdir(state.dir.get()))) {
+            if (state.ent->d_type == DT_REG and !filename_is_gcode(state.ent->d_name)) {
```
vorner, line 62: Can we get things like links here now? REG means regular file, DIR is directory… are there others?

lib/WUI/nhttp/file_info.cpp
```diff
@@ -33,6 +35,63 @@ namespace nhttp::printer {
...
+                continue;
```
vorner, line 63: It probably isn't part of this task, strictly speaking, but eventually, we want to be showing all the non-Gcode files too. Is the only thing missing for that removing of this `continue`? 😇

lib/WUI/nhttp/gcode_upload.cpp
```diff
@@ -222,22 +222,67 @@ UploadHooks::Result GcodeUpload::data(std::string_view data) {
...
+        return make_tuple(Status::Ok, nullptr);
+    }
+
+    UploadHooks::Result make_dir(string_view dir) {
+        char path[USB_MOUNT_POINT_LENGTH + dir.length() + 1];
+
+        auto error = prepend_usb_path(dir, path, sizeof(path));
+        if (std::get<0>(error) != Status::Ok)
+            return error;
+        if (mkdir(path, 777) != 0) {
```
vorner, line 245: Again, braces… (below too).

vorner: Also… what if the folder already exists. Is that OK to create it „again“?

lib/WUI/nhttp/gcode_upload.cpp
```diff
@@ -222,22 +222,67 @@ UploadHooks::Result GcodeUpload::data(std::string_view data) {
...
+    // for whatever reason, so for simplicity and performance
+    // this is probably fine
+    std::variant<bool, UploadHooks::Result> check_is_needed(string_view path) {
+        size_t pos = path.find_last_of('/');
+        if (pos == path.npos)
+            return true;
+        char last_dir[pos + USB_MOUNT_POINT_LENGTH + 1];
+        auto error = prepend_usb_path(path.substr(0, pos), last_dir, sizeof(last_dir));
+        if (std::get<0>(error) != Status::Ok)
+            return error;
```
vorner, line 279: Don't we have some function somewhere for this? `basename` or so? :thinking:

lib/WUI/nhttp/gcode_upload.cpp
```diff
@@ -222,22 +222,67 @@ UploadHooks::Result GcodeUpload::data(std::string_view data) {
...
         return make_tuple(Status::Ok, nullptr);
     }
 
+    //Note: if we would have weird name chars, it would fail
+    // only at the end of upload, but if we would create all the
+    // dirs to check, then we would have to also remove them
+    // otherwise we might create leftovers if the upload failed
+    // for whatever reason, so for simplicity and performance
+    // this is probably fine
+    std::variant<bool, UploadHooks::Result> check_is_needed(string_view path) {
```
vorner, line 272: The name `check_is_needed` doesn't really suggest what exactly it checks…

src/common/http/filename_type.cpp
```diff
@@ -0,0 +1,24 @@
+#include "filename_type.hpp"
+#include "../gcode_filename.hpp"
```
vorner, line 2: Would it be better to just unify everything under one (renamed) file instead of having one half in the gcode_filename and another in filename_type? The code is kind of similar to each other…

src/common/http/filename_type.hpp
```diff
@@ -0,0 +1,9 @@
+#pragma once
+
+#include <dirent.h>
```
vorner, line 3: Does `dirent` have to be in the _header_?

lib/WUI/nhttp/file_info.cpp
```diff
@@ -82,7 +149,45 @@ JsonResult FileInfo::DirRenderer::renderState(size_t resume_point, JsonOutput &o
...
+    get_SFN_path(filename);
+    JSONIFY_STR(filename);
+    char long_name[FILE_NAME_BUFFER_LEN];
+    get_LFN(long_name, sizeof long_name, filename);
+    // Keep the indentation of the JSON in here!
+    // clang-format off
+    JSON_START;
+    JSON_OBJ_START;
+
+        JSON_FIELD_STR("name", basename_b(filename_escaped)) JSON_COMMA;
```
vorner, line 169: This seems a little weird. Why are you taking a basename of the _escaped_ version? And, the JSON_FIELD_STR AFAIK already _does_ the escaping, so this might turn out double-escaped.

lib/WUI/nhttp/file_info.cpp
```diff
@@ -82,7 +149,45 @@ JsonResult FileInfo::DirRenderer::renderState(size_t resume_point, JsonOutput &o
...
+        JSON_FIELD_BOOL("ro", false) JSON_COMMA;
+        JSON_FIELD_STR("type", file_type_by_ext(long_name)) JSON_COMMA;
+        JSON_FIELD_INT("m_timestamp", state.m_timestamp) JSON_COMMA;
+        JSON_FIELD_INT("size", state.size) JSON_COMMA;
+        JSON_FIELD_STR("display_name", long_name) JSON_COMMA;
+        if (filename_is_gcode(filename_escaped)) {
```
vorner, line 175: Again, we probably should use the FS functions on the original, not on a JSON-escaped version.

lib/WUI/nhttp/file_info.cpp
```diff
@@ -82,7 +149,45 @@ JsonResult FileInfo::DirRenderer::renderState(size_t resume_point, JsonOutput &o
...
+            JSON_FIELD_OBJ("refs");
+                JSON_CUSTOM("\"icon\":\"/thumb/s%s\",", filename_escaped);
+                JSON_CUSTOM("\"thumbnail\":\"/thumb/l%s\",", filename_escaped);
+                JSON_FIELD_STR("download", filename_escaped);
```
vorner, line 179: Double-escaped.

lib/WUI/nhttp/gcode_upload.cpp
```diff
@@ -17,6 +18,7 @@ extern "C" {
 // Inject for tests, which are compiled on systems without it in the header.
 size_t strlcpy(char *, const char *, size_t);
 size_t strlcat(char *, const char *, size_t);
+int mkdir(const char *path, mode_t mode);
```
vorner, line 21: Prefer to use real headers for these? Or is there some problem, like mkdir not being available on some common systems?

vorner (commented):
Apart from the inline notes. You specialize the rendering functions as completely separate ones. How different are they from each other, would it be better to have just local ifs for variations instead of two similar methods?

HelclM:
> Apart from the inline notes. You specialize the rendering functions as completely separate ones. How different are they from each other, would it be better to have just local ifs for variations instead of two similar methods?

:thinking: I think there would be 3-4 ifs in the function needed to merge those two, it can be done, but personally I feel like this is far more readable and when the time comes to stop supporting the old one, we can just erase the one function.

lib/WUI/nhttp/gcode_upload.cpp
HelclM: `basename` returns the name of the file, meaning everything _after_ the last `/`, I want the parent path (everything _before_ `/`), I have not been able to find any function, that would do that, but I might have missed it somewhere...

lib/WUI/nhttp/gcode_upload.cpp
HelclM: AFAIK it is, it just fails with `errno == EEXIST` for which I check.

src/common/http/filename_type.cpp
HelclM: My thoughts were, that the functions `file_type_by_ext` and `file_type` are specific to networking (the return value in this form does not seem to me to be likely used anywhere else), so I though I would put it inside `common/http`, but the gcode function is usable (and used already) in different places in the FW, :thinking:  But by this logic the `filename_is_firmware` should probly be also in `common` because that one might be used somewhere else.

Or we can just put it all in `src/common/filename_type.(cpp|hpp)` :man_shrugging:

lib/WUI/nhttp/gcode_upload.cpp
HelclM: For some reason I was sure this is needed for tests to pass, but looking at it now, it isn't :sweat_smile:

lib/WUI/nhttp/file_info.cpp
HelclM: :thinking:  Kind of yes, we won't send the firmware metadata, but other than that, it is fine. But I asked and Link is not prepared for it as of now, so we should probably leave it here for now and remove it once Link can actually deal with other files.

src/common/http/filename_type.cpp
vorner: Yes, something like that (`src/common/filename_type`) was what I've had in mind.

lib/WUI/nhttp/file_info.cpp
```diff
@@ -33,6 +35,75 @@ namespace nhttp::printer {
...
+    JSON_OBJ_START;
+
+        JSON_CONTROL("\"type\":\"FOLDER\"") JSON_COMMA;
+        JSON_FIELD_BOOL("ro", false) JSON_COMMA;
+        // Note: we can only get a timestamp for folders that are not
+        // /usb/ because that is the root of FATFS and stat() don't work on it
+        if (stat(state.filepath, &st) == 0) {
+            // Note: We need the timestamp to be persistent betweeen resumes,
+            // so we save it to state.
+            state.base_folder_timestamp = st.st_mtime;
```
vorner, line 60: Well, this is one solution. But that seems to be the more complicated (and larger to store long-term) solution.

I was simply thinking of putting the `stat` before the JSON_START… but that would complicate the situation of the `if(stat(...))` :thinking: … which would require a JSON_FIELD_INT_G… hmm, maybe we can leave it as it is.

lib/WUI/nhttp/gcode_upload.cpp
vorner: So, why is it still here? 😈

vorner (commented):
Just few minor leftovers…

PR #2971, opened 2023-02-05
Connect filesystem manipulation

HelclM:
[BFW-3257](https://dev.prusa3d.com/browse/BFW-3257)

### DIFF
--- a/src/connect/command.cpp
+++ b/src/connect/command.cpp
@@ -101,6 +101,7 @@ Command Command::parse_json_command(CommandId id, const string_view &body, Share
             T("CANCEL_PRINTER_READY", CancelPrinterReady)
             T("START_CONNECT_DOWNLOAD", StartConnectDownload)
             T("DELETE_FILE", DeleteFile)
+            T("DELETE_FOLDER", DeleteFolder)
             return;
         }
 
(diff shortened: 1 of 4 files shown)


src/connect/command.cpp
```diff
@@ -133,19 +145,15 @@ Command Command::parse_json_command(CommandId id, const string_view &body, Share
...
-            // Missing parameters
-            data = BrokenCommand {};
-        }
+        get_path(start->path);
+    } else if (auto *del_file = get_if<DeleteFile>(&data); del_file != nullptr) {
+        get_path(del_file->path);
+    } else if (auto *del_folder = get_if<DeleteFolder>(&data); del_folder != nullptr) {
+        get_path(del_folder->path);
+    } else if (auto *create_folder = get_if<CreateFolder>(&data); create_folder != nullptr) {
+        get_path(create_folder->path);
```
vorner, line 156: :+1: for the deduplication.

In theory, we could have something like a „cycle“ over a typelist… but I don't think such abominations really belong to a real code.

src/connect/planner.cpp
```diff
@@ -72,14 +73,56 @@ namespace {
...
+    bool dir_exists(const char *path) {
+        struct stat st = {};
+        // This could give some false negatives, in practice rare (we don't have permissions, and such).
+        return stat(path, &st) == 0 and S_ISDIR(st.st_mode);
     }
 
     template <class>
     inline constexpr bool always_false_v = false;
+
+    optional<const char *> delete_file(const char *path) {
```
vorner, line 91: I'm not sure if `optional<char *>` isn't a bit of overcomplicated… `char *` has a `nullptr` value 😈

vorner (commented):
Not sure if you agree with my `optional<char *>` observation… if you disagree, I'm fine giving the approve in this version.

vorner (approved):

PR #2979, opened 2023-02-07
Connect STOP_TRANSFER command.

HelclM:
[BFW-3258](https://dev.prusa3d.com/browse/BFW-3258)


HelclM:
> Two things:
> 
>     * I think the `stopped` can be set to true, but is never set back to false (notice that the variable is inside the monitor, which is long-living).
>

It is set back to false in `allocate` so whenever you starting a new transfer it is set to false.

vorner (approved):
✔

PR #2983, opened 2023-02-09
Large_remaining_time_fix

emause1793:
[BFW-3376](https://dev.prusa3d.com/browse/BFW-3376)

### DIFF
--- a/src/connect/render.cpp
+++ b/src/connect/render.cpp
@@ -126,8 +126,11 @@ namespace {
                 // These are not included in the fingerprint as they are changing a lot.
                 if (printing) {
                     JSON_FIELD_INT("time_printing", params.print_duration) JSON_COMMA;
-                    JSON_FIELD_INT("time_remaining", params.time_to_end) JSON_COMMA;
                     JSON_FIELD_INT("progress", params.progress_percent) JSON_COMMA;
+                    //Send only valid time_to_end value
+                    if(params.time_to_end != (uint32_t)-1){
+                        JSON_FIELD_INT("time_remaining", params.time_to_end) JSON_COMMA;
+                    }
                 }
 
                 if (update_telemetry) {


vorner (commented):
Little nitpick.

Also, similar one is needed for the Connect part too. Can be a separate PR, of course.

src/connect/render.cpp
```diff
@@ -126,8 +126,11 @@ namespace {
                 // These are not included in the fingerprint as they are changing a lot.
                 if (printing) {
                     JSON_FIELD_INT("time_printing", params.print_duration) JSON_COMMA;
-                    JSON_FIELD_INT("time_remaining", params.time_to_end) JSON_COMMA;
                     JSON_FIELD_INT("progress", params.progress_percent) JSON_COMMA;
+                    //Send only valid time_to_end value
+                    if(params.time_to_end != (uint32_t)-1){
+                        JSON_FIELD_INT("time_remaining", params.time_to_end) JSON_COMMA;
+                    }
```
vorner, line 133: Style nitpicks:
* The whole JSON_* macro monster is excluded from auto-formatting because indentation. Nevertheless, the rest of that we probably might want to follow, and that includes an empty space between `if` and the parens.
* This is C++, so we shall either prefer C++-style casts (`static_cast<uint32_t>(-1)`) or, even better, `numeric_limits<uint32_t>::max()`.

vorner (changes requested):
The style nitpicks would not be blocking the review, nevertheless:
* It doesn't pass the tests. I'm not aware of having a test case with `-1` time to end value, so the failure is a bit surprising.
* Considering we _do_ have tests for rendering in connect (unlike in link), it may be worth the effort to add a dedicated test case for time to end being unknown and therefore not present at all.
* Please clean up the git history, we probably should not merge this „mess“ ‒ things in pull requests probably should contain only the commits relevant to the pull request and, if not absolutely necessary for some reasons, not contain any merge commits.

src/connect/render.cpp
vladamatena: I am just looking into similar time estimate problems on XL (-1 interpreted as regular time) and there is even `TIME_TO_END_INVALID` constant defined in
https://github.com/prusa3d/Prusa-Firmware-Buddy-Private/blob/private/src/common/marlin_server_shared.h#L9
using the numeric limits.

emause1793: Thanks for the note. I have not noticed the constant as it was not used in any code relevant to my task and therefore I had no idea someone have already introduced it. I was thinking of doing the same thing.

I'll also use it in other places that have been using (uint32_t) casting until now.

PR #3012, opened 2023-03-01
connect: Support resuming encrypted transfers

vorner:
If the transfer fails, be able to recover with the Range header (and, if not supported by the server, just starting from the beginning).

BFW-3502.

### DIFF
--- a/src/connect/planner.cpp
+++ b/src/connect/planner.cpp
@@ -12,9 +12,12 @@
 
 using http::HeaderOut;
 using std::holds_alternative;
+using std::is_same_v;
+using std::make_tuple;
 using std::min;
 using std::nullopt;
 using std::optional;
+using std::tuple;
 using std::unique_ptr;
 using std::visit;
 using transfers::Decryptor;
@@ -58,6 +61,8 @@ namespace {
     // window. Safety measure, as it may be related to that specific event and
     // we would never recover if the failure is repeateble with it.
     const constexpr uint8_t GIVE_UP_AFTER_ATTEMPTS = 5;
+
+    const constexpr uint8_t MAX_DOWNLOAD_RETRIES = 5;
 
     optional<Duration> since(optional<Timestamp> past_event) {
         // optional::transform would be nice, but it's C++23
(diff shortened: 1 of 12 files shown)


src/connect/planner.cpp
```diff
@@ -690,17 +715,60 @@ void Planner::background_done(BackgroundResult result) {
...
-    assert(observed_transfer.has_value()); // Because download still holds the slot.
-    download.reset();
+    char url[enc_url_len];
+    make_enc_url(url, download->orig_iv);
+
+    char range[6 /* bytes = */ + 10 /* 2^32 in text */ + 1 /* - */ + 1 /* \0 */];
+    snprintf(range, sizeof range, "bytes=%" PRIu32 "-", position);
+    HeaderOut hdrs[2] = {
+        { "Range", range, nullopt },
+        { nullptr, nullptr, nullopt },
```
HelclM, line 753: Just a thought, not really related to this PR, but maybe it would be nice to have this "null" terminating line added automatically and not having to put it in every extra header out there :thinking: .

HelclM (approved):

src/connect/planner.cpp
vorner: Any suggestion on _how_? I mean, I could just _omit_ it and rely on the fact there's an extra element in the array and that it would be auto-zeroed, but that IMO hurts readability and resilience of the code against changes (if you add another header and this doesn't fit, it should IMO fail to compile instead of doing some random walking of stack memory).

HelclM: My thought was to automatically add the line, while actually sending the request. So adding it somehow in the `send_request` function and the caller wouldn't need to worry about it at all, since it is needed every single time.

vorner: That line is there to mark the end of the array. We either need a „sentinel entry“ or a size. At that time, sentinel value looked nicer.

That is, the function doesn't _need_ an empty header for an actual purpose, it needs to know where the headers end. Adding it automatically doesn't work, since the function doesn't know where ‒ and if it did, it wouldn't have to add it, as there would be no further purpose.

PR #3014, opened 2023-03-02
De-escape json while parsing it.

HelclM:
It is done in place, so we save memory, but it means, that we mess up the string and it should not be used afterwards.

### DIFF
--- a/src/common/json_encode.c
+++ b/src/common/json_encode.c
@@ -88,3 +88,41 @@ const char *jsonify_bool(bool value) {
         return json_false;
     }
 }
+
+size_t unescape_json_i(char *in, size_t size) {
+    char *write = in;
+    char *read = in;
+    const char *end = &in[size];
+    while (read < end) {
+        if (read + 5 < end && strncmp(read, "\\u0000", 6) == 0) {
+            *write = '\0';
+            write++;
+            read += 6;
+            size -= 5;
+        } else if (*read == '\\') {
+            // to prevent escaping if the \ is the last char we should escape
+            // otherwise we would touch the input after size
+            if (read + 1 == end) {
+                *write++ = *read++;
+                break;
+            }
+            bool escaped = false;
+            for (size_t j = 0; j < sizeof special_chars / sizeof *special_chars; j++) {
+                if (*(read + 1) == special_chars[j].escape) {
+                    escaped = true;
+                    *write = special_chars[j].input;
+                    write++;
+                    read += 2;
+                    size--;
+                    break;
+                }
+            }
+            if (!escaped) {
+                *write++ = *read++;
+            }
+        } else {
+            *write++ = *read++;
+        }
+    }
+    return size;
+}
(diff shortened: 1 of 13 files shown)


tests/unit/common/json_encode_tests.cpp
```diff
@@ -50,3 +50,62 @@ TEST_CASE("Escape len") {
...
+    char json[] = "\\\"abc\\\\12345";
+    int i = unescape_json_i(json, 4);
+
+    INFO("json: " + std::string(json));
+    const char *expected = "\"ab";
+    REQUIRE(i == 1);
+    REQUIRE(strncmp(json, expected, 3) == 0);
+}
+
+TEST_CASE("null unescape") {
```
vorner, line 93: The null is not the only thing that can be encoded as a „number“. Anything (including „nice“ printable characters) can. How hard would be supporting these (and testing)?

tests/unit/connect/command.cpp
```diff
@@ -14,11 +14,11 @@ namespace {
 SharedBuffer buffer;
 
 template <class D>
-D command_test(const string_view cmd) {
+D command_test(char *cmd, size_t size) {
```
vorner, line 17: Wouldn't it be kind of easier to just make a copy of the string here instead of having to add another `char cmd[] =` to each one test?

src/common/json_encode.c
```diff
@@ -88,3 +88,39 @@ const char *jsonify_bool(bool value) {
...
+            replace_first_char(seq, size, special_chars[i].input, 2);
+            break;
+        }
+    }
+}
+
+int unescape_json_i(char *in, size_t size) {
+    int removed_chars = 0;
+    for (size_t i = 0; i < size; i++) {
+        if (strncmp(&in[i], "\\u0000", 6) == 0) {
```
vorner, line 110: What if you're near the end and there _aren't_ 6 chars available?

src/common/json_encode.c
```diff
@@ -88,3 +88,39 @@ const char *jsonify_bool(bool value) {
         return json_false;
     }
 }
+
+void replace_first_char(char *data, size_t data_size, char replace_with, size_t to_replace_size) {
+    data[0] = replace_with;
+    // shift the rest of the string by the amount of replaced chars
+    memmove(data + 1, data + to_replace_size, data_size - (to_replace_size));
```
vorner, line 95: This'll make the whole de-escaping algorithm an `O(n²)`  ☹. Can't you make it that you have a „read pointer“ and „write pointer“ (or offset or whatever) and shovel the decoded / read ones to the write one?

vorner (commented):
I wonder if there's a ready-made solution for the actual deescaping. 

I've seen this one, but I'm not sure about the licence and haven't read it thoroughly. https://gist.github.com/schneidersoft/ca11078da9e0607501b85a0cdef317bc

vorner (approved):

PR #3024, opened 2023-03-07
Fix double preview screen after gcode upload.

HelclM:
There were two seperate mechanisms by which the preview screen was triggered after uploading gcode, so there were two preview screens. This deletes one of them, so there si only one.

BFW-3350

### DIFF
--- a/src/gui/screen_home.cpp
+++ b/src/gui/screen_home.cpp
@@ -60,7 +60,6 @@ const char *labels[7] = {
 };
 
 bool screen_home_data_t::usbWasAlreadyInserted = false;
-uint32_t screen_home_data_t::lastUploadCount = 0;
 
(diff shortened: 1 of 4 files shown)


src/gui/screen_home.cpp
```diff
@@ -298,13 +298,6 @@ void screen_home_data_t::printBtnDis() {
     w_labels[0].SetText(_(labels[labelNoUSBId]));
 }
 
-bool screen_home_data_t::moreGcodesUploaded() {
-    const uint32_t total = wui_gcodes_uploaded();
-    const bool result = total != lastUploadCount;
-    lastUploadCount = total;
```
vorner, line 304: Is `lastUploadedCount` used for anything right now?

vorner (approved):
Always celebrate a diff that removes code :-)

PR #3100, opened 2023-06-12
i2c2: Apply busy workaround at start

RadekVana:
BFW-3158


RadekVana:
This approach is wrong, we need to clock-out message from I2C devices
This workaround worked only because it accessed SCL pin, and it helped only sometimes.

Issue #2325, opened 2022-07-30
[BUG] Can't connect to SSID starting with space


Crocmagnon:
# Bug description

**Printer type** - [MINI]

**Printer firmware version** - 4.4.0
**Original or Custom firmware** - Original

**Optional upgrades** - Filament Runout Sensor

**USB drive or USB/Octoprint**
N/A

**Describe the bug**
I can't connect the Prusa Mini to my wifi network. Its SSID starts with a space.

**How to reproduce**
* Create a wifi network with an SSID starting with a space
* Follow the [guide](https://help.prusa3d.com/guide/how-to-setup-wi-fi-mini-mini_316781) to setup the wifi on the Prusa MINI
* Configure your wifi credentials in the `prusa_printer_settings.ini` file

I've tried the following config entries (with and without quotes):

```ini
[wifi]
ssid= REDACTED SSID
key_mgmt=WPA
psk=REDACTED_PSK
```

```ini
[wifi]
ssid=" REDACTED SSID"
key_mgmt=WPA
psk="REDACTED_PSK"
```

**Expected behavior**
The printer should connect to the wifi network.

I tried connecting it to another AP not starting with a space - tethering with my phone, and it works fine.

**G-code**
N/A

**Crash dump file**
[dump.bin.zip](https://github.com/prusa3d/Prusa-Firmware-Buddy/files/9227182/dump.bin.zip)

# Additional analysis
According to the docs, the code parsing INI files strips whitespaces:

https://github.com/prusa3d/Prusa-Firmware-Buddy/blob/215bcf5b377220655c831c730940324d4dbdbf47/lib/inih/ini.h#L42-L55

> Parse given INI-style file. May have [section]s, name=value pairs (**whitespace stripped**)

The code that strips the whitespaces from the value is here:

https://github.com/prusa3d/Prusa-Firmware-Buddy/blob/215bcf5b377220655c831c730940324d4dbdbf47/lib/inih/ini.c#L195-L196

Also, from what I can see, there is no handling of quotes to delimit strings.

If I take my two config examples:
```ini
[wifi]
ssid= REDACTED SSID
key_mgmt=WPA
psk=REDACTED_PSK
```
This is parsed as:
```json
{
    "ssid": "REDACTED SSID",
    "psk": "REDACTED_PSK",
    "key_mgmt": "WPA"
}
```
(missing leading whitespace)

```ini
[wifi]
ssid=" REDACTED SSID"
[...]

PR #3158, opened 2023-07-24
Support quoted INI string values

nmschulte:
support singly and doubly quoted INI strings, for e.g. leading and trailing spaces in WiFi SSID/PSK field values

addresses #2325


nmschulte:
I see now this parser is actually a copy of https://github.com/benhoyt/inih, which attempts to align with a Python behavior as noted in this interesting report:

https://github.com/benhoyt/inih/issues/51

It would be ideal if this was supported, so please give this PR some consideration.  Otherwise, I guess the ultimate solution to my goals is to support BSSID specification, which I'm happy to add instead, but would like to know the status of this comment first: https://github.com/prusa3d/Prusa-Firmware-Buddy/issues/2325#issuecomment-1268233166

nmschulte:
bump -- @dragomirecky, et al

nmschulte:
Bump.

nmschulte:
Bump.

douglsmith:
Adding my use case in that I am dealing with a long-existing Wi-Fi network having a PSK that ends in a space, which is allowed in the Wi-Fi specifications. Would love to see this PR considered as a solution.

lib/inih/ini.c
```diff
@@ -51,6 +51,18 @@ static char* lskip(const char* s)
     return (char*)s;
 }
 
+/* Strip quotes from single and double quoted strings */
+static char* unquote(char* s)
+{
+    int l = strlen(s);
+    if (s[0] == s[l] && (s[0] == '\'' || s[0] == '"'))
```
nmschulte, line 58:
```suggestion
    if (s[0] == s[l - 1] && (s[0] == '\'' || s[0] == '"'))
```

In review, I think this should be `s[l - 1]` versus `s[l]`.

danopernis: Empty case should be handled as well.

danopernis (changes requested):
@nmschulte sorry for taking so long. Thanks for the patch and also for the digging of related issues in INIH parser, it is interesting reading indeed. Your code seems to be almost working, so let's fix the special case and hopefully get this merged.

lib/inih/ini.c
nmschulte: Empty case refers to `''` and `""`?
How should the empty case be handled?
Currently this will make `s` point to `\0` (really, `\0\0`); an empty c-string.

lib/inih/ini.c
```diff
@@ -194,6 +206,7 @@ int ini_parse_stream(ini_reader reader, void* stream, ini_handler handler,
 #endif
                 value = lskip(value);
                 rstrip(value);
+                unquote(value);
```
danopernis, line 209: Well, this still won't work because you are discarding return value from the `unquote` function.

nmschulte: Pushed a commit to correct this; thanks.

PR #3305, opened 2023-09-23
fix: correct typo in http 500 error text

HumbertoL:
I noticed a small typo on the HTTP 500 error. It should be [500 Internal Server Error](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/500)

Instead, it responds with "Infernal" server error.

image


thijstriemstra:
Let's keep infernal! 😆

Issue #3287, opened 2023-09-20
[BUG] Unable to login to PrusaLink on Safari (macOS/iOS/iPadOS )


pyrho:
Please, before you create a new bug report, please make sure you searched in open and closed issues and couldn't find anything that matches.

**Printer type** - [ALL]

**Printer firmware version** - [>= 5.0.0]

**Original or Custom firmware** - [Original]

**Optional upgrades** - [N/A]

**USB drive or USB/Octoprint** - USB drive

**Describe the bug**
The `nonce` generated by the backend for HTTP Digest authentication (used by PrusaLink) is only valid for 5 seconds (as defined in the code [here](https://github.com/prusa3d/Prusa-Firmware-Buddy/blob/bfb0ffc745ee6546e7efdba618d0e7c0f4c909cd/src/common/http/types.h#L151)).

Safari (contrary to the other browsers it seems) does not re-submit the credentials with a fresh nonce upon receiving a 401 response code. So each time Safari displays its Digest Auth modal window, you **only have 5 seconds** to provide valid credentials, after which the `nonce` will become stale and any login attempt will fail.

This seems to be specific to Safari, as it would seem like the other browsers work around this by re-submitting the authentication request when the `nonce` is stale (using the freshly provided one).

Short of submitting a feature request to Safari, our only hope (assuming we're sticking with Digest Auth),  is to increase the `nonce`'s lifetime to something more sensible. [jvasileff](https://github.com/jvasileff) mentioned in https://github.com/prusa3d/Prusa-Link-Web/issues/386 that [Apache's default `nonce` lifetime is 300](https://httpd.apache.org/docs/2.4/mod/mod_auth_digest.html#authdigestnoncelifetime). This document also states the following:

> this should probably never be set to less than 10 seconds

This timeout is too short, especially on iOS where typing is slower than on a desktop, making it impossible to login in the 5 second time frame allowed.

**How to reproduce**
Try to connect to PrusaLink (FW >= 5.0.0) using Safari (iOS, macOS, or iPadOS), take more than 5 seconds to enter your login and password when prompted; the login attempt will fail.

**Expected behavior**
Users are able to log into PrusaLink using the Safari web browser.

**How to fix**

Patch provided by  [jvasileff](https://github.com/jvasileff) in https://github.com/prusa3d/Prusa-Link-Web/issues/386#issuecomment-1727946657

```diff
diff --git i/src/common/http/types.h w/src/common/http/types.h
[...]

PR #3371, opened 2023-10-18
fix(prusalink): increase nonce valid period

pyrho:
The initial valid period for the HTTP Digest Auth nonce was too low (at 5 seconds).
This make it impossible to log into PrusaLink when using Safari, a saner value of 300 (chosen because it's Apache's default value) does not hinder security and fixes the issue with Safari-based browsers (macOS, ipadOS, iOS).

See https://github.com/prusa3d/Prusa-Firmware-Buddy/issues/3287 and https://github.com/prusa3d/Prusa-Link-Web/issues/386 for more details.


lib/WUI/nhttp/req_parser.cpp
```diff
@@ -297,13 +297,11 @@ bool RequestParser::nonce_valid(uint64_t nonce_to_check) const {
     uint32_t random = static_cast<uint32_t>(nonce_to_check >> 32);
     uint32_t time = nonce_to_check & 0xffffffff;
     uint32_t age = ticks_s() - time;
-    // Make valid period for POST and PUT longer, to avoid infinit uploading
-    // loops if nonce get stale for upload request.
-    uint32_t max_valid_age = has_body(method) ? http::extended_valid_nonce_period : http::valid_nonce_period;
```
pyrho, line 302: This ternary is no longer needed with a `max_valid_age` of 300.

src/common/http/types.h
```diff
@@ -144,11 +144,12 @@ static const size_t MAX_URL_LEN = 100;
 using Url = std::array<char, MAX_URL_LEN>;
 
 // # of seconds after which nonce becomes stale for digest authentication
-// the extended version is used for requests with body, so that PrusaLink
-// hopefully never gets stale nonce for request uploading a gcode, which
-// can cause an infinit upload loop, if the browser does not read errors
-// before sending the whole body.
-static const uint32_t valid_nonce_period = 5;
-static const uint32_t extended_valid_nonce_period = 8;
```
pyrho, line 152: The `extended_valid_nonce_period` variable is no longer needed, since the only value used in now `300`.

pyrho:
> Thank you for finding the cause (I'd still tend to claim the root cause is on safari's side, since the RFC claims the client may wish to retry with new nonce without reprompting the user…).
> 
> I'm going to rebase to get rid of conflicts and fix the tests, then accept it.

Great news, thank you for moving this forward 👌

vorner (approved):

PR #3089, opened 2023-05-28
gui: Fix French Translations

RJRP44:
During my prusa experience, I found some strange or missing French translations.
I hope this helps

PR #3467, opened 2023-11-12
Fixing french unloading filament warning

alecharp:
When changing the filament, the first warning is to validate the filament was successfully removed.
The french translation is in fact asking if the filament was correctly loaded.

This is fixing that.


bkerler:
Please close this PR, it's already included in PR #3089

Issue #3550, opened 2023-12-02
[ENHANCEMENT] Improve time offset to include GMT+13


Kingston802:
**Printer type** - MK4

Currently, New Zealand is in the GMT+13 timezone, but the maximum offset on my MK4 is +12 meaning I can't have an accurate time on my printer, would be great if the offset could account for this edge case

**Describe the expected functionality**
  Time Zone Offset allows for values up to 13

tracernz:
This is the first and only real issue I encountered after assembling my new Mk4 kit. Nice to see it’s already reported 👍.

Prusa-Support:
This issue is solved by FW [6.0.0-RC1](https://github.com/prusa3d/Prusa-Firmware-Buddy/releases/tag/v6.0.0-RC1), which is a Release Candidate and therefore considered to be a rather stable pre-release.

Feel free to give it a go and share further feedback.

Michele Moramarco
Prusa Research

PR #3565, opened 2023-12-07
Fix timezone to reflect correct timezone range from -12 to 14 

bkerler:
Fix timezone to reflect correct timezone range from -12 to 14 as already defined in "src/guiapi/src/menu_spin_config_basic.cpp".

This fix resolves issue #3550.


bkerler:
Closed in favor of commit #3570, which adds additional improvements

Issue #3152, opened 2023-07-21
[BUG] Missing newline in the output of `M46 M` (Show the assigned IP address and MAC address)


ondt:
**Printer type** 
MK4

**Printer firmware version** 
4.7.1

**Original or Custom firmware**
Original

**Optional upgrades**
Nothing

**USB drive or USB/Octoprint**
USB

**Describe the bug**
A newline is missing at the end of the output of `M46 M`.

**How to reproduce**
Send `M46 M` to the printer over USB serial. It returns this:

```
192.168.100.100
de:ad:be:ef:12:34ok
```

**Expected behavior**
It should return this:
```
192.168.100.100
de:ad:be:ef:12:34
ok
```

Note that `M46` (without `M`) works fine:
```
192.168.100.100
ok
```

**G-code**
`M46 M`

**Crash dump file**
Not needed

**Video**
Not needed

Prusa-Support:
Interesting.
Thanks for reporting, we are taking notes.

Michele Moramarco
Prusa Research

bkerler:
Fix has been merged, thus the issue can be closed.

Prusa-Support:
Sure. Thanks for the notice.

Michele Moramarco
Prusa Research

PR #3566, opened 2023-12-08
Add missing newline to mac output on M46 command

bkerler:
This adds a missing newline char to the output of the gcode command M46 (output of ip and mac) and resolves issue #3152.

### DIFF
--- a/src/marlin_stubs/host/M46.cpp
+++ b/src/marlin_stubs/host/M46.cpp
@@ -8,14 +8,13 @@ void GcodeSuite::M46() {
     netdev_get_ipv4_addresses(netdev_get_active_id(), &ethconfig);
     uint8_t *ipp = (uint8_t *)&ethconfig.addr_ip4;
     char ip4[17] = { 0 };
-    sprintf(ip4, "%u.%u.%u.%u\n", ipp[0], ipp[1], ipp[2], ipp[3]);
+    snprintf(ip4, sizeof(ip4), "%u.%u.%u.%u\n", ipp[0], ipp[1], ipp[2], ipp[3]);
     serialprintPGM(ip4);
-
     if (parser.seen('M')) {
         mac_address_t mac;
+        char mac_buffer[19] = { 0 };
         get_MAC_address(&mac, netdev_get_active_id());
-        // Append a newline character to the mac parameter
-        strcat(mac, "\n");
+        snprintf(mac_buffer, sizeof(mac_buffer), "%s\n", mac);
         serialprintPGM(mac);
     }
 }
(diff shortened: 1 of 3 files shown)


bkerler:
The build test failed not because of code issues, but because the prusa holly server complains about "No space left on device" :/

src/marlin_stubs/host/M46.cpp
```diff
@@ -14,6 +14,8 @@ void GcodeSuite::M46() {
     if (parser.seen('M')) {
         mac_address_t mac;
         get_MAC_address(&mac, netdev_get_active_id());
+        // Append a newline character to the mac parameter
+        strcat(mac, "\n");
```
vorner, line 18: Prove be wrong, but this seems not to fit into the buffer. Looking at mac_addr_t, it's 18 chars long. The MAC address is 17 chars long, you add `\n` and then there's `\0` that overflows now.

Also, `strcat` is the unfortunate combination of unsafe (doesn't check the size) and linear walk through the whole string. Can I ask for either placing the `\n` and `\0` at the right positions manually or using `strlcat`? (yes, we prefer the one with `l` over `strncat`).

vorner (changes requested):

src/marlin_stubs/host/M46.cpp
bkerler: Sure :) Let me check on this one

bkerler: I'm also not really happy about the strcat and also about the sprintf without size check on the ip and the mac. So I've come up with a different approach using snprintf instead.

src/marlin_stubs/host/M46.cpp
```diff
@@ -8,14 +8,13 @@ void GcodeSuite::M46() {
     netdev_get_ipv4_addresses(netdev_get_active_id(), &ethconfig);
     uint8_t *ipp = (uint8_t *)&ethconfig.addr_ip4;
     char ip4[17] = { 0 };
-    sprintf(ip4, "%u.%u.%u.%u\n", ipp[0], ipp[1], ipp[2], ipp[3]);
+    snprintf(ip4, 17, "%u.%u.%u.%u\n", ipp[0], ipp[1], ipp[2], ipp[3]);
```
vorner, line 11: Small nitpick 😇: It would be better to either use constant or `sizeof ip4` instead of repeating that 17. The chance of the buffer size changing and getting out of sync is pretty low and they are close together, but it'd still be somewhat nicer.

Same goes for the MAC.

vorner (commented):
Just a tiny nitpick 😇

src/marlin_stubs/host/M46.cpp
bkerler: Sure, no problem :) Should be fine now :)

vorner (approved):

bkerler:
Saw that serialprintPGM(mac); should be instead  serialprintPGM(mac_buffer); I've added a PR #3608 to fix that.

Issue #3141, opened 2023-07-17
[ENHANCEMENT] Summertime / Timezone


CommanderRedYT:
Please, before you create a new feature request, please make sure you searched in open and closed issues and couldn't find anything that matches.

Enter what type of printer or upgrade the enhancement applies to.
**Printer type** - MK4
**Optional upgrades** - None

**Describe the expected functionality**
  A clear and concise description of expected functionality.

Currently the time menu only has an UTC offset, but this means one would have to change that everytime when time switches from summertime to Wintertime. This could be automatically done by including a list of timezones. Optionally one could find out the time zone via Public IP if the printer is connected to the Network and automatically set that

Prusa-Support:
Thanks for your suggestion.

This issue partially duplicates the existing issue https://github.com/prusa3d/Prusa-Firmware-Buddy/issues/639 as an automatic timezone detection would remove the need for selecting timezones and virtually address this issue too.
However, that was an initial implementation of the feature on the Prusa MINI, so I'd keep both issues open for now.

Our developers will look into this.

Michele Moramarco
Prusa Research

Prusa-Support:
This issue is solved by FW [6.0.0-RC1](https://github.com/prusa3d/Prusa-Firmware-Buddy/releases/tag/v6.0.0-RC1), which is a Release Candidate and therefore considered to be a rather stable pre-release.

Feel free to give it a go and share further feedback.

Michele Moramarco
Prusa Research

CommanderRedYT:
@Prusa-Support I do not consider this issue closed

Prusa-Support:
Not sure if the timezone can be detected automatically yet but the wintertime/summertime parameter should be implemented.

@CommanderRedYT Please check the description of PR https://github.com/prusa3d/Prusa-Firmware-Buddy/pull/3570 and test the even newer (and closer to a definitive release) FW [6.0.0-RC2](https://github.com/prusa3d/Prusa-Firmware-Buddy/releases/tag/v6.0.0-RC2).

Michele Moramarco
Prusa Research

Issue #1779, opened 2021-12-17
[ENHANCEMENT] Addition of timezones not a full hour from UTC


Gabriel-06:
Please, before you create a new feature request, please make sure you searched in open and closed issues and couldn't find anything that matches.

Enter what type of printer or upgrade the enhancement applies to.
**Printer type** - MINI(+)
**Optional upgrades** - Filament Runout Sensor

**Is your suggested improvement related to an existing problem? Please describe.**

Currently, in the timezone settings, it is impossible to specify a timezone that is not a full hour offset from the UTC time (e.g. +9:30). As I live in Adelaide and our timezone is +9:30 or +10:30 depending on whether daylight savings is active, the time is always wrong for me.

**Describe the expected functionality**

It would be great to be able to specify timezones that are not a full hour from UTC, either in the form of hours and minutes (9:30) or in just hours (9.5)

I have looked at the code, but I am way too inexperienced with C to implement it myself. I was hoping it was as easy as changing the 1 in `const SpinConfigInt SpinCnf::timezone_range = { { -12, 14, 1 } };` (in src/guiapi/src/menu_spin_config_basic.cpp)  to a 0.5, but as the function is built for working with integers it cannot accept the float value of 0.5. I'd love to help out and make the changes myself but I don't think I have the skills for it sadly.

Prusa-Support:
Thanks for the suggestion.
Recently we have been reworking the clock features so it may be a good chance for me to remind our developers about this.

Michele Moramarco
Prusa Research

Gabriel-06:
Any updates on this @Prusa-Support?

Prusa-Support:
I'm sorry, recently the clock features have been improved in various ways but the timezones haven't been updated yet,
For the time being this issue will remain open.

Michele Moramarco
Prusa Research

dbr:
This would be nice - I'm also in South Australia and the print ETA's are wrong because of this.

Note there's also some +45min timezones, most notably in Nepal https://en.wikipedia.org/wiki/UTC%2B05:45 (and [also some small parts of Australia](https://en.wikipedia.org/wiki/UTC%2B08:45), which seems to love obscure timezone offsets)

The suggestion in https://github.com/prusa3d/Prusa-Firmware-Buddy/issues/639#issuecomment-657608143 of selecting by region-name may solve this better, along with also possibly solving the daylight-saving offsets

Prusa-Support:
[...]

PR #3570, opened 2023-12-09
gui: Add option to set timezone minutes offset & Fix wrong timezone hour offset & Add timezone dst support

bkerler:
This fixes the wrong timezone hour offset range (did include only +12 but not up to +14 hours) and also introduces a timezone minutes option to select a timezone offset in minutes (0 min, 30 min and 45 min).

This addresses the issues #3550, #1779 and #3141.

As the timezone eeprom storage value is only uint8_t, the data could be further optimized and compressed into a bitstream as it also saves space, which would be possible.

My recommendation for storing the timezone data compressed in a uint8_t:
struct tz_offset{
         uint8_t hours:4;            // 0..14 hours offset
         uint8_t mins:2;               // 0=0, 1=30, 2=45 minutes offset
         uint8_t summertime:1; // 0=wintertime, 1=summertime
         uint8_t signed:1;          // 0=negative gmt hour offset, 1=positive gmt hour offset
};

If that is wanted/needed, I can also add these needed eeprom changes as well to use a compressed structure instead.

### DIFF
--- a/src/persistent_stores/store_instances/config_store/store_definition.hpp
+++ b/src/persistent_stores/store_instances/config_store/store_definition.hpp
@@ -61,6 +61,8 @@ struct CurrentStore : public journal::CurrentStoreConfig<journal::Backend, backe
     StoreItem<std::array<char, lan_hostname_max_len + 1>, defaults::net_hostname, journal::hash("LAN Hostname")> lan_hostname;
 
     StoreItem<int8_t, defaults::lan_timezone, journal::hash("LAN Timezone")> timezone; // hour difference from UTC
+    StoreItem<time_tools::TimeOffsetMinutes, defaults::timezone_minutes, journal::hash("Timezone Minutes")> timezone_minutes; // minutes offset for hour difference from UTC
+    StoreItem<time_tools::TimeOffsetSummerTime, defaults::timezone_summer, journal::hash("Timezone Summertime")> timezone_summer; // Summertime hour offset
 
     // WIFI settings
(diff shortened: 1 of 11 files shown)


bkerler:
> Not sure how space-limited this firmware is, but embedding something like https://github.com/evq/utz would be even nicer so we would be able select the time zone and have all offsets + DST applied automatically.

I tried to also implement a list-based approach using a list of countries / timezone settings (without libraries). The problem is, that this would be static (353 and more timezones that change quite often) and very firmware space consuming. Especially for the mini printers it would cause huge issues because of missing space.

src/persistent_stores/store_instances/config_store/store_definition.hpp
```diff
@@ -61,6 +61,8 @@ struct CurrentStore : public journal::CurrentStoreConfig<journal::Backend, backe
     StoreItem<std::array<char, lan_hostname_max_len + 1>, defaults::net_hostname, journal::hash("LAN Hostname")> lan_hostname;
 
     StoreItem<int8_t, defaults::lan_timezone, journal::hash("LAN Timezone")> timezone; // hour difference from UTC
+    StoreItem<time_tools::TimeOffsetMinutes, defaults::lan_timezone_min, journal::hash("LAN Timezone Minutes")> timezone_min; // minutes offset for hour difference from UTC
```
vorner, line 64: I'm afraid the `lan_` prefix (both in the name and the default) is wrong (this has nothing to do with Local Area Network configuration). This mistake slipped through the migration and since its hash is used to identify it in the EEPROM, we can't change it now.

But I would like to avoid bringing the same error into the new items.

bkerler: I've removed the LAN prefix from both variable names and hash names as you recommended.

vorner (approved):

PR #3581, opened 2023-12-12
gcode: Add option to set/show filament color and filament type for filament change using M600 command

bkerler:
Although it was already prepared in the firmware, setting a color to show the filament color on filament change on gcode command M600 was missing and instead a std::nullptr was hardcoded. 

Hereby an option "C" is added to the M600 command in order to set a color value which is being displayed on filament change in form of a spool color.

Example usage would be "M600 C 65280", which corresponds to the RGB value of 0x00FF00, which translates to a green color.

Also an option "T" was added to the M600 command to allow changing the filament type on filament change, which also was prepared but not yet added.
Example usage would be `M600 S"PLA" C"RED"` or `M600 S"PLA" C 16711680`.

### DIFF
--- a/src/marlin_stubs/pause/M600.cpp
+++ b/src/marlin_stubs/pause/M600.cpp
@@ -77,6 +77,9 @@ static void M600_manual();
  *  B[count]    - Number of times to beep, -1 for indefinite (if equipped with a buzzer)
  *  T[toolhead] - Select extruder for filament change
  *  A           - If automatic spool join is configured for this tool, do that instead, if not, do manual filament change
+ *  C[color]    - Set color for filament change (color rgb value as integer)
+ *  C"color"    - Set color for filament change (color name as string)
+ *  S"filament" - Set filament type for filament change. RepRap compatible.
  *
  *  Default values are used for omitted arguments.
  */
@@ -116,10 +119,48 @@ void GcodeSuite::M600() {
 }
 
 void M600_execute(xyz_pos_t park_point, int8_t target_extruder,
-    xyze_float_t resume_point, std::optional<float> unloadLength, std::optional<float> fastLoadLength, std::optional<float> retractLength,
-    pause::Settings::CalledFrom);
+    xyze_float_t resume_point, std::optional<float> unloadLength, std::optional<float> fastLoadLength,
+    std::optional<float> retractLength, std::optional<filament::Colour> filament_colour,
+    std::optional<filament::Type> filament_type, pause::Settings::CalledFrom);
 
 void M600_manual() {
+    char filamenttype[16] = { '\0' };
+    char colourtype[16] = { '\0' };
+
+    auto filament_to_be_loaded = filament::Type::NONE;
+    const char *text_begin = 0;
+    if (parser.seen('S')) {
+        text_begin = strchr(parser.string_arg, '"');
+        if (text_begin) {
+            ++text_begin; // move pointer from '"' to first letter
+            const char *text_end = strchr(text_begin, '"');
+            if (text_end) {
+                auto filament = filament::get_type(text_begin, text_end - text_begin);
+                if (filament != filament::Type::NONE) {
+                    filament_to_be_loaded = filament;
+                }
+            }
+        }
+    }
+
+    if (parser.seenval('C')) {
+        const char *colourtype_ptr = nullptr;
+        if ((colourtype_ptr = strstr(parser.string_arg, " C")) != nullptr) {
+            if (*(colourtype_ptr + 2) == '"' || *(colourtype_ptr + 2) == ' ') {
+                colourtype_ptr++;
+            }
+            if (*(colourtype_ptr + 2)) {
+                strncpy(colourtype, colourtype_ptr + 2, sizeof(colourtype));
+                for (char *fn = colourtype; *fn; ++fn) {
+                    if (*fn == '"' || *fn == ' ') {
+                        *fn = '\0';
+                        break;
+                    }
+                }
+            }
+        }
+    }
+
     const int8_t target_extruder = GcodeSuite::get_target_extruder_from_command();
     if (target_extruder < 0) {
         return;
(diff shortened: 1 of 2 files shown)


src/marlin_stubs/pause/M600.cpp
```diff
@@ -116,10 +119,48 @@ void GcodeSuite::M600() {
...
+                    filament_to_be_loaded = filament;
+                }
+            }
+        }
+    }
+
+    if (parser.seenval('C')) {
+        const char *colourtype_ptr = nullptr;
+        if ((colourtype_ptr = strstr(parser.string_arg, " C")) != nullptr) {
+            if (*(colourtype_ptr + 2) == '"' || *(colourtype_ptr + 2) == ' ') {
```
vorner, line 149: I'm a bit nervous about these pointer arithmetics with `+2` (here and 3 lines below). What guarantees this is in-range? If there's a guarantee from something, a comment here explaining it would be nice.

src/marlin_stubs/pause/M600.cpp
```diff
@@ -160,11 +201,15 @@ void M600_manual() {
         parser.seen('U') ? std::make_optional(parser.value_axis_units(E_AXIS)) : std::nullopt,
         parser.seen('L') ? std::make_optional(parser.value_axis_units(E_AXIS)) : std::nullopt,
         parser.seen('E') ? std::make_optional(std::abs(parser.value_axis_units(E_AXIS))) : std::nullopt,
+        parser.seen('C') ? std::make_optional(filament::Colour::from_string(filamenttype)) : std::nullopt,
```
vorner, line 204: Should it be `colourtype` instead of `filamenttype`?

src/marlin_stubs/pause/M600.cpp
```diff
@@ -116,10 +119,48 @@ void GcodeSuite::M600() {
...
+    auto filament_to_be_loaded = filament::Type::NONE;
+    const char *text_begin = 0;
+    if (parser.seen('S')) {
+        text_begin = strchr(parser.string_arg, '"');
+        if (text_begin) {
+            ++text_begin; // move pointer from '"' to first letter
+            const char *text_end = strchr(text_begin, '"');
+            if (text_end) {
+                auto filament = filament::get_type(text_begin, text_end - text_begin);
+                if (filament != filament::Type::NONE) {
```
vorner, line 139: Is this particular `if` needed for anything?

src/marlin_stubs/pause/M600.cpp
```diff
@@ -116,10 +119,48 @@ void GcodeSuite::M600() {
...
 
 void M600_execute(xyz_pos_t park_point, int8_t target_extruder,
-    xyze_float_t resume_point, std::optional<float> unloadLength, std::optional<float> fastLoadLength, std::optional<float> retractLength,
-    pause::Settings::CalledFrom);
+    xyze_float_t resume_point, std::optional<float> unloadLength, std::optional<float> fastLoadLength,
+    std::optional<float> retractLength, std::optional<filament::Colour> filament_colour,
+    std::optional<filament::Type> filament_type, pause::Settings::CalledFrom);
 
 void M600_manual() {
+    char filamenttype[16] = { '\0' };
```
vorner, line 127: I think nothing ever assigns into this one.

src/marlin_stubs/pause/M600.cpp
```diff
@@ -116,10 +119,48 @@ void GcodeSuite::M600() {
...
+    }
+
+    if (parser.seenval('C')) {
+        const char *colourtype_ptr = nullptr;
+        if ((colourtype_ptr = strstr(parser.string_arg, " C")) != nullptr) {
+            if (*(colourtype_ptr + 2) == '"' || *(colourtype_ptr + 2) == ' ') {
+                colourtype_ptr++;
+            }
+            if (*(colourtype_ptr + 2)) {
+                strncpy(colourtype, colourtype_ptr + 2, sizeof(colourtype));
```
vorner, line 153: The preferred one in our source codes is `strlcpy` (the problem with strncpy is, the result isn't guaranteed to have a trailing `\0` in it ‒ which even seems to trigger a possible out-of-range access in this particular code).

vorner (changes requested):

src/marlin_stubs/pause/M600.cpp
bkerler: Right, I should add a range check for the whole string length.

bkerler: The +2 referred to " C" which is indeed always 2, but might lead to other issues. So I decided to restructure the whole part. Solved with commit [d631509](https://github.com/prusa3d/Prusa-Firmware-Buddy/pull/3581/commits/d6315094ec575102dc032859100f6114a28a341d)

src/marlin_stubs/pause/M600.cpp
bkerler: Indeed :)  Solved with commit [d631509](https://github.com/prusa3d/Prusa-Firmware-Buddy/pull/3581/commits/d6315094ec575102dc032859100f6114a28a341d)

src/marlin_stubs/pause/M600.cpp
bkerler: Exactly. Removed.  Solved with commit [d631509](https://github.com/prusa3d/Prusa-Firmware-Buddy/pull/3581/commits/d6315094ec575102dc032859100f6114a28a341d)

src/marlin_stubs/pause/M600.cpp
bkerler: Yes, will make sure to use strlcpy in the future.  Solved with commit [d631509](https://github.com/prusa3d/Prusa-Firmware-Buddy/pull/3581/commits/d6315094ec575102dc032859100f6114a28a341d)

src/marlin_stubs/pause/M600.cpp
bkerler: Yes, it is needed. It does make sure that the filament type was used correctly.
`M600 S"PLA"` is fine. However if `M600 S"PLA` is sent without quote for whatever reason, it shouldn't continue.

vorner: My point here is ‒ if the result is `Type::NONE` and the `filament_to_be_loaded` is initialized to `Type::NONE`, if the condition wasn't there, it'd simply overwrite the old `NONE` with a new `NONE` 😇.

But that doesn't matter.

vorner (dismissed):

vorner (approved):

bkerler:
@giacomosiena at first look looks like the code was merged but not updated internally. Will track it down.

giacomosiena:
@bkerler any update?

bkerler:
It seems to be included in the current master (6.5.1)

PR #3590, opened 2023-12-14
doc: Add step-by-step tutorial for setting up vscode development + simulator

bkerler:
This adds a detailed step-by-step tutorial for setting up vscode development + simulator for mk4/xxl/mini using ubuntu 22.04.3 and 24.04 LTS


danopernis (commented):
I like the general idea, but there are a few nitpicks. This feels more like a wiki article (which we do not have setup at the moment). Also, significant portion of the guide seems to be dealing with setting up the visual studio on Ubuntu and general usage of the IDE itself. But maybe I am just used to it working on my machine so I forgot how difficult it was to actually set it up when I first started.

What I really like about this patch is the part where you setup the Launch target for Visual Studio. In my opinion, this should be actually merged into the `.vscode/launch.json`

I really dislike having to patch the selftest and wonder if that part is still relevant. I guess I will have to try that. Maybe it is justan old version of the simulator that requires this workaround? @vintagepc would you know?

Also, I am not sure how I feel about having the images commited to the repository.

bkerler:
@danopernis sadly the patch is still required, as the simulator has issues with the precise timing of rpm (@vintagepc can give more insight into that). And I disliked it more to add a specific launch target and integrate that patch into the source code, so having a patch solely for the people using the simulator felt like the best option for now. 

I fully agree that we might put the documentation (and other tutorials) into a wiki or maybe even better, separate repository solely for documentation purposes for developers.

vintagepc:
The fan self-test was unintentionally fixed in the `more-xl-work` branch rather than `main` for the sim. That branch can probably be merged now but of course once again github's CI has changed and broken things behind the scenes so I need to fix that first. :roll_eyes:

bkerler:
I'll wait for the official release 6.3 and then I'm going to update this PR. Seems other issues have emerged right now, like "requests" needed for compilation etc.

Issue #3592, opened 2023-12-14
[BUG] "Mass (W) must be greater than 0" error


jerrygrey:
**Printer type** - MINI+

**Printer firmware version** - 5.1.0

**Original or Custom firmware** - Original

**Describe the bug**
PrusaSilcer emits `M74 W0` on first layer resulting in the error, "Mass (W) must be greater than 0". According to the [comment for M74](https://github.com/prusa3d/Prusa-Firmware-Buddy/blob/15d5ae0b317a37b69b8e7a4720711b962c25b57d/lib/Marlin/Marlin/src/gcode/feature/input_shaper/M74.cpp#L44), the g-code sets "the total mass in grams of everything that is currently sitting on the bed." It seems like a value of zero is valid and expected. The [comparison](https://github.com/prusa3d/Prusa-Firmware-Buddy/blob/15d5ae0b317a37b69b8e7a4720711b962c25b57d/lib/Marlin/Marlin/src/gcode/feature/input_shaper/M74.cpp#L49) should be changed to allow for zero values. 

**G-code**
`M74 W0`

bkerler:
The error "Mass (W) must be greater than 0" is actually related to the M74 gcode command and not the M572 command afaiks: https://github.com/prusa3d/Prusa-Firmware-Buddy/blob/15d5ae0b317a37b69b8e7a4720711b962c25b57d/lib/Marlin/Marlin/src/gcode/feature/input_shaper/M74.cpp#L54

bkerler:
After testing the firmware, the error "Mass (W) must be greater than 0" does only appear with M74 gcode command if the value is 0. The Smooth time (between 0 and 0.2) would instead indicate the error "Pressure advance smooth time (W) value out of range (0-0.2)". I don't see any relation between M74 and M572 in the code to be honest ...

jerrygrey:
@bkerler I double checked and you're correct the two are not related.

jerrygrey:
Updated, related [PR#3594](https://github.com/prusa3d/Prusa-Firmware-Buddy/pull/3594)

bkerler:
I've updated the PR and can confirm that prusaslicer indeed ALWAYS adds "M74 W0" with IS, so the PR does make sense except if prusaslicer shouldn't set the weight to 0 at all.

bkerler:
It's now merged so this issue can be closed :)

PR #3594, opened 2023-12-14
Fix float compare issue in M74

bkerler:
This fixes a int versus float comparison error that will print "Mass (W) must be greater than 0" on a value < 0.49_.
The value m, being float, should be compared against a float value instead. This PR is related to issue #3592.

### DIFF
--- a/lib/Marlin/Marlin/src/gcode/feature/input_shaper/M74.cpp
+++ b/lib/Marlin/Marlin/src/gcode/feature/input_shaper/M74.cpp
@@ -48,7 +48,7 @@ void GcodeSuite::M74() {
 
     if (parser.seen('W')) {
         const float m = parser.value_float();
-        if (m > 0) {
+        if (m > 0.f) {
             params.mass = m;
         } else {
             SERIAL_ECHO_MSG("?Mass (W) must be greater than 0");


bkerler:
Seems that the value m is compared correctly, the error seems to be related otherwise.

bkerler:
It's still worth fixing it, as it's imho better code style to compare a float against a float value and not against an integer value.

jerrygrey:
PrusaSilcer does emit `M74 W0` on the start of first layer. According to the [comment for M74](https://github.com/prusa3d/Prusa-Firmware-Buddy/blob/15d5ae0b317a37b69b8e7a4720711b962c25b57d/lib/Marlin/Marlin/src/gcode/feature/input_shaper/M74.cpp#L44), the g-code sets "the total mass in grams of everything that is currently sitting on the bed." It seems like a value of zero is also valid and expected.

bkerler:
I've corrected the behaviour to be equal or greater than 0. Either the fw is wrong about the weight or prusaslicer :)

vorner (commented):
I'm definitely for the change to a float. I'll try to check that the value isn't used in some way where zero could cause issues first…

vorner (approved):

Issue #3605, opened 2023-12-18
[BUG] Cannot set nozzle temperature higher than 275°C


d4nix-cz:
**Printer type** - [MINI]

**Printer firmware version** - 5.1.2

**Original or Custom firmware** - Original

**Optional upgrades** - Filament Runout Sensor

**USB drive or USB/Octoprint**
  USB flash drive or USB

**Describe the bug**
  When I had FW version 4.4.1 the Mini let me set nozzle temperature to at least 290°C. From version 5.1.0 I'm able to set temperature only at least to 275°C, no matter if I try it from menu or use temperature set in gcode.
As Prusament PC Blend Carbon Fibre filament is recommend to print at 280°C then it's now not possible to print this material on Mini. I've printed several spools of PCCF at 288°C on my Mini in the past without any issue.

**How to reproduce**
  Go to menu > Control > Nozzle > set temperature as high as possible

**Expected behavior**
  Mini should let me set nozzle temperature to 290°C

**G-code**
  _not relevant_

**Crash dump file**
  _not relevant_

**Video**
[link to video](https://photos.app.goo.gl/rymZmoFHrioNbLFc7)

bkerler:
Actually the mini heater max temperature is 290°C and the decrease is due to a satety margin of 15°C. Not really a bug but a safety requirement in order to prevent components from overheating.

Th3Rom3:
Still there is aa small conflict since the MINI is advertised with 280°C max nozzle temp:

https://www.prusa3d.com/product/original-prusa-mini-semi-assembled-3d-printer-4/

Some official clarification/comment  would be nice.

kozross:
[Duplicate of 3489](https://github.com/prusa3d/Prusa-Firmware-Buddy/issues/3489). 

@bkerler - this is wrong, and wildly so. This has never been a problem prior to the 5.0.0 line, and, as the linked issue specifies, this behaviour is in direct contradiction with _at least_:

* The Mini's product page
* The Mini's official documentation in at least one place
* The _rest of the Mini's firmware_ in at least one place (nylon loading routine)

It's a bug, not a feature.

bkerler:
@kozross That's just what the source code says, not my opinion. 

https://github.com/prusa3d/Prusa-Firmware-Buddy/blob/cf299fc7e84532f60ba8615074ea3d201701536c/include/marlin/Configuration_MINI.h#L406

[...]

Issue #3489, opened 2023-11-19
[BUG] Cannot raise nozzle temperature to hot tighten according to instructions


kozross:
Please, before you create a new bug report, please make sure you searched in open and closed issues and couldn't find anything that matches.

**Printer type** - MINI

**Printer firmware version** - 5.1.0-beta

**Original or Custom firmware** - Original

**Optional upgrades** - Filament Runout Sensor

**USB drive or USB/Octoprint**

USB drive (not relevant here)

**Describe the bug**

According to [Prusa's own instructions](https://help.prusa3d.com/article/changing-replacing-the-nozzle-mini_134235#procedure), you need to heat the nozzle to 280C to hot tighten a new nozzle. I have done this before plenty of times without issue on previous firmware versions. 

However, on 5.1.0-beta, I cannot set a temperature any higher than 275 for the nozzle: if I try to go higher, the setting won't change. This is a bizarre state of affairs: either the documentation is no longer correct (which I doubt, since this is a procedure that isn't dependent on software in any way), or there is a bug in the firmware. This is _doubly_ puzzling because the [Mini's own product page](https://www.prusa3d.com/product/original-prusa-mini-semi-assembled-3d-printer-4/) specifies a hotend temperature cap of 280.

**How to reproduce**

1. Select the 'Control' menu from the main screen.
2. Select the 'Temperature' sub-menu.
3. Scroll to the 'Nozzle' temperature and try to raise it.
4. Observe that at 275, the value no longer rises, no matter how much you turn the click wheel.

**Expected behavior**

Either of the following:

* The documentation for the Mini, both technical and product page, should be updated to reflect its new 275C minimum; or
* The firmware should allow setting temperatures as they used to be.

**G-code**

No G-code was used here.

**Crash dump file**

No crash dump file was produced for this.

**Video**

No video is needed to reproduce this problem.

allengregoryiv:
This needs to be fixed, we can't load nylon filament properly either as it asks for 285C at the nozzle in the menu.

kozross:
> This needs to be fixed, we can't load nylon filament properly either as it asks for 285C at the nozzle in the menu.

This is still a problem even in the release version of the firmware.

allengregoryiv:
It just gets to 275 and lets you load but continues to try to reach 285 even though it can't.

[...]

PR #3613, opened 2023-12-18
configuration: Correct the Max Temperature for Prusa Mini

bkerler:
Lower the Max Temp safety margin for Prusa Mini in order to reach advertised nozzle temp of 280°C, related to issue #3489 and #3605.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -9,7 +9,7 @@ assignees: ''
 
 Please, before you create a new bug report, please make sure you searched in open and closed issues and couldn't find anything that matches.
 
-**Printer type** - [MINI]
+**Printer type** - [MINI, MK4, XL]
 
 **Printer firmware version** - [e.g. 4.0.5, ...]
 
@@ -33,7 +33,9 @@ Please, before you create a new bug report, please make sure you searched in ope
   Please attach the G-code you had trouble with. This will make it easier for us to replicate the error.
 
 **Crash dump file**
-  Please attach the crash dump file. This will make it easier for us to investigate the bug.
+  Please send the crash dump file to Prusa by emailing it to reports@prusa3d.com. Sharing this file is important and helps us investigate the bug.
+
+  Do not share the file publicly, as the crash dump contains a raw snapshot of the printer's memory and may include unencrypted sensitive information.
 
 **Video**
   Please attach a video. It usually helps to solve the problem.
--- a/README.md
+++ b/README.md
@@ -5,7 +5,8 @@
 This repository includes source code and firmware releases for the Original Prusa 3D printers based on the 32-bit ARM microcontrollers.
 
 The currently supported models are:
-- Original Prusa MINI
+- Original Prusa MINI/MINI+
+- Original Prusa MK3.9
 - Original Prusa MK4
 - Original Prusa XL
 
@@ -59,15 +60,9 @@ The build process of this project is driven by CMake and `build.py` is just a hi
 - [Eclipse, STM32CubeIDE](doc/editor/stm32cubeide.md)
 - [Other LSP-based IDEs (Atom, Sublime Text, ...)](doc/editor/lsp-based-ides.md)
 
-#### Formatting
+#### Contributing
 
-All the source code in this repository is automatically formatted:
-
-- C/C++ files using [clang-format](https://clang.llvm.org/docs/ClangFormat.html),
-- Python files using [yapf](https://github.com/google/yapf),
-- and CMake files using [cmake-format](https://github.com/cheshirekow/cmake_format).
-
-If you want to contribute, make sure to install [pre-commit](https://pre-commit.com) and then run `pre-commit install` within the repository. This makes sure that all your future commits will be formatted appropriately. Our build server automatically rejects improperly formatted pull requests.
+If you want to contribute to the codebase, please read the [Contribution Guidelines](doc/contributing.md).
 
 #### XL and Puppies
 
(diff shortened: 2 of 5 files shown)


danopernis:
Proposed solution is simple, but it may be dangerous. We should properly test this. The general consensus is that the current MINI temperature controller should be an improvement to the original PID controller and should be able to regulate temperature, so the fix should work.

Proposed testing scenario (thanks @mkbel)
* Start with cold extruder, request maximum temperature (280°C, with this patch)
* Check that we do not overshoot near 290°C
  * Overshooting to 287°C (temporarily) is OK, anything more is NOK
* Now the extruder reached 280°C
* Gradually increase print fan speed until temperature starts dropping
* Stop the print fan
* Check that we do not overshoot near 290°C
  * Overshooting to 287°C (temporarily) is OK, anything more is NOK

bkerler:
Yes, fully agree. It's a safety functionality, so intense testing is needed. If 10 cannot be reached as safety margin (with allowance for 280), and 15 has to stay as safety margin, then the sales description has to be corrected to 275°C asap. Or an option for allowance is going to be added that the user is accepting that there will be no warranty for the extruder or parts of the extruder if the safety margin is set to 10.

kozross:
> Proposed solution is simple, but it may be dangerous. We should properly test this. The general consensus is that the current MINI temperature controller should be an improvement to the original PID controller and should be able to regulate temperature, so the fix should work.
> 
> Proposed testing scenario (thanks @mkbel)
> 
>     * Start with cold extruder, request maximum temperature (280°C, with this patch)
> 
>     * Check that we do not overshoot near 290°C
>       
>       * Overshooting to 287°C (temporarily) is OK, anything more is NOK
> 
>     * Now the extruder reached 280°C
> 
>     * Gradually increase print fan speed until temperature starts dropping
> 
>     * Stop the print fan
> 
>     * Check that we do not overshoot near 290°C
>       
>       * Overshooting to 287°C (temporarily) is OK, anything more is NOK

Is this the official word from Prusa? Because if so, I would _love_ to know how this problem stood quiescent for nearly _four years_, and then got dropped on our heads with nary a word of announcement. If there were indeed safety concerns, these should have been announced, broadly and clearly, with an explanation of what fix has been undertaken, in detail, not smuggled in amid a bunch of other unrelated changes with nary a word.

I'd also like to add that, if indeed the Mini cannot sustain max temperatures of 280C, then the product I've been sold no longer meets the specifications it is supposed to, and some compensation is in order. I've literally paid for a feature I can no longer use.

danopernis:
Let's not be overly dramatic here. I will try to explain how we got there. Bear in mind that I am nobody official, just a developer trying his best here, hoping that his openness and honesty will be appreciated.

During the development of MK4 firmware we stumbled upon the inconsistency in setting of the safety margins and maximum temperatures. At that point in time, we were sure we are not releasing new firmware for MINI any time soon. To stay on the safe side, we went with increasing the safety margin. This change apparently introduced the bug and ended up in MINI release later, where it affects you.

Please be mindful that this is a bug we are trying to fix and I am trying to be honest about it. Proposed fix is very likely sufficient. I am just trying to be extra careful here, given the opportunity to properly test this.

I am pretty confident this will be eventually fixed in future release. In the meantime, if you absolutely need to print with these temperatures before we get this sorted out, I suggest using the older firmware. Harassing the developers online is not going to bring you that fix any sooner.

For the official word from Prusa, I suggest contacting the support.

parsingphase:
It is hard to see this polite, if frustrated, complaint as either "overly dramatic" or "hassling the developers". Admittedly, no developer likes seeing problems with their work or company discussed in public. However, given that this selfsame company explicitly makes its code and issue trackers public, that is where the discussion has to take place, and I would personally be very happy if the last 30 years' complaints about my own work had all been phrased with such restraint as this.

bkerler:
Please keep discussions in the issues, not in the PR, people. It's already hard to keep track of PRs. 

Do keep in mind it's not like we're not trying to find an acceptable solution for you MINI owners (I don't even own a Mini, but still pushed some fixes for it, btw. and I'm doing it in my spare time).

e-dreyer:
I think some transparency is required here. Mini users are not able to rune their own PID autotune or set it easily. These PID values determine the rate of the heating as well as the max temperature. 

The printer's actual "allowed printing temperature by Prusa" is different from the "actual advertised printing temperature by Prusa". What is the reason for this?

Currently, the topic is being evaded, shunned and kept on the down low. I am starting to think, in order to bring the printer's price down, Prusa cheaped out on the power supply. 150W! Which is just pure insanity. 

Is the Mini not able to achieve this max temperture safely due to the actual construction of the hotend, such as materials possibly melting, or can you not heat it safely with the stock power supply? So now you are blind siding us and make us feel as if we are wearing tin foil hats. Explain why we can't heat our printers to the temperature that you advertised and that we bought it for. If I bought the printer for that specific temperature, I cannot use it for my use case.

Did the specification change during the design phase and you did not update it, or were you not able to achieve it and chose to lie to us?

Bambu stole a lot from Prusa, but damn, you guys are now stealing false advertising and shady business practices from their playbook.

bkerler:
@e-dreyer This isn't a discussion board here, it's a pull request. PID autotune isn't even related to this pull request. If you want a discussion, there's a forum for that. Thanks.

danopernis:
@bkerler I have run some tests on MINI and did some digging in the code history. Looks like there was a bug introduced when adjusting the code to be in sync with the spec. I did not observe any problems with thermal regulation when maintaining those 280C with various print fan speeds, so we should be safe here.

Can you please update the PR accordingly? Max temp should be 295 and safety margin should be 15, giving the maximum of 280 in accordance with the spec.

Also, let's remove the PA from the menu in separate PR in order to not clutter this already cluttered PR. Thanks :pray:

bkerler:
Sure, that's great ! Let's have a look

bkerler:
Done :)

PR #3642, opened 2023-12-27
Preheat before checking thermistor wires

jackdesert:


jackdesert:
Targeted wrong repo, sorry

Issue #3643, opened 2023-12-28
[FEATURE REQUEST] PHA filament support


pbozeman:
Both ColorFabb and BeyondPlatics produce a PHA based filament that is more compostible than PLA and PLA based filament blends.  PHA can be broken down in a backyard composter, v.s. needing special industrial composters like PLA.  PHA also contains no microplastics.

I almost print exclusively in PHA due to to being more environmentally friendly.

There's already an open pull request for PHA support: https://github.com/prusa3d/Prusa-Firmware-Buddy/pull/2210, but since it is so old, it isn't clear that old pull requests are being triaged.  This feature request is an attempt to bump this for your attention.  It would be great if you could either land the request, or do something to support PHA on your side. 

Thanks - because I'm sure you all care about the planet too ;)

bkerler:
Right now mainly the profiles are added that are commonly used. But there has already been a discussion to add an option for custom filament type slots instead.

bkerler:
I added a new PR which allows usage of custom filament profiles, which is even more convenient :)

TinkerNinja:
+1 for this - have some 100 hours experience with PHA printing if you need info/support/beta testing

Issue #3633, opened 2023-12-21
[FEATURE REQUEST] no NGEN filament type


metebalci:
**Printer type** - [MK3.9]

**Printer firmware version** - [5.1.2]

**Original or Custom firmware** - [Original]

**USB drive or USB/Octoprint**
USB flash drive

**Describe the bug**
There is no NGEN filament option in the printer but there is in the slicer. When a gcode with NGEN filament type is sent, the printer always issues a warning since NGEN filament has to be loaded as PETG etc.

**How to reproduce**
Choose an NGEN filament in the Slicer, load NGEN filament (but select PETG as there is no NGEN option). The printer issues filament type mismatch warning.

**Expected behavior**
It should be possible to choose an NGEN filament type in the printer.

bkerler:
Hi,
in order to write a PR for your request, can you please
a) confirm that NGEN works on the MK3.9 without issues / modifications
b) the temperature profile you used (recommended values for NGEN are nozzle=240, heatbed=90) ... you can try the PP (prefered but with heatbed set to 90 instead of 100) for example

Thanks

metebalci:
I just started trying an NGEN filament and my understanding is it is like PETG so I do not think there is a major problem, it prints as expected. I used ColorFabb nGen profile in the Slicer.

bkerler:
Ok, it's also officially listed on the Prusa homepage, so the PR should be fine then. Still let me know if you hit any problems with temp (sticking to the bed or stringing).

bkerler:
Oh and please change the subject from bug to feature request. Thanks :)

bkerler:
I added a new PR which allows usage of custom filament profiles, which is even more convenient :)

metebalci:
> I added a new PR which allows usage of custom filament profiles, which is even more convenient :)

Great!

Issue #3589, opened 2023-12-13
[BFW-5513] [FEATURE REQUEST] PCTG as new Material


Keridos:
If it makes sense, enter what type of printer or upgrade the feature request applies to.
**Printer type** - MK4

**Describe the expected functionality**
Add a new material profile for PCTG:
Nozzle temperature: 260°C
Heatbed termperature: 100°C

Two examples for this type of filament
https://www.3djake.com/fiberlogy/pctg-black
https://www.3djake.com/3djake/pctg-black-1

bkerler:
PCTG seems to have the same Temp specs as ASA. Shall it still be added ?

bkerler:
I've added PR #3591 for this request. It would be good if you can confirm that you successfully printed with that material on the MK4.

Keridos:
I printed successfully. Have not fully worked out the retraction settings yet as my 3DJake sample was a bit to moist after shipping (suspect the packaging was faulty). But after drying it prints ok. A bit of stringing, but that's it.

dstulken:
> PCTG seems to have the same Temp specs as ASA. Shall it still be added ?

Yes.  It's much more straightforward from the user perspective.  And combining them ("ASA/PCTG" in the menus, etc) would be a strange as they are completely different filament compositions.   Leaving them as separate items, even if they have similar values under the hood, also allows the temps and profile characteristics to diverge over time if tuning and feedback suggest they print differently and do need different treatment.

bkerler:
OK, then the PR should be fine for that. PCTG is a separate item.

bkerler:
I added a new PR which allows usage of custom filament profiles, which is even more convenient :)

silvvero:
I would also like to requst PCTG as new filament in the mk4.
PrusaSilcer support it in mk4 profiles so the machine itself should also support it.

kvtorp:
Also the XL could use the ability to load PCTG. PCTG is available in PrusaSlicer, but not in the material load menu, leading to lots of warnings when sending a print as the materials do not match

CZDanol:
Hey, I will try to push for adding support for custom material types. Unifying the related issues under the internal ticket BFW-5513

CZDanol:
Hello, we've released a 6.2.0-alpha1 firmware with user material types support: https://github.com/prusa3d/Prusa-Firmware-Buddy/releases/tag/v6.2.0-alpha1

[...]

PR #3657, opened 2024-01-02
[BFW-5513] gui: Add custom filament temperature settings

bkerler:
Refers to issues #3643, #2210,  #3633, #3589, #3667, #3786 and for sure even more.
As many users requested to add custom filament settings for PHA, NGEN etc. and it is forseeable that many more filament types will appear, it does make sense to add custom modifiable filament temperature settings.

Here's an example of the implementation:

Preheat

It was fully tested to be GUI conform on the MK4 and Mini. It also compiled without issues on the XL.

The 4 custom filament slots can be set via "prusa_printer_settings.ini", for example:
```
[custom_filament_1]
name=PCTG
nozzle_preheat=170
nozzle=260
heatbed=100

[custom_filament_2]
name=PHA
nozzle_preheat=170
nozzle=210
heatbed=0

[custom_filament_3]
name=NGEN
nozzle_preheat=170
nozzle=240
heatbed=90

[custom_filament_4]
name=CPE
nozzle_preheat=275
nozzle=210
heatbed=90
```

Copy the prusa_printer_settings.ini to usb stick, then load via "Settings->System->Load settings from file" and it will be available immediately on the printer.


bkerler:
Rebased to v5.1.2 as requested

FVerbaas:
The above procedure would yield 'anonymous' filament names, and post-it notices on the printer. Sequence number for a filament will be different for each printer.
I have no deep insight in the things yet, but the printer does throw warnings if the filament type loaded does not match the filament type specified for the BGCODE files, so there must be a way to identify a material. I suggest to let the printer accept filament type name and temperature for bed and hotend as specified in the BGCODE, and on encounter add the option to add this to the list.
Multiple 'new' materials could be added in one go with a single GCODE file in which a dummy object is printed with a layer in each material.

bkerler:
What do you mean by sequence number ? The filament can be identified and set using the filament name, in this case "CUSTOM_1" instead of "PLA".

FVerbaas:
By sequence number I mean the number behind the "CUSTOM" string. 
I was not aware there is a limit in number of slots. Anyway 4 strikes as a bit low in this respect, but admitted one can jot the local mapping on a post-it. 
Also not sure if the approach now painted avoids the warning by the printer that the filament currently loaded is not the one specified in the GCODE. Of course the warning could be suppressed when the current material is one of the 'CUSTOM' slots, but not sure if that is a desirable situation because this is exactly the situation when such a verification and warning can be useful. 

I do not understand why you say extra gcodes would be neded. The present GCODE files already contain all the necessary info. 


```
; generated by PrusaSlicer 2.7.1+win64


; printer_model = MK4IS
; filament_type = PETG                 <====
; nozzle_diameter = 0.4
; bed_temperature = 90               <====
; brim_width = 0
; fill_density = 15%
; layer_height = 0.2
; temperature = 245                    <====
; ironing = 0
; support_material = 0
; extruder_colour = ""
; max_layer_z = 2.00
; filament used [mm] = 200.92
; filament used [g] = 0.61
; filament cost = 0.02
; filament used [cm3] = 0.48
; total filament used for wipe tower [g] = 0.00
; estimated printing time (normal mode) = 2m 28s
    
```

In short I suggest the printer takes these data for material, hotend temp and bed temp and stores this in the 'custom' slot   indicated by the user, to be used like any other material from that moment on. 
I imagine that for practical purposes the slots get a name like 'CUSTOM_PCGT'.

Another question while I am 'at it' Does the info in these custom slots survive a firmware update?

bkerler:
The sequential number you are refering to is the slot number. It's not possible to give custom slot names such as "CUSTOM_PCGT" as the main issue is, that these values have to be stored in the eeprom as well and the variable needs to be in a hardcoded array that is being stored in the eeprom. It would be possible to store a name for the slot, but still PrusaSlicer would have to configure a slot, as the names are hardcoded in the firmware. I don't see any easy way to change that behaviour without changing huge portions of the firmware, sorry :(

Of course you could also save the custom name to the slots in the eeprom and compare against the filament_type in the gcode, but that would require a lot of eeprom space as you would need to reserve a fixed space like 16 chars for each slot (4) and also the user would need to enter the name manually into the printer, which I haven't seen a functionality in the printer to do that so far. Would be doable, but would also take a lot of time and testing (and also might not be possible for the mini printer).

4 Slots are the max limit due to a max of 16 items on the screen, and 12 menu items are already hardcoded. Changing that limit might cause big trouble for the mini printer fw due to hw restrictions and does also mean huge changes to the current firmware as far as I could see so far.

FVerbaas:
Thank you Björn for the extensive answer. I understand the issue better now. The LCD screen thing is not just a monitor and peripheral for the buddy board. This may be explained from a historical background but too bad a new printer type launched less than a year ago now already runs into such limitations. 
I see that on the short term the proposed patch solution will probably be the best one feasible.  
At present there is a comparison between the material name specified in the GCODE and the material saved in the printer as 'currently loaded'. If there is a discrepancy the printer issues a warning. If the currently loaded material is one of the 'custom' types, could one avoid that warning by editing the name of the material in the GCODE, so replace the string 'PCTG' by 'CUSTOM_1'?

bkerler:
I will seek into that, but not sure if there will be an easy way to do that, @FVerbass :D

FVerbaas:
I just defined custom material CUSTOM_1 in Prusaslicer (saved a copy of Fiberlogy PCGT) and made a GCODE file with that material type.  I attached a zipped copy. 
I do hope this would pass the printer's checks without complaints. It is just up to me to keep that Post-it on my printer to remind me that my CUSTOM_1 material is PCGT. ;-)

[Shape-Box_0.4n_0.15mm_CUSTOM_1_MK4IS_29m.zip](https://github.com/prusa3d/Prusa-Firmware-Buddy/files/13929155/Shape-Box_0.4n_0.15mm_CUSTOM_1_MK4IS_29m.zip)

bkerler:
Thanks, will try the gcode on the simulator :)

bkerler:
Seems there is an issue as the gcode preview does truncate the filament type to "CUSTOM_", missing the slot. Gonna need to have a look if CUSTOM_1 needs to be replaced by "CUST_1" or if the length can be increased.

bkerler:
indeed, there is a limitation in [filament_buff](https://github.com/prusa3d/Prusa-Firmware-Buddy/blob/dfc2bc6235505eacee24846e1659d693ed040e80/src/common/gcode/gcode_info.hpp#L79) to have a max size of 8 chars

bkerler:
I've solved it now by increasing the char buffer size to 9 :) Tested and it seems to work fine printing your gcode @FVerbaas.

FVerbaas:
Excellent!
The actual name is immaterial of course, as long as it can be read and
lines up with what is hardcoded.

Thanks and regards,

Frans

Op za 13 jan. 2024 23:55 schreef Bjoern Kerler ***@***.***>:

bkerler:
Frans, yes, having the filament name instead of "CUSTOM_1" would be great, but I don't see any easy way to set a custom name via the user interface right now, except we introduce a new gcode to allow setting custom names.But that wouldn't be very userfriendly and may introduce security issues. Guess I need to discuss that with the prusa guys, maybe it can be added to the web interface.

FVerbaas:
There is a workaround for the short term but yes please raise it with them. It strikes as somewhat unprofessional. One buys a 1000+EURO printer; the launched less than a year ago top line model of a top notch manufacturer, and then one has to go exploiting loopholes in the manufacturer's own slicer app (writing in what appears to be intended a pure drop-down selection box) to make a clone of a known material definition only to give it a special name so it can be recognized by the printer.

bkerler:
I had a chat with the devs and there is currently a development for text entry. So    I will prepare the commit for adding custom names to the slot, then the user shouldn't have to remember custom names ;)

FVerbaas:
Great!
Thx.

Op di 16 jan. 2024 09:18 schreef Bjoern Kerler ***@***.***>:

FVerbaas:
Great!
Thanks for letting me know.

Frans

Op zo 21 jan. 2024 16:48 schreef Bjoern Kerler ***@***.***>:

KlausKraemer:
I just loaded Willowflex filament into my XL. Recommended nozzle temperature is from 180°C to 200°C.

Nearest filament to load Willowflex is PLA with 215°C. At this temperature, Willowflex is nearly flowing through the nozzle by itself. Way too liquid. 215°C ist too hot for Willowflex.

I expect problems when unloading, as retraction leads a still overheated filament back to the extruder.

**We really need common filament settings!** The sooner, the better! And please reconsider to place these custom settings editable on the USB Stick. I wouldnt like to fiddle around with the printer display.

bkerler:
Hi KlausKraemer, right now the filament slots aren't in the prusasettings.ini, but I might have a look if it can be added to my PR.

bkerler:
@KlausKraemer Code was updated to now use prusa_printer_settings.ini for setting up the custom filament slots, including names.

KlausKraemer:
@bkerler Thanks - still I'm not sure that solves my and others problem. I suppose the "prusasettings.ini" is used for compiling a FW version? That would sadly be way beyond my skills.

My idea had been a file stored on the USB-Stick (e.g. "customfilaments.txt"), where (if found on the stick) the printer looks for special filament requirements. That way any user could easily add a special filament, if needed. Ideally editing/creating and transferring the file could be done in Prusa Connect later?

bkerler:
@KlausKraemer No, it's way easier. Just create a "prusa_printer_settings.ini" with this content on the usb stick (adopt for your settings), and then use Settings->System->Load settings to load it immediately to eeprom:

`[custom_filament]
name_1=PCTG
nozzle_preheat_1=170
nozzle_1=260
heatbed_1=100

name_2=PHA
nozzle_preheat_2=170
nozzle_2=210
heatbed_2=0

name_3=NGEN
nozzle_preheat_3=170
nozzle_3=240
heatbed_3=90

name_4=CPE
nozzle_preheat_4=275
nozzle_4=210
heatbed_4=90`

KlausKraemer:
My XL has FW v ersion5.1.2 13478

The root directory of the USB Stick contains the new file "prusa_printer_settings.ini" 
image

When I try to load via "Settings -> System -> Load settings from file" absolutely nothing at all seems to happen - there is no feedback about having read something - nada ...

Despite getting no feedback, I also tried to see the new defined custom filament in "preheat", but there's still no WFLX 

File content:
image

File is encoded in UTF-8.

bkerler:
@KlausKraemer yeah ... same for me if the prusa_printer_settings.ini is only having the short entry. Let me check why this happens.

bkerler:
@KlausKraemer I've updated the commit. Seems there is a bug in the ini parser which I need to sort out next. But with the current commit it should work fine.

bkerler:
custom_filament_wlx

KlausKraemer:
Looks great.

As I'm not quite familiar with github, what's a commit? Is it something like a FW beta?

Sorry for my stupid questions, but I suppose I should download another FW version?

bkerler:
PRs are pull requests. It means that it is custom code that needs to be reviewed and merged into the main code. Then, once fully tested it might or might not be released in the release firmware :)
A commit is a change of the code. So if you want to use it on your printer, you either choose to break your warranty by breaking the fuse, compile the firmware and flash it to your printer, or you have to wait (hopefully not long) for it to make it into the release firmware.

KlausKraemer:
Thanks for the enlightening, your patience and especially your work!! 🙋‍♂️
Klaus

bkerler:
Also wrote a PR #3769 for giving user feedback in the future for "Load settings", @KlausKraemer. It's indeed annoying to get no feedback at all if loading the settings worked or not.

FVerbaas:
Looks good. Any idea when the release of this firmware update can be expected?

tkircher:
Starting to regret getting the MK4 instead of the X1-Carbon.

bkerler:
@tkircher please do not hijack this PR for your opinions. This shall stay strictly code-related. Thanks for your understanding !

KlausKraemer:
Does a filament change from a high temperature filament to a low temperature filament remember to first purge at high temperature? Maybe some higher temperature differences should ask for loading another filament first if the temperature difference is too high?

And I think retraction of the new low temperature filament should be delayed until the nozzle has cooled down enough. As I mentioned before, I actually had problems to remove Willowflex, when the unloading took place with PLA temperatures. The filament wouldn't fit back into the bowden tube anymore as a blob had built and some of the filament wrapped around the extruder gear.

TOMASNetr:
Hi, 
I tried to edit the "prusa_printer_settings.ini" file with the first lines being the wifi configuration and then adding a custom-filament. Wifi got through to the printer just fine, but Custom-filament didn't. Does it still have to be enabled somewhere or what should the entire record be, including the wifi record? Or should I load the wifi (the printer will remember without a flash drive) and make a new "prusa_printer_settings.ini" file with just the custom-filament settings?
thanks

KlausKraemer:
> Hi, I tried to edit the "prusa_printer_settings.ini" file with the first lines being the wifi configuration and then adding a custom-filament. Wifi got through to the printer just fine, but Custom-filament didn't. Does it still have to be enabled somewhere or what should the entire record be, including the wifi record? Or should I load the wifi (the printer will remember without a flash drive) and make a new "prusa_printer_settings.ini" file with just the custom-filament settings? thanks

See @bkerler 's comment above: https://github.com/prusa3d/Prusa-Firmware-Buddy/pull/3657#issuecomment-1951442158

bkerler:
> Does a filament change from a high temperature filament to a low temperature filament remember to first purge at high temperature? Maybe some higher temperature differences should ask for loading another filament first if the temperature difference is too high?
> 
> And I think retraction of the new low temperature filament should be delayed until the nozzle has cooled down enough. As I mentioned before, I actually had problems to remove Willowflex, when the unloading took place with PLA temperatures. The filament wouldn't fit back into the bowden tube anymore as a blob had built and some of the filament wrapped around the extruder gear. Better, there is some oozing than a calamity ,,,

There is a related PR already for this as well ;)

KlausKraemer:
I had it in mind, but i believe both issues are thoroughly interwoven and connot be separated. Also setting the filament type without loading / unloading could make life easier.

bryn51:
This development is great, and would suit my own immediate needs,  but I suggest it can only be half of the story.
My objections to this method of approach:
1.  It includes no changes to the operation of Prusa Slicer, which would need to have matching alterations to support the filament types.  PS generates GCode for each model and it includes filament type string that is recognised by the printer and matched with filament loaded in the printer at the time of printing.  The user is delivered an error message (which can be ignored) if the filament type in the gcode does not match what filament is loaded.  
I already implemented PMMA filament type in PS, using PETG as the filament type (as it sorta matches PETG in temp characteristics), but this already gives up such an error message.
So, the loop is not closed on this development, its one end of the solution only.
2. I object to the use of flat text ".ini" format in the settings file.  I do not see why Prusa do not use properly structured industry standard XML format. Use of XML leads to the possibility at some later date of being able to edit the file using a simple editor that would obviate the possibility of simple typos causing problems loading and parsing the file.
The ini file already exists and the type of information that could be within it is not documented.  The information relevant to this added development needs to form part of such documentation.  It needs to appear first in the Release notes for whatever firmware version in which this feature appears.
3. The solution calls for greater technical/computer savvy than held by the majority of users. It's a programmer's solution, not one for the masses of folks who can't even read readily available solutions available online for basic printing issues, and not sufficiently savvy in 3D printing to diagnose their own printing issues, let alone understanding how to edit an ini file.  And yet the range of available filaments keeps growing.  It's one thing to develop a mechanism for handling of additional filament types under the hood, but further development is needed to make it accessible to the masses.  Using Prusa Printers ought not to be confined to a tech savvy elite.
4. You might say that all of the above development is outside of your own scope.  That's fine, but SOMEBODY needs to put this together as a complete technical package, with changes for PS included in a parallel release.

bkerler:
@bryn51 Although I do understand your opinions do keep in mind this is a PR and no issue. 

But regarding your points:
1. That is happening in the PrusaSlicer repo after the commit was accepted and this is the first of many commits to follow.
2. Agree. But adding xml or json adds more complexity and also will make use of more firmware space, which for example the Prusa Mini simply doesn't have. Also this is not PR related, as this is about adding custom filaments.
3. Yes, but there is already additional work done internally by the Prusa team in their repo, this is just the open source repo. I'm just adapting to the things already published by them.
4. Yup, I'm doing that in my spare time, but I do coordinate with the prusa devs and you can be sure that this is a very small part of many many people involved that are even paid for maintaining the code. But some of the code isn't published yet as it is being tested and developed internally and people want a solution right now and to do not want to wait several months or years.
5. That's a limitation due to a max of 16 items in the XLCD menu structure and might change in the future with new board revisions with more ram and flash memory but the limitation is here right now.

bryn51:
> @bryn51 Although I do understand your opinions do keep in mind this is a PR and no issue. 
> 
> 
> 
> But regarding your points:
> 
> 1. That is happening in the PrusaSlicer repo

>> great to hear it. But Prusa Slicer is in itself long overdue for a complete review. Its UI is dated, illogical and overly complex. They have their work cut out for them.

> 
> 2. Agree. But adding xml or json adds more complexity and also will make use of more firmware space, which for example the Prusa Mini simply doesn't have. Also this is not PR related, as this is about adding custom filaments.

>> Here's a suggestion: You already have code for parsing of an ini file and an xml parser is somewhat similar in form and function.
Why not put the entire list of supported filaments out to the same xml file and have it generated by PS, so your own copy is shared between printer and slicer. 
There already exist (for 20+ years) efficient parser libraries for structured xml that could replace the ini parser. 

Supporting the mini ? Why should that entry level printer dictate capabilities for the printers having more resources? Compiler switches could be used to control code elements for each printer, as they already do.
   
> 
> 3. Yes, but there is already additional work done internally by the Prusa team in their repo, this is just the open source repo. I'm just adapting to the things already published by them.
> 
>> Ahh ok. So you aren't a Prusa staffer, and this code base is not necessarily what we see in production release versions. unless prusa like it and include it, I guess. All folks doing work on code such as this should be paid IMHO.
  
> 4. Yup, I'm doing that in my spare time, but I do coordinate with the prusa devs and you can be sure that this is a very small part of many many people involved that are even paid for maintaining the code. But some of the code isn't published yet as it is being tested and developed internally and people want a solution right now and to do not want to wait several months or years.
> 
> 5. That's a limitation due to a max of 16 items in the XLCD menu structure and might change in the future with new board revisions with more ram and flash memory but the limitation is here right now.

> >> Sounds like a great reason to put all of the filament definitions into a single xml file. Folks could add stuff they use within the limit, and remove those they do not use.
> 

Thanks for your effort on this. Given that it's voluntary it is laudable.
There had to be a reason why your post was so informative and civil.

bkerler:
@bryn51 yes of course and there are already defines just for the mini. But you need to see the overhead that is produced if you have to mix ini and json/xml in the documentation for each printer .. and the issues it does cause with maintaining two different code bases. But be assured, I will discuss the ini stuff with the crew.

dpenezic:
> @bryn51 yes of course and there are already defines just for the mini. But you need to see the overhead that is produced if you have to mix ini and json/xml in the documentation for each printer .. and the issues it does cause with maintaining two different code bases. But be assured, I will discuss the ini stuff with the crew.

I understand the issue connected with two code bases, but not understanding that the world around you is changing is another thing, At the same time have you ever heard for if command? :) heawiliy use when you create code for different types of environment

bkerler:
@dpenezic not sure I do get your point. There are two different code bases, one is public (here), but the main development is happening internally and non-public. So I have to wait for new releases and check the tag to see what has changed. But no matter what, in the end it's Prusa's decision what they add and what they don't :)

Kyle0654:
Just watching this, but if it's worth anything, I've got a mini and would really appreciate being able to add custom filament types. I've got a filament that wants zero bed temperature, and it's a pain trying to calibrate a new sheet with it (and then have to pretend it's PLA otherwise e.g. when loading).

bkerler:
Currently broken as v6.0 seems to have missing merged https://github.com/prusa3d/Prusa-Firmware-Buddy/pull/3769/commits/474a2743255ec7b881fbfd3a299d68b588097a91

bkerler:
Should work with v6 again :)

muesli:
Just to chime in how badly this is needed. There's a whole range of filaments below 200C that are currently unsupported by the firmware.

CZDanol (changes requested):
This is not mergeable, sorry.
* Heavy code repetition
* This needs to be coordinated with the Slicer team for filament type checks and such
* There are some conceptual changes that we want to do to prevent this change introducing more technical debt - for example not using responses in the preheat menu

We do want to have this feature, don't take me wrong.

CZDanol:
@KlausKraemer discussion here is meant to be only about this PR. Please re-post it in some of the github issues, or create a new one.

That being said, I unfortunately don't understand your request, please try rephrasing it a little.

bkerler:
> This is not mergeable, sorry. While this would possibly work, it introduces a code that is quite hard to manage.
> 
>     * Heavy code repetition
> 
>     * This needs to be coordinated with the Slicer team for filament type checks and such
> 
>     * There are some conceptual changes that we want to do to prevent this change introducing more technical debt - for example not using responses in the preheat menu
> 
>     * The UI design for this feature is possibly worth revising as well
> 
> 
> We do want to have this feature, don't take me wrong.

@CZDanol 

Thx for the feedback. Regarding your points:

- Heavy code repetition: Yes, I also was quite unhappy, so I restructured the code. Let me know anything else that you think should be optimized / handled differently.

- This needs to be coordinated with the Slicer team for filament type checks and such: This shouldn't be needed as it doesn't interfer with the Slicer but instead uses the values that the Slicer is setting in the gcode

- There are some conceptual changes that we want to do to prevent this change introducing more technical debt - for example not using responses in the preheat menu: Yep, any ideas for that ?

- The UI design for this feature is possibly worth revising as well: If you are referring to the XL, yes indeed, there are already many issues regarding the footer and status design

CZDanol:
Internal ticket for this feature: BFW-5513

FVerbaas:
An intermediate solution would be to work from the settings in the GCODE file that is to be printed. The header of the file, at least those generated by PrusaSlicer, contain the printing temperature. The printer could use that for loading filament. 
I do not know how standard this is, so it may not work for all GCODE files, but if it at least works with PrusaSlicer, the current (embarassing) gap between the slicer and the printer would be closed. A stand-alone loading sequence for a specific material can be defined in a GCODE file.

bkerler:
thx @CZDanol. Highly appreciated.

muesli:
Now that we even have official Prusa filament with a recommended temperature below 200C, I really hope this becomes a priority.

image

bryn51:
Its great to see this is progressing. Several times I have used unusual filaments but found it a pain having to mess around with temps or compromise "similar" filament type selections in the slicer and during load/unload. Better to identify the material correctly.

May I suggest the Prusa Connect folks might need to have a look at this too, because the statistics screen identifies filaments by type.

CZDanol:
@bryn51 we are coordinating with the Connect guys, but it will probably be first implemented on the printers and then receive full Connect support some time later.

FVerbaas:
Good to hear this now gets attention. It really is a shame that we have to
use all these tricks to fool the printer and the software, even for
filament sold by Prusa themselves.

Op vr 26 jul. 2024 08:37 schreef Danol (Daniel Čejchan) <
***@***.***>:

Lyude:
> > There are some conceptual changes that we want to do to prevent this change introducing more technical debt - for example not using responses in the preheat menu: Yep, any ideas for that ?
> 
> Yes, I do. But those changes need to be implemented internally after discussion with the whole team and such. Not suitable for an external PR, sorry. I will push for doing it.

????
Could there be more explanation on this? I have to assume I'm missing some context here, maybe this is something that needs to happen internally due to safety certification requirements or something that just needs to happen in sync with services that Prusa hosts like PrusaConnect?

CZDanol:
> > > There are some conceptual changes that we want to do to prevent this change introducing more technical debt - for example not using responses in the preheat menu: Yep, any ideas for that ?
> > 
> > 
> > Yes, I do. But those changes need to be implemented internally after discussion with the whole team and such. Not suitable for an external PR, sorry. I will push for doing it.
> 
> ???? Could there be more explanation on this? I have to assume I'm missing some context here, maybe this is something that needs to happen internally due to safety certification requirements or something that just needs to happen in sync with services that Prusa hosts like PrusaConnect?

Yes, you are correct. Filament type support is definitely something that needs to be consulted with the Connect, and the Slicer teams. I've mentioned that just above your comment :)

I was also responding to "for example not using responses in the preheat menu" - and that is a very technical thing. Basically, that is a rather large conceptual change in the codebase that requires consultations across the dev team, so it's not something that can be just handed to us by an external developer.

Lyude:
> > > > There are some conceptual changes that we want to do to prevent this change introducing more technical debt - for example not using responses in the preheat menu: Yep, any ideas for that ?
> > > 
> > > 
> > > Yes, I do. But those changes need to be implemented internally after discussion with the whole team and such. Not suitable for an external PR, sorry. I will push for doing it.
> > 
> > 
> > ???? Could there be more explanation on this? I have to assume I'm missing some context here, maybe this is something that needs to happen internally due to safety certification requirements or something that just needs to happen in sync with services that Prusa hosts like PrusaConnect?
> 
> Yes, you are correct. Filament type support is definitely something that needs to be consulted with the Connect, and the Slicer teams. I've mentioned that two comments above yours :)
> 
> I was also responding to "for example not using responses in the preheat menu" - and that is a very technical thing. Basically, that is a rather large conceptual change in the codebase that requires consultations across the dev team, so it's not something that can be just handed to us by an external developer. We would need to integrate him in the team deeply and well, then he wouldn't be external anymore.

Ah, apologies I missed that comment above - thank you for the clarification! My main reason for asking was because I (and almost everyone I've talked to) bought into Prusa's ecosystem to help support open source development. This being said - an easy solution to the problems described here (I mean, except for PrusaConnect of course) would simply be to default to having discussions around this kind of design in the open wherever possible - as doing so dramatically reduces the complexity in working across teams with outside contributors and makes this sort of thing a non-issue. It's also kind of the standard in FOSS

Would a compromise of some sort here be acceptable in the mean time? Even being able to only configure a custom temperature through an ini file would be quite useful, and seeing as prusaslicer is open source and on github already this seems like something that would be pretty easy to coordinate. Functionality that Prusa in the GUI that prusa wants to keep in sync with services like PrusaConnect could certainly come later and I think a lot of people would still be pretty happy with that. This is also basically the way that Wi-Fi passwords were handled in Prusa's firmware up until now, and that didn't seem to be too much of an issue.

CZDanol:
@Lyude No need for compromises :) We've just released a 6.2.0 alpha1 FW with a fully featured user material support :) https://github.com/prusa3d/Prusa-Firmware-Buddy/releases/tag/v6.2.0-alpha1

Closing this PR.

bryn51:
- [ ] Please advise what is related ticket number on Prusa Slicer. Will change releases be coordinated ?

CZDanol:
@bryn51  There are no changes needed on the PrusaSlicer side.

CZDanol:
image

CZDanol:
Oh, actually, I forgot - yes, there will be some small changes in the Slicer. They will be adding fields to configure filament abrasivity and extruder hardened/high flow nozzles

bryn51:
> @bryn51  There are no changes needed on the PrusaSlicer side. Everything should work out of the box.

Thanks for your response, but it begs the question:
Apropos PS generates gcode that includes short mnemonic generic recognition code of filament, seen in dropdown list on PS filament profile you depicted, and this is (I believe) used on printer firmware to alert the user when filament loaded does not match the gcode. It's part of default filename too. The filament code also finds its way to Prusa Connect, and used for statistical purpose and in a couple of other places.
So on PS, the expected change would be to add several "custom" filament codes into the dropdown list.
 Ideally the filament type recorded in Connect would be an mnemonic of the material not "custom_1" etc. 
Thus changes, albeit small ones, might be expected on both Connect and PS.  
That said the ability to have custom type added to the printer simplifying filament load will be a significant improvement by itself, so long as unexpected errors or behaviors do not occur. I guess that must be a part of testing regime.
I'm wondering if I should perhaps bring my own experience to bear on this (normally I steer clear of test versions).  I'm definitely interested in the outcome.

CZDanol:
@bryn51
* The idea is not that you use "CUSTOM1" for your custom filaments, but rather the mnemonic for it. For example "PCTG".
* In PrusaSlicer, you are not limited to the options in the dropdown, you can enter any string you want (as demonstrated in the screenshot above).
* So for example, you add a custom profile for "PCTG" in the printer, you set up a PCTG filament profile in the slicer, and everything will work (even Connect should, though your filament profiles will not be in the preheat list, at least for now).
* The same goes if you have say glow in the dark PLA. You'd simply create something like "GLPLA"

bryn51:
> Oh, actually, I forgot - yes, there will be some small changes in the Slicer. They will be adding fields to configure filament abrasivity and extruder hardened/high flow nozzles

Thats also a big deal and welcome. unclear what behavior change on PS or the printer might occur, but speculatively one might expect it to influence maximum Flow rate and motion speed.(for CHT).

CZDanol:
> Thats also a big deal and welcome. unclear what behavior change on PS or the printer might occur, but speculatively one might expect it to influence maximum Flow rate and motion speed.(for CHT).

Flow rate and motion speed is determined by the printer profile in the slicer. These fields will be intended for compatibility checks.

bryn51:
> @bryn51
> 
> * The idea is not that you use "CUSTOM1" for your custom filaments, but rather the mnemonic for it. For example "PCTG".
> 
> * In PrusaSlicer, you are not limited to the options in the dropdown, you can enter any string you want (as demonstrated in the screenshot above).
> 
> * So for example, you add a custom profile for "PCTG" in the printer, you set up a PCTG filament profile in the slicer, and everything will work.
> 
> * Connect should be able to crunch your custom mnemonics as well, though they won't be in the preheat menu list, at least for now.
> 
> * The same goes if you have say glow in the dark PLA. You'd simply create something like "GLPLA"

I get it. So "custom_x" just used as example. makes sense.

CZDanol:
@bryn51 the CUSTOM_X was used as an example for this PR - but this PR was not merged, we implemented the things differently. So the demo image you see on the top has nothing to do with how things work on the printer.

bkerler:
Yeah, custom_1 .. custom_4 was just at the beginning of the PR, but it was modified to use any filament name instead afterwards. Prusa has even added more functionality and so far I'm pretty happy about it (it looks pretty familiar to what I did) :D

KlausKraemer:
@bkerler Thank you for all the energy, you invested to create a solution!

Issue #3656, opened 2024-01-02
0.6 Nozzel is set to the Printer - no warning if you send a file with a 0.4 Profile!


beroliv:
**Printer type** - [ MK4]

**Printer firmware version** - [5.1.2.]

**Original or Custom firmware** - [Original]

**USB drive or USB/Octoprint**
 Pusa Link or Prusa Connect

**Describe the bug**
 0.6 Nozzel is set to the Printer - no warning if you send a file with a 0.4 Profile! The opposide way the Printer give a warning the Nozzel diameter are not pass!

**How to reproduce**
  Set the Printer to 0.6 Nozzle send a file with 0.4 Profile

**Expected behavior**
  The Printer should give a warning Message as he do the opposide way!

dpark12300:
Are the settings set to warn you if the nozzle is incorrect? Not sure where the setting is though.

beroliv:
Yes! 
The opposide way the Printer give a warning the Nozzel diameter are not pass!

dpark12300:
Hmm. I don't know what to do.

beroliv:
set your printer to 0.6 Nozzle, try to print a file with 0.4 Profile! Did a warning appear? If yes nothing to do! if not something is wrong.

danopernis:
This should be fixed since v6.0.0 via commit https://github.com/prusa3d/Prusa-Firmware-Buddy/commit/3840f9a4f388e82eef08d26540d78fab8af6460e

@beroliv can you check if the issue is fixed for you?

PR #3663, opened 2024-01-06
common: Fix nozzle diameter check (Float vs. Double issue)

bkerler:
This PR simplifies the code for comparing the nozzle diameter (positive or negative) and also addresses an issue with 
`double get_nozzle_diameter([[maybe_unused]] size_t idx)`, whereas the nozzle_diameter is float being substracted from get_nozzle_diameter being double (not float) [here](https://github.com/prusa3d/Prusa-Firmware-Buddy/blob/4bea923810302d654b932291ba324eaedff072fc/src/gui/screen_tools_mapping.cpp#L302).
Related to issue #3656.

### DIFF
--- a/src/common/marlin_print_preview.cpp
+++ b/src/common/marlin_print_preview.cpp
@@ -166,8 +166,8 @@ auto PrintPreview::check_tools_mapping_validity(const ToolMapper &mapper, const 
             return true;
         }
 
-        float nozzle_diameter_distance = gcode.get_extruder_info(gcode_tool).nozzle_diameter.value() - get_nozzle_diameter(physical_extruder);
-        if (nozzle_diameter_distance > 0.001f || nozzle_diameter_distance < -0.001f) {
+        float nozzle_diameter_distance = std::abs(static_cast<float>(gcode.get_extruder_info(gcode_tool).nozzle_diameter.value()) - static_cast<float>(get_nozzle_diameter(physical_extruder)));
+        if (nozzle_diameter_distance > 0.001f) {
             return false;
         }
 
--- a/src/gui/screen_tools_mapping.cpp
+++ b/src/gui/screen_tools_mapping.cpp
@@ -144,11 +144,11 @@ void disable_radio(RadioButton &radio) {
     radio.Invalidate();
 }
 
-double get_nozzle_diameter([[maybe_unused]] size_t idx) {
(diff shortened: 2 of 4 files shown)


src/common/marlin_print_preview.cpp
```diff
@@ -166,8 +166,8 @@ auto PrintPreview::check_tools_mapping_validity(const ToolMapper &mapper, const
             return true;
         }
 
-        float nozzle_diameter_distance = gcode.get_extruder_info(gcode_tool).nozzle_diameter.value() - get_nozzle_diameter(physical_extruder);
-        if (nozzle_diameter_distance > 0.001f || nozzle_diameter_distance < -0.001f) {
+        float nozzle_diameter_distance = std::abs((float)gcode.get_extruder_info(gcode_tool).nozzle_diameter.value() - (float)get_nozzle_diameter(physical_extruder));
```
vorner, line 169: The overall preference is to use C++-style casts (`static_cast` here) where possible. If you find some C-style casts around, it's mostly just leftovers from ancient times 😇

src/gui/screen_tools_mapping.cpp
```diff
@@ -162,7 +162,7 @@ void print_right_tool_into_buffer(size_t idx, std::array<std::array<char, ToolsM
 
     if (drawing_nozzles) {
         const auto cur_strlen = strlen(text_buffers[idx].data());
-        snprintf(text_buffers[idx].data() + cur_strlen, ToolsMappingBody::max_item_text_width - cur_strlen, " %-4.2f", get_nozzle_diameter(idx));
+        snprintf(text_buffers[idx].data() + cur_strlen, ToolsMappingBody::max_item_text_width - cur_strlen, " %-4.2f", (double)get_nozzle_diameter(idx));
```
vorner, line 165: Similar case

vorner (commented):
Thank you, looks good, there's only some style nitpick.

src/common/marlin_print_preview.cpp
bkerler: Is resolved :D

src/gui/screen_tools_mapping.cpp
bkerler: Is resolved as well :D

vorner (dismissed):

vorner (approved):

PR #3665, opened 2024-01-06
nhttp: Support unquoted Authorization header parameters

the-ress:
### Problem

When authenticating using the Digest scheme, some HTTP clients (e.g. [httpx](https://github.com/encode/httpx)) send the `algorithm` parameter even when it was not explicitly specified by the server in the `WWW-Authenticate` header. The value is being sent without quotes as required by [RFC 7616](https://datatracker.ietf.org/doc/html/rfc7616#section-3.4). This trips up the parsing which only expects quoted parameters and the server return error 400.

Example header:
```
Authorization: Digest username="maker", realm="Printer API", nonce="[redacted]", uri="/api/version", response="[redacted]", algorithm=MD5
```

Additionally, I noticed the scheme part (`Digest`) wasn't consumed separately and got attached to the first param instead. In a header like this, with `nonce` as the first parameter, its name would be understood as `Digest nonce` instead of just `nonce` and the value would get ignored.
```
Authorization: Digest nonce="860bd2ef00017e3a", ...
```

### Changes

1. Unquoted parameters are now supported (both known and unknown).
2. The initial `Digest` scheme at the start is consumed before parsing the parameters. The actual value isn't checked, following the existing pattern. (_We only really read nonce and response, the rest is assumed for performance reasons._)
3. The `separator -> read_par` transition is now fallthrough so it doesn't eat the first char of the value (`"`, first char of an unquoted value, etc.)

### DIFF
--- a/utils/gen-automata/http.py
+++ b/utils/gen-automata/http.py
@@ -287,7 +287,7 @@ def content_encryption_mode_header():
     for t in terminals:
         terminals[t].mark_enter()
         terminals[t].set_name(encryption_modes[t])
-    # Eat spaces before the content type
+    # Eat spaces before the encryption mode
     start = tr.start()
     start.loop('HorizWhitespace', LabelType.Special)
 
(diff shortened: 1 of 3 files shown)


utils/gen-automata/http.py
```diff
@@ -322,23 +322,46 @@ def create_folder_header():
...
 
 def auth_value(name):
+    name_unquoted = f"{name}Unquoted" if name != None else None
+
     auto = Automaton()
     start = auto.start()
-    value = auto.add_state()
+    quote = auto.add_state()
+    value_quoted = auto.add_state()
+    value_quoted = auto.add_state()
```
vorner, line 331: Do we need two value_quoted?

vorner (commented):
Just this little nitpick, but thank you for discovering this.

utils/gen-automata/http.py
the-ress: Oh, that's a typo I missed. I removed it.

vorner (approved):

vorner:
Thank you. I expect this'll get released in the next non-patch release (5.2, not 5.1.3), though I can't guarantee it.

the-ress:
thanks

Issue #3671, opened 2024-01-08
[ENHANCEMENT] Option to set time manually


anghelos:
**Printer type** - MK4

Related: #3523 
There should be an option to set the time manually.
I'm in a school setting where connecting the printers to the WiFi is impossible. Even if I manage to connect them to the internet through Ethernet, the school uses its own time servers and blocks all other requests.

bkerler:
You can already set the time manually. Or do you plan to use a local time server ? If so, please change the topic accordingly, thanks :)

mcavoya:
@bkerler, please elaborate. @anghelos and I want to manually set the time on our MK4 printers. Is that what you are referring to?

Searching for "mk4 set time" in the Prusa knowlege base yeilds no helpful results.

anghelos:
> You can already set the time manually.

How? From what I've found, you can only set the time zone offset, not the actual time.

bkerler:
Yes, I was wrong. Sorry for that. I will look into what's possible and try to write a PR.

bkerler:
PR #3679 doesn't fully solve your request, but at least you can send the gcode command `M1702 I "192.168.0.1"` to set your own ntp server, which could be set up using an raspberry pi for example. 

I had a look and changing the local time is more challenging as expected in the current firmware, as the time is being taken from the RTC, which is set up using an fixed NTP server on the printer startup. Of course the time could be set manually (in a rather inconvienient scroll bar matter), but the problem would be to somehow override the ntp server and the initialization. I will have a look, hopefully I find a way to sort this out.

anghelos:
While your PR looks great, and might somehow solve my problem, it'd be great to also have an option to set the time on printers that will always be offline.

bkerler:
The problem is that there is currently no gui method to enter the time / date in a single line approach. It would be possible using multi lines, but wouldn't be very user friendly.

dpark12300:
I second this. Maybe using the touchscreen hardware in the mk4?

dpark12300:
Or something like the iOS time entry system (roller that can be scrolled with knob)

Issue #3523, opened 2023-11-26
[FEATURE REQUEST] Mini - use a local time server


Gisi0:
**Printer type** - MINI
**Optional upgrades** - Filament Runout Sensor

**Describe the expected functionality**
Normally my mini has no connection to the internet, so it happened, that I have no actual time on it.
Please make it possible to set a local time server.

Thank you, great work on the mini
Gisi

nvtkaszpir:
Right now the address is hardcoded to `195.113.144.238`
https://github.com/prusa3d/Prusa-Firmware-
Buddy/blob/c2d7108782c8b84fd4ff318926461c7ae99a4543/lib/WUI/sntp/sntp_client.c

Thankfully you can use destination NAT rule on the router firewall to redirect destination address `195.113.144.238` (which is `tak.cesnet.cz`)  your preferred IP address which is serving time.

For example for [Mikrotik]( https://help.mikrotik.com/docs/display/ROS/NAT#NAT-DestinationNAT) assuming your NTP is on `10.0.0.3` then it would be something like below:
```shell
/ip firewall nat add chain=dstnat action=dst-nat dst-address=195.113.144.238 to-addresses=10.0.0.3
```

bkerler:
This PR should make it much more easier and adds support for a local time server: https://github.com/prusa3d/Prusa-Firmware-Buddy/pull/3679 :)

PR #3679, opened 2024-01-13
network/time: Add option to set ntp ip address or to set ntp via dhcp

bkerler:
Refers to issues #3523 and partly #3671.

This commit adds two new features in order to set the network time automatically :

1. Default: Use the hardcoded ntp ip (as it was before).
2. Setting the ntp ip address via dhcp server instead, selectable in the GUI via Settings -> Network.
3. Add an option to set the ntp ip address manually. 
   Store this as "prusa_printer_settings.ini" and upload to usb_drive:
```
[network]
ntp4=195.113.144.238
```
or
```
[network]
ntp4=pool.ntp.org
```

Then load via "Settings -> System -> Load Settings" in the GUI.

The PR also adds a label in "Settings -> Network -> NTP Address" to show the current NTP address.

The static ntp address "pool.ntp.org" is being stored in the eeprom. 
This also enables the option in the future for a menu entry in the gui itself in order to set the ntp address manually using scroll wheel (which isn't yet implemented).

ntp


bkerler:
Option 3 (via gcode) will be replaced with configuration file as gcode doesn't feel right for setting up the static ntp ip. Thanks @vorner for the hint.

vorner (commented):
I'm a bit worried about the setting of the NTP to gateway, otherwise there are some nitpicks.

src/gui/MItem_lan.cpp
```diff
@@ -83,3 +85,20 @@ MI_IP4_GWAY::MI_IP4_GWAY()
...
+// MI_NTP_VIA_DHCP
+MI_NTP_VIA_DHCP::MI_NTP_VIA_DHCP()
+    : WI_ICON_SWITCH_OFF_ON_t(bool(config_store().lan_ntp_via_dhcp.get()), _(label), nullptr, is_enabled_t::yes, is_hidden_t::no) {}
+
+void MI_NTP_VIA_DHCP::OnChange(size_t /*old_index*/) {
+    bool enabled = config_store().lan_ntp_via_dhcp.get();
+    config_store().lan_ntp_via_dhcp.set(!enabled);
+    ip4_addr_t ntp_ipv4 { .addr = config_store().lan_ntp_ip4_addr.get() };
+    sntp_client_stop();
+    sntp_client_step(config_store().lan_ntp_via_dhcp.get(), &ntp_ipv4);
```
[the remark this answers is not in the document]
bkerler: Now it's thread-safe, done.

thess:
I think this is a good approach to resolving most of the issues with network time setting. Having a fixed IP for NTP is useful in a number of situations but as a global solution it falls short. I found that enabling DNS and settting the default server to [`pool.ntp.org`](https://www.ntppool.org) gives the best result for the majority of your installations - worldwide. See #3730

bkerler:
@thess I'm going to add your commit into mine, if that's ok ?

bkerler:
Integrated PR #3730 (Pool as default & DNS support)

thess:
Looking this over, I don't think it works the way you are expecting. I see a number of problems with the configuration and code flow. Notably:
* No need to use `LWIP_IPV4` conditional in `sntp_client.c` (If no LWIP then no need for SNTP, DNS or any network config)
* SNTP_MAX_SERVERS is 1 (set by `LWIP_DHCP_MAX_NTP_SERVERS`) -- therefore setting indexes other than 0 do nothing.
* lwip_gethostbyname() can take a string argument with an IP address and return correct info. No need to differentiate.
* Need to store NTP server/pool name or NTP address as a string in config_store. It should only be an address if it is a fixed local/private server. Public servers addresses WILL change - use DNS when possible. (Default: `pool.ntp.org`)
* LWIP_DHCP_GET_NTP_SRV is not set, so using DHCP for NTP server is not functional.

I also object to putting a hard-coded IP address into the firmware. It will bite you someday.

I can re-work my PR to encompass your proposed functionality and re-submit it if you would like - your choice.

bkerler:
- I'm using the LWIP_IPV4 define intentionally as currently, IPV6 support is prepared.
- Thanks for pointing out the SNTP_MAX_SERVERS, will fix that
- Ok, will then use lwip_gethostbyname for ip as well
- Storing the ntp server/pool name is a problem, as the space is limited ... need to discuss that with the devs
- LWIP_DHCP_GET_NTP_SRV .... seems I missed that one, thanks

I will add the fixes and test them on the hardware and simulator :)

bkerler:
The commit does now reflect the needed changes.

thess:
> * Storing the ntp server/pool name is a problem, as the space is limited ... need to discuss that with the devs

Network configuration can get messy. You can support 95% of situations with well-thought-out defaults. The last 5% costs time and space - no avoiding it. You could probably get by with 48-64 chars instead of 128!
Hope you have space for IPv6.

> * LWIP_DHCP_GET_NTP_SRV .... this one isn't needed as SNTP_GET_SERVERS_FROM_DHCP is already defined in lwipopts.h as 1

Without LWIP_DHCP_GET_NTP_SRV set for lwIP build, there will be no NTP server (DHCP type 042) found and therefore no callback. Routers in general don't supply this unless specifically configured for something like ActiveDirectory or a private time server.

bkerler:
> Network configuration can get messy. You can support 95% of situations with well-thought-out defaults. The last 5% costs time and space - no avoiding it. You could probably get by with 48-64 chars instead of 128! Hope you have space for IPv6.

Indeed, the char length is something I need to discuss with the Prusa Devs .. including if they will allow it at all. IPv6 is already tested and it works with only minor changes needed. However it does need a clear separation of the web services like PrusaConnect ... but that's on my current toDo list.
 
> > * LWIP_DHCP_GET_NTP_SRV .... this one isn't needed as SNTP_GET_SERVERS_FROM_DHCP is already defined in lwipopts.h as 1
> 
> Without LWIP_DHCP_GET_NTP_SRV set for lwIP build, there will be no NTP server (DHCP type 042) found and therefore no callback. Routers in general don't supply this unless specifically configured for something like ActiveDirectory or a private time server.

It should be defined as 1 using the LWIP_DHCP_GET_NTP_SRV but in my tests it didn't work. I will try to trace back if it is set, so thanks again for your hint. The devs said that it might not be added at all, but I wanted to have it included at least (if it does work), as it was requested to be used in special environments where custom routers with own time servers are needed (like schools).

bkerler:
DNS name length for the eeprom has now been limited to 60 chars as 128 chars was too long.

include/buddy/lwipopts.h
```diff
@@ -9,26 +9,30 @@ extern "C" {
...
+#define WITH_RTOS                    1
+#define MEM_LIBC_MALLOC              0
+#define CHECKSUM_BY_HARDWARE         0
+#define LWIP_DHCP                    1
+#define MEM_ALIGNMENT                4
+#define MEMP_NUM_SYS_TIMEOUT         8
+#define LWIP_ETHERNET                1
+#define LWIP_DNS_SECURE              7
+#define DNS_MAX_NAME_LENGTH          128
+#define DNS_MAX_NAME_LENGTH_EEPROM   60
```
vorner, line 21: Is this the right place for the constant? This thing is mostly for configuring LwIP itself.

lib/WUI/sntp/sntp_client.c
```diff
@@ -1,30 +1,50 @@
...
+void sntp_client_dhcp_init(const char *ntp_address) {
+    sntp_setoperatingmode(SNTP_OPMODE_POLL);
+    sntp_servermode_dhcp(1);
+#if LWIP_IPV4
+    #ifdef SNTP_SERVER_DNS
+    sntp_setservername(1, "pool.ntp.org");
+    #else
+    sntp_setservername(1, ntp_address);
+    #endif
+#endif /* LWIP_IPV4 */
```
vorner, line 24: This looks… weird. Both the LWIP_IPV4 thing, but more so the SNTP_SERVER_DNS thing… so in that case we ignore the provided address and use something hardcoded? And why there's no such thing in the `static init` thing?

lib/WUI/sntp/sntp_opts.h
```diff
@@ -78,7 +78,7 @@
  * \#define SNTP_SERVER_ADDRESS "pool.ntp.org"
  */
 #if !defined SNTP_SERVER_DNS || defined __DOXYGEN__
-    #define SNTP_SERVER_DNS 0
+    #define SNTP_SERVER_DNS 1
```
vorner, line 81: These options probably should go to `include/buddy/lwipopts.h` (see the top of this file, it's included from there and this is just creating a default).

lib/WUI/sntp/sntp_client.h
```diff
@@ -5,8 +5,10 @@
 extern "C" {
 #endif
 
-void sntp_client_init(void);
-void sntp_client_step(void);
+void sntp_client_static_init(const char *ntp_address);
+void sntp_client_dhcp_init(const char *ntp_address);
```
vorner, line 9: These inits probably aren't called from the outside, so maybe they should be static / anonymous namespace and not exported.

lib/WUI/wui_api.cpp
```diff
@@ -202,6 +208,7 @@ void load_net_params(ETH_config_t *ethconfig, ap_entry_t *ap, uint32_t netdev_id
         ethconfig->dns2_ip4.addr = config_store().wifi_ip4_dns2.get();
         ethconfig->lan.msk_ip4.addr = config_store().wifi_ip4_mask.get();
         ethconfig->lan.gw_ip4.addr = config_store().wifi_ip4_gateway.get();
+        strlcpy(ethconfig->ntp, config_store().lan_ntp_addr.get_c_str(), DNS_MAX_NAME_LENGTH_EEPROM + 1);
```
vorner, line 211: This looks weird, we are in the wifi section…

lib/WUI/wui_api.h
```diff
@@ -48,6 +48,7 @@ typedef enum {
     ETHVAR_DNS1_IP4, // ip_addr_t, dns1_ip4
     ETHVAR_DNS2_IP4, // ip_addr_t, dns2_ip4
     ETHVAR_MAC_ADDRESS, // is not included in ethconfig (used in stringifying for screen)
+    ETHVAR_NTP_ADDRESS, // char[256+1], hostname or ip
```
vorner, line 51: 256 seems like a lot.

src/persistent_stores/store_instances/config_store/store_definition.hpp
```diff
@@ -58,12 +58,15 @@ struct CurrentStore : public journal::CurrentStoreConfig<journal::Backend, backe
...
     StoreItem<uint32_t, defaults::uint32_t_zero, journal::hash("LAN IP4 DNS1")> lan_ip4_dns1; // X.X.X.X address encoded
     StoreItem<uint32_t, defaults::uint32_t_zero, journal::hash("LAN IP4 DNS2")> lan_ip4_dns2; // X.X.X.X address encoded
+    StoreItem<std::array<char, DNS_MAX_NAME_LENGTH_EEPROM + 1>, defaults::ntp_server, journal::hash("NTP Address")> lan_ntp_addr; // X.X.X.X address encoded or string
     StoreItem<std::array<char, lan_hostname_max_len + 1>, defaults::net_hostname, journal::hash("LAN Hostname")> lan_hostname;
 
     StoreItem<int8_t, defaults::lan_timezone, journal::hash("LAN Timezone")> timezone; // hour difference from UTC
     StoreItem<time_tools::TimeOffsetMinutes, defaults::timezone_minutes, journal::hash("Timezone Minutes")> timezone_minutes; // minutes offset for hour difference from UTC
     StoreItem<time_tools::TimeOffsetSummerTime, defaults::timezone_summer, journal::hash("Timezone Summertime")> timezone_summer; // Summertime hour offset
 
+    StoreItem<bool, defaults::bool_false, journal::hash("NTP via DHCP")> lan_ntp_via_dhcp; // use dhcp server for ntp
```
vorner, line 68: The lan prefix seems a bit wrong in both of these, as they are global settings.

vorner (commented):
I still feel this must have some weird corner-cases in various situations.

Notice that both interfaces are brought up if possible, the setting in the menu is only which one is the default. And in case one of them is through DHCP, the other is through static IP, will it do DHCP based one, or static one? Will it try to initialize the ntp service twice?

include/buddy/lwipopts.h
bkerler: normally it's DNS_MAX_NAME_LENGTH as already defined in the LWIP stack. I tried to use it in the constants.hpp but that isn't used by the lwip stack on init (and cannot be imported).

lib/WUI/sntp/sntp_opts.h
bkerler: Yes, I will have a look.

lib/WUI/sntp/sntp_client.h
bkerler: yep.

lib/WUI/wui_api.h
bkerler: I need to update the comment as it is now 60+1 in fact.

src/persistent_stores/store_instances/config_store/store_definition.hpp
bkerler: it's being read from the lan area in the configuration ini, not sure where you would store it instead

lib/WUI/sntp/sntp_client.h
bkerler: sntp_client_static_init is used in wui.cpp, thus cannot be static. But I made sntp_client_dhcp_init static.

lib/WUI/sntp/sntp_opts.h
bkerler: Changed that.

lib/WUI/wui_api.cpp
bkerler: This is used in the prusa_settings.ini, so should be fine I think

include/buddy/lwipopts.h
bkerler: another option (maybe better) would be to use DNS_MAX_NAME_LENGTH-68 instead.

bkerler: Used now DNS_MAX_NAME_LENGTH-68.

lib/WUI/sntp/sntp_client.c
bkerler: I removed the LWIP_IPV4 define for now.

lib/WUI/wui_api.cpp
bkerler: I removed the lan_ prefix, as ntp is universal and not limited to lan or wifi

lib/WUI/wui_api.h
bkerler: Updated the comment

koenkooi:
I've been trying to get this to work on a more recent codebase and I managed to get it working on v6.6.1. The problem I ran into was that any DHCP supplied NTP server was getting overwritten after the DNS lookup for `prusa3d.pool.ntp.org` finished. 
The debugging was mostly done by my mecha buddy CoPilot, I'm cleaning up the patch and should be able to post it here Soon(TM).

koenkooi:
[0001-network-time-Add-option-to-set-ntp-ip-address-or-to-.patch](https://github.com/user-attachments/files/29796834/0001-network-time-Add-option-to-set-ntp-ip-address-or-to-.patch)

Attached patch "Works For Me" on MK3.5, CoreOne and Mini+, I squashed all AI changes into @bkerler 's original work. I hope this is useful for someone :)

Issue #3585, opened 2023-12-12
[BUG] Filament Should Purge At Higher-Temp Filament's Temperature When Changing - MK4


adamz5k:
**Printer type** - Prusa Mk4

**Printer firmware version** - 5.1

**Original or Custom firmware** - Original

**Optional upgrades** - ObXidian Nextruder Nozzle / Nextruder V6 Adapter with hardened steel nozzle

**USB drive or USB/Octoprint**
USB

**Describe the bug**
I switch between PETG and PLA probably more frequently than most and have discovered that my nozzle tends to get clogs when switching between PLA and PETG. I've never had this issue on the MK3 or the Mk2 and though the nextruder is different, I'm even having this issue with the V6 adapter. I believe I've narrowed down the issue to the filament change procedure. On the Mk3 & Mk2, the filament unloads at the temperature based on user input and I've always selected PETG as the filament type when changing. 

However, on the MK4, due to the fact that it "remembers" the filament type, it always asks what the newly inserted filament is, and purges at that temperature. What this means is that, when switching from PETG to PLA, the remaining PETG in the nozzle is being purged at PLA temperatures; most is purged, but occasionally not all, and what's left tends to harden in the nozzle. This can cause occasional build-ups of PETG and partial clogs. I haven't tried this with higher-temp filaments, but expect the issue would either be similar or worse.

I tested this by switching between black PETG and white PLA until I noticed curling on the filament, indicating a partial clog. Once the partial clog was identified, I executed a cold pull and found that the partial clog appeared to have been caused by black filament still left in the nozzle.

**How to reproduce**
Switch between PETG and PLA several times. 
To test: Switch from black PETG to white PLA, purge, then do a cold pull to see the remaining PETG.

**Proposed Solution**
I propose that the MK4 emulates the behavior of my Mk2 & 3 by identifying the hotter of the two filaments, and then purging at that temperature. At the moment, this is possible via workaround from either selecting the incorrect filament and changing it later or manually adjusting the temperatuer, but building it in to the firmware would be significantly more conventient and make things less likely to fail for other users.

kozross:
I [requested this previously](https://github.com/prusa3d/Prusa-Firmware-Buddy/issues/3083), but got bounced, because apparently Prusa don't consider it a problem. I wonder if, like with Octoprint support, this will suddenly get attention because it's an MK4 issue as well now...

Tupson444:
[...]

Issue #3083, opened 2023-05-26
[BUG] Swapping filaments can trigger thermal runaways


kozross:
**Printer type** - MINI

**Printer firmware version** - 4.4.1

**Original or Custom firmware** - Original

**Optional upgrades** - 

**USB drive or USB/Octoprint**

USB drive.

**Describe the bug**

This issue can occur when changing from a higher-temperature filament to a lower-temperature filament (for example, going from ABS being loaded to PLA being loaded). In order to avoid clogging, the process to follow is like so:

1. Unload ABS.
2. Load PLA filament using the ABS setting so that the ABS is purged from the melt zone.
3. Unload PLA.
4. Load PLA filament using the PLA setting.

Failing to do this can cause partial clogs (and has for me in the past), so to be safe, I always do this when going from higher-to-lower-temperature filaments. At the same time, when going from step 2 to 3, or step 3 to 4, I can occasionally trigger a thermal runaway, due to the enforced cooldown to 170 after a load. This can sometimes happen mid-load, causing a forced reboot of the printer, and in bad cases requiring an extruder disassembly to clear out.

**How to reproduce**

1. Have a higher-temperature filament loaded, like ABS.
2. Follow the procedure described above to try and load PLA.
3. Note the occasional annoying thermal runaway false positives.

**Expected behavior**

I can see good solutions as any of the following:

* Allow loading of a filament at a higher temperature than its minimum.
* Don't force a drop to 170 on the nozzle post-load.
* Allow a special 'to colder' change setting that automates the process above.
* A more carefully-tuned thermal model to avoid this happening.

I consider these to be in rough order of complexity and adjacent inconvenience.

**G-code**

None relevant.

**Crash dump file**

None present.

**Video**

None exists.

Prusa-Support:
Thank you for your suggestion.

When a filament is unloaded, it is slightly purged and then withdrawn at the material temperature. This is usually enough but a little residue from the previous material may actually remain.

We already have an open issue about this.
I hope you don't mind me closing this issue as a duplicate of https://github.com/prusa3d/Prusa-Firmware-Buddy/issues/1821.

Michele Moramarco
Prusa Research

Issue #1821, opened 2022-01-13
[BFW-7089] [ENHANCEMENT] Filament change temps


DanidCH:
I’m wondering if for a filament change it wouldn’t be better to use the hottest of two temperatures (between last filament and new filament) during the purge. This would probably make the change easier.

Here is a specific case to illustrate how this would help: let’s say I used ASA for my print where hotend is set to 260c and I next want to use PLA, while purging the hot end will be set to 215c (recommendation for PLA) when there is probably some ASA left.

Probably setting the hotend temp to the higher temperature will help the purging process.

Enter what type of printer or upgrade the enhancement applies to.
**Printer type** - [MINI]

zoltan-l:
Hi @DanidCH ,
thanks for your input, we will look at it

TerTerro:
Agree on this. I always do, load PLA after ASA, but specify I'm loading ASA, purging 3-5 times and then unload and load again same PLA with PLA temps. Would be great option to have, switching to purge at old filament temps.

JdogAwesome:
Couldn't agree more on this, would love to see this as a feature. Often when I'm printing PC and go to print PLA I will try to do it as quickly as possible to keep the nozzle temp high to get all the PC out of the hotend.

SnickCH:
+1

kozross:
It's been over 18 months since this was reported, and I just had a more current issue closed in favour of this one, along with two others in the last four months. Is there any ETA on this being dealt with? Given that we can't manually load filament into the MINI+ without going through this system, and I personally have clogged my printer a few times this way already, it'd be good to have a roadmap, or even an explicit statement that we shouldn't wait for a solution.

Prusa-Support:
I'm sorry, no estimated time available and I'm not sure this is in the current development plans.
The issue will stay open.

Michele Moramarco
Prusa Research

Tupson444:
One of the basic advices for 3D printing is e.g. "when changing between PLA and ABS, set the temperature for ABS". This is an obvious overlook that should be fixed.

mjablon1:
@Prusa-Support can you try again putting this into the development roadmap, please?

bkerler:
I wrote a PR #3681 for it. Only tested so far in the simulator, but it should work fine :)

Prusa-Support:
I'll try my best to escalate. 🙂
[...]

PR #3681, opened 2024-01-13
gcode: Fix issue on unloaded filament temperature being higher temperature than loaded filament (lower temperature)

bkerler:
This PR adds a configuration store to remember the past filament and does compare if the temperature is lower than the filament that has been unloaded before, and if, does set the last filament temperature only for loading until purging has finished, thus preventing the nozzle to clog.

Fix is related to issue #3585, #3083 and #1821

### DIFF
--- a/src/marlin_stubs/pause/M701_2.cpp
+++ b/src/marlin_stubs/pause/M701_2.cpp
@@ -47,6 +47,11 @@ static bool load_unload(Pause::LoadType load_type, pause::Settings &rSettings) {
     float disp_temp = marlin_vars().active_hotend().display_nozzle;
     float targ_temp = Temperature::degTargetHotend(rSettings.GetExtruder());
 
+    float previous_filament_temp = config_store().get_previous_filament_temp(rSettings.GetExtruder());
+    if (previous_filament_temp > disp_temp) {
+        targ_temp = (float)((double)targ_temp + (((double)previous_filament_temp - (double)targ_temp) * 0.6));
+    }
+
     if (disp_temp > targ_temp) {
         thermalManager.setTargetHotend(disp_temp, rSettings.GetExtruder());
     }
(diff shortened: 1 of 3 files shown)


nazar-pc:
Not sure where to comment, but these tags on GitHub are extremely annoying for those who subscribe to releases via Atom feed. Please consider an alternative approach there. I'm talking about the likes of https://github.com/prusa3d/Prusa-Firmware-Buddy/releases/tag/build%2F6.0.0-BETA%2B7062.PR3681.B4

danopernis:
@nazar-pc thanks for bringing that up. The CI settings that got merged to the master branch after 6.0.0 release seems to contain some changes causing builds to fail and CI system to generate those tags. I will try to fix that tommorow.

Meanwhile I manually deleted those tags.

FYI @bkerler

spooknik:
I appreciate the work done on this PR and I just wanted to chime in with my input. 

I think it would be better to apply a purge temp offset derived from the difference in filament load temperatures. So if loading PLA (215) after PETG (240) the purge temp would be 230. I explained my line of thinking [here](https://github.com/prusa3d/Prusa-Firmware-Buddy/issues/1821#issuecomment-2549364522).

But basically the formula is: 

```
Previous Filament Temp - Current Filament Temp = Temp Delta 

Temp Delta * 0.6 = Purge Temp Offset

Current Filament Temp + Purge Temp Offset = Filament Purge Temp
```

spooknik:
> @spooknik I don't think a delta of 0.6 will be good as for some filament types this will clearly be too low. Guess some testing will be needed to find a reasonable factor. But I will consider your recommendation :)

Yea I could see a problem going from PC to PLA for example. I guess you can apply an offset multiplier based on the delta. Bigger delta = bigger multiplier.

matoushybl:
This seems to be fixed in unreleased 6.5.4: https://github.com/prusa3d/Prusa-Firmware-Buddy/commit/3e0eaf03bef9e67ec8a9e95e710445fadbf31412

Issue #3697, opened 2024-01-20
[ENHANCEMENT] Add 1.2mm per-tool nozzle size selection (XL)


tg73:
Enter what type of printer or upgrade the enhancement applies to.
**Printer type** - [XL]

**Describe the expected functionality**
Under Settings->Tools->Tool->Nozzle Diameter, only sizes up to 1.0mm can be selected. 1.2mm nozzles are available. Please add 1.2mm to the list.

bkerler:
Just wrote PR #3699 for your request.

tg73:
Actually, I now own a Slice Engineering GammaMaster 2.4mm nozzle (yes, two point four millimetres). Please support sizes up to 2.4mm.

danopernis:
New version of the firmware 6.1.2 should contain the fix, can you test that?

tg73:
I can't find where you set the nozzle diameter in 6.1.2. The setting seems to have disappeared. I'll try to get through to live chat and ask them.

tg73:
@danopernis testing is pending #4152. But from code inspection, the range appears to be limited to max 1.2mm, so for example the 2.4mm nozzle noted above cannot be properly set up. Please extend the range accordingly.

bkerler:
@tg73 I could add 2.4mm as well as a PR, but I don't see that this will make it in the release, as it is no "common" nozzle size and thus won't be officially supported and considered to be a "mod".

bkerler:
I assume that this issue can be closed by now ?

Issue #3468, opened 2023-11-12
[ENHANCEMENT] Add support for 0.2mm and 1.2mm nozzle sizes


nazar-pc:
Please, before you create a new feature request, please make sure you searched in open and closed issues and couldn't find anything that matches.

Enter what type of printer or upgrade the enhancement applies to.
**Printer type** - XL
**Optional upgrades** - None

**Is your suggested improvement related to an existing problem? Please describe.**
Currently nozzle size is limited to 0.25-1.0mm range, while I have 0.2mm and 1.2mm nozzles that I'd like to try. I was able to print successfully with 0.2mm nozzle while configuring 0.25mm on the printer, but it complains that gcode doesn't match installed nozzle size.

**Describe the expected functionality**
I'd like to see nozzle size range increased in future firmware versions, probably to 0.2-1.2mm range.

nazar-pc:
Those are independent and I have already opened https://github.com/prusa3d/PrusaSlicer-settings/issues/220

Even if I create a slicer profile with a smaller/bigger nozzle size, printer will still complain about mismatch because I can't select a proper size in the firmware, which this issue is requesting to change.

MichaelJLew:
Oh, I see. Sorry.

danopernis:
This is scheduled for next minor (not patch) version of the firmware.

danopernis:
New version of the firmware 6.1.2 should contain the fix, can you test that?

nazar-pc:
I have no idea where the option is anymore. It was under Settings->Tools->Tool 1 in 6.0.4 firmware, but now there is no nozzle size option there and I don't see it in any other menus :man_shrugging:

bkerler:
@nazar-pc Should be under Settings->Hardware->Nozzle Diameter. In my test, the firmware 6.1.2 seems to work fine, nozzle size up to 1.2 mm is now supported.

nazar-pc:
Probably true on single-tool XL, but I recently upgraded mine to dual and nozzle diameter is in a different menu there (for whatever reason) and it is not there anymore, see https://github.com/prusa3d/Prusa-Firmware-Buddy/issues/4151

I have no reason to believe it wouldn't be there if menu wasn't broken, so I guess the issue can be closed as resolved, but it didn't help me much just yet.

nazar-pc:
Confirm as fixed on dual-head XL in 6.1.3, thanks!

PR #3699, opened 2024-01-21
gui: Allow nozzle size up to 1.2mm and down to 0.2mm

bkerler:
This commit increases the max nozzle size of 1.0mm to 1.2mm and adds min nozzle size of 0.2mm (0.25mm before). Solves issue #3697 and #3468.


vorner:
Internal reference: BFW-5037

danopernis (approved):
We are going to merge this. It is not clear if the heater can keep up with the increased nozzle size, so it may be necessary to reduce flow rate. But it won't hurt to at least provide the option.

Issue #3708, opened 2024-01-25
[BUG] Manual preheat for PC not going over 170C


m-cas:
Please, before you create a new bug report, please make sure you searched in open and closed issues and couldn't find anything that matches.

**Printer type** - [MINI, MK4, XL]
MK4

**Printer firmware version** - [e.g. 4.0.5, ...]
5.1.2

**Original or Custom firmware** - [Original/Custom]
Original - kit

**Optional upgrades** - [e.g. Filament Runout Sensor]
prusa enclosure, silicon socket

**USB drive or USB/Octoprint**
USB flash drive 

**Describe the bug**
Previous print with PC blend finished. I wanted to raise the temp to clean the nozzle, so I selected "preheat -> PC"; the target temp (250) is shown but the actual temp never exceeds 170C; reset didn't work in making it go above 170. I tried with "change filament" and the temp went above 170.

**How to reproduce**

**Expected behavior**
 Achieve the 250C

**G-code**

**Crash dump file**

**Video**

murk-sy:
This is intended behavior, see #3074 
Its main purpose is to preheat for printing (temperatures intended to start bed leveling in a print)

m-cas:
sorry I allow myself to disagree :D  If (and the linked issue has the same problem) the purpose it to raise the temp to 170 (that is the temp for the bed routines) why on hearth it shows the various materials (pets, pc, etc) ? It is confusing and it takes up space on the menu and code for nothing; it should just read "preheat the bed for start of print operations"!!!

murk-sy:
Well I agree in fact, and made a feature request in regards to this, see #1795 
If you've got any ideas for improving it further, I'd appreciate sharing your thoughts there

m-cas:
I think it should go in one of two possible directions:
1. leave the materials list and allow the temps to go up to the material's limit BUT add one entry for "only warm the table for print operations"
2. drop the whole menu and add one entry somewhere that say "warm the table for print operations"

As it stands at the moment it doesn't make any sense as it allows a choice that is limited at 170 for all entries - waste of space and confusing like crazy.

-

DoubleStrike:
Yeah - this is definitely a UI/UX bug.  Users should not have to read reference documentation to understand what a simple option on a screen does.  If preheat only goes to 170, it should clearly state that on the screen - there is literally no point listing 230 or other temperatures if the preheat does not achieve them.

[...]

Issue #1795, opened 2022-01-04
[BFW-5879] [ENHANCEMENT] Improve Preheat menu usability


murk-sy:
Since you generally don't need to preheat for non-loaded materials, I propose the preheat menu layout is changed to this or similar:
- Preheat for printing PLA (170-215/60)
- Heat nozzle for PLA (215)
- Heat bed for PLA (60)
- Cooldown

Preheat is intended for printing preparation since it also heats the bed, though to be very honest I would actually prefer to have a function to fully heat just the nozzle for cleaning/nozzle change/filament change (which shouldn't heat the bed either in my opinion, but that's just my particular process).

Using preheat to clean the nozzle without accidentally touching the bed and burning my hand would be a lot easier with an extra option like that. Currently, the only way to achieve nozzle-only heating is to remember the temperature required (woe is me, I know) and then go manually change it. It's just a bit of a drag when it could literally be just 2 clicks and make maintenance a lot easier.

Displaying 170°c in that menu for preheat would be nice since that's what it *actually* heats to, but I understand there's limited space. 

With the above suggestion,
- Preheat for current material is accessible by just double clicking on preheat
- Cooldown is still accessible by clicking preheat and scrolling down
- An easy way to heat up just the nozzle is added
- Since you don't *really* need to preheat for non-loaded material, other filament options should be removed. Once filament is changed, the printer is preheated for that material automatically.
 
If you use pre-heat for non-loaded material, please post about it, because it's not something I've encountered before.
(Except as workaround for bug #1400 )

Since this firmware will be/is similar for Prusa XL, doing it this way would allow quality of life improvement for preheating those:
- Tool 1 (PLA)
  - Preheat for printing PLA (170-215/60)
  - Heat nozzle for PLA (215)
  - Auto-clean tool? (assuming that is an issue and not in the filament/tool menu)
  - Other materials
    - Preheat PLA
    - Preheat PETG
    - Etc (the standard menu)
  - Cooldown
- Tool 2 (PETG)
  - Preheat for printing PETG (170-230/80)
  - Heat nozzle for PETG (230)
  - Auto-clean tool?
  - Other materials
    - Preheat PLA
    - Preheat PETG
    - Etc (the standard menu)
  - Cooldown
- Bed
  - Bed specific options (quadrant/all heating, will likely be too complicated to efficiently control directly except for troubleshooting)
- Cooldown All

[...]

PR #3709, opened 2024-01-26
gui: Fix wrong temp display in preheat menu, add new preheat menu

bkerler:
Fix missing separation between preheating and filament load/unload leading to wrong ui information, displaying nozzle temperature for the "preheat" menu instead of the nozzle_preheat temperature.

Adds a new preheat menu for preheating heat and bed (displaying nozzle temp of nozzle_preheat correctly which was display wrong before), heating up nozzle to target temp (according to selected filament nozzle temp) and heating up only heatbed to preheat temperature separately.

Avoids confusion and makes maintenance easier (nozzle change, cleaning etc.).

Related to issues #3708, #1795.

image

image


murk-sy:
I still think the modified preheat menu should display the temperatures, and ideally, the loaded material - see original feature request. Maybe having them listed right-aligned like in material selection would make sense. Ideally in a way so that the slashes are aligned.

I'm not sure if what I used would be the best terminology since the space is limited, but not including the temperatures makes it also confusing (but in the other direction compared to before).

vorner:
Internal reference: BFW-5040

CZDanol:
Hey, we've consulted this in the company and decided to extend your proposal with a bit more functionality. So - we will implement something quite similar to this, but we need to develop it internally. Also, the preheat menu has already been refactored, so this PR is not mergeable anymore.

So I'm closing this PR. Thank you for your work and sorry that it hasn't been taken as-is.

bkerler:
@CZDanol No worries, at least it was helpful in terms of "inspiration" :D

Issue #3559, opened 2023-12-06
[ENHANCEMENT] Prusa Link Password edit/remove


sydios:
Hello,

it would be great if you could change the password yourself or disable it completely. 
Not all printers are in an open network and need protection. It's a bit annoying especially because it's done with an .htaccess kind of protection. On mobile, it blocks the whole browser if you don't enter the correct password.

Gompje:
Second this. 

its annoying as hell. especially given it's done with a .htaccess file like you said which eliminates any browser saving or password manager. Please let us decide if we need this or not. 

I don't care if it's on by default, but for the love of God let us remove it.

tkircher:
This is part of why most of us want Octoprint support. Even if Prusa had an army of devs working on their custom interface, there's only so much they can do with the hardware. I'm pulling the built-in wifi module as soon as I can.

CZDanol:
Internal ticket: BFW-5639

CZDanol:
Note: Completely disabling the password is not doable, it would make the printer way more vulnerable to attacks.

Btw, you can apparently use http://user:password@ip/ to connect if you want to skip the obnoxious auth dialog.

CZDanol:
Also note: There's a quite nice integration of PrusaLink with HomeAssistant :)

danopernis:
Also note: user:password@ip doesn't work because of "secure" `window.fetch` https://stackoverflow.com/questions/75539407/why-fetch-doesnt-allow-credentials-in-the-url

See also https://github.com/prusa3d/Prusa-Firmware-Buddy-Private/pull/5067

danopernis:
> Btw, you can apparently use http://user:password@ip/ to connect if you want to skip the obnoxious auth dialog.

This doesn't work because `window.fetch()` is "secure" (see [stack overflow](https://stackoverflow.com/questions/75539407/why-fetch-doesnt-allow-credentials-in-the-url)) but maybe could work if PrusaLink extracted the fields from URL and sent them using headers.

The-Wurst:
> Note: Completely disabling the password is not doable, it would make the printer way more vulnerable to attacks.

I don't mean that it's disabled by default but that every user can choose to do so if they want. I have a secure network and don't need the PW. Or at least make the PW editable so you can choose your own. Forced security is always annoying, especially on a pro machine like this.

CZDanol:
> I don't mean that it's disabled by default but that every user can choose to do so if they want. 

[...]

Issue #3213, opened 2023-08-21
[FEATURE REQUEST] PrusaLink embedded - Allow setting password for "maker" user


GRIFFCOMMca:
- Mk4 (and presumably the Mini and XL)
- PrusaLink (embedded)

- PrusaLink has a default user of "maker", the password is preset, unable to set to a password you may require.

- Need to be able to change the password for the user "maker" to anything the owner decides
- Side note, would be nice to be able to create new users (and passwords)

nocakeinc:
I'd like to support this.
I do not need a super secure password, the printer is contained in my LAN, barred from WAN access. When accessing, I don't always want to have to fire up my password tool. It reduces the usability of PrusaLink.

wseifert:
I also support this request

danopernis:
Although this is technically older issue than https://github.com/prusa3d/Prusa-Firmware-Buddy/issues/3559 let me close this one as a duplicate in order to not clutter the list of open issue :pray:

Issue #3161, opened 2023-07-24
[ENHANCEMENT] Redesign PrusaLink login/password credentials and interface


neuron-whisperer:
**Printer type** - MK4 built from kit

**Optional upgrades** - None

**Is your suggested improvement related to an existing problem? Please describe.**

Currently, PrusaLink is accessible via username/password. The username is hard-coded as "maker" and password is a randomly generated, fixed-length alphanumeric string that is difficult to remember and type. The UI allows the user to re-roll the password, but does not allow a user to set a preferred password. (Obviously, it is desirable to allow the user to specify both username and password, but that is not really the point of this submission.)

Most modern web applications that require username/password credentials are coded to embed username and password fields in a login web page. Those fields are tagged with metadata that allows password managers (like 1Password) to recognize them, to record credentials when they are first entered, and to inject them for subsequent login. Password managers vastly improve both user convenience (due to not having to remember and type passwords for websites) and security (due to promoting unique passwords for each site, in lieu of reusing passwords across multiple sites).

**Problem:** The MK4 web interface does not request credentials this way. Rather, the MK4 web interface requests these credentials via Basic HTTP Authentication. Most browsers (including Google Chrome and Apple Safari for iOS) handle Basic HTTP Authentication by popping up a separate window or tab with Username and Password fields, and a "Submit" button that transmits entered credentials to the web interface. Unfortunately, Basic HTTP Auth is not compatible with password managers like 1Password - it can neither record entered credentials nor inject them for subsequent login. Because of this design choice, users must manually type in (or, at best, copy-and-paste) both the "maker" username and the randomly-selected alphanumeric password.

This problem is not only a user convenience issue - it is also a security issue: forcing the "maker" username for all user accounts, and forcing all passwords to conform to a fixed-length alphanumeric format, vastly reduce the entropy of user credentials and make intrusion via brute-force password-guessing much more feasible. An attacker who succeeds in brute-force guessing the password could interrupt a print, inject malware that invades the user's LAN, or **even start a fire** by misconfiguring the heating elements of the MK4.

**Describe the expected functionality**

(1) PrusaLink must allow users to choose their own username (instead of forcing everyone to use "maker") and also allow users to specify their own passwords (instead of forcing everyone to use randomly generated, fixed-length alphanumeric passwords).

(2) PrusaLink must encode the username and password fields in a login web page like most modern web applications, instead of using Basic HTTP Authentication, which is not compatible with password managers.

[...]

Issue #222, opened 2020-02-16
Stop button inactive during loading of filament 4.0.3-RC1 


simonbugler:
During the loading of filament the Stop butting is inactive (greyed out).

As the printer is doing some quite fast movements if anything is wrong then being able to Stop is important.

If this is not doable - then the Stop button should be removed from the screen.

As I was testing to make sure I was correct with this issue - I accidentally loaded the filament that was wrapped through the reel, it tried to drag the reel through - I really needed the stop button to be working.

loglow:
In this kind of situation, isn't this precisely what the hardware panic (reset) button is for?

I do strongly agree that if the "stop" button doesn't do anything then it should be removed from the interface.

PR #3731, opened 2024-02-08
[BFW-5934] PrusaLink: Differentiate user authentication password from generated api-key (REST API)

thess:
Following some discussions here (#3559, #3213, #3161 and others #222) and in the [forum](https://forum.prusa3d.com/forum/english-forum-original-prusa-i3-mk4-hardware-firmware-and-software-help/mk4-ui-issue-prusalink-authentication-issues/#post-663752), I propose the attached PR to make a new WebUI authentication password separate from the randomly generated X-API-KEY for the PrusaLink REST API. Support for this change includes:
* Add the ability to set the user-password and api-key via a new [prusalink] section in the `prusa_printer_settings.ini` file.
* REST API uses legacy password,  renamed to api-key.
* WebUI authentication uses new User Password.
* Display user-password on PrusaLink menu.
* Change `Generate Password` to `Update Keys` and re-read settings if available.
* Fallback to legacy behavior (both passwords are the same random value).
* Persist new password in EEPROM
* Add ability to edit passwords in V6.1 firmware using DialogTextInput screen.

Note: PrusaSlicer (tested) works with either key by user selection.

### DIFF
--- a/doc/prusa_printer_settings.ini
+++ b/doc/prusa_printer_settings.ini
@@ -52,3 +52,8 @@ hostname = buddy-a.connect.prusa3d.com
 token = 1234567890
 port = 443
 tls = true
+
+# Password length is limited to 15 chars
+[prusalink]
+apikey=Key1337sp34k
+psw=UserSecretCode
--- a/lib/WUI/http_lifetime.cpp
+++ b/lib/WUI/http_lifetime.cpp
@@ -45,7 +45,8 @@ private:
 
 public:
     virtual const Selector *const *selectors() const override { return selectors_array; }
-    virtual const char *get_password() const override { return wui_get_password(); }
+    virtual const char *get_apikey() const override { return wui_get_apikey(); }
+    virtual const char *get_user_password() const override { return wui_get_user_password(); }
     virtual altcp_pcb *listener_alloc() const override {
         /*
          * We know we are in the part where ALTCP is turned off. That means the
--- a/lib/WUI/nhttp/req_parser.cpp
+++ b/lib/WUI/nhttp/req_parser.cpp
@@ -223,7 +223,7 @@ Step RequestParser::step(string_view input, bool terminated_by_client, uint8_t *
         return Step { 0, 0, Continue() };
     }
 
-    api_key = server->get_password();
+    api_key = server->get_apikey();
     if (api_key && api_key[0] == '\0') {
         // An empty password means "login disabled".
         // (can be a result of generator failure)
@@ -385,7 +385,7 @@ namespace {
 bool RequestParser::check_digest_auth(uint64_t nonce_to_use) const {
     if (auto digest_params = get_if<DigestAuthParams>(&auth_status)) {
 
-        const char *pass = server->get_password();
+        const char *pass = server->get_user_password();
         if (pass == nullptr || pass[0] == '\0') {
             // Login disabled
             return false;
--- a/lib/WUI/nhttp/server.h
+++ b/lib/WUI/nhttp/server.h
@@ -70,7 +70,8 @@ public:
      *   deleting the old one is a data race/UB. We need to find a solution,
      *   but for now, changing the key is very rare and happens in-place.
      */
-    virtual const char *get_password() const = 0;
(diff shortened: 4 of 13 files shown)


thess:
Rebased to master (v6.0.0) in case someone wants to try this and set their own PrusaLink passwords.

thess:
Rebased again to master (v6.1.0).
Looking at possible use of new `DialogTextInput` for setting password (and maybe API-key).

thess:
Added ability to edit passwords on-screen (New in v6.1.0)

danopernis (commented):
Thanks for the contribution @thess

I have created an internal ticket and @vorner volunteered to review this. There are security concerns and I am not a security expert, but hopefully this will get approved.

thess:
These enhancements do not change the 'security' of the product but they do greatly enhance its usability :wink:

The use of Basic Authentication vs Digest mode in the PrusaLink WebUI or how a client would obtain a token using the API-Key could be a different discussion - but this is not that.

vorner:
Hello

First, let me get into some background of the thing. Originally, the only authentication mechanism was the api key header. This was later deemed suboptimal and we moved to digest ‒ which has the advantage of not sending the password in plain text (even though it is local network). During the move, we decided to keep accepting the old header, as a transition measure (not to immediately break any scripts or integrations users could have). There isn't ‒ and never was ‒ any intention of differentiating the web UI access vs some kind of API access. They are _the same_, the web UI is just one possible „default“ API client from all practical points of view. More importantly, the support of the API key header was intended to be eventually removed ‒ and until now, there never was a good trigger for it.

In that sense, we really would _not_ want to waste additional precious space of EEPROM to store two passwords based on the mechanism. Instead, this might be a good time to _drop_ the API key header support.

As for using an INI file or the on-screen input to set ones own password (as opposed to using auto-generated one) ‒ this actually _does_ have security consequences. The generated password is unique, used only on that printer. The fact we store it in plain text in EEPROM or shown on the screen of the printer poses no additional risk ‒ to extract it from either, the attacker would need to have access to the printer already. But users are likely to use a password that might be used on some other account or device already. Storing such password in plain text or showing it on screen is _not_ OK, because if the attacker gains access to the printer, they _also_ get access to password of additional things.

So while we _wanted_ to make the password configurable (like a lot other things are now configurable, for example the hostname), we decided to do it properly. This means:
* At the initialization (or with the regenerate button), we generate the password just as we do now. We store it in plain text, show it to the user on the screen, etc.
* We let the user set the password, but we store it _hashed_ (the Digest-MD5 algorithm allows for storing the password hashed), which adds some slight protection. We also remember a flag that it is hashed and do _not_ show it in the menu.
* At that point, we no longer support the API key header (it wasn't decided if we keep it in the first scenario, where the password is plain text).

Also, skimming the code, you seem to be loading the ini file quite aggressively on many circumstances. And even in case it is not deemed valid (eg. the loading routine eventually returns a failure), it might or might not store one or the other. Both of these things could be potentially confusing, especially in context where users might share USB drives with models and accidentally leaving their ini file there ‒ and the other user accidentally loading it through some other action. Furthermore, if I read it correctly, this does _not_ preserve the original API key on firmware upgrade and generates a new one…

Overall, making the password configurable is desired (though often falls through priorities in our team) and we will be happy to have an external PR in that regard, but we would prefer for it to adhere to the above plan.

thess:
@vorner - Thanks for the insightful reply. I understand your position on this subject. My desire is to be able to set a password I can remember and easily enter without cut-n-paste (which is not possible). I would also like to have all my printers have the same password and have client software to interface the printers via the REST API. The PrusaLink Web UI on the MK4 is only marginally useful as it now stands.

>Furthermore, if I read it correctly, this does not preserve the original API key on firmware upgrade and generates a new one…

It does not and, it preserves the previously randomly generated one and duplicates it to both keys.

Let me see if I can work up a solution that meets both of our needs - including the setting/storing of the username and remove the use of an .INI file.

vorner:
It's alright to read from the ini file, I'd just want to do it only in the „Load …“ menu item explicitly, not on all the other occasions.

Issue #2956, opened 2023-01-25
[ENHANCEMENT] Wifi module fw info


krupkaj:
Please, could you add info about WIFI module firmware version into to screen Info -> System Info?

Prusa-Support:
Thanks for your suggestion.

Our developers will look into this.

Michele Moramarco
Prusa Research

bkerler:
I just wrote a PR for this, enjoy :)

CZDanol:
Internal ticket BFW-5075

CZDanol:
Hello, after consultation within the company, we've decided what we will not implement this.
* The ESP FW number is basically an internal number that doesn't have much significance to the user.
* Each printer FW has its own required ESP firmware and it will not work properly with any other.
* In the upcoming release, there will be a different UI element added showing the status of the Wi-Fi module, including the "Wrong version" (or something along those lines) status.

PR #3736, opened 2024-02-09
gui: Add wifi firmware version information to Info/Version Info

bkerler:
Adds a gui element that displays the current wifi module firmware version as requested in issue #2956.

wififirmware


src/gui/screen_menu_lan_settings.cpp
```diff
@@ -23,7 +23,14 @@ ScreenMenuConnectionBase::ScreenMenuConnectionBase(uint32_t dev_id, const char *
 
 void ScreenMenuConnectionBase::refresh_addresses() {
     char str[ADDR_LEN];
-
+    char fwver[4];
```
CZDanol, line 26: `fwver` should be defined inside the if then branch.

src/gui/screen_menu_lan_settings.cpp
```diff
@@ -23,7 +23,14 @@ ScreenMenuConnectionBase::ScreenMenuConnectionBase(uint32_t dev_id, const char *
...
+    const uint8_t fw_version = espif_firmware_version();
```
CZDanol, line 27: You can put this inside the if statement: `if(const auto fw_version = espif_firmware_version())`

src/gui/screen_menu_lan_settings.cpp
```diff
@@ -23,7 +23,14 @@ ScreenMenuConnectionBase::ScreenMenuConnectionBase(uint32_t dev_id, const char *
...
+    if (fw_version > 0) {
+        snprintf(fwver, sizeof(fwver), "%d", espif_firmware_version());
```
CZDanol, line 29: You're calling the function here again even though you have the data stored in a variable

src/gui/screen_menu_lan_settings.cpp
bkerler: Done.

src/gui/screen_menu_lan_settings.cpp
bkerler: Done.

src/gui/screen_menu_lan_settings.cpp
bkerler: Done.

src/gui/screen_menu_lan_settings.cpp
```diff
@@ -24,6 +24,13 @@ ScreenMenuConnectionBase::ScreenMenuConnectionBase(uint32_t dev_id, const char *
 void ScreenMenuConnectionBase::refresh_addresses() {
     char str[ADDR_LEN];
 
+    if (const auto fw_version = espif_firmware_version() > 0) {
```
CZDanol, line 27: This would set the fw_version to bool of `espif_firmware_version() > 0`, so it would always report `1` as the version.
You can do two things:
* Omit the `> 0`. The fw_version type is `uint8_t`, so the cast to bool is equivalent to it anyway.
* Make it `if (const auto fw_version = espif_firmware_version(); fw_version > 0) {`

bkerler: Indeed. Fixed.

CZDanol (approved):

CZDanol:
Internal ticket BFW-5075

CZDanol:
We've decided not to implement this feature. See conversation in https://github.com/prusa3d/Prusa-Firmware-Buddy/issues/2956

Issue #3737, opened 2024-02-10
[BUG] M27 does not auto-report the print status when S parameter is used


jerrygrey:
**Printer type** - MINI

**Printer firmware version** - 5.1.2

**USB drive or USB/Octoprint**
USB flash drive

**Describe the bug**
The `S` parameter to set the interval for auto-reporting is ignored when using `M27`.

**Expected behavior**
The print status should auto-report at the interval set by the S parameter when using `M27`.

**G-code**
`M27 S10`

bkerler:
Auto-report isn't implemented for M27 (Print sd card status) at this moment. Please change topic to feature request (no bug).

bkerler:
I've added the PR #3739 in order to add support for the requested functionality.

PR #3739, opened 2024-02-10
[BFW-7620]gcode: Add gcode M27 auto report support

bkerler:
This addresses issue #3737 with a missing auto report functionality for gcode M27 (SD card print status).


bkerler:
Feature has already been included.

PR #3749, opened 2024-02-13
sensor: Fix missing board temperature display on MK3.5

bkerler:
This fixes the missing init of the board temperature value for the MK3.5.


mody (approved):

src/gui/screen_menu_sensor_info_MK3.5.cpp
```diff
@@ -19,6 +19,12 @@
 
 void ScreenMenuSensorInfo::windowEvent(EventLock /*has private ctor*/, window_t *sender, GUI_event_t event, void *param) {
     if (event == GUI_event_t::LOOP) {
+
+#if HAS_TEMP_BOARD
```
CZDanol, line 23: This file is specific to MK3.5. The ifdef is not necessary here.

bkerler: Has been changed. Thanks.

CZDanol (approved):

CZDanol:
Internal ticket number: BFW-5074

Issue #3747, opened 2024-02-12
[BUG] [MK3.5] Wrong machine type and extruder count reported on M115 (MACHINE_TYPE:Prusa-mini EXTRUDER_COUNT:5)


arekm:
Please, before you create a new bug report, please make sure you searched in open and closed issues and couldn't find anything that matches.

**Printer type** - MK3.5
**Printer firmware version** - 5.2.0
**Original or Custom firmware** - Original

**Describe the bug**

Printer reports on M115:

````
FIRMWARE_NAME:Prusa-Firmware-Buddy 5.2.0+13881 (Github) SOURCE_CODE_URL:https://github.com/prusa3d/Prusa-Firmware-Buddy PROTOCOL_VERSION:1.0 MACHINE_TYPE:Prusa-mini EXTRUDER_COUNT:5 UUID:...
````

**How to reproduce**

Issue ````M115```` command over a serial connection.

**Expected behavior**

MACHINE_TYPE should be probably MK3.5
EXTRUDER_COUNT should be 1

There was a similar bug on MK4 firmware but that one got fixed: https://github.com/prusa3d/Prusa-Firmware-Buddy/issues/3153

CZDanol:
Internal ticket: BFW-5081

CZDanol:
Fix merged :)

PR #3750, opened 2024-02-13
gcode: M115 displays wrong firmware info for MK3.5 / Use printer model for machine type and add XL improvements

bkerler:
This fixes the wrong firmware identification for gcode M115 on MK3.5. Addresses issue #3747.

### DIFF
--- a/src/marlin_stubs/host/M115.cpp
+++ b/src/marlin_stubs/host/M115.cpp
@@ -23,6 +23,7 @@
 #include "../gcode.h"
 #include <option/has_modularbed.h>
 #include <option/has_dwarf.h>
+#include <option/has_mmu2.h>
 
 #if ENABLED(EXTENDED_CAPABILITIES_REPORT)
 static void cap_line(PGM_P const name, bool ena = false) {
@@ -39,18 +40,25 @@ static void cap_line(PGM_P const name, bool ena = false) {
  * M115: Capabilities string
  */
 void GcodeSuite::M115() {
-
     SERIAL_ECHOPGM("FIRMWARE_NAME:Prusa-Firmware-Buddy ");
     SERIAL_ECHOPGM(project_version_full);
     SERIAL_ECHOPGM(" (Github) SOURCE_CODE_URL:https://github.com/prusa3d/Prusa-Firmware-Buddy");
-#if PRINTER_IS_PRUSA_MK4
-    SERIAL_ECHOLNPGM(" PROTOCOL_VERSION:" PROTOCOL_VERSION " MACHINE_TYPE:"
-                     "MK4"
-                     " EXTRUDER_COUNT:" STRINGIFY(1) " UUID:" MACHINE_UUID); // extruder hardcoded due to mmu, should be dynamic in the future
-#else
-    SERIAL_ECHOLNPGM(" PROTOCOL_VERSION:" PROTOCOL_VERSION " MACHINE_TYPE:" MACHINE_NAME " EXTRUDER_COUNT:" STRINGIFY(EXTRUDERS) " UUID:" MACHINE_UUID);
-#endif
+    SERIAL_ECHOPGM(" PROTOCOL_VERSION:" PROTOCOL_VERSION " MACHINE_TYPE:" PRINTER_MODEL);
 
+    uint8_t extruder_count = 0;
+#if HAS_MMU2()
+    // Devices with MMU2/MMU3 only have one physical extruder, whereas EXTRUDERS refers to total amount of virtual extruders
+    // As only physical extruders are reported here, we hardcode the value to 1
+    extruder_count = 1;
+#elif HAS_TOOLCHANGER()
+    extruder_count = PrusaToolChangerUtils::get_num_enabled_tools();
+#else
+    extruder_count = EXTRUDERS;
+#endif
+    SERIAL_ECHOPGM(" EXTRUDER_COUNT:");
+    SERIAL_ECHO(extruder_count);
+
+    SERIAL_ECHOLNPGM(" UUID:" MACHINE_UUID);
 #if ENABLED(EXTENDED_CAPABILITIES_REPORT)
 
     // SERIAL_XON_XOFF
--- a/src/marlin_stubs/host/M115.cpp
+++ b/src/marlin_stubs/host/M115.cpp
@@ -24,6 +24,11 @@
 #include <option/has_modularbed.h>
 #include <option/has_dwarf.h>
 #include <option/has_mmu2.h>
+#include <option/has_toolchanger.h>
+
+#if HAS_TOOLCHANGER()
+    #include "../../module/prusa/toolchanger.h"
+#endif
 
 #if ENABLED(EXTENDED_CAPABILITIES_REPORT)
 static void cap_line(PGM_P const name, bool ena = false) {
@@ -43,7 +48,7 @@ void GcodeSuite::M115() {
     SERIAL_ECHOPGM("FIRMWARE_NAME:Prusa-Firmware-Buddy ");
     SERIAL_ECHOPGM(project_version_full);
     SERIAL_ECHOPGM(" (Github) SOURCE_CODE_URL:https://github.com/prusa3d/Prusa-Firmware-Buddy");
-    SERIAL_ECHOPGM(" PROTOCOL_VERSION:" PROTOCOL_VERSION " MACHINE_TYPE:" PRINTER_MODEL);
+    SERIAL_ECHOPGM(" PROTOCOL_VERSION:" PROTOCOL_VERSION " MACHINE_TYPE:Prusa-" PRINTER_MODEL);
 
     uint8_t extruder_count = 0;
 #if HAS_MMU2()
@@ -51,7 +56,7 @@ void GcodeSuite::M115() {
     // As only physical extruders are reported here, we hardcode the value to 1
     extruder_count = 1;
 #elif HAS_TOOLCHANGER()
-    extruder_count = PrusaToolChangerUtils::get_num_enabled_tools();
+    extruder_count = prusa_toolchanger.get_num_enabled_tools();
 #else
     extruder_count = EXTRUDERS;
 #endif


CZDanol (changes requested):
Since we're digging into this, this would better be done more conceptually. I've talked about this with @bkerler in the PMs.

src/marlin_stubs/host/M115.cpp
```diff
@@ -43,9 +43,13 @@ void GcodeSuite::M115() {
     SERIAL_ECHOPGM("FIRMWARE_NAME:Prusa-Firmware-Buddy ");
     SERIAL_ECHOPGM(project_version_full);
     SERIAL_ECHOPGM(" (Github) SOURCE_CODE_URL:https://github.com/prusa3d/Prusa-Firmware-Buddy");
-#if PRINTER_IS_PRUSA_MK4
+#ifdef PRINTER_IS_PRUSA_MK4 || PRINTER_IS_PRUSA_MK4
```
CZDanol, line 46: Remove this ifdef, print this no matter the printer

src/marlin_stubs/host/M115.cpp
```diff
@@ -43,9 +43,13 @@ void GcodeSuite::M115() {
...
     SERIAL_ECHOLNPGM(" PROTOCOL_VERSION:" PROTOCOL_VERSION " MACHINE_TYPE:"
+    #if PRINTER_IS_PRUSA_MK4
```
CZDanol, line 48: There is a macro for printer name somewhere, use it instead of ifdefs

src/marlin_stubs/host/M115.cpp
```diff
@@ -43,9 +43,13 @@ void GcodeSuite::M115() {
...
                      "MK4"
+    #elif PRINTER_IS_PRUSA_MK3_5
+                     "MK3.5"
+    #endif
                      " EXTRUDER_COUNT:" STRINGIFY(1) " UUID:" MACHINE_UUID); // extruder hardcoded due to mmu, should be dynamic in the future
```
CZDanol, line 53: * For XL, this needs to be read dynamically somewhere from the variables to see how many extruders are actually connected
* Otherwise, check against HAS_MMU2() and if so, report 1
* Otherwise, report `EXTRUDERS`

CZDanol (changes requested):

src/marlin_stubs/host/M115.cpp
bkerler: Solved.

src/marlin_stubs/host/M115.cpp
bkerler: Solved.

src/marlin_stubs/host/M115.cpp
bkerler: Solved.

include/marlin/Configuration_MINI.h
```diff
@@ -135,7 +135,7 @@
 
 // Optional custom name for your RepStrap or other custom machine
 // Displayed in the LCD "Ready" message
-//#define CUSTOM_MACHINE_NAME "3D Printer"
+#define CUSTOM_MACHINE_NAME "Mini"
```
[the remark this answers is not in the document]
bkerler: resolved.

src/marlin_stubs/host/M115.cpp
```diff
@@ -43,9 +43,8 @@ void GcodeSuite::M115() {
     SERIAL_ECHOPGM("FIRMWARE_NAME:Prusa-Firmware-Buddy ");
     SERIAL_ECHOPGM(project_version_full);
     SERIAL_ECHOPGM(" (Github) SOURCE_CODE_URL:https://github.com/prusa3d/Prusa-Firmware-Buddy");
-#if PRINTER_IS_PRUSA_MK4
-    SERIAL_ECHOLNPGM(" PROTOCOL_VERSION:" PROTOCOL_VERSION " MACHINE_TYPE:"
-                     "MK4"
+#ifdef HAS_MMU2()
```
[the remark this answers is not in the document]
bkerler: Solved. As for the XL, I will try to find a solution.

src/marlin_stubs/host/M115.cpp
```diff
@@ -43,14 +43,14 @@ void GcodeSuite::M115() {
...
     SERIAL_ECHOPGM(project_version_full);
     SERIAL_ECHOPGM(" (Github) SOURCE_CODE_URL:https://github.com/prusa3d/Prusa-Firmware-Buddy");
-#if PRINTER_IS_PRUSA_MK4
-    SERIAL_ECHOLNPGM(" PROTOCOL_VERSION:" PROTOCOL_VERSION " MACHINE_TYPE:"
-                     "MK4"
-                     " EXTRUDER_COUNT:" STRINGIFY(1) " UUID:" MACHINE_UUID); // extruder hardcoded due to mmu, should be dynamic in the future
+    SERIAL_ECHOLNPGM(" PROTOCOL_VERSION:" PROTOCOL_VERSION " MACHINE_TYPE:" PRINTER_MODEL
+                     " EXTRUDER_COUNT:"
+#ifdef HAS_MMU2()
+        STRINGIFY(1)
```
CZDanol, line 49: This would deserve an explaining comment

bkerler: I've added comments.

CZDanol: The comment is not descriptive enough, sorry. It should also explain why EXTRUDERS has a different value.

src/marlin_stubs/host/M115.cpp
```diff
@@ -43,14 +43,18 @@ void GcodeSuite::M115() {
...
-                     "MK4"
-                     " EXTRUDER_COUNT:" STRINGIFY(1) " UUID:" MACHINE_UUID); // extruder hardcoded due to mmu, should be dynamic in the future
+    SERIAL_ECHOLNPGM(" PROTOCOL_VERSION:" PROTOCOL_VERSION " MACHINE_TYPE:" PRINTER_MODEL
+                     " EXTRUDER_COUNT:"
+#ifdef HAS_MMU2()
+        // Devices with MMU2/MMU3 only have one physical extruder
+        STRINGIFY(1)
+#elif HAS_TOOLCHANGER()
+        // Retrieve the amount of physical extruders for devices with Toolchanger
+        PrusaToolChangerUtils::get_num_enabled_tools()
```
CZDanol, line 53: This code will not compile, because you're trying to concatenate C string literals with this.

CZDanol (changes requested):

src/marlin_stubs/host/M115.cpp
CZDanol: Also the comment above this is pointless.

bkerler: Ok, I removed the comment for the xl. It did compile, but it wouldn't have printed out the number in human-readable form, used stringify now.

src/marlin_stubs/host/M115.cpp
bkerler: Ok, I've added a more descriptive comment, however that is my opinion as I understand the code and might not reflect how it should be understood.

src/marlin_stubs/host/M115.cpp
```diff
@@ -39,18 +40,26 @@ static void cap_line(PGM_P const name, bool ena = false) {
  * M115: Capabilities string
  */
 void GcodeSuite::M115() {
-
+    uint8_t extruders = 0;
+    char extruder[4] = { 0 };
```
CZDanol, line 44: Wrong variable naming. Some suggestions:
* `extruders` and `extruders_str`
* `extruder_count` and `extruder_count_str`

CZDanol: Also, it doesn't make sense to have them declared in the beginning in the file. Declare them just before you use them.
They don't need to be initialized (at elast wewould have a warning if something is wrong.
Ideally, I'd create a dedicated `{}` scope for them and have these variables only in the scope where you're using them (so the `SERIAL_ECHOPGM(extruder);` would be the last thing in the scope).

src/marlin_stubs/host/M115.cpp
```diff
@@ -39,18 +40,26 @@ static void cap_line(PGM_P const name, bool ena = false) {
...
     SERIAL_ECHOPGM("FIRMWARE_NAME:Prusa-Firmware-Buddy ");
     SERIAL_ECHOPGM(project_version_full);
     SERIAL_ECHOPGM(" (Github) SOURCE_CODE_URL:https://github.com/prusa3d/Prusa-Firmware-Buddy");
-#if PRINTER_IS_PRUSA_MK4
-    SERIAL_ECHOLNPGM(" PROTOCOL_VERSION:" PROTOCOL_VERSION " MACHINE_TYPE:"
-                     "MK4"
-                     " EXTRUDER_COUNT:" STRINGIFY(1) " UUID:" MACHINE_UUID); // extruder hardcoded due to mmu, should be dynamic in the future
+    SERIAL_ECHOPGM(" PROTOCOL_VERSION:" PROTOCOL_VERSION " MACHINE_TYPE:" PRINTER_MODEL
+                   " EXTRUDER_COUNT:");
+#ifdef HAS_MMU2
```
CZDanol, line 50: Should be `HAS_MMU2()`

src/marlin_stubs/host/M115.cpp
```diff
@@ -39,18 +40,26 @@ static void cap_line(PGM_P const name, bool ena = false) {
...
+    extruders = 1;
+#elif HAS_TOOLCHANGER()
+    extruders = PrusaToolChangerUtils::get_num_enabled_tools();
 #else
-    SERIAL_ECHOLNPGM(" PROTOCOL_VERSION:" PROTOCOL_VERSION " MACHINE_TYPE:" MACHINE_NAME " EXTRUDER_COUNT:" STRINGIFY(EXTRUDERS) " UUID:" MACHINE_UUID);
+    extruders = EXTRUDERS;
 #endif
-
+    snprintf(extruder, sizeof(extruder), "%d", extruders);
+    SERIAL_ECHOPGM(extruder);
```
CZDanol, line 61: `PGM` stands for "program memory". `extruder` is not in the program memory, so using this function is not valid.

CZDanol: You also don't need to use the snprintf at all, there is the `SERIAL_ECHO_F` function. To make the code readable, I'd do `SERIAL_ECHO_F(" EXTRUDER_COUNT:%d", extruder_count);`

src/marlin_stubs/host/M115.cpp
```diff
@@ -39,18 +40,26 @@ static void cap_line(PGM_P const name, bool ena = false) {
...
     SERIAL_ECHOPGM(project_version_full);
     SERIAL_ECHOPGM(" (Github) SOURCE_CODE_URL:https://github.com/prusa3d/Prusa-Firmware-Buddy");
-#if PRINTER_IS_PRUSA_MK4
-    SERIAL_ECHOLNPGM(" PROTOCOL_VERSION:" PROTOCOL_VERSION " MACHINE_TYPE:"
-                     "MK4"
-                     " EXTRUDER_COUNT:" STRINGIFY(1) " UUID:" MACHINE_UUID); // extruder hardcoded due to mmu, should be dynamic in the future
+    SERIAL_ECHOPGM(" PROTOCOL_VERSION:" PROTOCOL_VERSION " MACHINE_TYPE:" PRINTER_MODEL
+                   " EXTRUDER_COUNT:");
+#ifdef HAS_MMU2
+    // Devices with MMU2/MMU3 only have one physical extruder, whereas EXTRUDERS refers to total amount of virtual + physical extruders
```
CZDanol, line 51: Not "virtual + physical", only virtual.

src/marlin_stubs/host/M115.cpp
```diff
@@ -39,18 +40,26 @@ static void cap_line(PGM_P const name, bool ena = false) {
...
+    // Example: MK4 (EXTRUDERS: 1) / MK4 with MMU3 (EXTRUDERS: 5)
```
CZDanol, line 52: Invalid example. HAS_MMU2 is always enabled on the MK4 target.

CZDanol (changes requested):

CZDanol:
Internal ticket: BFW-5081

CZDanol (dismissed):

CZDanol (approved):

Issue #3387, opened 2023-10-22
[BUG] Z Stepper Motors won't shut off - remain at 40°C


IBNobody:
**Printer type** - MK4

**Printer firmware version** - 5.0.0

**Original or Custom firmware** - Original

**Optional upgrades** - None

**USB drive or USB/Octoprint** - USB

**Describe the bug**
  Both Z Stepper Motors are not shutting off when idle. They are constantly energized and are warm to the touch (40°C), even hours after a print. The X, Y, and extruder stepper motors are all room temperature.

**How to reproduce**
  Print a model, wait an hour.

**Expected behavior**
  The Z stepper motors should be powered off and sit at room temperature.

**G-code** - N/A

**Crash dump file** - N/A

**Video** - I will post images taken with my thermal camera later.

IBNobody:
I will be monitoring this to try and dial in the exact moment or mechanism that causes the steppers to remain energized. My next test will be a Z axis movement.

faunris:
The same problem. Just power off and on the printer. 
MK4 firmware 5.1.3

bkerler:
As stated by the Prusa devs, this is intentional and needed for the MK3.5 and MK4, so it is no bug. Please close this issue.

Gl-Mk4:
After switching off the MK4 for two weeks, the Z axis still is not down at the bed. So when a print is **not paused** but completed, I see no reason for burning 11 Watts for hours! 
Furthermore, at least the existing menu option "Disable motors" should do so.

danopernis:
This duplicates https://github.com/prusa3d/Prusa-Firmware-Buddy/issues/3339 and will be closed.

PR #3760, opened 2024-02-15
configuration: Disable z being enabled all the time on MK4/MK3.5

bkerler:
The setting DISABLE_INACTIVE_Z being false means that at the end of the print, the Z axis cannot be disabled and stays enabled. However that behaviour is only valid for printers where the bed itself or the extruder on the z axis are so heavy that they could move down just by its weight AND if the slicer would disable the Z extruder via M84 gcode. This isn't the case for regular MK3.5 and MK4. As stated correctly in issue #3387 for the MINI this setting is already true as it should be on MK3.5 and MK4 as well.

### DIFF
--- a/include/marlin/Configuration_MK3.5_adv.h
+++ b/include/marlin/Configuration_MK3.5_adv.h
@@ -543,7 +543,7 @@
 #define DEFAULT_STEPPER_DEACTIVE_TIME 120
 #define DISABLE_INACTIVE_X true
 #define DISABLE_INACTIVE_Y true
-#define DISABLE_INACTIVE_Z false // set to false if the nozzle will fall down on your printed part when print has finished.
+#define DISABLE_INACTIVE_Z true // set to false if the nozzle will fall down on your printed part when print has finished.
 #define DISABLE_INACTIVE_E true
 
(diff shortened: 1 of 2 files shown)


CZDanol:
Nope.

bkerler:
The issue was closed as the MK3.5/MK4 x gantry is heavier than the MINI one and thus enabling Z all the time on the MK3.5/MK4 is required.

arekm:
> Nope. This is intended behavior and will stay this way.

Could you explain more like "why" it needs that? MK3 didn't do that but MK3.5 does but weight of gantry didn't change between these.

CZDanol:
I personally don't know the details behind this decision, sorry.

arekm:
Could you ask teammates who know?

danopernis:
@arekm This is quite an archeological task. I wasn't able to find any written sources, so I had to resort to oral tradition and I asked the elders :smile:

The only answer I was able to get was that it prevents z-axis misalignment, so we do not have to go up all the way each time we start a print. That would also explain why it is only enabled for MK3.5 and MK4 (MINI only has one axis and XL starts with the plate up)

Personally I do not find that answer satisfactory (what happens on  poweroff or reset?) but this is all I got.

PR #3761, opened 2024-02-15
mmu2: Prevent possible buffer overflow in ProtocolLogic::LogRequestMsg

bkerler:
**ProtocolLogic::LogRequestMsg**
The txbuff size in ProtocolLogic::LogRequestMsg could be larger than the modules::protocol::Protocol::MaxRequestSize. In that scenario, if the size of the txbuff is larger than the MaxRequestSize, the later on initialized tmp buffer with size of MaxRequestSize can be overwritten, leading to a buffer overflow.

In order to prevent a possible buffer overflow, the size of txbuff being used shall not be larger than the MaxRequestSize.

**MMU2::rx_str_P**
If the rx_buffer is filled and the content of the rx_buffer isn't null-terminated, the strlen can lead to a possible overflow of the rx_buffer as more bytes are being read into the rxbuffer via mmuSerial.read().

### DIFF
--- a/lib/Marlin/Marlin/src/feature/prusa/MMU2/protocol_logic.cpp
+++ b/lib/Marlin/Marlin/src/feature/prusa/MMU2/protocol_logic.cpp
@@ -759,6 +759,7 @@ void ProtocolLogic::LogRequestMsg(const uint8_t *txbuff, uint8_t size) {
         MMU2::LogRequestMsg(tmp);
     }
     strncpy(lastMsg, tmp, rqs);
+    lastMsg[rqs - 1] = '\0';
 }
 
 void ProtocolLogic::LogError(const char *reason_P) {
--- a/tests/unit/lib/Marlin/MMU2/stubs/mmu2_log.cpp
+++ b/tests/unit/lib/Marlin/MMU2/stubs/mmu2_log.cpp
@@ -1,5 +1,6 @@
 #define PROGMEM /**/
 #include <mmu2_log.h>
+#include "protocol.h"
 #include "stub_interfaces.h"
 
 MarlinLogSim marlinLogSim;
@@ -13,9 +14,11 @@ void LogEchoEvent_P(const char *msg_P) {
 }
 
 void LogRequestMsg(const char *msg) {
+    REQUIRE((strlen(msg) + 1) < (uint8_t)mp::Protocol::MaxRequestSize() + 1);
     marlinLogSim.log.push_back(msg);
 }
 void LogResponseMsg(const char *msg) {
+    REQUIRE((strlen(msg) + 1) < (uint8_t)mp::Protocol::MaxResponseSize() + 1);
     marlinLogSim.log.push_back(msg);
 }
 


lib/Marlin/Marlin/src/feature/prusa/MMU2/protocol_logic.cpp
```diff
@@ -736,7 +736,11 @@ void ProtocolLogic::FormatLastResponseMsgAndClearLRB(char *dst) {
 }
 
 void ProtocolLogic::LogRequestMsg(const uint8_t *txbuff, uint8_t size) {
-    constexpr uint_fast8_t rqs = modules::protocol::Protocol::MaxRequestSize() + 1;
+    constexpr unsigned char mrs = modules::protocol::Protocol::MaxRequestSize();
```
CZDanol, line 739: If you investigate the code, you'll find out that the size of the requests is rather fixed, so it won't exceed the maximum size by design and these checks are not necessary. An assert would be more fitting here.

CZDanol (changes requested):

lib/Marlin/Marlin/src/feature/prusa/MMU2/protocol_logic.cpp
bkerler: thx, added an assert :)

lib/Marlin/Marlin/src/feature/prusa/MMU2/mmu2.cpp
```diff
@@ -345,6 +345,11 @@ bool MMU2::rx_start() {
 bool MMU2::rx_str_P(const char *str) {
     uint8_t i = strlen(rx_buffer);
 
+    // In case rx_buffer isn't null-terminated, prevent an over-read
+    if (i > sizeof(rx_buffer) - 2) {
```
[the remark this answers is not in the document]
bkerler: yes, changed that.

lib/Marlin/Marlin/src/feature/prusa/MMU2/mmu2.cpp
```diff
@@ -346,13 +346,12 @@ bool MMU2::rx_str_P(const char *str) {
     uint8_t i = strlen(rx_buffer);
 
     while (mmuSerial.available()) {
-        rx_buffer[i++] = mmuSerial.read();
-        rx_buffer[i] = '\0';
-
-        if (i == sizeof(rx_buffer) - 1) {
+        if (i == sizeof(rx_buffer) - 2) {
```
CZDanol, line 349: Since we're doing a safety check here, it would make more sense to use `>=`.

lib/Marlin/Marlin/src/feature/prusa/MMU2/mmu2.cpp
```diff
@@ -346,13 +346,12 @@ bool MMU2::rx_str_P(const char *str) {
...
             DEBUG_ECHOLNPGM("rx buffer overrun");
             break;
         }
+        rx_buffer[i++] = mmuSerial.read();
+        rx_buffer[i] = '\0';
```
CZDanol, line 354: The terminating null set can as well be put outside of the loop.

lib/Marlin/Marlin/src/feature/prusa/MMU2/mmu2.cpp
```diff
@@ -363,7 +362,7 @@ bool MMU2::rx_str_P(const char *str) {
 
     str += len;
 
-    while (len--) {
+    while (len-- && i < sizeof(rx_buffer)) {
```
CZDanol, line 365: Have you noticed that the i is being decreased inside the loop?

bkerler: yes. if i is 0 and the loop continues, i is decreased and will have the value 0xFF on the next run (being uint8_t), which is bigger than the size of rx_buffer.

lib/Marlin/Marlin/src/feature/prusa/MMU2/mmu2.cpp
bkerler: indeed.

lib/Marlin/Marlin/src/feature/prusa/MMU2/mmu2.cpp
bkerler: absolutely. good point.

lib/Marlin/Marlin/src/feature/prusa/MMU2/mmu2.cpp
CZDanol: And have you also noticed that `i` always will be larger than `len` because of the theck two lines above?

bkerler: yep, you're right.

lib/Marlin/Marlin/src/feature/prusa/MMU2/mmu2.cpp
```diff
@@ -346,14 +346,13 @@ bool MMU2::rx_str_P(const char *str) {
     uint8_t i = strlen(rx_buffer);
 
     while (mmuSerial.available()) {
-        rx_buffer[i++] = mmuSerial.read();
-        rx_buffer[i] = '\0';
-
-        if (i == sizeof(rx_buffer) - 1) {
+        if (i >= sizeof(rx_buffer) - 2) {
```
DRracer, line 349: this source file is irrelevant - that's the old MMU2 implementation originating from upstream Marlin

lib/Marlin/Marlin/src/feature/prusa/MMU2/protocol_logic.cpp
```diff
@@ -758,7 +760,7 @@ void ProtocolLogic::LogRequestMsg(const uint8_t *txbuff, uint8_t size) {
     } else {
         MMU2::LogRequestMsg(tmp);
     }
-    strncpy(lastMsg, tmp, rqs);
+    strlcpy(lastMsg, tmp, rqs);
```
DRracer, line 763: There is no strlcpy on 8bit FW and this source file must remain 8bit compatible ... but I do agree that strncpy's behaviour is highly "unpredictable" so probably adding an explicit zero terminator into `lastMsg` would do the trick

lib/Marlin/Marlin/src/feature/prusa/MMU2/protocol_logic.cpp
```diff
@@ -737,6 +737,8 @@ void ProtocolLogic::FormatLastResponseMsgAndClearLRB(char *dst) {
 
 void ProtocolLogic::LogRequestMsg(const uint8_t *txbuff, uint8_t size) {
     constexpr uint_fast8_t rqs = modules::protocol::Protocol::MaxRequestSize() + 1;
+    assert((size + 1) < (uint8_t)rqs);
```
DRracer, line 740: Is it really possible for `(size+1 >= rqs)` to happen? If yes and if it cannot be checked at compile time, the correct spot for this check are unit tests in the Prusa-Firmware-MMU repo.

bkerler: Yes, should be added to the unit tests then, or better get rid completely of this function / code.

lib/Marlin/Marlin/src/feature/prusa/MMU2/protocol_logic.cpp
bkerler: ok, will look into that.

lib/Marlin/Marlin/src/feature/prusa/MMU2/protocol_logic.cpp
DRracer: Logging is an important part of debugging, it must not be removed.

lib/Marlin/Marlin/src/feature/prusa/MMU2/mmu2.cpp
bkerler: I've reverted the patch on mmu2 as it's not being used.

lib/Marlin/Marlin/src/feature/prusa/MMU2/protocol_logic.cpp
bkerler: the assert is now added to the unit tests.

tests/unit/lib/Marlin/MMU2/stubs/mmu2_log.cpp
```diff
@@ -13,9 +14,11 @@ void LogEchoEvent_P(const char *msg_P) {
 }
 
 void LogRequestMsg(const char *msg) {
+    assert((strlen(msg) + 1) < (uint8_t)mp::Protocol::MaxRequestSize() + 1);
```
DRracer, line 17: use REQUIRE() in unit tests and the related stubs - it is an "advanced" assert playing nicely with the unit test infrastructure

bkerler: Thanks for pointing that out.

DRracer (approved):
nice :+1:

Issue #3054, opened 2023-04-19
[BUG] Bring back support for long file names and timestamps in M20 L T g-code command


arekm:
Please, before you create a new bug report, please make sure you searched in open and closed issues and couldn't find anything that matches.

**Printer type** - MK4, MK3.5

**Printer firmware version** - 4.5.4, 5.2.0

**Original or Custom firmware** - Original

**Describe the bug**

M20 command issued over serial doesn't list long file names and doesn't list timestamps.
This feature existed for many years in MK3 firmware but is missing from MK4 firmware. 

**How to reproduce**

Issue
````
M20 L T
````
over serial.

````
Send: M20 L T
Recv: Begin file list
Recv: T-NUT_~1.GCO
Recv: End file list
````

**Expected behavior**

Long file names and timestamps (in form of hex numbers) are shown.

````
Send: M20 L T
Recv: Begin file list
Recv: T-NUT_~1.GCO 1061513 0x5690a164 "T-nut_M3_x12_20m.gcode"
Recv: End file list
````

**More info**
Code for that exists in MK3 firmware and also Marlin 2.1.2+  (
guarded by LONG_FILENAME_HOST_SUPPORT
https://github.com/MarlinFirmware/Marlin/blob/2.1.2.2/Marlin/src/sd/cardreader.cpp#L304-L314
https://github.com/MarlinFirmware/Marlin/blob/2.1.2.2/Marlin/src/gcode/sd/M20.cpp#L46

and M20_TIMESTAMP_SUPPORT defs,
https://github.com/MarlinFirmware/Marlin/blob/2.1.2.2/Marlin/src/sd/cardreader.cpp#L327C3-L336C8
https://github.com/MarlinFirmware/Marlin/blob/2.1.2.2/Marlin/src/gcode/sd/M20.cpp#L47

so will be easier to see what's needed for backporting).

Prusa-Support:
Thanks for reporting.

Our developers will look into this.

Michele Moramarco
Prusa Research

bkerler:
I just wrote a PR #3765 for this issue.

danopernis:
This shouldn't have been closed since there are several PRs basically solving it.

timur-tabi:
Just is yet another feature that is missing in the MK4 firmware that was present in MK3.

timur-tabi:
> Thanks for reporting.
> 
> Our developers will look into this.

[...]

PR #3765, opened 2024-02-18
gcode: Update M20 to fit common marlin behaviour

bkerler:
This addresses the issue #3054.

It improves the gcode M20 to:
1. Support listing of subdirectories and files
2. Differentiate between directories and files
3. Adds support for timestamp ("T") and long filenames ("L") to match the marlin firmware behaviour

### DIFF
--- a/src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
+++ b/src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
@@ -4,9 +4,75 @@
 #include <dirent.h>
 
 #include "../../lib/Marlin/Marlin/src/gcode/gcode.h"
+#include "libs/hex_print_routines.h"
 #include "marlin_server.hpp"
 #include "media.hpp"
 #include "marlin_vars.hpp"
+#include "str_utils.hpp"
+
+struct TimeFlags {
+    bool print_lfn : 1;
+    bool print_timestamps : 1;
+};
+
+static void time_to_m20(const struct stat stat_buf, uint16_t *m20_date, uint16_t *m20_time) {
+    time_t t = stat_buf.st_mtim.tv_sec;
+    struct tm lt;
+    localtime_r(&t, &lt);
+    *m20_date = (lt.tm_year + 1900 - 1980) << 9 | (lt.tm_mon + 1) << 5 | lt.tm_mday;
+    *m20_time = lt.tm_hour << 11 | lt.tm_min << 5;
+    *m20_time |= int((lt.tm_sec - (lt.tm_sec % 2)) / 2);
+}
+
+static void print_file_info(const char *const prefix, const char *const path, const char *const long_path, const struct TimeFlags tf) {
+    // we skip the /usb/ prefix on print as it is only used internally
+    SERIAL_ECHO(&path[strlen_constexpr(prefix) + 1]);
+    SERIAL_CHAR(' ');
+    struct stat stat_buf;
+    int rc = stat(path, &stat_buf);
+    if (rc == 0) {
+        SERIAL_ECHO(stat_buf.st_size);
+        if (tf.print_timestamps) {
+            uint16_t m20_date;
+            uint16_t m20_time;
+            time_to_m20(stat_buf, &m20_date, &m20_time);
+            SERIAL_ECHOPGM(" 0x");
+            print_hex_word(m20_date);
+            print_hex_word(m20_time);
+        }
+        if (tf.print_lfn) {
+            SERIAL_CHAR(' ');
+            SERIAL_CHAR('"');
+            SERIAL_ECHO(&long_path[strlen_constexpr(prefix) + 1]);
+            SERIAL_CHAR('"');
+        }
+    }
+    SERIAL_EOL();
+}
+
+static void print_listing(const char *const prefix, const char *const prefix_long, const struct TimeFlags tf) {
+    DIR *dir;
+    dir = opendir(prefix);
+    if (dir != NULL) {
+        struct dirent *entry;
+        while ((entry = readdir(dir)) != NULL && entry->d_name[0]) {
+            char *path = nullptr;
+            char *long_path = nullptr;
+            int path_length = strlen_constexpr(prefix) + 1 + strlen_constexpr(entry->d_name) + 1;
+            int long_path_length = strlen_constexpr(prefix_long) + 1 + strlen_constexpr(entry->lfn) + 1;
+            path = (char *)malloc(path_length);
+            snprintf(path, path_length, "%s/%s", prefix, entry->d_name);
+            long_path = (char *)malloc(long_path_length);
+            snprintf(long_path, long_path_length, "%s/%s", prefix_long, entry->lfn);
+            if (entry->d_type != DT_DIR) {
+                print_file_info(prefix, path, long_path, tf);
+            }
+            free(path);
+            free(long_path);
+        }
+        closedir(dir);
+    }
+}
 
 /** \addtogroup G-Codes
  * @{
@@ -16,16 +82,15 @@
  * M20 - List SD card on serial port
  */
 void GcodeSuite::M20() {
-    SERIAL_ECHOLNPGM(MSG_BEGIN_FILE_LIST);
-    DIR *dir;
-    dir = opendir("/usb/");
-    if (dir != NULL) {
-        struct dirent *entry;
-        while ((entry = readdir(dir)) != NULL && entry->d_name[0]) {
-            SERIAL_ECHOLN(entry->d_name);
-        }
-        closedir(dir);
+    TimeFlags tf;
+    if (parser.seen('L')) {
+        tf.print_lfn = 1;
     }
+    if (parser.seen('T')) {
+        tf.print_timestamps = 1;
+    }
+    SERIAL_ECHOLNPGM(MSG_BEGIN_FILE_LIST);
+    print_listing("/usb", "/usb", tf);
     SERIAL_ECHOLNPGM(MSG_END_FILE_LIST);
 }
 
--- a/src/marlin_stubs/host/M115.cpp
+++ b/src/marlin_stubs/host/M115.cpp
@@ -218,6 +218,9 @@ void GcodeSuite::M115() {
     #endif
     );
 
+    // PARSE FILES ON USB (M20)
+    cap_line(PSTR("EXTENDED_M20"), true);
+
 #endif // EXTENDED_CAPABILITIES_REPORT
 }
 


arekm:
EXTENDED_M20 capability needs to be reported, too

https://reprap.org/wiki/Firmware_Capabilities_Protocol#Cap:EXTENDED_M20

src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
```diff
@@ -5,18 +5,57 @@
 #include "media.h"
 #include "marlin_vars.hpp"
 
-// M20 - List SD card
-void GcodeSuite::M20() {
-    SERIAL_ECHOLNPGM(MSG_BEGIN_FILE_LIST);
+static long GetFileSize(std::string filename) {
```
danopernis, line 8: This doesn't conform to coding style. Also, `std::string` allocates and is unacceptable here. Also, returning -1 or size seems wrong. Especially when we are printing the value.

src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
```diff
@@ -5,18 +5,57 @@
...
+static void printListing(const char *const prepend, const char *const prependLong, bool timestamps, bool longfilenames) {
     DIR *dir;
-    dir = opendir("/usb/");
+    dir = opendir(prepend);
     if (dir != NULL) {
         struct dirent *entry;
         while ((entry = readdir(dir)) != NULL && entry->d_name[0]) {
-            SERIAL_ECHOLN(entry->d_name);
+            std::string dirname(std::string(prepend) + "/" + entry->d_name);
+            std::string longdirname(std::string(prependLong) + "/" + entry->lfn);
```
danopernis, line 21: Again, `std::string` is unacceptable

src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
```diff
@@ -5,18 +5,57 @@
...
+            if (entry->d_type == DT_DIR) {
+                printListing(dirname.c_str(), longdirname.c_str(), timestamps, longfilenames);
```
danopernis, line 23: Recursion may quickly lead to stack overflow here.

src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
```diff
@@ -5,18 +5,57 @@
...
+            } else {
+
+                SERIAL_ECHO(dirname.substr(5, dirname.length()).c_str());
+                SERIAL_CHAR(' ');
+                SERIAL_ECHO(GetFileSize(dirname));
+                if (timestamps) {
+                    char timestampstr[11];
+                    snprintf(timestampstr, sizeof(timestampstr), " 0x%llX", entry->time);
```
danopernis, line 31: Hexadecimal timestampstr seems fishy. Is this standard way of doing this? Is there some spec for M20?

danopernis (changes requested):

src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
bkerler: According to marlin source code [here](https://github.com/MarlinFirmware/Marlin/blob/9e879a5b1f801e7572e7948be38a6dad16ad35d8/Marlin/src/sd/cardreader.cpp#L338-L339) that's how they implement it. However they also separate between date and time, which I haven't done yet.

src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
bkerler: Agreed. I wasn't sure as std:string would be the easiest to cope with concat strings.

src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
arekm: > Hexadecimal timestampstr seems fishy. Is this standard way of doing this? Is there some spec for M20?

That format was invented by Prusa in 8bit firmware (AFAIK as first firmware handling that). Marlin adhered to "standard" introduced by Prusa. OctoPrint also did that and handles this format.

arekm: > However they also separate between date and time, which I haven't done yet.

These aren't just hex representation of just time. Examples of calculations:
https://github.com/OctoPrint/OctoPrint/blob/1.10.0rc2/src/octoprint/util/files.py#L321-L380

bkerler: Yeah, I will have a closer look to implement the hex timestamp in a similar manner.

src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
bkerler: That's how it was handled in the past. I'll come up with a for loop instead.

bkerler: Seems like dir parsing introduces more issues than expected (opendir will hang indefinitely), so I remove parsing subdirectories for this commit.

bkerler:
image

Implementation has been updated.

src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
bkerler: Done, using mainly char now, no more std::string

src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
bkerler: Done

src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
```diff
@@ -1,22 +1,77 @@
...
+static void print_file_size(const char *const filename) {
+    struct stat stat_buf;
+    int rc = stat(filename, &stat_buf);
+    if (rc == 0) {
+        SERIAL_ECHO(stat_buf.st_size);
+    }
+}
+
+static void print_file_info(const char *const dirname, const char *const long_dirname, bool timestamps, bool long_filenames) {
+    SERIAL_ECHO(&dirname[5]);
```
danopernis, line 18: What is `dirname[5]` ? Also seen that magical 5 at other places...

src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
```diff
@@ -1,22 +1,77 @@
...
+    SERIAL_CHAR(' ');
+    print_file_size(dirname);
+    if (timestamps) {
+        struct stat stat_buf;
+        int rc = stat(dirname, &stat_buf);
```
danopernis, line 23: Calling `stat()` twice seems wasteful...

src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
```diff
@@ -1,22 +1,77 @@
...
+        if (rc == 0) {
+            time_t t = stat_buf.st_mtim.tv_sec;
+            struct tm lt;
+            localtime_r(&t, &lt);
+            uint16_t m20_date = (lt.tm_year + 1900 - 1980) << 9 | (lt.tm_mon + 1) << 5 | lt.tm_mday;
+            uint16_t m20_time = lt.tm_hour << 11 | lt.tm_min << 5;
+            m20_time |= int((lt.tm_sec - (lt.tm_sec % 2)) / 2);
```
danopernis, line 30: Isn't this just `(lt.tm_sec >> 1)`? Entire thing would deserve to be factored out and unit tested. Why is there `1900 - 1980` ?

And I am also not sure if localtime_r is right function to call, but I was not looking into that too much. Just noting that timezones are always fun to deal with.

src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
```diff
@@ -1,22 +1,77 @@
...
+static void print_listing(const char *const prepend, const char *const prepend_long, bool timestamps, bool long_filenames) {
     DIR *dir;
-    dir = opendir("/usb/");
+    dir = opendir(prepend);
     if (dir != NULL) {
         struct dirent *entry;
         while ((entry = readdir(dir)) != NULL && entry->d_name[0]) {
-            SERIAL_ECHOLN(entry->d_name);
+            char dirname[5 + 512];
+            char long_dirname[5 + 512];
```
danopernis, line 52: Surely there are some named constants for this.

src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
```diff
@@ -1,22 +1,77 @@
...
+        SERIAL_ECHO(&long_dirname[5]);
+        SERIAL_CHAR('"');
+    }
+    SERIAL_EOL();
+}
+
+static void print_listing(const char *const prepend, const char *const prepend_long, bool timestamps, bool long_filenames) {
     DIR *dir;
-    dir = opendir("/usb/");
+    dir = opendir(prepend);
```
danopernis, line 47: Suggestion: `prepend` -> `prefix`

src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
```diff
@@ -1,22 +1,77 @@
...
+    bool longfilenames = false;
+    bool timestamps = false;
+    if (parser.seen('L')) {
+        longfilenames = true;
+    }
+    if (parser.seen('T')) {
+        timestamps = true;
+    }
+    SERIAL_ECHOLNPGM(MSG_BEGIN_FILE_LIST);
+    print_listing("/usb", "/usb", timestamps, longfilenames);
```
danopernis, line 74: Instead of bools which are interchangable, I would rather see something like

```
struct Flags {
    bool print_lfn : 1;
    bool print_timestamps : 1;
};
```

src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
```diff
@@ -1,22 +1,77 @@
...
-    SERIAL_ECHOLNPGM(MSG_BEGIN_FILE_LIST);
+static void print_file_size(const char *const filename) {
+    struct stat stat_buf;
+    int rc = stat(filename, &stat_buf);
+    if (rc == 0) {
+        SERIAL_ECHO(stat_buf.st_size);
+    }
+}
+
+static void print_file_info(const char *const dirname, const char *const long_dirname, bool timestamps, bool long_filenames) {
```
danopernis, line 17: Just realized that dirname is actually filename.

danopernis (changes requested):

src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
danopernis: @arekm thanks for the insight!

src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
bkerler: Thats the length of "/usb/", the prefix.

src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
bkerler: It's not just filename, it does hold the prefix "/usb/" as well, wo maybe I should name it path instead

src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
bkerler: This is taken from octoprint and tested. 1900 is the basis of timestruct whereas the value expected is value -1980 which would come from the fat entry. Tested to be correct, so I'll leave it this way.

src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
bkerler: Yeah, will have a look.

src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
bkerler: yeah, will change the naming.

src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
bkerler: Will have a look, but at first sight, 512 was used.

src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
bkerler: Done.

src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
bkerler: Done.

src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
bkerler: Done.

src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
bkerler: It's now named path and long_path, which is appropriate.

arekm (changes requested):
Have no way to test this on a printer but tested time_to_m20() implementation on Linux and it correctly translates dates/times that fit into 32bit (tests from OctoPrint test suite, beside 64bit one).

I would move printing LFN inside stat() success path because other implementations parse fat dir directly and always know the size (thus always return it). OctoPrint parser was made with that in mind and won't be able to parse long file name from just
````shortfilename "longfilename"````
output. It needs at least size in 3 element lines (so "shortfilename size timestamp" or "shortfilename size longfilename)

For reference:
https://github.com/OctoPrint/OctoPrint/blob/1.10.0rc2/src/octoprint/util/comm.py#L6451-L6472

**Comment by bkerler** [comment]
> Have no way to test this on a printer but tested time_to_m20() implementation on Linux and it correctly translates dates/times that fit into 32bit (tests from OctoPrint test suite, beside 64bit one).
> 
> I would move printing LFN inside stat() success path because other implementations parse fat dir directly and always know the size (thus always return it). OctoPrint parser was made with that in mind and won't be able to parse long file name from just `shortfilename "longfilename"` output. It needs at least size in 3 element lines (so "shortfilename size timestamp" or "shortfilename size longfilename)
> 
> For reference: https://github.com/OctoPrint/OctoPrint/blob/1.10.0rc2/src/octoprint/util/comm.py#L6451-L6472

Agree. I've moved it into the stat() success path.

**Review by arekm — APPROVED** [review]

File: src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp [hunk 12]
```diff
@@ -1,22 +1,80 @@
...
+    struct tm lt;
+    localtime_r(&t, &lt);
+    *m20_date = (lt.tm_year + 1900 - 1980) << 9 | (lt.tm_mon + 1) << 5 | lt.tm_mday;
+    *m20_time = lt.tm_hour << 11 | lt.tm_min << 5;
+    *m20_time |= int((lt.tm_sec - (lt.tm_sec % 2)) / 2);
+}
+
+static void print_file_info(const char *const path, const char *const long_path, const struct TimeFlags tf) {
+    // &path[5] and &long_path[5], because we skip the /usb/ prefix on print as it is only used internally
+    SERIAL_ECHO(&path[5]);
```
danopernis, line 25: There's a `strlen_constexpr()` in `str_utils.hpp`, we should be able to use that. Not that I am expecting the internal path to change, it just looks more readable/maintainable to me.

src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
```diff
@@ -1,22 +1,80 @@
...
     DIR *dir;
-    dir = opendir("/usb/");
+    dir = opendir(prefix);
     if (dir != NULL) {
         struct dirent *entry;
         while ((entry = readdir(dir)) != NULL && entry->d_name[0]) {
-            SERIAL_ECHOLN(entry->d_name);
+            char path[5 + sizeof(entry->d_name) * 2];
+            snprintf(path, sizeof(path), "%s/%s", prefix, entry->d_name);
+            char long_path[5 + sizeof(entry->lfn) * 2];
```
danopernis, line 57: I still don't understand these buffer sizes. Why `*2` ? We are only supporting root directory, right? I would imagine we need "/usb/" + respective filename (sfn or lfn) + possibly a null terminator. What am I missing?

danopernis (changes requested):
Starting to shape up, good job :wink:

src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
bkerler: Indeed, good idea.

src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
bkerler: The idea was that sizeof(entry->d_name) has the same size as prefix. But yeah, I've changed that completely because it wasn't clean and safe. Should be fine now.

bkerler:
Closing in favor of #3976, which has additional support

PR #3769, opened 2024-02-18
gui: "Settings->System->Load settings from File": Provide Feedback

bkerler:
This PR will add user feedback to the function "Settings->System->Load settings from File".
There was no feedback given and users cannot see if their "prusa_printer_settings.ini" failed or not.

load_settings

### DIFF
--- a/src/common/str_utils.cpp
+++ b/src/common/str_utils.cpp
@@ -261,6 +261,22 @@ void StringBuilder::append_string(const char *str) {
     *current_pos_ = '\0';
 }
 
+void StringBuilder::append_string_view(string_view_utf8 str) {
+
+    while (true) {
+        if (is_problem()) {
+            return;
+        }
+
+        char b = str.getbyte();
+        if (b == '\0') {
+            return;
+        }
+
+        append_char(b);
+    }
+}
+
 void StringBuilder::append_printf(const char *fmt, ...) {
     va_list args;
     va_start(args, fmt);
--- a/src/gui/MItem_menus.cpp
+++ b/src/gui/MItem_menus.cpp
@@ -1,5 +1,6 @@
 #include "MItem_menus.hpp"
 #include "ScreenHandler.hpp"
+#include "str_utils.hpp"
 #include <option/buddy_enable_connect.h>
 #if BUDDY_ENABLE_CONNECT()
     #include <connect/marlin_printer.hpp>
@@ -403,16 +404,27 @@ MI_LOAD_SETTINGS::MI_LOAD_SETTINGS()
 }
 
 void MI_LOAD_SETTINGS::click(IWindowMenu & /*window_menu*/) {
-    // FIXME: Some error handling/reporting
-    // TODO: Loading other things than just network
-    if (netdev_load_ini_to_eeprom()) {
+    auto build_message = [](StringBuilder &msg_builder, string_view_utf8 name, bool ok) {
+        msg_builder.append_string_view(name);
+        msg_builder.append_string(": ");
+        msg_builder.append_string_view(ok ? _("Ok") : _("Failed"));
+        msg_builder.append_char('\n');
+    };
+    std::array<char, 150> msg;
+    StringBuilder msg_builder(msg);
+    msg_builder.append_string_view(_("\nLoading settings finished.\n\n"));
+
+    const bool network_settings_loaded = netdev_load_ini_to_eeprom();
+    if (network_settings_loaded) {
         notify_reconfigure();
     }
+    build_message(msg_builder, _("Network"), network_settings_loaded);
 
-// FIXME: Error handling
 #if BUDDY_ENABLE_CONNECT()
-    connect_client::MarlinPrinter::load_cfg_from_ini();
+    build_message(msg_builder, _("Connect"), connect_client::MarlinPrinter::load_cfg_from_ini());
 #endif
+
+    MsgBoxInfo(string_view_utf8::MakeRAM((const uint8_t *)msg.data()), Responses_Ok);
 }
 
 /**********************************************************************************************/
--- a/src/common/str_utils.hpp
+++ b/src/common/str_utils.hpp
@@ -6,6 +6,7 @@
 #include <cstdint>
 #include <algorithm>
 #include <assert.h>
+#include "../lang/string_view_utf8.hpp"
 
 inline constexpr char CHAR_SPACE = ' ';
 inline constexpr char CHAR_NBSP = '\xA0'; /// Non Breaking Space
@@ -383,6 +384,8 @@ public:
     void append_char(char ch);
 
     void append_string(const char *str);
+
+    void append_string_view(string_view_utf8 str);
 
     /// Appends text to the builder, using vsnprintf under the hood.
     void append_printf(const char *fmt, ...);
--- a/src/lang/string_view_utf8.hpp
+++ b/src/lang/string_view_utf8.hpp
@@ -145,6 +145,11 @@ public:
         , s(0xff) {}
     ~string_view_utf8() = default;
 
+    /// @returns one uint8_t from the input data
+    uint8_t getbyte() {
+        return getbyte(attrs);
+    }
+
     /// @returns one UTF-8 character from the input data
     /// and advances internal pointers (in derived classes) to the next one
     unichar getUtf8Char() {


vorner:
Good morning

Agreed it would be nice to have _some_ feedback coming from this menu item. It's always been this very out of the way „geek“ access.

Nevertheless, the current solution seems both a bit overkill (a complete new screen) and not really looking nice. Do you think a simple message box (possibly with slightly less detailed output, maybe just success/failure) would suffice?

An inspiration can be taken here: https://github.com/prusa3d/Prusa-Firmware-Buddy/blob/master/src/gui/screen_menu_connect.cpp#L35

bkerler:
@vorner Couldn't agree more. I tried a more minimalistic approach based on your code recommendation, I hope you like it :)

src/gui/MItem_menus.cpp
```diff
@@ -402,17 +402,37 @@ MI_LOAD_SETTINGS::MI_LOAD_SETTINGS()
     : WI_LABEL_t(_(label), nullptr, is_enabled_t::yes, is_hidden_t::no) {
 }
 
+struct SettingsLoadState {
+    bool is_network : 1,
+        is_connect : 1;
+};
```
vorner, line 408: Tiny nitpick: I think this is the only place in the whole codebase that declares two fields in one statement in a struct; we prefer to repeat the type, eg:

```
struct SettingsLoadState {
   bool is_network;
   bool is_connect;
}
```

Also, having a struct for this is kind of complete overkill ‒ I know you're trying to save 1 byte, but it's on stack and this is probably not the deepest stack around. You can just have them as two variables on stack and everything will be fine (and more readable 😇).

vorner (commented):
Yes, I like it better, with just a small nitpick.

src/gui/MItem_menus.cpp
bkerler: I had the custom filament PR in mind, but true, it would also only add another bool and nothing more. I've changed it.

src/gui/MItem_menus.cpp
```diff
@@ -402,17 +402,37 @@ MI_LOAD_SETTINGS::MI_LOAD_SETTINGS()
     : WI_LABEL_t(_(label), nullptr, is_enabled_t::yes, is_hidden_t::no) {
 }
 
+struct SettingsLoadState {
+    bool is_network;
```
CZDanol, line 406: Naming the variables `is_` is very confusing here. Much better would be `network_settings_loaded` and `connect_settings_loaded`.

Also, you can make these use a single bit to save space by `bool network_settings_loaded : 1;`

src/gui/MItem_menus.cpp
```diff
@@ -402,17 +402,37 @@ MI_LOAD_SETTINGS::MI_LOAD_SETTINGS()
...
+    bool is_connect;
+};
+
 void MI_LOAD_SETTINGS::click(IWindowMenu & /*window_menu*/) {
-    // FIXME: Some error handling/reporting
-    // TODO: Loading other things than just network
+    SettingsLoadState settings_loaded = SettingsLoadState { .is_network = true, .is_connect = true };
```
CZDanol, line 411: Avoid this initialization by setting default values in the struct definition.

Also, I'd prefer having them initialized to false and not true, doesn't make much sense like this to me.

src/gui/MItem_menus.cpp
```diff
@@ -402,17 +402,37 @@ MI_LOAD_SETTINGS::MI_LOAD_SETTINGS()
...
+
     if (netdev_load_ini_to_eeprom()) {
         notify_reconfigure();
+    } else {
```
CZDanol, line 415: You can avoid the else branch if the settings are default initialized to false.

src/gui/MItem_menus.cpp
```diff
@@ -402,17 +402,37 @@ MI_LOAD_SETTINGS::MI_LOAD_SETTINGS()
...
 
-// FIXME: Error handling
 #if BUDDY_ENABLE_CONNECT()
-    connect_client::MarlinPrinter::load_cfg_from_ini();
+    if (!connect_client::MarlinPrinter::load_cfg_from_ini()) {
+        settings_loaded.is_connect = false;
+    }
+#endif
+    if (settings_loaded.is_connect & settings_loaded.is_network) {
+        MsgBoxInfo(_("All settings loaded."), Responses_Ok);
```
CZDanol, line 425: * When you say "all", it's quite unclear what it encompasses.
* When multiple settings fail to load, multiple dialogs will be shown, which would probably be annoying.
* I'd argue that settings not loading should not necessarily raise "FAIL FAIL". The settings might simply not be in the ini file and that would be okay.

So in the end, I'd probably always show a single dialog with a text akin to:
> Loading settings finished.
> Network: (Not) loaded
> Connect: (Not) loaded

CZDanol: Edit: checked the code and settings load returns false if the config is somehow malformed or if the ini file is missing. So it is probably okay to call it  "failed".

src/gui/MItem_menus.cpp
CZDanol: Also, as @vorner previously remarked, you don't really need a structure here at all. You would simply do with two local variables inside MI_LOAD_SETTINGS::click.

src/gui/MItem_menus.cpp
bkerler: I've adapted to it, but used "+" and "-" instead, as the Dialog max msg length is rather limited and more messages will be added in the future.

src/gui/MItem_menus.cpp
bkerler: Done, removed the struct

src/gui/MItem_menus.cpp
bkerler: Removed, so no longer relevant

src/gui/MItem_menus.cpp
```diff
@@ -402,17 +402,32 @@ MI_LOAD_SETTINGS::MI_LOAD_SETTINGS()
...
     }
 
-// FIXME: Error handling
 #if BUDDY_ENABLE_CONNECT()
-    connect_client::MarlinPrinter::load_cfg_from_ini();
+    if (connect_client::MarlinPrinter::load_cfg_from_ini()) {
+        is_connect = true;
+    }
+#endif
+    char msg[MSG_MAX_LENGTH] = "\nLoading settings finished.\n\n";
```
CZDanol, line 425: Why is there a newline at the beginning of the text?

CZDanol, line 425: Why are you using MSG_MAX_LENGTH for the limit?

CZDanol, line 425: The next needs to be put through the translation process.

src/gui/MItem_menus.cpp
```diff
@@ -402,17 +402,32 @@ MI_LOAD_SETTINGS::MI_LOAD_SETTINGS()
     : WI_LABEL_t(_(label), nullptr, is_enabled_t::yes, is_hidden_t::no) {
 }
 
+static void add_load_response(const char *name, bool flag, char *msg, int msglen) {
+    char tmp[strlen(name) + 5];
+    snprintf(tmp, sizeof(tmp), "%s: %s\n", name, flag ? "+" : "-");
```
CZDanol, line 407: "+" and "-" are not very telling. If you want to make it short, make it for example "Ok" and "Fail". We're not that limited in memory.

Also, the "Ok" and "Fail" need to be translated

src/gui/MItem_menus.cpp
```diff
@@ -402,17 +402,32 @@ MI_LOAD_SETTINGS::MI_LOAD_SETTINGS()
...
     if (netdev_load_ini_to_eeprom()) {
         notify_reconfigure();
+        is_network = true;
     }
 
-// FIXME: Error handling
 #if BUDDY_ENABLE_CONNECT()
-    connect_client::MarlinPrinter::load_cfg_from_ini();
+    if (connect_client::MarlinPrinter::load_cfg_from_ini()) {
+        is_connect = true;
```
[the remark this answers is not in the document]
bkerler: Agreed

src/gui/MItem_menus.cpp
bkerler: That's for optical reasons, looks better this way.

src/gui/MItem_menus.cpp
```diff
@@ -402,17 +402,32 @@ MI_LOAD_SETTINGS::MI_LOAD_SETTINGS()
     : WI_LABEL_t(_(label), nullptr, is_enabled_t::yes, is_hidden_t::no) {
 }
 
+static void add_load_response(const char *name, bool flag, char *msg, int msglen) {
+    char tmp[strlen(name) + 5];
+    snprintf(tmp, sizeof(tmp), "%s: %s\n", name, flag ? "+" : "-");
+    strncat(msg, tmp, msglen);
```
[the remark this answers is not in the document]
bkerler: Removed the function, by using StringBuilder instead

src/gui/MItem_menus.cpp
```diff
@@ -402,17 +402,32 @@ MI_LOAD_SETTINGS::MI_LOAD_SETTINGS()
     : WI_LABEL_t(_(label), nullptr, is_enabled_t::yes, is_hidden_t::no) {
 }
 
+static void add_load_response(const char *name, bool flag, char *msg, int msglen) {
+    char tmp[strlen(name) + 5];
```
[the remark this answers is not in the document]
bkerler: Using StringBuilder instead

src/gui/MItem_menus.cpp
```diff
@@ -402,17 +402,32 @@ MI_LOAD_SETTINGS::MI_LOAD_SETTINGS()
...
 
-// FIXME: Error handling
 #if BUDDY_ENABLE_CONNECT()
-    connect_client::MarlinPrinter::load_cfg_from_ini();
+    if (connect_client::MarlinPrinter::load_cfg_from_ini()) {
+        is_connect = true;
+    }
+#endif
+    char msg[MSG_MAX_LENGTH] = "\nLoading settings finished.\n\n";
+    add_load_response("Network", is_network, msg, MSG_MAX_LENGTH - strlen(msg));
```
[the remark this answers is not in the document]
bkerler: Good to know. Didn't see any usage of it before.

src/gui/MItem_menus.cpp
```diff
@@ -402,17 +402,32 @@ MI_LOAD_SETTINGS::MI_LOAD_SETTINGS()
     : WI_LABEL_t(_(label), nullptr, is_enabled_t::yes, is_hidden_t::no) {
 }
 
+static void add_load_response(const char *name, bool flag, char *msg, int msglen) {
```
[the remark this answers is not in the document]
bkerler: Removed the function, by using StringBuilder instead

src/gui/MItem_menus.cpp
bkerler: Ok

src/gui/MItem_menus.cpp
bkerler: Changed that.

src/common/str_utils.cpp
```diff
@@ -261,6 +261,23 @@ void StringBuilder::append_string(const char *str) {
...
 
+void StringBuilder::append_string_view(string_view_utf8 str) {
+
+    while (true) {
+        if (is_problem()) {
+            return;
+        }
+
+        char b;
+        str.copyToRAM(&b, 1);
```
CZDanol, line 272: Uff, haven't noticed this before. copyToRam is quite heavy for this, I'd suggest making getbyte public instead.

src/gui/MItem_menus.cpp
```diff
@@ -403,16 +404,26 @@ MI_LOAD_SETTINGS::MI_LOAD_SETTINGS()
 }
 
 void MI_LOAD_SETTINGS::click(IWindowMenu & /*window_menu*/) {
-    // FIXME: Some error handling/reporting
-    // TODO: Loading other things than just network
-    if (netdev_load_ini_to_eeprom()) {
+    bool network_settings_loaded = netdev_load_ini_to_eeprom();
```
CZDanol, line 407: Please add all the code using network_settings_loaded close together

src/gui/MItem_menus.cpp
```diff
@@ -403,16 +404,26 @@ MI_LOAD_SETTINGS::MI_LOAD_SETTINGS()
...
+    if (network_settings_loaded) {
         notify_reconfigure();
     }
-
-// FIXME: Error handling
+    auto build_message = [](StringBuilder &msg_builder, string_view_utf8 name, bool ok) {
+        msg_builder.append_string_view(name);
+        msg_builder.append_string(": ");
+        msg_builder.append_string_view(ok ? _("Ok") : _("Failed"));
+        msg_builder.append_string("\n");
```
CZDanol, line 416: Nitpick: You could use more efficient append_char here.

src/common/str_utils.cpp
bkerler: it's now lightweighter :)

src/gui/MItem_menus.cpp
bkerler: aye

src/gui/MItem_menus.cpp
bkerler: Done.

src/common/str_utils.cpp
```diff
@@ -261,6 +261,22 @@ void StringBuilder::append_string(const char *str) {
...
 }
 
+void StringBuilder::append_string_view(string_view_utf8 str) {
+
+    while (true) {
+        if (is_problem()) {
+            return;
+        }
+
+        char b = str.getUtf8Char();
```
CZDanol, line 271: getUtf8Char does decoding the string and returns a full unicode. You cannot use it here. You have to use getbyte

src/gui/MItem_menus.cpp
```diff
@@ -403,16 +404,27 @@ MI_LOAD_SETTINGS::MI_LOAD_SETTINGS()
...
+        msg_builder.append_string_view(name);
+        msg_builder.append_string(": ");
+        msg_builder.append_string_view(ok ? _("Ok") : _("Failed"));
+        msg_builder.append_char('\n');
+    };
+    std::array<char, 150> msg;
+    StringBuilder msg_builder(msg);
+    msg_builder.append_string_view(_("\nLoading settings finished.\n\n"));
+
+    bool network_settings_loaded = netdev_load_ini_to_eeprom();
```
CZDanol, line 417: Nitpick: This can be made `const bool`

bkerler: Done.

src/common/str_utils.cpp
bkerler: getbyte was private, so I added a public getbyte function

CZDanol (dismissed):

vorner:
Hmm, I've noticed the unit tests don't build now. Can you have a look? This one should be easy.

bkerler:
@vorner Tests build fine now. Everything looks perfect.

vorner (approved):

Issue #3748, opened 2024-02-12
[ENHANCEMENT] [OctoPrint support] Printer state information is not passed to host over serial


arekm:
**Printer type** - MK3.5
**Printer firmware version** - 5.2.0
**Original or Custom firmware** - Original

**USB drive or USB/Octoprint**
USB flash drive - that's important. **Prints are started from printer LCD screen, from USB drive**.

**Describe the bug**
MK3S+ firmware informed host (OctoPrint) over serial about what is the state of the printer.
For example if you started print from sd card then OctoPrint knew that. When print finished - the same. On pause - the same.

OctoPrint with such knowledge was able to start timelapse recording when print was started (from lcd) and finish it when it ended. And do much more like turn on LED lighting at print start, turn off at end of print.  Send notification about printer state to the user phone (using OctoPod plugin) and more.

All that functionality is lost with 32bit firmware.

**How to reproduce** / **Expected behavior**
Connect Octoprint to MK3.5/MK4 printer. Start print from LCD. OctoPrint should see that print has started. Stop the print from LCD. OctoPrint should also see that fact.

**Logs**

Example from MK3S+ log for a print started from MK3S+ LCD:
````
2024-02-01 12:18:00,997 - Recv: Not SD printing
2024-02-01 12:18:00,998 - Recv: ok
2024-02-01 12:18:01,791 - Send: M27
2024-02-01 12:18:01,796 - Recv: Not SD printing
2024-02-01 12:18:01,796 - Recv: ok
2024-02-01 12:18:02,484 - Recv: echo:Now fresh file: gasket~1.gco
2024-02-01 12:18:02,492 - Recv: File opened: /GASKET_0.2mm_FLEX_MK3S_21m.gcode Size: 125448
2024-02-01 12:18:02,495 - Recv: File selected
2024-02-01 12:18:02,514 - Recv: LCD status changed
2024-02-01 12:18:02,793 - Send: M27
2024-02-01 12:18:02,943 - Recv: LCD status changed
2024-02-01 12:18:02,947 - Recv: echo:enqueing "M23 gasket~1.gco"
2024-02-01 12:18:02,947 - Recv: echo:enqueing "M24"
2024-02-01 12:18:02,983 - Recv: echo:Now fresh file: gasket~1.gco
2024-02-01 12:18:02,992 - Recv: File opened: /GASKET_0.2mm_FLEX_MK3S_21m.gcode Size: 125448
[...]

PR #3770, opened 2024-02-18
serial: Add missing serial file responses for Octoprint on file open and print finished

bkerler:
Related to issue #3748. In octoprint, for running a gcode file, several responses are required over serial port. See ["Done printing file"](https://github.com/OctoPrint/OctoPrint/blob/65a5533c5d645f7033f01056c8ef735a938339f0/src/octoprint/util/comm.py#L3120-L3122) and ["File opened"](https://github.com/OctoPrint/OctoPrint/blob/65a5533c5d645f7033f01056c8ef735a938339f0/src/octoprint/util/comm.py#L3085). This PR adds this missing functionality.

### DIFF
--- a/src/common/media.cpp
+++ b/src/common/media.cpp
@@ -354,6 +354,10 @@ void media_print_start() {
         media_gcode_position = media_current_position = 0;
         media_print_state = media_print_state_PRINTING;
         media_print_size_estimate = media_print_file->get()->get_gcode_stream_size_estimate();
+
+        // Do not remove, needed for 3rd party tools such as octoprint to get status about the gcode file being opened
+        SERIAL_ECHOLNPAIR(MSG_SD_FILE_OPENED, marlin_vars()->media_SFN_path.get_ptr(), " Size:", media_print_size_estimate);
+
         gcode_filter.reset();
         osSignalSet(prefetch_thread_id, PREFETCH_SIGNAL_START);
     } else {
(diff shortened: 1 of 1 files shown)


danopernis (changes requested):
This doesn't seem like the right place to put those serial messages into. If you do a quick
```
git grep SERIAL_ECHO |cut -f 1 -d : |sort |uniq
```
you will see that output to serial line is mostly done in `lib/Marlin/Marlin/src/` or `src/marlin_stubs/` (with some exceptions, which should frankly be eradicated)

I would look for some better place to do that. Maybe somewhere near serial printing code?

Also, please indicate somehow that those string actually need to be verbatim because some program outside of our codebase depends on them. This will prevent somebody from removing/changing them, unintentionally breaking some 3rd party code.

bkerler:
The problem here is that the gcode command M23 https://github.com/prusa3d/Prusa-Firmware-Buddy/blob/b3ee47aa573f4d9b4af82d7a6963f43ab37ccc08/src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp#L41
is only setting the marlin_vars()->media_SFN_path. As the file is only opened in the media handler, there is no other way to send the serial message (at least none that would make sense to me). Same applies for stopping the print of the gcode file. (We could use the spot for finishing print at all, which has also serial messages, but this doesn't make sense as it wouldn't be usb gcode file related). I will add some comments that they relate to 3rd party code.

Issue #3775, opened 2024-02-20
[BUG] M0 command does not display the message string


DoubleStrike:
**Printer type** - XL 5 Toolhead

**Printer firmware version** - 5.1.2+13478

**Original or Custom firmware** - Original

**Describe the bug**
  Sending an M0 unconditional stop command does not display the message associated with the gcode.  While this is minimally functional, the message may be important and should always be shown.

**How to reproduce**
  Add the following line to any gcode:
`M0 You should see this message`

**Expected behavior**
  The printer should pause and display the user message. This is the expectation for a functional Marlin firmware, as per this page: https://marlinfw.org/docs/gcode/M000-M001.html.

bkerler:
I wrote the PR #3779 for your issue.

Prusa-Support:
Good news! The Pull Request was merged and just implemented in [FW 6.0.0-RC3](https://github.com/prusa3d/Prusa-Firmware-Buddy/releases/tag/v6.0.0-RC3).

Michele Moramarco
Prusa Research

PR #3779, opened 2024-02-20
gcode: Add string argument for cmd M0

bkerler:
This adds a custom message for gcode command M0. Resolves issue #3775.

Example: `M0 This is a test msg !`

m0_msg

### DIFF
--- a/src/gui/dialogs/window_dlg_quickpause.cpp
+++ b/src/gui/dialogs/window_dlg_quickpause.cpp
@@ -11,7 +11,7 @@
 
 constexpr static const char quick_pause_txt[] = N_("Waiting for the user. Press \"Resume\" once the printer is ready.");
 
-DialogQuickPause::DialogQuickPause(fsm::BaseData)
+DialogQuickPause::DialogQuickPause(fsm::BaseData data)
     : AddSuperWindow<IDialogMarlin>(GuiDefaults::RectScreenBody)
     , icon(this, GuiDefaults::MessageIconRect, &img::warning_48x48)
     , text(this, GuiDefaults::MessageTextRect, is_multiline::yes, is_closed_on_click_t::yes, _(quick_pause_txt))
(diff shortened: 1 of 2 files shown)


danopernis (commented):
While this certainly works, it permanently eats 423 bytes of RAM. Is it worth it? Especially MINI might have problems with that, we are already hitting some OOM situations due to heathshrink decompression and what not...

I am dreaming of being able to transport pointer to the memory via the FSM mechanism. Instead of using the empty here
```
    FSM_HOLDER_WITH_DATA__LOGGING(QuickPause, PhasesQuickPause::QuickPaused, fsm::PhaseData());
```
we could use the pointer which will just fit the 4 bytes available in the PhaseData. There might be some race condition when destroying the dialog though, not sure about that. Granted, it will require more engineering that this hotfix. On the other hand, we may merge this and fix it properly later. I will need to discuss this inside the team.

bkerler:
Thanks @danopernis. Yes, this was meant as quick-fix. I looked into the FSM mechanism but couldn't find any easy way (or at all) without modifying the whole QuickPause class which would lead to much more issues. We could also decrease the size of max chars, I'm open for ideas.

bkerler:
@danopernis I've improved the code now based on your suggestions. A better solution would be of course to introduce a message field in BaseData, but that change would required much larger changes and might introduce issues for sure.

src/gui/dialogs/window_dlg_quickpause.cpp
```diff
@@ -24,5 +24,12 @@ DialogQuickPause::DialogQuickPause(fsm::BaseData)
         marlin_vars()->media_LFN.copy_to(buff, FILE_NAME_BUFFER_LEN, lock);
         gcode_name.SetText(_(buff));
     }
+
+    const char *msg;
+    memcpy(&msg, (uint32_t *)data.GetData().data(), sizeof(uint32_t));
+    if (msg) {
+        text.SetText(_(msg));
```
danopernis, line 31: We do not want to translate this string. Use `string_view_utf8::MakeRAM()` for construction.

danopernis (changes requested):

src/gui/dialogs/window_dlg_quickpause.cpp
bkerler: Done.

danopernis (approved):
Thanks a lot!

Issue #3711, opened 2024-01-27
[ENHANCEMENT] Mk4 xBuddy GPIO gcode accessibility (M42, M226 specifically)


Fil4ment:
**Printer type** - [MK4]
**Optional upgrades** - [Lack enclosure]

**Is your suggested improvement related to an existing problem? Please describe.**
  I'm trying to build an enclosure temperature control system (arduino) which will be commanded by the G-Code running on the printer, allowing a separate processor controlling chamber temperature, with the printer sending only "desired temperature" in the pre-print code, then waiting for a "temperature reached" response, before continuing to print. I know that I can do this using M42 and M226 commands, which are both supported by the firmware.
  
  But only 3 pins (on a stock Mk4) are defined for "analogWrite" commands, which come from the M42 command: FAN, FAN1, and HEAT0.

**Describe the expected functionality**
  I would like to see some way of setting unused pins to be input, output, input_pullup, or output_pullup (if available in the controller), before addressing them via GCode, for external accessory use.

  There really should be a way of using "M42 P76 T1" to set Pin E12 (unused pin on expansion port) as an output, then run "M42 P76 S128" to drive pin E12 to 50% duty cycle. "M42 P15 T0" to set Pin A15 (TDI(?) on expansion port) as an input, then run "M226 S0" to wait for the arduino to pull the pin low (assuming the arduino is driving the pin high, until temperature is reached).
  
  The M42 and M226 functionality is already in the firmware. You just need to open up the unused pins to user/gcode access.

me25542:
I have a very similar plan to Fil4ment, I would also like to see a firmware in which all unused pins (like the ones described here [https://help.prusa3d.com/article/accessory-connectors-mk4_622345](url) are unlocked for the M42 and M226 gcode commands.
Thanks!

bkerler:
In fact, it does make more sense to use the specified extension headers in the documentation for mods / third party accessories using the spi or i2c interfaces as recommended in the article and the firmware than the gpios directly for multiple reasons. So to use i2c or spi interfaces as listed [here](https://help.prusa3d.com/article/accessory-connectors-mk4_622345).

Fil4ment:
That doesn't list J9, and there's not (as far as I'm aware) an easy gcode command to "wait for a command" over i2c or SPI, and the SPI pins available on J9 aren't usable by me, anyway, as the firmware doesn't allow SPI via Gcode.

[...]

PR #3840, opened 2024-03-19
[BFW-5420] Feature/i2c m260 m261

rbuehlma:
This adds support for the m260 m261 commands and makes it possible to control an I2C device attached to the I2C socket.

Related to https://github.com/prusa3d/Prusa-Firmware-Buddy/issues/3711

Since I'm not used to developing c++ and even less for a firmware, please let me know if something needs to be changed to make it possible to include this change.

Please not that this change requires EXPERIMENTAL_I2CBUS to be enabled to be effective and therefore requires a custom firmware build.

### DIFF
--- a/lib/Marlin/Marlin/src/feature/twibus.cpp
+++ b/lib/Marlin/Marlin/src/feature/twibus.cpp
@@ -1,6 +1,6 @@
 /**
  * Marlin 3D Printer Firmware
- * Copyright (c) 2019 MarlinFirmware [https://github.com/MarlinFirmware/Marlin]
+ * Copyright (c) 2020 MarlinFirmware [https://github.com/MarlinFirmware/Marlin]
  *
  * Based on Sprinter and grbl.
  * Copyright (c) 2011 Camiel Gubbels / Erik van der Zalm
@@ -16,7 +16,7 @@
  * GNU General Public License for more details.
  *
  * You should have received a copy of the GNU General Public License
- * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  *
  */
 
@@ -28,11 +28,24 @@
 
 #include <Wire.h>
 
+#include "../libs/hex_print.h"
+
+TWIBus i2c;
+
 TWIBus::TWIBus() {
   #if I2C_SLAVE_ADDRESS == 0
-    Wire.begin();                  // No address joins the BUS as the master
+
+    #if PINS_EXIST(I2C_SCL, I2C_SDA) && DISABLED(SOFT_I2C_EEPROM)
+      Wire.setSDA(pin_t(I2C_SDA_PIN));
+      Wire.setSCL(pin_t(I2C_SCL_PIN));
+    #endif
+
+    Wire.begin();                   // No address joins the BUS as the master
+
   #else
-    Wire.begin(I2C_SLAVE_ADDRESS); // Join the bus as a slave
+
+    Wire.begin(I2C_SLAVE_ADDRESS);  // Join the bus as a slave
+
   #endif
   reset();
 }
@@ -43,43 +56,32 @@ void TWIBus::reset() {
 }
 
 void TWIBus::address(const uint8_t adr) {
-  if (!WITHIN(adr, 8, 127)) {
+  if (!WITHIN(adr, 8, 127))
     SERIAL_ECHO_MSG("Bad I2C address (8-127)");
-  }
 
   addr = adr;
 
-  #if ENABLED(DEBUG_TWIBUS)
-    debug(PSTR("address"), adr);
-  #endif
+  debug(F("address"), adr);
 }
 
 void TWIBus::addbyte(const char c) {
   if (buffer_s >= COUNT(buffer)) return;
   buffer[buffer_s++] = c;
-  #if ENABLED(DEBUG_TWIBUS)
-    debug(PSTR("addbyte"), c);
-  #endif
+  debug(F("addbyte"), c);
 }
 
 void TWIBus::addbytes(char src[], uint8_t bytes) {
-  #if ENABLED(DEBUG_TWIBUS)
-    debug(PSTR("addbytes"), bytes);
-  #endif
+  debug(F("addbytes"), bytes);
   while (bytes--) addbyte(*src++);
 }
 
 void TWIBus::addstring(char str[]) {
-  #if ENABLED(DEBUG_TWIBUS)
-    debug(PSTR("addstring"), str);
-  #endif
+  debug(F("addstring"), str);
   while (char c = *str++) addbyte(c);
 }
 
 void TWIBus::send() {
-  #if ENABLED(DEBUG_TWIBUS)
-    debug(PSTR("send"), addr);
-  #endif
+  debug(F("send"), addr);
 
   Wire.beginTransmission(I2C_ADDRESS(addr));
   Wire.write(buffer, buffer_s);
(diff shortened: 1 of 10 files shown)


rbuehlma:
I did rebase the code to the current release v6.0.0

danopernis:
Hi @rbuehlma thanks for the rebase and sorry about the force-push and the wait.

Looking at the code I don't think this would work. Buddy platform is providing Arduino-compatible API for Marlin to use, See `src/common/hwio*` files and `digitalWrite()` function and those i2c pins are not exposed. But even if they were, bit banging the i2c would not work because the bits are in i2c mode controlled by the stm32f4 processor and they should be controled via STM HAL Interface (`HAL_I2C_*()` functions). Special care must be taken to not cause any deadlock and not trigger the main task watchdog. What is more, you do not want to enable just any gcode to communicate via i2c as there are other peripherals attached there and a malicious attacker could craft a special gcode to for example destroy EEPROM constents.

While I like the general idea, it needs to be reimplemented entirely to conform with the buddy architecture and safety checks must be in place.

P.S. Holly/Jenkins build system is broken right now, do not worry about it.

3d-gussner:
Please don't forget to update https://reprap.org/wiki/G-code#M260:_i2c_or_Modbus_Send_Data

rbuehlma:
Thanks for the feedback. I'm not really used to develop firmwares and am therefore happy for any input :)

Technically, this change is working (I'm using it to control a laser header through an I2C port expander but this required additional changes (see below)). The calls to the i2c bus are with this patch done through  HAL_I2C_Master_Transmit. Did I still miss something there?

I agree that malicious gcode has an additional attack vector by enabling this feature but taking gcode from an untrusted source has already been a risk before. By blacklisting already used I2C addresses, I believe it would be possible to fix this.

During my testings, I never had deadlocks caused by this code, but I guess the watchdog could be triggered if some device takes too long to respond (this never happened to me).

If you are interested in what else I did to get my laser header working correctly, this is the corresponding branch:
https://github.com/rbuehlma/Prusa-Firmware-Buddy/tree/feature/laser
Don't be afraid, I won't create a PR for this one as I had to dig too much into the movement processing to get the timing of the laser header to be in sync with the movement :) I'm sure there must be a much simpler way.

danopernis:
Oh shoot you are right, I was somehow only looking into the first commit :facepalm: 

This already looks much better :slightly_smiling_face: and I will open an internal ticket [BFW-5420] for us to see what we can do about addressing the security issues raised and getting this merged.

We will also need to solve which i2c HAL handle to use. There are several of them and each has different devices attached, differing between MINI, MK4 and XL depending on the particular electronics. For this, I will have to look into electronic schemas.

rbuehlma:
Great, thanks. Just FYI: The first commit is simply a C&P from the current Marlin Repo to have their current I2C commands implementation. The second commit is what I did to get it running with the Buddy firmware.

rbuehlma:
Thanks @michalrudolf 

I have updated the PR according to your suggested changes.

lib/Marlin/Marlin/src/feature/twibus.cpp
```diff
@@ -26,174 +26,216 @@
...
+
+bool TWIBus::address(const uint8_t adr) {
   if (!WITHIN(adr, 8, 127)) {
     SERIAL_ECHO_MSG("Bad I2C address (8-127)");
+    return false;
+  }
+
+  if (isRestrictedAddress(const uint8_t addr)) {
+    SERIAL_ECHO_MSG("Restricted I2C address.");
+    return false
```
michalrudolf, line 65: missing ";"

lib/Marlin/Marlin/src/feature/twibus.cpp
```diff
@@ -26,174 +26,216 @@
...
+  return read_buffer[read_buffer_pos++];
+}
+
+bool TWIBus::address(const uint8_t adr) {
   if (!WITHIN(adr, 8, 127)) {
     SERIAL_ECHO_MSG("Bad I2C address (8-127)");
+    return false;
+  }
+
+  if (isRestrictedAddress(const uint8_t addr)) {
```
michalrudolf, line 63: Should be `isRestrictedAddress(adr)`

lib/Marlin/Marlin/src/feature/twibus.cpp
```diff
@@ -26,174 +26,216 @@
...
-    debug(PSTR("send"), addr);
-  #endif
-
-  Wire.beginTransmission(I2C_ADDRESS(addr));
-  Wire.write(buffer, buffer_s);
-  Wire.endTransmission();
-
+  debug(F("send"), addr);
+  
+  i2c::Result ret = i2c::Transmit(&hi2c2, addr << 1, buffer, buffer_s, 100);
```
michalrudolf, line 109: First parameter of i2c::Transmit is a reference (not a pointer), so it should be without "&".

lib/Marlin/Marlin/src/feature/twibus.cpp
```diff
@@ -26,174 +26,216 @@
...
-  while (bytes-- && Wire.available()) SERIAL_CHAR(Wire.read());
-  SERIAL_EOL();
+bool TWIBus::check_hal_response(i2c::Result response) {
+  if (response == i2c::Result::ok) {
+    return true;
+  }
+
+  switch (response)
+  {
+  case i2c::Result::busy_after_retries::error:
```
michalrudolf, line 122: `i2c::Result::error:`

lib/Marlin/Marlin/src/feature/twibus.cpp
```diff
@@ -26,174 +26,216 @@
...
+    SERIAL_ERROR_MSG("TWIBus::send failed with: ERROR");
+    break;
+  case i2c::Result::busy_after_retries:
+    SERIAL_ERROR_MSG("TWIBus::send failed with: BUSY");
+    break;
+  case i2c::Result::busy_after_retries::timeout:
```
michalrudolf, line 128: `i2c::Result::timeout:`

lib/Marlin/Marlin/src/feature/twibus.cpp
```diff
@@ -26,174 +26,216 @@
...
-  if (Wire.requestFrom(addr, bytes) == 0) {
-    #if ENABLED(DEBUG_TWIBUS)
-      debug("request fail", addr);
-    #endif
     return false;
   }
 
+  flush();
+
+  i2c::Result ret = i2c::Receive(&hi2c2, addr << 1, read_buffer, bytes, 100);
```
michalrudolf, line 193: Also here, pass `hi2c2` without `&`

lib/Marlin/Marlin/src/feature/twibus.cpp
```diff
@@ -26,174 +26,216 @@
...
   addr = adr;
 
-  #if ENABLED(DEBUG_TWIBUS)
-    debug(PSTR("address"), adr);
-  #endif
+  debug(F("address"), adr);
+  return true;
+}
+
+bool TWIBus:isRestrictedAddress(uint8_t addr) {
```
michalrudolf, line 74: `TWIBus::`

michalrudolf (changes requested):
Compiler would not be happy if EXPERIMENTAL_I2CBUS was enabled :)
Please address these little changes and then we are good to go.

rbuehlma:
Hell indeed, I had the flag disabled while compiling this version. Now it works with the flag as well.

Would it be a good idea to enable the flag by default such that users of the official firmware can make use of this feature?

michalrudolf:
Yes, you can uncomment it.

lib/Marlin/Marlin/src/feature/twibus.cpp
```diff
@@ -26,174 +26,235 @@
...
+     case 0x37:
+     case 0x38:
+     case 0x39:
+     case 0x3A:
+     case 0x3B:
+     case 0x3C:
+     case 0x3D:
+     case 0x3E:
+     case 0x3F:
+      // IO Extender
```
michalrudolf, line 102: I'm sorry, some of the addresses I gave you were already bitshifted with read/write bit included, so I had to recalculate them:

```
EEPROM:
    0x53, 0x57
USBC:
    0x22, 0x23
IO Extender:
    0x18, 0x19, 0x1A, 0x1B, 0x1C, 0x1D, 0x1E, 0x1F
```

rbuehlma:
I2C is now enabled. I did additionally verify that it still works after these changes.

lib/Marlin/Marlin/src/feature/twibus.cpp
michalrudolf: Replace all restricted addresses with the ones on this list.

michalrudolf:
Enable for XL, not for XL_Dwarf or XL_Dev_Kit (atleast for now).

rbuehlma:
All done.

michalrudolf:
Looks like our CI have a problem. There is a simple solution:

Remove this from Marlin.h
```
#if ENABLED(EXPERIMENTAL_I2CBUS)
  #include "feature/twibus.h"
#endif
```

and this from Marlin.cpp
```
#if ENABLED(EXPERIMENTAL_I2CBUS) && I2C_SLAVE_ADDRESS > 0

  void i2c_on_receive(int bytes) { // just echo all bytes received to serial
    i2c.receive(bytes);
  }

  void i2c_on_request() {          // just send dummy data for now
    i2c.reply("Hello World!\n");
  }

#endif
```

Could you please squash the fixup commits, so we have a few well defined commits in this PR?

rbuehlma:
All done :)

michalrudolf (approved):
Thanks for the contribution :rocket: You can merge it with 'Rebase and merge'

Fil4ment:
Please forgive me for being a noob at github, but I am, so I ask: How do I tell what release this commit has been/is going to be merged into, or when it's in the public release?

danopernis:
@Fil4ment it is good that you ask, it is not so obvious and we are still trying to create a workflow that would work best for us.

We are generally trying to track public pull requests in the release notes. Other than that, we cherry-pick bugfixes to patch versions. Features usually do not make it to patch versions, but eventually end up in minor/major versions. Regardless of the release they actually end up in, we always merge them to the master branch here.

Some day, we should write some nice documentation about that, along with some general info for contributors. It gets complicated because lot of development happens behind closed doors (for reasons we developers don't really have control over)

Issue #3847, opened 2024-03-21
[BFW-5432] [BUG] G90 not working as for MK3 in MK3-compatibility mode


thomasloven:
Please, before you create a new bug report, please make sure you searched in open and closed issues and couldn't find anything that matches.

**Printer type** - MK3.9

**Printer firmware version** - 5.1.2+13478

**Original or Custom firmware** - Original

**Optional upgrades** - E3d Revo heater and nozzle

**USB drive or USB/Octoprint**
  Flash drive

**Describe the bug**
When trying to print gcode sliced for a MK3 printer the MK3.9 tells me that it will run in "MK3-compatibility mode".
However, G90 resets the extruder to absolute positioning, which does not happen on MK3 resulting in a failed purge line.

**How to reproduce**
Print the attached G-code on a MK3.9 or MK4 printer.
Accept the compatibility warning stating that the file will be run in MK3-compatibility mode.
During printing of the purge line, only half will be printed followed by a 21 mm retraction of the filament - causing a failed first layer of the print.

```gcode
M83  ; extruder relative mode
G90 ; use absolute coordinates
G28 W ; home all axes without mesh bed leveling

; several lines omitted

; purge line
; several lines omitted
G1 X240.0 E25.0 F2200.0
; several lines omitted
G1 X55.0 E25 F1400.0
; several lines omitted
G1 X5.0 E4.0 F1000.0
```

**Expected behavior**
  G90 in MK3-compatibility mode behaves the same as G90 on a MK3 printer - not resetting the Extruder stepping mode.

**G-code**
  https://drive.google.com/file/d/172TxwGLs-DS_bLN3n-HEgjLDg-ojv_xd/view?usp=sharing

**Crash dump file**
  N/A

**Video**
  N/A

bkerler:
Just to fully understand the issue: 
1. I assume that you have homed all axes with mesh bed leveling before you intent to print in mk3 mode and you just use G28 W to prevent the mbl to be done again ? 
2. The G90 should issue to use absolute coordinates, so if you say the MK4 does that, but the MK3 doesn't do that, isn't then the MK3 G90 gcode to blame which worked incorrectly before ?
Or do you mean that the MK4 G90 gcode does reset the x,y,z homing values ?

thomasloven:
I did not slice the file, so I cannot answer about the intent.

On MK4 G90 sets all axes (X, Y, Z, E) to absolute coordinates.
[...]

PR #3865, opened 2024-03-29
gcode: Add mk3 compatibility to G90 command

bkerler:
This addresses issue #3847.

When in mk3 compatibility mode, the G90 command should behave exactly as in https://github.com/prusa3d/Prusa-Firmware/blob/MK3/Firmware/Marlin_main.cpp#L5120.

However the MK4 normally does also set absolute values for the E-Axis, which the MK3 didn't do before (which is also documented), as you can see here: 
https://github.com/prusa3d/Prusa-Firmware-Buddy/blob/master/lib/Marlin/Marlin/src/gcode/gcode.h#L314

This PR adds further support for the MK3 compatibility mode by removing setting absolute values for the E-Axis for gcode G90.

### DIFF
--- a/lib/Marlin/Marlin/src/gcode/gcode.h
+++ b/lib/Marlin/Marlin/src/gcode/gcode.h
@@ -323,7 +323,15 @@ public:
     return TEST(axis_relative, a);
   }
   static inline void set_relative_mode(const bool rel) {
-    axis_relative = rel ? _BV(REL_X) | _BV(REL_Y) | _BV(REL_Z) | _BV(REL_E) : 0;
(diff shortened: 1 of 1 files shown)


danopernis:
This mixes actual code changes with formatting changes and is quite difficult to review. It sill also be more difficult to merge. Can you please rework this so the change is minimal?

bkerler:
@danopernis Sorry, seems like clang-format automatically changed the whole file. It's now modded to only reflect the change.

danopernis (approved):
BFW-5432 for internal reference

PR #3919, opened 2024-04-16
Merge v6.0.0 into master

danopernis:


danopernis:
After some internal discussion we decided to do the force-push instead

PR #3944, opened 2024-04-23
Fix CI in public repo

danopernis:
Remote check is remove entirely. If you do not want to push to the remote accidentally, the do not add the remote to the config in the first place. This worked only when running cmake anyway and to keep this it would require tunneling the remote url through buildsystem.

Only allow creating tags on the private fork in order to not clutter the tags page and not spam users subscribed to releases.

Only allow running integration tests on private fork as they are expensive and quite flaky.


mody (approved):

mody (approved):

mody (approved):

danopernis:
BFW-5426 for future reference

PR #3954, opened 2024-04-25
Increase the lines searched for PNG image

dunkmann00:
Only effects plaintext gcode. The new QOI images make it necessary to look further into the gcode to find the PNG image for display in Prusalink (and maybe Prusaconnect? I've not used it so not sure).

Currently, if you have a plaintext gcode file with the default gcode thumbnails (for MK4 that means: 16x16/QOI, 313x173/QOI, 440x240/QOI, 480x240/QOI, 640x480/PNG) Prusalink can't load the thumbnail, because the printer fails to send it.

This fixes https://github.com/prusa3d/Prusa-Link-Web/issues/475

### DIFF
--- a/src/common/gcode/gcode_reader_plaintext.cpp
+++ b/src/common/gcode/gcode_reader_plaintext.cpp
@@ -26,7 +26,7 @@ bool PlainGcodeReader::stream_gcode_start(uint32_t offset) {
 }
 bool PlainGcodeReader::stream_thumbnail_start(uint16_t expected_width, uint16_t expected_height, ImgType expected_type, bool allow_larger) {
     // search for begining of thumbnail in file
-    static const size_t MAX_SEARCH_LINES = 2048;
+    static const size_t MAX_SEARCH_LINES = 4096;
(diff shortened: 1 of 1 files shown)


dunkmann00:
Okay, sounds good. I hadn't realized about the small thumbnails. Where are those shown on PrusaLink?

dunkmann00:
@danopernis I was just looking at this again and made a slight change. I know you said going with just QOI is a better solution, but since that is much more work (and something I presume the team would rather work on themselves vs accepting a contribution) I was hoping a temporary solution could be adopted.

This still expands the number of lines we are searching for, but if as we are checking thumbnails we find one larger than we are looking for, when we do not want a larger thumbnail, we stop looking. The intent here is to address the situation you mentioned with the `thumb/s/` endpoint.

I know it is a bit of a workaround, but I think it could work as a temporary solution so long as the thumbnails are ordered smallest to largest in the Prusaslicer config, and so long as the PNG at the end is not smaller than a QOI image before it.

PR #4006, opened 2024-05-17
Allow firmware files to be uploaded to sdcard remotely

mikec85:
Allow firmware files to be uploaded to sdcard remotely.

### DIFF
--- a/src/transfers/transfer.cpp
+++ b/src/transfers/transfer.cpp
@@ -227,7 +227,7 @@ void Transfer::init_download_order_if_needed() {
     }
     bool is_plain_gcode = filename_is_plain_gcode(slot.destination());
     bool has_sufficient_size = partial_file->final_size() >= PlainGcodeDownloadOrder::MinimalFileSize;
-    if (is_plain_gcode && has_sufficient_size) {
+    if (is_plain_gcode && has_sufficient_size || filename_is_firmware) {
         order = DownloadOrder(PlainGcodeDownloadOrder(*partial_file));
     } else {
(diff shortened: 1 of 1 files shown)


CZDanol (changes requested):
How is this supposed to fix anything? I don't understand why firmware shouldn't go through GenericFileDownloadOrder

pausethelogic:
There is no SD card on printers with the Buddy board. Are you sure you made a PR on the right repo?

Also, on printers with the buddy board, firmware can already be downloaded remotely to the printer with Prusa Link and Prusa Connect, so I'm not sure what problem this solves

pilot51:
> Also, on printers with the buddy board, firmware can already be downloaded remotely to the printer with Prusa Link and Prusa Connect

I just tried it in PrusaLink on my MK4 with 6.0.1 upgrading to 6.0.2 and got this error after the upload completed:
> 415: Unsupported Media Type
> Not a GCODE

I rebooted after the transfer and it did not prompt to install the update. I transferred again and then checked the USB drive on my computer and the firmware file did not make it over.

It would be nice to not have to move the USB drive to my PC to do firmware updates. And no, I have no need for Prusa Connect and I'm not going to use it just so I can do that.

danopernis:
@pilot51 Hmm, that is a valid request, however this PR doesn't work. It doesn't even compile. Maybe the author wanted to call the function, but who knows. The author didn't even check if the code works at all. Even then, why would we follow gcode path if the file contains firmware? It is certainly not printable. All in all, this doesn't work and we are closing this.

PR #4014, opened 2024-05-26
Update fast nozzle cooling

espr14:
Fast nozzle cooling used 'F' parameter in M109 but this collides with Marlin FW so 'C' should be used (as reserved in https://reprap.org/wiki/G-code)


espr14:
I've created and implemented that parameter but I've used already used parameter. So I've changed it to C. I'm not sure why it's back. Also people don't use it because it was not promoted.

bkerler:
@danopernis he used the wrong flag in [this commit](https://github.com/prusa3d/Prusa-Firmware-Buddy/commit/a6dce3720f2efa9c54a4455449e2527bf8497d17). It should be fixed in order to reflect the official documentation [here](https://reprap.org/wiki/G-code#M109:_Set_Extruder_Temperature_and_Wait) , which states that 'C' should be used.

Issue #3480, opened 2023-11-16
[BFW-6598] [BFW-5677] [BUG] XL Fan Test Inconsistent


JeremiahGillis:
**Printer type** - [XL Five-Head]

**Printer firmware version** - [5.1.0-alpha2]

**Original or Custom firmware** - [Original]

**Describe the bug**
The fan test produces inconsistent results. I tried this test 15 times and it would not pass. The fan test would tell me different fans failed each time.

A quick Google search shows a number of users with the same problem. Some users take the fan shroud off to get this test to pass.

![image](https://github.com/prusa3d/Prusa-Firmware-Buddy/assets/13179985/14039670-44cd-4443-86d4-3d05f3b94cb0)

![image](https://github.com/prusa3d/Prusa-Firmware-Buddy/assets/13179985/0dd3c651-cf7e-4329-b558-243f614b30f4)

**How to reproduce**
Power on printer, navigate to the fan test, and run it. In my specific case, this menu came up at power on because I just unboxed my five-head assembled machine, updated the firmware, then went through the printer setup. I ended up skipping this test, but the printer prompted that I had an unfinish test multiple times during the rest of the calibration.

**Expected behavior**
The fan test passes since all 10 fans were functioning visually. Or the fan test fails correctly when a fan is not functional.

This isn't a great experience when powering your machine up for the first time and you cannot get past the first test due to bugs.

JeremiahGillis:
I downgraded to 4.7.5. My fan test failed the first time. I tried a second time and it passed.

brettvitaz:
FWIW, I had the same problem on 4.7. Prusa Support said it would be fixed in a future firmware.

jmsether:
If you want to get it to pass while they are fixing it in the firmware you can remove all the part cooling fan shrouds and then run the test. I was getting this same error and removing all the fan shrouds allowed me to pass so I could get rid of that annoying message to complete calibration.

daavery:
Fan tests are air density sensitive.  At high altitudes the fans spin faster.  I'm at 8200 ft (2500 m) and full speed print fan runs at 7200 rpm which fails the fan test

M106 P0 S255
M123
 E0:0 RPM PRN1:7184 RPM E0@:0 PRN1@:255
 E0:0 RPM PRN1:7184 RPM E0@:0 PRN1@:255

JeremiahGillis:
[...]

PR #4028, opened 2024-06-03
SelfTest: BUG-3480: Fix XL Fan Test Inconsistent

deniolan:
This PR addresses [BUG-3480](https://github.com/prusa3d/Prusa-Firmware-Buddy/issues/3480)

Testing results from many XL users in various forums have shown that depending on shroud and altitude, the XL print fan can reach speeds of up to 7200 RPM.

Previous Self Test code was problematic for fans, because switched fan detection did not allow any overlap in fan min/max ranges.   This pull request attempts to add the notion of valid ranges for print and heatbreak fans as well as the notion of a "minimum overlap" to combat this issue.   The reality is showing that the previous ranges were insufficient to reliably self-test both fans AND perform switched fan detection.

This pull request allows the fans to have ranges that slightly overlap, and will appropriately abandon switched fan detection only during ranges where there is a possibility both fans are operating in the same RPM range (which is possible).   Outside of those allowed overlaps - switched fan detection still works as expected - and XL users can stop removing their fan shrouds to pass the fan self check.

In the end, fans that are not operating at all are a bigger issue from a self test perspective - and the extruders come assembled so the likelihood of a switched fan is most possible during a service after a fan has been broken in - and both fans operating within the minimum overlap range are unlikely unless the overlap range is great.

For printers where switched fan detection should not have overlap (MK4/Mini/MK3x) this PR implements a 0 rpm overlap.

### DIFF
--- a/lib/WUI/link_content/basic_gets.cpp
+++ b/lib/WUI/link_content/basic_gets.cpp
@@ -431,9 +431,9 @@ json::JsonResult get_job_v1(size_t resume_point, json::JsonOutput &output) {
             // Note: This modifies the buffer, so it has to be done after all other
             // uses of it above!!
             dirname(sfn_path);
-            JSON_FIELD_STR("path", sfn_path) JSON_COMMA;
+            JSON_FIELD_STR("path", sfn_path);
             if (has_stat) {
-                JSON_FIELD_INT_G(has_stat, "size", st.st_size) JSON_COMMA;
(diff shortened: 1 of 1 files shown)


larrrox (approved):

larrrox:
oops sorry

crabdancing:
Our lab is operating at quite high altitude, and unfortunately, even with this commit the print fan RPM test still fails. :/

PR #4031, opened 2024-06-06
[BFW-5652] mmu3: skip unloading filament on early abort

### LINKED ISSUE #3955: [BFW-5652] Needless Filament Unload Procedure, when MMU Print is cancelled in the Warm Up Phase
Printer: MKS 3.5 + MMU3
Printer FW: 6.0.0
MMU FW 3.0.3

Scenario:

- Starting a MMU Print
- Cancel the print in the heating phase, before filament is loaded for the first time

Printer proceeds the filament unload procedure (heating up, ramming, ...) although no filament is loaded.
The procedure is needless in this case.

Best regards
Jochen

krx:
checks if filament is loaded before queuing M702

fixes #3955

### DIFF
--- a/src/common/marlin_server.cpp
+++ b/src/common/marlin_server.cpp
@@ -482,6 +482,9 @@ void print_fan_spd() {
 /// can do this since this will be only called at the end of the print or when aborting.
 /// So it shouldn't overwrite any important gcodes.
 void safely_unload_filament_from_nozzle_to_mmu() {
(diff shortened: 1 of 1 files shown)


DRracer (approved):
:thinking: this basically implies, that the filament sensor and FINDA can be relied on. 
Not sure if we want that in all cases (esp. when resolving error states), but when the print has been started already, both fsensor and FINDA probably did their job correctly.

Issue #4070, opened 2024-07-03
[BFW-5923] [ENHANCEMENT] Add option to disable serial printing screen


PinkPandaKatie:
### Printer model

All models

### Describe the enhancement

I use Octoprint solely as a means of external control. I have a USB number pad next to my printer which I have connected to a laptop, which runs a program that sends GCode commands via Octoprint when keys are pressed. This allows me to quickly set temperatures, move the head around, control the fans, etc.

The annoying thing is, though, any codes that move the head cause it to enter "serial printing" mode, blocking access to the UI. Worse, there have been a few times where I unloaded the filament but didn't pull it out of the sensor right away, then I moved the head up with the numpad, and pulling the filament out triggered a filament runout sequence that I could not cancel without resetting the printer.

### Expected functionality

Two possible solutions:

* Add a toggle setting in "User Interface" to enable/disable the serial printing screen. SerialPrinting::serial_command_hook can check this setting and do nothing if it is disabled.
* Add a GCode toggle that Octoprint could send which would indicate that following commands are not part of a print, but are just control moves. This would be more complex and require allocating and documenting the GCode as well as an enhancement to Octoprint.

I have implemented the first option myself and can submit a PR for it.

PinkPandaKatie:
PR is here: https://github.com/prusa3d/Prusa-Firmware-Buddy/pull/4071

I expected it to auto-link back but it didn't

danopernis:
Merged. This is not a bugfix, so I expect this to be released in the next non-patch version. Keeping the issue open to track progress.

PR #4071, opened 2024-07-03
[BFW-5923] Make serial printing screen a menu option (Implements #4070)

PinkPandaKatie:
This adds a toggle menu option under User Interface to enable/disable the serial printing screen.


danopernis (changes requested):
Hi @PinkPandaKatie thanks for the contribution.

Since we recently changed the master branch, I have taken a liberty of rebasing your PR and resolving conflicts.

Overall I quite like the change and find it useful for expert users. IIRC the serial printing screen exists to prevent interaction with the printer via the LCD. Interactions may lead to failed prints, akin to inserting random G-codes to the instruction stream. I would be happy to merge this if you moved the menu item to the "Experimental Settings" menu which is more difficult for regular users to access. This way I wouldn't need to go through the product department to get their approval. What do you think?

Rant: if only there was a "Start printing" and "End printing" G-code which would enable us to clearly differentiate between various printer states instead of using heuristics like "Is it a movement G-code?"

danopernis (approved):
Looks good to me.

Regarding your question about development (green) menu items: no, you can't enable them on the production firmware. There is a CMake option DEVELOPMENT_ITEMS_ENABLED which translates to a compile-time constant GuiDefaults::ShowDevelopmentTools which in turn controls if the menu item is displayed.

Issue #4081, opened 2024-07-06
[ENHANCEMENT]  INCREASE THE FIRST LAYER CALIBRATION RANGE TO -3.000


antimix:
### Printer model

MK3.5 REVO

### Describe the enhancement

Currently the distance between the nozzle and the bed during the first layer calibration has a range of +2.000  <=< -2.000

### Expected functionality

With a E3D REVO nozzle , some time, the actual range may not be enough after installing a new nozzle.
It would be better to have the possibility to have +2.000 <=< -3.000

jlandegent:
i had the same issue. simple solution would be to place the pindaprobe further away from the bed. i just used a thicker tiewrap for calibration.

bryn51:
This looks like something that a user could handle themselves by adjusting the printer physical settings to cope with the fact they chose to deploy a different hotend than the standard.
I vote this enancement in the negative.

Levi--G:
This is definitely not just a custom hotend issue, i run into the same issue on a stock mini on the textured PEI sheet. I need -2.010 but it stops at -2. Its been this way since it came pre-assembled. So definitely not an end user issue!

danopernis:
Basically what @jlandegent said :+1: 

@Levi--G on MINI, you can follow [this guide](https://help.prusa3d.com/guide/height-adjustment-of-m-i-n-d-a-superpinda-sensor-mini-_154547) to adjust probe height.

@antimix I don't know how the REVO hotend looks like, but it seems to me that you can follow similar process on MK3.5

If you have any further questions, [feel free to contact our support](https://help.prusa3d.com/article/customer-support_2287). We are not planning to increase the live Z adjust range, 4 mm should already give you enough leeway. This is the same range as the 8-bit firmware for MK3 has.

Levi--G:
> Basically what @jlandegent said 👍
> 
> @Levi--G on MINI, you can follow [this guide](https://help.prusa3d.com/guide/height-adjustment-of-m-i-n-d-a-superpinda-sensor-mini-_154547) to adjust probe height.
> 
> @antimix I don't know how the REVO hotend looks like, but it seems to me that you can follow similar process on MK3.5
> 
> If you have any further questions, [feel free to contact our support](https://help.prusa3d.com/article/customer-support_2287). We are not planning to increase the live Z adjust range, 4 mm should already give you enough leeway. This is the same range as the 8-bit firmware for MK3 has.

[...]

PR #4087, opened 2024-07-17
zaxis:Decrease first layer calibration live-z from -2000 to -3000

bkerler:
This fixes issue #4081. Allows the value for the live z-axis calibration to be decreased to -3000 instead of just -2000 in order to allow to setup third-party extruders like the Revo 3D on MK 3.5.

### DIFF
--- a/src/gui/footer/footer_item_live_z.cpp
+++ b/src/gui/footer/footer_item_live_z.cpp
@@ -19,7 +19,7 @@ int FooterItemLiveZ::static_readValue() {
 
 string_view_utf8 FooterItemLiveZ::static_makeView(int value) {
     static std::array<char, 7> buff; //"-2" - "0", longest "-0.001"
-    int value_clamped = std::clamp(value, -2000, 0);
+    int value_clamped = std::clamp(value, -3000, 0);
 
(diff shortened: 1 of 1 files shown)


danopernis (changes requested):
I rebased your PR onto current master since we recently released 6.1.0

I don't think this change actually solves the linked issue though. This seems to only address the value displayed in the footer. The actual limit is somewhere else in the code. Arguably, this should be replaced by some named constant which should be shared by this code and the code which actually handles live Z adjustment.

Even still, we are not going to merge this change. +-2mm is 4mm total which is plenty of leeway. I doubt anybody uses sheets as thick or frame as skewed. PINDA position is adjustable on stock MK3.5 hotend. I have never seen any REVO so I can't comment on that.

antimix:
@danopernis from your comments I understand that the first layer calibration code was written in a rather complex manner, and finding and changing the lower limit, may require too much efforts and can introduce bugs. So it is easier to physically raise the SUPERPINDA from the bed 1 mm more, and the calibration limit should be enough.
I have already used that trick on my MK3.5 and I was able to calibrate the first layer as I wanted.

bkerler:
Ok, so I assume I can close this PR then, unless it is really required, then it does need more work as @danopernis already stated, then I'll reopen this issue again..

Issue #4042, opened 2024-06-13
[BFW-5918] [ENHANCEMENT] Don't remove trailing zeroes of the Footer's axis value for static footer appearance


puhitaku:
### Printer model

All

### Describe the enhancement

Hi, thanks for the cool firmware!

The current implementation of footer items of [axis position](https://github.com/prusa3d/Prusa-Firmware-Buddy/blob/3e63d0e8ec1dcefeac7e128c198538586d83f01a/src/gui/footer/footer_item_axis.hpp#L36-L38), [gcode position](https://github.com/prusa3d/Prusa-Firmware-Buddy/blob/3e63d0e8ec1dcefeac7e128c198538586d83f01a/src/gui/footer/footer_item_axis.hpp#L72-L74), and [live z](https://github.com/prusa3d/Prusa-Firmware-Buddy/blob/3e63d0e8ec1dcefeac7e128c198538586d83f01a/src/gui/footer/footer_item_live_z.cpp#L29-L32) removes trailing zero. This makes sense, for aesthetic reasons, but in reality this makes the footer items wobble around when the value is like `2.50` that will actually be `2.5` on the screen.

I personally prefer NOT removing zero since not removing it will result in constant displayed width of footer items and constant positioning. I'm ready to contribute when this idea is accepted. Thank you!

danopernis:
Hi @puhitaku thanks for the generous offer. I talked to the product department and they like your proposal.

Before implementing it, please wait for the next minor release. The footer was recently refactored in our private fork so integrating your changes if you start developing now would involve solving a lot of git conflicts.

puhitaku:
Thanks @danopernis. Then, I'll begin watching this repository for the upcoming release. Please ping me if I seem to forget here, just in case.

puhitaku:
@danopernis Apologies for taking your time, but if I can know the estimated release date of the next minor release, I'd be happy

danopernis:
@puhitaku I am expecting several weeks, but don't quote me on that :slightly_smiling_face: it wouldn't be the first delay I've seen in my software engineering career.

Perhaps the merge conflict wouldn't be as problematic as I was originally imagining it, so if you want to, go ahead :+1:

puhitaku:
I've filed a PR for this issue (and I've invalidated the warranty of my MK4 to see if this change works haha 😁)
Looking forward to seeing the upcoming review and merger!

danopernis:
[...]

PR #4129, opened 2024-08-08
[BFW-5918] gui: don't remove trailing zeroes of the Footer's axis value

puhitaku:
This PR implements the enhancement I proposed in #4042 . I confirmed it works as expected on my Prusa MK4.
I set `master` as the base branch since other PRs do this. Please switch this if necessary. Thank you!

# Changes
- `src/gui/footer/footer_item_axis.hpp`, `footer_item_live_z.cpp`: else-if blocks are removed to stop removing trailing zeroes after the decimal point.

# Fixes
- Added `setuptools` to the Holly's dependency
   - Since Python 3.12.3, venv won't pre-install setuptools
   - I used Python 3.12.3 to build the firmware and suffered an error due to the lack of it


danopernis (approved):
LGTM. I rebased it onto current master branch, turns out we already have the setuptools patch.

PR #4136, opened 2024-08-15
[BFW-5927] Removed "Plenk"

foorschtbar:
Spaces/Blanks ("[Plank](https://de.m.wikipedia.org/wiki/Plenk)") before punctuation marks are grammatically incorrect in German


danopernis (approved):
Thanks @foorschtbar

Issue #4117, opened 2024-08-04
[BUG] really implement M104.1


Ro3Deee:
### Printer model

XL

### Firmware version

6.0.4

### Describe the bug

The current implementation of M104.1 is to ignore the P and Q params ( time in seconds till the temperature S is required)  simply process this command as if it was standard M104. 

M104.1 GCode was introduced to Prusa-Firmware-Buddy by this commit:
https://github.com/prusa3d/Prusa-Firmware-Buddy/commit/652a65f8c176db7069917d13bb51dab6b40769ae
to early set hotend temperature so that the printer can start preheating the tool in advance.

PrusaSlicer started to use M104.1 in place of M104 with this commit: 
https://github.com/prusa3d/PrusaSlicer/commit/c916249f2b0e2d9b50077cf3c2402115426ef4f4

OrcaSlicer started to use M104.1 with this commit:
https://github.com/SoftFever/OrcaSlicer/commit/7d0fb4f42a3f6a1f43503a601491071b8e29052c
and issues it even on current tool to set a delayed future temp. 

As the current implementation is to simply process this command as if it was standard M104, M104.1 is wrongly used. It should be implemented because it is usefull to preheat future tools so the printer doesn't wait to heat the tool when picking it up.

CZDanol:
I believe the current implementation of M104.1 is not problematic. In #4115, you mentioned that "So the fw should look ahead for  the relevant M109", but that is not economical to do on the printer - and that is why we asked the Slicer guys to do this for us and emit it to the gcode through M104.1.

CZDanol:
Okay, that is a bug - and that seems to be covered by #4115

I still don't see a point to the current issue, so unless you convince me otherwise, I'd close this issue.

CZDanol:
The first three paragraphs - that we are already doing. I still don't understand your point.

And as I said before, the forward gcode scanning is not economical, that is why we have the M104.1 emitted by the Slicer.

CZDanol:
It is not a problem that those parameters are ignored. Slicer starts emitting those gcodes only a few minutes in advance. It is equivalent to a condition `P < 300` (for example)

CZDanol:
The #4115 is a fair report and we should fix it, yeah

Ro3Deee:
M104.1 , used by prusa slicer for XL, should be implemented

Issue #4116, opened 2024-08-04
[BFW-6887] [BUG] M104.1 immediately  sets the temp of future M104 


Ro3Deee:
### Printer model

XL 

### Firmware version

6.0.4

### Printing from...

USB

### Describe the bug

In my example, I have two tools (  T0 and T1). T1 is purged first, T0 second.  A very strange effect (probably by fw planner) I found out (watching the display of my Prusa XL) that issuing a M104.1 on current tool has the effect of immediately setting the tool to the temperature that will be set in the future (but not the one given as its S parameter). In the picture bellow (also video below), M104.1 at line 489 (sets a temp of 225) has the effect of immediately setting T1 temperature to 199, as set by M104 at line 500:
![image](https://github.com/user-attachments/assets/4e4ca1b5-e59c-447a-b833-bdd142746467)

### Video
https://youtu.be/8H7SEEuuDPg?feature=shared 
.

### How to reproduce

```
;
; purge second tool
;
G1 F24000
P0 S1 L2 D0; park the tool
M109 T1 S230
T1 S1 L0 D0; pick the tool
G92 E0 ; reset extruder position

G0 X150 Y-7 Z10 F24000 ; move close to the sheet's edge
M104.1 T1 P36 S225
G0 E10 X140 Z0.2 F50 ; purge while moving towards the sheet
G0 X110 E9 F80 ; continue purging and wipe the nozzle
M73 P28 R23
G0 X107 Z0.05 F80 ; wipe, move close to the bed
G0 X104 Z0.2 F80 ; wipe, move quickly away from the bed
M73 P26 R23
G1 E-1.2 F2400 ; retract
 ; update slicer internal retract variable
G92 E0 ; reset extruder position

M104 S199 T1
```

from the attached g-code, sets the temp to 199 before purging the tool. See the video also

### Expected behavior

the temp should be set to 199 after purging

### Files

G-Code:
[5cube_0.2mm_PLA_32m17s_noHeatQforT1.zip](https://github.com/user-attachments/files/16488999/5cube_0.2mm_PLA_32m17s_noHeatQforT1.zip)

bkerler:
Please don't reopen the same issue. Please close this issue in favor of #4115.

bkerler:
@petrubecheru Thanks for pointing that out, however for me it seems to be the same issue, as P and Q aren't implemented and thus no timing can be handled. It would be better if you point out all three relations in one issue instead.

[...]

Issue #4115, opened 2024-08-04
[BFW-6887] [BUG] M104.1 should not default to M104 for current tool


Ro3Deee:
### Printer model

XL 

### Firmware version

6.0.4

### Describe the bug

M104.1 GCode was introduced to Prusa-Firmware-Buddy by this commit:
https://github.com/prusa3d/Prusa-Firmware-Buddy/commit/652a65f8c176db7069917d13bb51dab6b40769ae
to early set hotend temperature so that the printer can start preheating the tool in advance.

PrusaSlicer started to use M104.1 in place of M104 with this commit: 
https://github.com/prusa3d/PrusaSlicer/commit/c916249f2b0e2d9b50077cf3c2402115426ef4f4

But the current implementation of M104.1 is to ignore the P and Q params ( time in seconds till the temperature S is required)  simply process this command as if it was standard M104. 

Because an M104 command on current tool immediately starts to modify the current temperature, an  M104.1 (that is M104 in fact) on current tool can destroy a print if used with P and Q params in mind.

 So, until m104.1 is properly implemented as requested here #4117 , it shouldn't do anything on current tool.

bkerler:
As the source code already states [here](https://github.com/prusa3d/Prusa-Firmware-Buddy/blob/f5a498ab8d2a42341d0dbeb969b7ae047783e860/src/marlin_stubs/M104_1.cpp#L16-L24), the P and Q parameters are planned to be developed already and meanwhile, if preheating is required, M104 is being run instead. So it isn't implemented right now and thus no bug.

bkerler:
@petrubecheru yes, I agree. Imho as of now, this issue has to be handled within PrusaSlicer. The changes within the printer firmware seem to be pretty huge to match the expected behaviour vs. the current behaviour.

CZDanol:
Is there any example where this actually mattered?

CZDanol:
Got an answer to my previous question here:  https://github.com/prusa3d/Prusa-Firmware-Buddy/issues/4117#issuecomment-2709801423
Internal ticket: BFW-6887

PR #4143, opened 2024-08-19
gcode: M104.1, add support for P and Q options

bkerler:
This commit makes sure that M104 isn't run immediately on preheating, but in fact given options P and Q are referred to for adding a dwell and waiting for heating up instead.

Refers to issue #4115 (issue #4116 / issue #4117)

### DIFF
--- a/src/marlin_stubs/M104_1.cpp
+++ b/src/marlin_stubs/M104_1.cpp
@@ -5,6 +5,7 @@
 #include <marlin_vars.hpp>
 #include <Marlin/src/gcode/parser.h>
 #include <gcode/gcode.h>
+#include "../../module/planner.h"
 
 #ifdef PRINT_CHECKING_Q_CMDS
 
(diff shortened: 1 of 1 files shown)


danopernis (commented):
You definitely do not want to dwell here. That is incorrect implementation and is essentially equal to combo M104+G4 (or M109 for that matter)

bkerler:
Ok, so I assume it would require a timer then instead of dwelling ? Wasn't sure if running it in parallel, dwelling would have the same effect in fact as using a timer.

bkerler:
> PrusaSlicer puts _both_ P and Q params in G-Code: `P` - - time in seconds till the temperature S is required (in standard mode) `Q` - - time in seconds till the temperature S is required (in stealth mode)
> 
> So the PR is incomplete. My proposal pseudo code:
> 
> ```
>       millis_t dwell_ms = 0;
>         if (parser.seen('Q')) {//default Q because is longer
>             dwell_ms = parser.ulongval('Q');
>         } 
>        if ( !in_stealth_mode && parser.seen('P')) {//if _printer NOT in stealth_ and has P param
>             dwell_ms = parser.ulongval('P');
>         }
> ```

Actually should_start_preheating does mean that P is only used when not in stealth mode and Q if in stealth mode, so it wouldn't make sense to add that additionally.

bkerler:
ok, the problem is that P and Q mean that the actual temperature should be reached **until** the period ends. So neither timer nor dwelling does make sense. It would require that the temperature is constantly monitored by using a separate thread. Not sure if there is any easy way to implement this. Closing for now until I come up with a PR that could be implemented properly.

Issue #4123, opened 2024-08-05
[BUG] MK4 with Octoprint - M486 A<long name> crashes the printer


DerMouse:
### Printer model

MK4

### Firmware version

6.0.1=14848

### Printing from...

USB/Octoprint

### Describe the bug

PruscaSlicer writes M486 A<Object name> into the gcode file. If <object name> is longer thant 83 characters, sending the file to print *through Octoprint USB connection* will crash/reboot the printer. No problem if same file printed through regular USB key.

### How to reproduce

Slice a (very long named) STL file. Send it to MK4 through Octoprint. No need to have any real printing commands in the Gcode file.

### Expected behavior

Should not crash anything.

DerMouse:
Cannot upload a file, so here is some Gcode that triggers the problem.

; length 80
M486 A01234567890123456789012345678901234567890123456789012345678901234567890123456789
; length 81
M486 A012345678901234567890123456789012345678901234567890123456789012345678901234567890
; length 82
M486 A0123456789012345678901234567890123456789012345678901234567890123456789012345678901
; length 83
M486 A01234567890123456789012345678901234567890123456789012345678901234567890123456789012
; length 84
M486 A012345678901234567890123456789012345678901234567890123456789012345678901234567890123
; length 85
M486 A0123456789012345678901234567890123456789012345678901234567890123456789012345678901234
; length 86
M486 A01234567890123456789012345678901234567890123456789012345678901234567890123456789012345

danopernis:
@DerMouse thanks for the report. I just tried this and wasn't able to reproduce it. I don't know much about octoprint, I just followed [this guide](https://hub.docker.com/r/octoprint/octoprint). I also tried reproducing this via serial port terminal and also didn't get any crash.

bkerler:
I also can't reproduce any "crash" related to gcode "M486 A" so far. @DerMouse did you see a BSOD or Red screen when the crash happened  ? And if yes, can you please take a picture of the screen ? or was the serial port just not responsive ?

DerMouse:
Sorry for the late answer.
If I remember proberly (not close to the printer yet), it was a red screen with (maybe) some kind of QR code. I'll check this afternoon and will try to take a picture.

DerMouse:
So, I did the test again.
As soon as I ask Octoprint to print the file, the printer reboots and, after reboot, displays a red screen :

[...]

PR #4168, opened 2024-08-29
serial: Indicate gcode cmd length longer than allowed.

bkerler:
Refers to issue #4123.

If a gcode command being sent via serial is longer than the MAX_CMD_SIZE and the gcode command does contain a checksum, the checksum given in the command may no longer be part of the command buffer and thus leads to a checksum error although the checksum is valid.

This code adds an error if the command length exceeds the cmd buffer size and thus ensures that the checksum no longer is validated.


CZDanol (dismissed):

o-vesely:
Closing this, took inspiration from this and implemented this. Will be in the next realease, thanks 🚀

Issue #3936, opened 2024-04-20
[BUG] Material in Footer of MK4 not updated during multi-material print with MMU3


unclej84:
**Printer type** - [MK4]

**Printer firmware version** - [6.0.0]

**Original or Custom firmware** - [Original]

**Optional upgrades** - [MMU3]

**USB drive or USB/Octoprint**
USB drive

**Describe the bug**
During a multi-material print with the MMU3 the material in the footer does not get updated according to currently used material although it was correctly defined by advanced loading

**How to reproduce**
Define multi-material print for MK4 with MMU3 using PETG in slot 1 and PLA in slot 2 - footer always stays at the same/first material

**Expected behavior**
[top-bolt-x-2_0.4n_0.2mm_PETG,PLA_MK4ISMMU3_1h16m.zip](https://github.com/prusa3d/Prusa-Firmware-Buddy/files/15050104/top-bolt-x-2_0.4n_0.2mm_PETG.PLA_MK4ISMMU3_1h16m.zip)

Display currently used material - when slot 1 is in use, display "PETG" - when slot 2 is in use, display "PLA"

**G-code**
[top-bolt-x-2_0.4n_0.2mm_PETG,PLA_MK4ISMMU3_1h16m.zip](https://github.com/prusa3d/Prusa-Firmware-Buddy/files/15050104/top-bolt-x-2_0.4n_0.2mm_PETG.PLA_MK4ISMMU3_1h16m.zip) - contains 3mf project and sliced bgcode file

VladaDHS:
Have same issue. all Fialments are PLA but down is shown PETG.

seifer69er:
After some more testing i think I know what the issue is,
When loading each filament into the MMU3, there is no filament type option on the MK4, so it relies on what it thinks is in that slot,
For me, slot 1 was PLA, 2 was  ---, 3 ----, 4 ---- and 5----
so when i use prusaslicer  ive told the slicer what tool is which filament.
the issue is the filament type not displaying correctly is that the Mk4 relies on EEPROM, for this data and not prusa connect

I did manage to select the correct filament type on the device, and it managed to load just fine, but otherwise its a fun bug to debug :)

unclej84:
The filament type selection is available with advanced loading via the menu - I used this and specified the material correctly, so the MMU/MK4 knew the type but still showed the wrong type

unclej84:
Is there any update from Prusa side?

danopernis:
[...]

PR #4169, opened 2024-08-30
footer: Fix display of filament type footer for mmu2

bkerler:
This aims to resolve issue #3936.

As for toolchanger/printers in general, the active_extruder does point to the tool hotend being actively used. However for mmu2, the active_extruder stays 0 all the time. So we need to take special care for mmu2 and make sure that it shows the current tool filament type instead.

BFW-6014

### DIFF
--- a/src/gui/footer/footer_item_filament.cpp
+++ b/src/gui/footer/footer_item_filament.cpp
@@ -7,7 +7,10 @@
 #include "img_resources.hpp"
 #include "filament.hpp"
 #include <config_store/store_instance.hpp>
-
+#include <option/has_mmu2.h>
+#if HAS_MMU2()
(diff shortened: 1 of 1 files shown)


CZDanol:
Internal ticket: BFW-6014

CZDanol:
The code contains a bug (and merge conflicts with out private branch - but that's on us): `MMU2::mmu2.get_current_tool()` returns -1 when no filament is in the nozzle, which was causing out of range.  Going another round with the author of the PR would take way more effort than if I just fix it myself.

So, this PR cannot be merged in it's current state due to the bug. I will be merging the fixed commit into our private repository and mark @bkerler as an author, so the fix will get to the public repo eventually through our standard development process. Closing this PR.

bkerler:
@CZDanol Great, thx !

PR #4170, opened 2024-08-30
pause: Make sure that the filament type is updated on mmu_change as well

bkerler:
This change does make sure that not only on CommonLoadType::mmu the filament type is getting updated, but instead as well on state CommonLoadType::mmu_change, which is called by Pause::loop_load_mmu_change in Pause::FilamentChange.

BFW-6015


CZDanol:
Internal ticket: BFW-6015

bkerler:
Closing as it seems it was integrated into v6.3 in a different way :)

Issue #3049, opened 2023-04-16
MK4 PrusaLink extremely slow upload over Wi-Fi


H0nzik1:
**Printer type**
MK4

**Printer firmware version** 
4.5.4

**Original or Custom firmware** - [Original/Custom]
Original 

**Optional upgrades** 
Nothing

**USB drive or USB/Octoprint**
 USB drive (PrusaLink)

**Describe the bug**
Upload speed over PrusaSlicer or Web interface to the MK4 printer is like 0.1MB/s max. So 10MB file takes forever to upload. MK3S+ with RPi Zero with PrusaLink was like 10x faster.

**How to reproduce**
  Just upload file to MK4 printer. My Wi-Fi router Asus AX68U is 2 meters from printer only.

**Expected behavior**
I expect at least 1MB/s upload speed like with MK3S+ with RPi Zero

**G-code**
  Any
**Crash dump file**
  does not exist any

**Video**

arekm:
(same for wired ethernet - 0.3MB/s)

H0nzik1:
> (same for wired ethernet - 0.3MB/s)

OMG. I have tried also different USB drive which can do 100MB/s write speed, but it did not helped either.

arekm:
For wifi:

https://github.com/prusa3d/Prusa-Firmware-Buddy/blob/e688bdf53b0296267a67c31f21960c937193d89d/lib/WUI/espif.cpp#L71-L72
comment suggests that maximal possible wifi speed is about 550KB/s (4.5Mbps) and it's not done because of some corruption.

For wired ethernet the driver copies between buffers:
https://github.com/prusa3d/Prusa-Firmware-Buddy/blob/e688bdf53b0296267a67c31f21960c937193d89d/lib/WUI/ethernetif.c#L241
which means it is not a zero-copy driver (like modernly written network drivers are, even for STM32 - https://github.com/lsgunth/lwip_contrib). Copying buffers will always be slow but no idea if that's the only reason for bad 300KB/s wired ethernet speed.

Fail-Fast-V0:
My upload speeds are unusable. I noticed the same problem with my Prusa Mini (also running a buddy board).

H0nzik1:
Downloading gcode via browser is also insanely slow.
![obrazek](https://user-images.githubusercontent.com/28353840/234519600-d590f449-45b9-46fd-af11-b606db51bbe1.png)

mrlinuxfish:
[...]

PR #4171, opened 2024-08-30
network: Network auto-negotiation duplex fix for LAN8742

bkerler:
On auto-negotiation using the external network phy LAN8742, the reference stm32f427 source code incorrectly uses the wrong PHY_SR (status register) addr of 0x10, leading to a false detection of the duplex mode, thus using half-duplex instead of duplex mode, resulting in retransmissions and overall lower throughput of data.

According to the [LAN8742a datasheet](https://ww1.microchip.com/downloads/en/DeviceDoc/8742a.pdf), page 89, "4.2.22 PHY Special Control/Status Register" the addr is defined as 31 (0x1E), with duplex/half-duplex speed indicator given in bit 4..2.

This patch fixes the duplex mode for the ethernetif.c ... it might be necessary to patch the legacy HAL_ETH_Init as well within the lib/Drivers/STM32F4xx_HAL_Driver/Src/Legacy/stm32f4xx_hal_eth.c.

See further details [here](https://community.st.com/t5/stm32-mcus-embedded-software/stm32f7-lwip-tcp-dupacks-and-delayed-retransmit-w-source-code/td-p/66204) as well for reference.

Somewhat related to issue #3049 (not for wifi, but ethernet part)

### DIFF
--- a/lib/WUI/ethernetif.c
+++ b/lib/WUI/ethernetif.c
@@ -16,6 +16,10 @@
 #include <stdbool.h>
 #include "buddy/priorities_config.h"
 
+#define PHY_REGNEW_SSR           (uint16_t)31 /* PHY Special Status Register */
+#define PHY_REGNEW_SSR_SPD_100M  ((uint32_t)1 << 3) /* Speed Indication, bits 4:2 */
+#define PHY_REGNEW_SSR_DUPL_FULL ((uint32_t)1 << 4) /* Speed Indication, bits 4:2 */
+
 /* The time to block waiting for input. */
 #define TIME_WAITING_FOR_INPUT (portMAX_DELAY)
 /* Stack size of the interface thread */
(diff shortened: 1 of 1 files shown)


bkerler:
Some statistics : 
* 6.1.3 without patch:
image
image
85 seconds for 16.9 MB PrusaLink gcode transfer

* 6.1.3 with network auto-negotiation patch:
image
image
70 seconds for 16.9 MB PrusaLink gcode transfer

bkerler:
@ivous Can you please elaborate which dhcp issues you mean ? I'm using the patch without any issues and can confirm the ethernet connection does work in full-duplex and dhcp works as well.

bkerler:
This PR has a followup in #4342. As @ivous stated, there were issues that needed to be addressed, thus the followup PR with the upgraded STMF4 Hal as recommended by @ivous :)

Issue #4162, opened 2024-08-28
[BUG] Lines that include # are treated as comments even when using quoted INI strings


Swizzler121:
### Printer model

MK3.5

### Firmware version

6.1.2

### Upgrades and modifications

None

### Printing from...

N/A

### Describe the bug

When filling out the prusa_printer_settings.ini if you are using a password (or any other key I assume) that begins or includes a # symbol, it's treated as a comment, even when in a quoted string (Feature added in latest update #3158).

Example config:
```
[wifi]
ssid = example ssid
psk = "#examplepsk"
```

I first noticed the issue in a previous firmware but had an alternative network at the time with a different psk that avoided the issue. When I saw the above mentioned feature I decided to try again, just to find this is still an issue.

This firmware update does include a way to manually enter the password using the screen, and using that you are then able to join networks that include a # in the login details, but compatibility with the prusa_printer_settings.ini file would be preferable.

### How to reproduce

Join a network with a PSK that includes a # sign by loading the prusa_printer_settings.ini file

### Expected behavior

It successfully loads the correct settings from the prusa_printer_settings.ini file and joins the wifi network successfully.

danopernis:
Yeah, I agree with you, this is a bug.

The thing is, the INI parser is based on https://github.com/benhoyt/inih and that is quite difficult to modify. Proper solution to solving this bug would most likely involve complete rewrite of the INI parser and that will take a lot of time, which would be better spent on solving other bugs and creating new features. Therefore, I don't think we will get to work on this anytime soon. If anybody wants to contribute proper parser based on proper tokenization and formal grammar, I would be happy to discuss details.

In the meantime, as a workaround, please use other means of entering the password, like the Prusa App or on-screen keyboard.

bkerler:
After debugging the issue, it seems the root cause is that inline comments are allowed. The issue is caused by this line: 
https://github.com/prusa3d/Prusa-Firmware-Buddy/blob/f5a498ab8d2a42341d0dbeb969b7ae047783e860/lib/inih/ini.c#L74

Imho, this doesn't make a lot of sense to allow inline comments in ini files and the easiest way to fix this is to just undefine INI_ALLOW_INLINE_COMMENTS [here](https://github.com/prusa3d/Prusa-Firmware-Buddy/blob/f5a498ab8d2a42341d0dbeb969b7ae047783e860/lib/inih/ini.h#L95). What do you think @danopernis ?

[...]

PR #4175, opened 2024-09-01
ini_parser: Remove inline comment parsing on ini handler for values

bkerler:
On rare occations when parsing the prusa_printer_settings.ini, a value after = is being interpreted as inline comment, although it shouldn't, leading to an invalid null termination of a value value string.

For example:
`field = #value`

The parser does only treat an inline comment as comment if a whitespace is followed by a hashtag char. In this case, field and value aren't treated as one, but as two lines. it is setting a nulltermination at "=" for parsing the field and then parses the value as " #value", assuming that an inline comment has occured.

By removing this inline comment check when parsing the value fields, we can rule out this problem.

Related to issue #4162, can relate to #2325 as well.

### DIFF
--- a/lib/inih/ini.c
+++ b/lib/inih/ini.c
@@ -199,11 +199,6 @@ int ini_parse_stream(ini_reader reader, void* stream, ini_handler handler,
                 *end = '\0';
                 name = rstrip(start);
                 value = end + 1;
-#if INI_ALLOW_INLINE_COMMENTS
-                end = find_chars_or_comment(value, NULL);
-                if (*end)
-                    *end = '\0';
-#endif
                 value = lskip(value);
                 rstrip(value);
                 value = unquote(value);


lib/inih/ini.c
```diff
@@ -199,11 +199,6 @@ int ini_parse_stream(ini_reader reader, void* stream, ini_handler handler,
                 *end = '\0';
                 name = rstrip(start);
                 value = end + 1;
-#if INI_ALLOW_INLINE_COMMENTS
```
nmschulte, line 202: https://github.com/prusa3d/Prusa-Firmware-Buddy/blob/master/lib/inih/README.md#compile-time-options

> **Inline comments**: By default, inih allows inline comments with the `;` character. To disable, add `-DINI_ALLOW_INLINE_COMMENTS=0`. You can also specify which character(s) start an inline comment using `INI_INLINE_COMMENT_PREFIXES`.
> 
> **Start-of-line comments**: By default, inih allows both `;` and `#` to start a comment at the beginning of a line. You can override this by changing `INI_START_COMMENT_PREFIXES`.

Did you consider simply tweaking either or both of these options, vs eliding this single block of code?

bkerler: We do not want to remove inline comments. Instead we just want to remove inline comments where inline comments shouldn't happen : On parsing values. That's why only this define is removed, but INI_ALLOW_INLINE_COMMENTS stays defined.

nmschulte: Ah okay, sounds like a bug in [inih](https://github.com/benhoyt/inih) then?  Or, is this an effect of the quoted string modification?  I think it's not given the comments from the issue this addresses but unsure.

Perhaps this is fixed in more recent inih?  https://github.com/benhoyt/inih

nmschulte (approved):

lib/inih/ini.c
bkerler: Sadly, it isn't fixed in the latest master, it's an issue by design of the parser. It would require to be rewritten and it would add more complexity. However this is only a temporary fix to solve current user issues until the inih is completely removed and replaced by a different parser library.

Issue #3155, opened 2023-07-22
[BUG] MK4 Network UI shows LAN as `[stat]` instead of `[static]` when the menu is re-entered


gudnimg:
**Printer type** - MK4
**Printer firmware version** - 4.7.1
**Original or Custom firmware** - Original
**USB drive or USB/Octoprint** - N/A
**Describe the bug**
Missing character in the LAN type setting.

**How to reproduce**

1. Go to Network settings
2. Go to Wifi settings
3. Change LAN from DHCP to Static --> Note the UI shows `[static]`
5. Leave Wifi settings menu
6. Enter Wifi settings menu and observe UI bug --> UI shows `[stat]`

![IMG_3748](https://github.com/prusa3d/Prusa-Firmware-Buddy/assets/8218499/d7b58583-0025-4dc1-baa0-b9ee72d722c4)

**Expected behavior**
  UI should show `[static]` always instead of `[stat]`

BastionNtB:
I'm seeing this on 5.0.0.

Also can't seem to get a static assignment to work either.

gudnimg:
Just checked using the latest MK4 firmware 6.0.1, the bug is still there:

![IMG_5206](https://github.com/prusa3d/Prusa-Firmware-Buddy/assets/8218499/219d78ec-335a-4a31-b2af-b994012539d1)

gudnimg:
Likely still an issue, I will wait for FW 6.1 for MK4 and look into submitting a PR. The fix is trivial if I remember correctly.

danopernis:
@gudnimg yep, still there, even in 6.1.x series :slightly_frowning_face:

bkerler:
I just had a look at the bug isn't "trivial" to fix at first sight. It is related to the refresh of the WI_SWITCH_t display class as it seems : 
https://github.com/prusa3d/Prusa-Firmware-Buddy/blob/f5a498ab8d2a42341d0dbeb969b7ae047783e860/src/gui/MItem_network.cpp#L176
I assume that the length of "DHCP" is used being the first option in the list although the length of "static" should be used.

gudnimg:
If I recall correctly, then doing something similar as here:

```cpp
MI_NET_INTERFACE_t::MI_NET_INTERFACE_t()
[...]

PR #4176, opened 2024-09-01
[BFW-6223] gui: Fix invalid display of "static" due to WI_SWITCH initialization in MI_NET_IP

bkerler:
This resolves issue #3155.

When setting the network ip address from "DHCP" to "static", on next reentry of Ethernet Settings -> LAN, the text "stat" is only being displayed. This is due to the default index on WI_SWITCH_t being 0 (DHCP). On setting the index, it should normally recalculate the string length to be 6 instead of 4, but due to the index already being set to 1 changeExtentionWidth is never run.

The problem is within a nasty bug as WI_SWITCH index should never be exposed public and MI_NET_IP didn't define index, thus leading to WI_SWITCH index being used as a global and not as a local variable. By defining a private index variable in WI_SWITCH, this problem is resolved.

!! **I strongly recommend to replace the public "index" variable of WI_SWITCH to be defined as private as there is already a function to set the index called "SetIndex" in WI_SWITCH to prevent similar issues in the future** !!

### DIFF
--- a/src/gui/MItem_network.cpp
+++ b/src/gui/MItem_network.cpp
@@ -176,12 +176,12 @@ MI_NET_IP::MI_NET_IP(NetDeviceID device_id)
     : WI_SWITCH_t(0, string_view_utf8::MakeCPUFLASH(label), nullptr, is_enabled_t::yes, is_hidden_t::no, string_view_utf8::MakeCPUFLASH(str_DHCP), string_view_utf8::MakeCPUFLASH(str_static))
     , device_id(device_id) //
 {
-    index = netdev_get_ip_obtained_type(this->device_id());
+    this->SetIndex(netdev_get_ip_obtained_type(this->device_id()));
 }
 
 void MI_NET_IP::OnChange([[maybe_unused]] size_t old_index) {
     const auto dev_id = device_id();
-    if (index == NETDEV_STATIC) {
+    if (this->GetIndex() == NETDEV_STATIC) {
         netdev_set_static(dev_id);
     } else {
         netdev_set_dhcp(dev_id);


CZDanol (changes requested):
This looks suspicious.
* Seems like you're shadowing the `ındex` for unknown reason

src/gui/MItem_network.cpp
```diff
@@ -177,6 +177,11 @@ MI_NET_IP::MI_NET_IP(NetDeviceID device_id)
     , device_id(device_id) //
 {
     index = netdev_get_ip_obtained_type(this->device_id());
+    if (index == NETDEV_STATIC) {
```
CZDanol, line 180: Setting `index` should be enough for the menu item to display the value correctly. If it doesn't we need to investigate that.

src/gui/MItem_network.cpp
```diff
@@ -177,6 +177,11 @@ MI_NET_IP::MI_NET_IP(NetDeviceID device_id)
...
+        this->SetIndex(1);
```
CZDanol, line 181: You're using 5 lines of code for something that could be solved by a ternary operator. Or even no ternary operator in this case possibly, bool implicitly casts to int.

src/gui/MItem_network.cpp
bkerler: In this case, yes, as NETDEV_STATIC is Index 1, as long as index isn't this->index. Will try if it does work with just setting index.

CZDanol:
Internal ticket: BFW-6223

CZDanol (approved):

PR #4177, opened 2024-09-01
[BFW-6709] WI_SWITCH_t : Ensure that variable "index" is private, preventing use outside the derived class

bkerler:
Reuse of the variable "index" within a WI_SWITCH_t derived class can lead to issues as the variable "index" might be used outside the WI_SWITCH_t class and could interfer with the derived class index value, thus leading to functionality issues. By redefining index as private variable in WI_SWITCH_t and making usage of the defined GetIndex/SetIndex functions, similar issues can be prevented in the future.

### DIFF
--- a/src/gui/MItem_crash.cpp
+++ b/src/gui/MItem_crash.cpp
@@ -81,7 +81,7 @@ MI_CRASH_SENSITIVITY_XY::MI_CRASH_SENSITIVITY_XY()
     : MenuItemSwitch(_(label), crash_sensitivity_items, stdext::index_of(crash_sensitivity_item_values, crash_s.get_sensitivity().x)) {}
 
 void MI_CRASH_SENSITIVITY_XY::OnChange([[maybe_unused]] size_t old_index) {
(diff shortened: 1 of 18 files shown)


CZDanol (changes requested):
See https://github.com/prusa3d/Prusa-Firmware-Buddy/pull/4176 for first commit issues. The correct solution would be to do:
```SetIndex(netdev_get_ip_obtained_type(this->device_id()));``` 
instead of
```index = netdev_get_ip_obtained_type(this->device_id());```

bkerler:
Updated, looks like it does build fine now :)

CZDanol (approved):

CZDanol:
Internal ticket: BFW-6709

michalrudolf:
I have removed the `GetIndex() -> get_index()` change from MMU code (unrelated change) and force-pushed the commit to your branch. Now we can merge it.

michalrudolf (approved):

PR #4221, opened 2024-09-24
Cleanup dependencies

### LINKED ISSUE #4195: [BUG] Bundled version of `clang-format` broken on ARM Apple devices
### Printer model

-

### Firmware version

Latest

None

### Printing from...

- 

### Describe the bug

When committing `pre-commit` hangs at the `clang-format` stage indefinitely. When running the command `./.dependencies/clang-format-16-83817c2f/clang-format` (as defined in the `.pre-commit-config.yaml`) we get the following output:

```bash
dyld[30989]: Library not loaded: /usr/local/opt/zstd/lib/libzstd.1.dylib
  Referenced from: <B970F465-3A4F-3100-ADE2-CBA39DB5AB8E> /Users/yorick/Developer/Personal/Prusa-Firmware-Buddy/.dependencies/clang-format-16-83817c2f/clang-format
  Reason: tried: '/usr/local/opt/zstd/lib/libzstd.1.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/usr/local/opt/zstd/lib/libzstd.1.dylib' (no such file), '/usr/local/opt/zstd/lib/libzstd.1.dylib' (no such file), '/usr/local/lib/libzstd.1.dylib' (no such file), '/usr/lib/libzstd.1.dylib' (no such file, not in dyld cache)
```

`clang-format@16` expects the `libzstd.1.dylib` to be at a different location than where `clang-format@18` expects it (there where brew installs it). Creating a symlink as suggested [in this thread](https://github.com/supermerill/SuperSlicer/issues/2488) proved no solution as the bundled `clang-format` expects a `x86` version, and not the now default `arm64` version.

```bash
'/usr/local/lib/libzstd.1.dylib' (mach-o file, but is an incompatible architecture (have 'arm64', need 'x86_64'))
``` 

Changing the `.pre-commit-config.yaml` hooks entry with id `clang-format` to the settings below allows the pre-commit hook to run. This makes it use the version installed via `brew install clang-format`.

```yaml
      entry: clang-format
      language: system
```

The solution would be to update the `clang-format` [dependency](https://github.com/prusa3d/Prusa-Firmware-Buddy/blob/f5a498ab8d2a42341d0dbeb969b7ae047783e860/utils/bootstrap.py#L66) to a later version.

### How to reproduce

Clone the repo on an Apple device with an ARM processor. Make a commit and run the pre-commit command.

### Expected behavior

_No response_

### Files

_No response_

ysmilda:
Closes #4195

This pull request updates the various dependencies to the latest versions. These updates bring support for using this codebase on ARM devices.

The following dependencies are updated:
 - clang-format 16 > 18.1.8 (19.1 has been released, but binaries not yet available)
 - ninja 1.10.2 > 1.12.1
 - cmake 3.28.3 > 3.30.3

An update for `gcc-arm-none-eabi` is available, but still working through some [issues](https://gitlab.kitware.com/cmake/cmake/-/issues/23105) when used from ARM devices.

This is still in development, but anyone wanting to help out and test the various integrations is more than welcome!

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,7 +14,7 @@ repos:
     - id: clang-format
       name: clang-format
       description: This hook automatically checks and reformats changed files using clang-format formatter.
-      entry: './.dependencies/clang-format-16-83817c2f/clang-format'
+      entry: './.dependencies/clang-format-18.1.8/clang-format'
       language: script
       files: \.(h\+\+|h|hh|hxx|hpp|cuh|c|cc|cpp|cu|c\+\+|cxx|tpp|txx)$
       args: ['-i', '-style=file']
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,7 +1,7 @@
 {
     "cmake.configureOnOpen": true,
     "cmake.copyCompileCommands": "${workspaceFolder}/compile_commands.json",
-    "cmake.cmakePath": "${workspaceFolder}/.dependencies/cmake-3.28.3/bin/cmake",
+    "cmake.cmakePath": "${workspaceFolder}/.dependencies/cmake-3.30.3/bin/cmake",
     "cmake.generator": "Ninja",
     "files.insertFinalNewline": true,
     "files.associations": {
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -13,7 +13,7 @@
             "cacheVariables": {
                 "CMAKE_MAKE_PROGRAM": {
                     "type": "FILEPATH",
-                    "value": "${sourceDir}/.dependencies/ninja-1.10.2/ninja"
+                    "value": "${sourceDir}/.dependencies/ninja-1.12.1/ninja"
                 },
                 "PRINTER": {
                     "type": "STRING",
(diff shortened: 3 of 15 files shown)


CZDanol:
Unittests CI failing:  `/work/.dependencies/cmake-3.28.3/bin/ctest: not found`

ysmilda:
That's weird. It seems to use an old config as I did update the [build-pr.jenkins](https://github.com/prusa3d/Prusa-Firmware-Buddy/pull/4221/files#diff-8f69169e72c1ef83df41e60c77b0d2878c60b973e903eb967a024216752644e3L245) file to target the new binaries.

CZDanol:
My guess would be that we have it hardcoded on some more places :/

ysmilda:
Can't check it anymore as the build seems to have been deleted. But from what I could gather from the output and the related jenkins file it did get stuck on the line I referenced above. Would it be possible to do a rerun and see if it picks up the changes?

CZDanol:
Allright, issuing a rerun :)

ysmilda:
Can't see the result, but it seems to have failed again (which isn't a big surprise). Is all the config for holly/jenkins contained within the repo or are there other sources that need updating?

CZDanol:
Afaik it should all be in the repo itself. @danopernis am I right?

PR #4256, opened 2024-10-16
Fix for SELFTEST-BUG 3480/BFW-5677.  Allowable overlap range on Switched Fan Detection.

deniolan:
Allowing for an overlap range on Fan Check to allow for the fact that sometimes switched fan detection is not possible, and on XL's this happens quite frequently.  The hot ends also come preassembled, so this alleviates the annoyance of having to remove the shroud just to do initial calibration or a factory reset, while still allowing for switched fan detection when ranges permit.

This addresses https://github.com/prusa3d/Prusa-Firmware-Buddy/issues/3480

### DIFF
--- a/src/common/selftest/selftest_MK4.cpp
+++ b/src/common/selftest/selftest_MK4.cpp
@@ -70,14 +70,16 @@ static constexpr SelftestFansConfig fans_configs[] = {
             ///  With MK4 shroud the values can be 6400 or so.
             .rpm_min = 5300,
             .rpm_max = 6799,
+            .ignore_min_overlap = 0,
         },
         .heatbreak_fan = {
             .rpm_min = 6800,
             .rpm_max = 8700,
+            .ignore_min_overlap = 0,
         },
     },
 };
-static_assert(fans_configs[0].print_fan.rpm_max < fans_configs[0].heatbreak_fan.rpm_min, "These cannot overlap for switched fan detection.");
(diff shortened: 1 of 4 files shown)


deniolan:
Can't load Holly to see what is going on - will try again later.

Tomcus:
@deniolan it is failing on format check. Make sure you have pre-commit installed on your system and that it is registered as a hook in your local repo.

deniolan:
> @deniolan it is failing on format check. Make sure you have pre-commit installed on your system and that it is registered as a hook in your local repo.

Thanks - any idea why I was getting a block page on holly?  Seemed to be an invalid certificate at least from what I could tell…  Got a Prusa block page telling me to use a VPN to view it.

Thanks!

Tomcus:
> > @deniolan it is failing on format check. Make sure you have pre-commit installed on your system and that it is registered as a hook in your local repo.
> 
> Thanks - any idea why I was getting a block page on holly? Seemed to be an invalid certificate at least from what I could tell… Got a Prusa block page telling me to use a VPN to view it.
> 
> Thanks!

Yes, IT moved most of our internal services behind VPN. That includes Holly. IT knows, that public developers can't see Holly and they plan to somehow to fix/remedy this issue. But I don't know when and how is it going to work.

deniolan:
Thank you.  I installed pre-commit and recreated the original commit/pull request.  Looks like it is green now.  Thanks for the info!

PR #4258, opened 2024-10-17
selftest: Add allowable fan RPM overlap ranges.

deniolan:
Allows for switched fan detection except for in overlapping ranges which can occur on the XL - Addresses BFW-5677.

Effectively, this change will still allow switched fan detection to work while allowing for an overlap range - which, for many of us (myself included) is a reality on the XL.   Switched fan detection is not possible only in this overlap range.

For the other printers using the same firmware base, the overlap has been defaulted to 0 (behavior unchanged), and printers not using the same fan test routine are not impacted.

The idea here is to allow many people, who like to factory reset any time there is a new firmware, to successfully re-run the self-tests without having to remove the shroud from one or more fans on the XL in order to pass the test.   This bug has been out there for quite some time and is very impactful to my workflow so I am hoping this can work its way in to a release and alleviate this issue for most.


CZDanol:
Hey, sorry, we've already fixed this in our codebase and forgot to inform you that there's no need to update the PR.

deniolan:
Any idea when this will be released?


On Oct 17, 2024, at 3:41 AM, Danol (Daniel Čejchan) ***@***.***> wrote:

﻿

Hey, sorry, we've already fixed this in our codebase and forgot to inform you that there's no need to update the PR.

CZDanol:
I'm not at liberty to give estimates, sorry :/

PR #4237, opened 2024-10-06
[BFW-6223] gui: Use marlin's cold extrusion temperature and allowance in extruder move screen

FiveTeethless:
As I found in issue #3674 the screen menu source code has a magic cold extrusion temperature defined as constexpr.
The Marlin core actually has a cold ectrusion tempereature and an allowance state, that can be modified with M302.

I modified the marlin_server to give access to this information and modified the GUI code to either ignore the cold extrusion temperature or to check the temperature against the limit of Marlin core.

I have not changed the material's "minimum" temerature check. If this temperature is above marlins cold-extrusion temperature, the material temp will be checked. Just as it was before.

Finally I did a small change in the CMakeLists.txt:
As I compile on a Windows machine, I need the Python executable infront of a Python file to run it.
I found the variable "${Python3_EXECUTABLE}" somewhere else in the file, so I used it here too.

PR #4265, opened 2024-10-20
[BFW-6223] cmake: Use Python executable to run gen_esp_parts.py

FiveTeethless:
As I mentioned in PR #4237 I only could build the source code on Windows, if I added the python executable to this cmake script.


Tomcus (approved):

CZDanol:
Internal ticket: BFW-6223

Issue #4330, opened 2024-12-03
[BUG] Incorrect reset value for extruder steps per unit in experimental settings


joshuafocht:
### Printer model

MINI+

### Firmware version

6.1.2

### Upgrades and modifications

filament runout sensor

### Printing from...

USB Drive, PrusaLink, and Prusa Connect

### Describe the bug

When resetting the extruder steps per unit under experimental settings it defaults to 1 not the stock value of 325 (for a mini/+).

### How to reproduce

1. Goto experimental settings by holding click on the Hardware menu of Settings for 2-3 seconds.
2. Click reset extruder steps per unit.
3. Extruder steps per unit gets set to 1.

### Expected behavior

Clicking reset extruder steps per second should set it to 325 not 1. (For the mini/+)

### Files

https://1drv.ms/v/s!ArmBR5ZkHUhmk8tYQyPg-7nA_JrzDQ?e=joetqc

bkerler:
@joshuafocht I'm looking into this issue.

So far I couldn't see any issue that might cause this at first sight, so I will try to emulate to see what goes wrong.

Just for reference for the other devs:
https://github.com/prusa3d/Prusa-Firmware-Buddy/blob/8a788ae4e23d0fc79cf15c4c2dde05168b95b6bb/src/gui/screen_menu_experimental_settings_release.cpp#L55

Being set here:
https://github.com/prusa3d/Prusa-Firmware-Buddy/blob/8a788ae4e23d0fc79cf15c4c2dde05168b95b6bb/src/persistent_stores/store_instances/config_store/store_definition.hpp#L409

Which is hardcoded to 325 here:
https://github.com/prusa3d/Prusa-Firmware-Buddy/blob/8a788ae4e23d0fc79cf15c4c2dde05168b95b6bb/include/marlin/Configuration_MINI.h#L678

bkerler:
Most probably caused by this :
https://github.com/prusa3d/Prusa-Firmware-Buddy/blob/8a788ae4e23d0fc79cf15c4c2dde05168b95b6bb/src/persistent_stores/store_instances/config_store/old_eeprom/eeprom_v10.hpp#L92

bkerler:
@joshuafocht Two questions:
1. Do you use the release or debug firmware ?
2. Did you try to reset the eeprom and the bug still consists ?
Thanks

joshuafocht:
I use the release firmware.
[...]

PR #4341, opened 2024-12-12
[BFW-6459] Fix invalid display of E0 Extruder settings

bkerler:
As default extruder settings can have negative values due to inversion, values below 0 are displayed incorrectly in the Experimental Settings display for RELEASE version (not the case for DEBUG version).

This solves issue #4330

### DIFF
--- a/src/gui/screen_menu_experimental_settings_release.cpp
+++ b/src/gui/screen_menu_experimental_settings_release.cpp
@@ -52,7 +52,7 @@ void ScreenMenuExperimentalSettings::windowEvent(window_t *sender, GUI_event_t e
         Invalidate();
         break;
     case ClickCommand::Reset_steps:
-        Item<MI_STEPS_PER_UNIT_E>().SetVal(config_store().axis_steps_per_unit_e0.default_val);
+        Item<MI_STEPS_PER_UNIT_E>().SetVal(std::abs(config_store().axis_steps_per_unit_e0.default_val));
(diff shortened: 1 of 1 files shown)


CZDanol:
Umm, 955 commits?

src/gui/screen_menu_experimental_settings_release.cpp
```diff
@@ -52,7 +52,7 @@ void ScreenMenuExperimentalSettings::windowEvent(window_t *sender, GUI_event_t e
         Invalidate();
         break;
     case ClickCommand::Reset_steps:
-        Item<MI_STEPS_PER_UNIT_E>().SetVal(config_store().axis_steps_per_unit_e0.default_val);
+        Item<MI_STEPS_PER_UNIT_E>().SetVal(abs(config_store().axis_steps_per_unit_e0.default_val));
```
danopernis, line 55: I would use `std::abs()` to be consistent with implementation in `src/gui/screen_menu_experimental_settings_debug.cpp`

danopernis (changes requested):
Other than that looks good to me, thank you for taking time to debug this.

src/gui/screen_menu_experimental_settings_release.cpp
bkerler: Sure, updated it accordingly :)

danopernis (approved):

Issue #4166, opened 2024-08-29
[BFW-6958] [BUG] Wrong display in the footer - MK3.5 - FW 6.1.2


LoakAudio:
### Printer model

MK3.5

### Firmware version

6.1.2

### Upgrades and modifications

none

### Printing from...

not relevant

### Describe the bug

Hi all,
I've just updated to 6.1.2.
Playing with the data showed in the footer of the main page, it appears that using "none" for a value leads to show the nozzle temperature instead of nothing.

### How to reproduce

Just play with the footer menu.

danopernis:
Thanks for the report @LoakAudio 

To be more specific, it doesn't invalidate the footer item, so the last selected footer item remains visible. This is a valid bug report, but it is not really critical so we are not going to work on it anytime soon. Still, this looks easy to fix so if anybody wants to create PR with the fix, I will be happy to merge it.

bkerler:
In fact, the footer item is invalidated, but as the existing item isn't removed, it stays active. I just wrote a PR #4345 to fix this behaviour.

bkerler:
Commit #4345 has just been updated and should be fine.

Prusa-Support:
The bugfix from @bkerler has been implemented in FW 6.3.3 for Prusa CO and should be ported to Prusa MK3.5 with the future universal firmware upgrade FW 6.4.

> Fixed a bug where the footer didn’t update correctly after selecting “none.” Thanks to @bkerler for the fix!

https://github.com/prusa3d/Prusa-Firmware-Buddy/releases/tag/v6.3.3

Michele Moramarco
Prusa Research

PR #4345, opened 2024-12-14
[BFW-6958] GUI: Fix footer update after selecting None

bkerler:
This fixes issue #4166.

After selecting None in the Footer settings menu, the existing footer item isn't erased, thus stays active although it has been disabled.

### DIFF
--- a/src/gui/footer/footer_line.cpp
+++ b/src/gui/footer/footer_line.cpp
@@ -44,9 +44,11 @@ bool FooterLine::Create(footer::Item item_id, size_t index) {
 
     auto &item = items[index];
 
-    const bool result = [&]<typename... Rec>(TypeList<Rec...>) -> bool {
+    bool result = [&]<typename... Rec>(TypeList<Rec...>) -> bool {
         return (((item_id == Rec::item && !std::holds_alternative<typename Rec::T>(item)) ? item.emplace<typename Rec::T>(this), true : false) || ...);
     }(footer::FooterItemMappings());
+
+    result |= (item_id == footer::Item::none && !std::holds_alternative<std::monostate>(item) ? item.emplace<std::monostate>(), true : false);
 
     if (result) {
         positionWindows();


src/gui/footer/footer_line.cpp
```diff
@@ -51,6 +51,8 @@ bool FooterLine::Create(footer::Item item_id, size_t index) {
 
     if (result) {
         positionWindows();
+    } else if (item_id == footer::Item::none) {
+        Erase(index);
```
CZDanol, line 55: This doesn't seem to be a correct fix. The result is set to true on any change, which should call positionWindows. That is, even on change to/from none. This thing might possibly work, but it's a bit random, plus it will call positionWindows each time (inside Erase), suboptimally.

By quickly looking at the code, the proper/elegant fix would probably be to add `FooterItemMappingRec<std::monostate, Item::none>` to `FooterItemMappings`

CZDanol (changes requested):

src/gui/footer/footer_line.cpp
bkerler: Trying to add FooterItemMappingRec as suggested doesn't work due to compiler error:
`error: static assertion failed: T must occur exactly once in alternatives
[build]  1115 |       static_assert(__detail::__variant::__exactly_once<_Tp, _Types...>,`.
As far as I could see, Erase doesn't call positionWindows and it only is erased if result is false, which happens if the item isn't registered (which is only the case for footer::item::none).

bkerler: I'm now using unregister directly and also moved it above the FooterItemsMapping function, which should be a more proper way to do it. If you have any ideas how to make the FooterItemMappingRec working with std::monostate, please let me know :)

src/gui/footer/footer_line.cpp
```diff
@@ -41,6 +41,9 @@ bool FooterLine::Create(footer::Item item_id, size_t index) {
 
     if (item_id >= footer::Item::_count) {
         item_id = footer::Item::none;
+    } else if (item_id == footer::Item::none) {
+        unregister(index);
```
CZDanol, line 45: This does not call positionWindows, which it probably should in some cases (at least according to the previous code).

src/gui/footer/footer_line.cpp
```diff
@@ -41,6 +41,9 @@ bool FooterLine::Create(footer::Item item_id, size_t index) {
...
+        return false;
```
CZDanol, line 46: It is rather debatable whether this should call `true` or `false`. From the previous code, it rather seems like the function should return `true` if an item is changed (and in that case, and only in that case, it should call reposition). This code does not follow that logic.

But looking around, it seems like the return value is not used anywhere, so it would perhaps be best to just turn it into a procedure.

CZDanol (approved):

CZDanol:
Internal ticket: BFW-6958

Issue #3848, opened 2024-03-22
[BUG] The XL reports chamber temperature despite not having one


smurfix:
**Printer type** - XL

**Printer firmware version** - 5.12.

**Original or Custom firmware** - Original

**Optional upgrades** - no

**USB drive or USB/Octoprint**
  Octoprint

**Describe the bug**
  The periodic temperature report from the printer contains a "C:-30/0" stanza. The printer doesn't have a chamber and no chamber sensor, thus reporting that temperature makes no sense and confuses Octoprint.

**How to reproduce**
  Connect to OctoPrint, look at its log entries. It complains about the C: every couple of seconds. This makes the SD card unhappy.

**Expected behavior**
  Don't send the chamber temperature if you don't have a chamber.

**G-code**
```
Send: M115
Recv: FIRMWARE_NAME:Prusa-Firmware-Buddy 5.1.2+13478 (Github) SOURCE_CODE_URL:https://github.com/prusa3d/Prusa-Firmware-Buddy PROTOCOL_VERSION:1.0 MACHINE_TYPE:Prusa-XL EXTRUDER_COUNT:6 UUID:cede2a2f-41a2-4748-9b12-c55c62f367ff
…
Recv: Cap:CHAMBER_TEMPERATURE:0

Send: M155 S2
Recv: ok
Recv:  T:6.00/0.00 B:20.39/0.00 C:-30.00/0.00 X5:6.00/36.00 A:33.56/0.00 T0:20.00/0.00 T1:19.00/0.00 T2:21.00/0.00 T3:21.00/0.00 T4:22.00/0.00 T5:6.00/0.00 @:0 B@:0 HBR@:0 @0:0 @1:0 @2:0 @3:0 @4:0 @5:0 B_0_0:20.30/0.00 B_1_0:20.40/0.00 B_2_0:20.30/0.00 B_3_0:20.20/0.00 B_0_1:20.30/0.00 B_1_1:20.40/0.00 B_2_1:20.50/0.00 B_3_1:20.30/0.00 B_0_2:20.30/0.00 B_1_2:20.50/0.00 B_2_2:20.50/0.00 B_3_2:20.20/0.00 B_0_3:20.50/0.00 B_1_3:20.60/0.00 B_2_3:20.60/0.00 B_3_3:20.30/0.00
```

smurfix:
*Sigh* 7 days are too short.

smurfix:
That auto-close is annoying.

[...]

PR #4347, opened 2024-12-14
serial: Fix serial temp chamber reported if not enabled

bkerler:
This solves issue #3848.

If the device doesn't have a chamber temp, then chamber temps shouldn't be reported via gcode M155.


bkerler:
Closed in favor of #4674 which fixes the root cause

Issue #3812, opened 2024-03-06
[BFW-5885] [FEATURE REQUEST] Add Ready Status as an option for ui Footer


Shinomorisama:
MK4 and probably mini and XL

It is currently possible to go to the control menu and set the printer as ready so it will grab the next item in queue or immediately start anything added to it's queue.  This is awesome and tremendous asset to productivity.   The issue is that the only way to tell if a printer is ready at the machine is to go into the control menu and see if the option is greyed out. Which is quite inconveniant.

The Feature I would request is something on the main screen to indicate the printer is ready.  This could be an icon on the top bar or as the title suggests and optional footer indicator.

murk-sy:
+1
See also / not really a duplicate but related #3315

ChrisHeerschap:
Alternately, having _any_ visual feedback when choosing "Set Ready" would be invaluable.

Shinomorisama:
Having worked with this more.  An indicator on the LED would also be excellent. Example: Red/yellow-errors or intervention needed depending on severity. Green-Ready or printing.  Blue-Not ready.

ChrisHeerschap:
> Having worked with this more. An indicator on the LED would also be excellent.

Yes, that would be excellent!

bkerler:
I just added a PR #4348 in order to add a footer with ready status

bkerler:
@Shinomorisama I recommend that your led feature request should be make as a new issue with "[FEATURE REQUEST] label :)

CZDanol:
Hello, I'm not sure if this is already out in 6.2.0-alpha2 or if will get out in the next firmware, but we are now displaying the connect status (plus ready status) in the header.

PR #4348, opened 2024-12-14
GUI: Add footer item for ready status

bkerler:
This adds a footer in order to display the printer ready status as requested in the issue #3812.

### DIFF
--- a/src/gui/footer/footer_item_ready_status.cpp
+++ b/src/gui/footer/footer_item_ready_status.cpp
@@ -1,0 +1,15 @@
+#include "footer_item_ready_status.hpp"
+#include "connect/marlin_printer.hpp"
+#include "img_resources.hpp"
+
+FooterItemReadyStatus::FooterItemReadyStatus(window_t *parent)
+    : FooterIconText_IntVal(parent, &img::ok_16x16, static_makeView, static_readValue) {
+}
+
+int FooterItemReadyStatus::static_readValue() {
+    return connect_client::MarlinPrinter::is_printer_ready();
+}
+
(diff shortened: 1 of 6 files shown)


src/gui/footer/footer_item_ready_status.cpp
```diff
@@ -0,0 +1,15 @@
...
+FooterItemReadyStatus::FooterItemReadyStatus(window_t *parent)
+    : FooterIconText_IntVal(parent, &img::ok_16x16, static_makeView, static_readValue) {
+}
+
+int FooterItemReadyStatus::static_readValue() {
+    return connect_client::MarlinPrinter::is_printer_ready();
+}
+
+string_view_utf8 FooterItemReadyStatus::static_makeView(int value) {
+    return string_view_utf8::MakeRAM(value == 1 ? (const uint8_t *)N_("YES") : (const uint8_t *)N_("NO"));
```
CZDanol, line 14: Yes/No wouldn't get translated

CZDanol (changes requested):
1. We already have an icon for this in the header that's always shown. Is this even needed/wanted anymore?
2. This is implemented as a single-trick footer item. It would make sense to make this into a general connect status footer item (offline/not registered/online/ready). But again - there is already a header icon for that.

bkerler:
> 1. We already have an icon for this in the header that's always shown. Is this even needed/wanted anymore?
> 
>     2. This is implemented as a single-trick footer item. It would make sense to make this into a general connect status footer item (offline/not registered/online/ready). But again - there is already a header icon for that.

Ok, guess then we can close this PR :) Yeah, there is a status icon for that.

Issue #4343, opened 2024-12-14
[ENHANCEMENT] Increase I2C M260 buffer size


me25542:
### Printer model

All that support I2C

### Describe the enhancement

The current implementation of the M260 command is great, and allows users to communicate with external devices. However, the maximum buffer size of 32 bytes can be limiting. I don't know if this is a hardware restriction, or in software. I did look in the `STM32F427ZIT6`'s datasheet, and couldn't find anything. If this is a hardware limitation, let me know and close the issue.

### Expected functionality

The print file would be able to add, say, up to 256 bytes to the buffer, and send it all in one command.

bkerler:
As we are discussing internally if sending 256 bytes in one round does make any sense instead of sending 8*32 byte packets in sequence, can you please let us know your specific use case ?

me25542:
I have a custom "smart" enclosure connected to the printer via I2C, and for most commands from the printer to the enclosure 32 bytes is enough. But, for setting a print name to be displayed on the enclosure, 32 bytes is often too short. I have implemented a workaround for this, enabling the data to be split into multiple 32-byte packets. However, it would be easier to send a single larger packet. This is low priority since implementing the workaround on my enclosure, but it would be nice. ☺️

bkerler:
Thanks. PR is already written, let's see how the dev gods will decide :D

me25542:
ok lol

PR #4396, opened 2025-01-09
I2C: Increase TWIBUS buffer size from 32 to 256

bkerler:
This covers issue #4343.

**For reference :** 
Originally the TWI Bus buffersize has been limited to 32 bytes. The reason mainly was the eeprom size limit. However, this doesn't seem to be a real hardware limitation but more a device dependent limitation, as it was stated on places all over the internet, for example see: [Arduino discussion on i2c twi buffer limitation](https://forum.arduino.cc/t/increase-32-byte-i2c-buffer-size-in-wire-library/54503)

**Alternative:**
In the past, if more than 32 bytes should be sent, a chain of multiple 32 byte buffers was sent in sequence instead.

**What this PR does:**
As this is no real limitation but in order to make sure that not too much ram is used for this feature, the buffer limit has been increased to only 256 bytes which should be sufficient for most edge cases.
For data larger then 256, multiple packet should be sent instead, further increasing the buffer size would unnecessarily fill the RAM.

### DIFF
--- a/lib/Marlin/Marlin/src/feature/twibus.h
+++ b/lib/Marlin/Marlin/src/feature/twibus.h
@@ -35,7 +35,7 @@
 typedef void (*twiReceiveFunc_t)(int bytes);
 typedef void (*twiRequestFunc_t)();
 
-#define TWIBUS_BUFFER_SIZE 32
+#define TWIBUS_BUFFER_SIZE 256
 
 /**
  * TWIBUS class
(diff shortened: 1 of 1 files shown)


CZDanol:
This change permanently increases RAM usage by 448 B. Every byte counts. I'm not really sure it's worth for us to increase it like this for a guy that managed to do it in a different way already.

bkerler:
Aye. So let's close this PR.

PR #4412, opened 2025-01-15
fix FW version check timeout comment

nmschulte:


CZDanol (commented):
Hey, thanks for your input. The comment is indeed wrong - but what about removing the comment altogether instead? It basically just duplicates the code, without really adding any value.

bkerler:
@nmschulte can you please rebase your code to the latest master and update your code according to what @CZDanol wrote ? Or close the PR if it is no longer needed. Thanks !

nmschulte:
> @nmschulte can you please rebase your code to the latest master and update your code according to what @CZDanol wrote ? Or close the PR if it is no longer needed. Thanks !

Done; thanks.

PR #4425, opened 2025-01-23
Fix comment: there are 26 letters in the alphabet

jedisct1:
Not 25, even if the serial number only uses 4 bits to produce each character of the 64-bit serial number.


danopernis:
Hi @jedisct1 thanks for bringing this to our attention. We decided that even better than fixing this comment would be to stop inventing serial numbers altogether. All our boards should have valid serial number programmed in OTP flash anyway. This was fixed in our private fork of the repository and will be present in some of the next releases. I am closing this PR now.

Issue #4293, opened 2024-11-11
[BUG] M21 does not respond properly


timur-tabi:
### Printer model

MK3.5s

### Firmware version

6.1.3+7898

### Upgrades and modifications

None

### Printing from...

USB

### Describe the bug

When a USB drive is inserted, the M21 command is supposed to respond with one of the following:

- SD card ok
- Card successfully initialized

However, all it does is:

```
2024-11-10 09:29:37,617 - Send: M21
2024-11-10 09:29:37,637 - Recv: ok

```
This makes Octoprint think that the SD card (technically the USB drive) is not available, even though it is.

timur-tabi:
See also https://github.com/OctoPrint/OctoPrint/issues/5077 for the debug trail.

timur-tabi:
It's still important to me.

timur-tabi:
I'm assuming that if the printer support USB but not SD, then the M21 command should work with the USB card as if it were an SD card.  So on the MK3.5, M21 should respond with "Card successfully initialized".

bkerler:
I wrote a PR #4428 for this issue.

CZDanol:
PR merged. Should appear in the 6.4 version.

timur-tabi:
I can confirm that this has been fixed in v6.4.  Thanks!

PR #4428, opened 2025-01-25
[BFW-6669] Serial: Emulate M21 SD card response for octoprint / third party

bkerler:
This PR is related to issue #4293. Third party software like octoprint expects the M21 command to return "SD card ok" on existing usb and "No SD card" if no usb is found, according to [Marlin M21 documentation](https://marlinfw.org/docs/gcode/M021.html).

Related to [Octoprint issue #5077](https://github.com/OctoPrint/OctoPrint/issues/5077).

### DIFF
--- a/src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
+++ b/src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
@@ -30,7 +30,12 @@ void GcodeSuite::M20() {
  * M21 - Initialize SD card
  */
 void GcodeSuite::M21() {
-    // not necessary - empty implementation
+    // required for Octoprint / third party tools to simulate an inserted SD card when using USB
+    if (marlin_vars()->media_inserted) {
+        SERIAL_ECHOLNPGM(MSG_SD_CARD_OK);
+    } else {
+        SERIAL_ECHOLNPGM(MSG_SD_CARD_FAIL);
+    }
 }
 
 /**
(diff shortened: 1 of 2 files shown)


src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
```diff
@@ -30,7 +30,14 @@ void GcodeSuite::M20() {
  * M21 - Initialize SD card
  */
 void GcodeSuite::M21() {
-    // not necessary - empty implementation
+    // required for Octoprint / third party tools to simulate an inserted SD card when using USB
+    DIR *dir;
+    dir = opendir("/usb/");
+    if (dir != NULL) {
+        SERIAL_ECHOLNPGM(MSG_SD_CARD_OK);
```
timur-tabi, line 37: You need to call closedir(), otherwise you will leak file handles.

bkerler: Good point, thx ! Patched it already.

src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
```diff
@@ -30,7 +30,15 @@ void GcodeSuite::M20() {
  * M21 - Initialize SD card
  */
 void GcodeSuite::M21() {
-    // not necessary - empty implementation
+    // required for Octoprint / third party tools to simulate an inserted SD card when using USB
+    DIR *dir;
+    dir = opendir("/usb/");
```
CZDanol, line 35: There's an easier way to check if the USB is present - ˙marlin_vars().media_inserted˙

bkerler: Perfect, exactly what I was searching for. Changed the PR accordingly, thx !

CZDanol:
Thanks for your work :) Created an internal ticket BFW-6669 so that someone would process the PR on our side

CZDanol (approved):

PR #4429, opened 2025-01-25
[BFW-6693] gcode: Add support for delete file via M30 gcode command

bkerler:
This partially refers to [Octoprint issue #5077] (https://github.com/OctoPrint/OctoPrint/issues/5077).

The gcode M30 for deleting files, required by octoprint and third party tools hasn't been implemented.

### DIFF
--- a/src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
+++ b/src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
@@ -1,9 +1,11 @@
 #include <dirent.h>
 
 #include "../../lib/Marlin/Marlin/src/gcode/gcode.h"
+#include "../src/common/print_utils.hpp"
 #include "marlin_server.hpp"
 #include "media.hpp"
 #include "marlin_vars.hpp"
+#include <str_utils.hpp>
 
 /** \addtogroup G-Codes
  * @{
@@ -119,7 +121,15 @@ void GcodeSuite::M29() {
 
 // M30 - Delete a file on the SD card
 void GcodeSuite::M30() {
-    // TODO
+    ArrayStringBuilder<FF_MAX_LFN> filepath;
+    filepath.append_printf("/usb/%s", parser.string_arg);
+    DeleteResult result = DeleteResult::GeneralError;
+    if (filepath.is_ok()) {
+        result = remove_file(filepath.str());
+    }
+    SERIAL_ECHOPGM(result == DeleteResult::Success ? "File deleted:" : "Deletion failed:");
+    SERIAL_ECHO(parser.string_arg);
+    SERIAL_ECHOLN(".");
 }
 
 // M32 - Select file and start SD print


src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
```diff
@@ -119,7 +120,18 @@ void GcodeSuite::M29() {
 
 // M30 - Delete a file on the SD card
 void GcodeSuite::M30() {
-    // TODO
+    char filepath[FILE_PATH_MAX_LEN];
+    strlcpy(filepath, "/usb/", FILE_PATH_MAX_LEN);
```
timur-tabi, line 124: I don't know if this violates any Marlin coding guidelines, but how about this:

char filepath[FILE_PATH_MAX_LEN] = "/usb/";

bkerler: yep, you're right. that's better.

src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
```diff
@@ -119,7 +120,17 @@ void GcodeSuite::M29() {
...
+    char filepath[FILE_PATH_MAX_LEN] = "/usb/";
+    strlcat(filepath, parser.string_arg, FILE_PATH_MAX_LEN);
+    DeleteResult result = remove_file(filepath);
+    if (result == DeleteResult::Success) {
+        SERIAL_ECHOPGM("File deleted:");
+        SERIAL_ECHOLN(parser.string_arg);
+    } else {
+        SERIAL_ECHOPGM("Deletion failed, File: ");
+        SERIAL_ECHO(parser.string_arg);
+        SERIAL_ECHOLN(".");
```
CZDanol, line 132: Thy is there "." at the end of one response and on the other one? If this was unified, you could put `SERIAL_ECHOLN(parser.string_arg);` outside of the branch and save some flash.

You could also possibly save some flash by doing `SERIAL_ECHOPGM(result == success ? "File deleted:": "Deletion failed:");`

src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
```diff
@@ -119,7 +120,17 @@ void GcodeSuite::M29() {
 
 // M30 - Delete a file on the SD card
 void GcodeSuite::M30() {
-    // TODO
+    char filepath[FILE_PATH_MAX_LEN] = "/usb/";
```
CZDanol, line 123: I'd maybe suggest using our StringBuilder for this, though your code is fine in this regard. Though theoretically, we could have a wrong file deleted if the string arg happens to get cropped.

bkerler: Yeah, good idea.

src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
bkerler: I was doing a cnp from the Prusa-Firmware, but you're right. Changing the code accordingly :)

danopernis: Just a note, I guess the octoprint will be very picky about string format, so we should try mimicking Marlin as much as possible.

bkerler: Yeah, I will check the octoprint code and make sure string format matches.

bkerler: Doesn't look like octoprint does check the response, instead it issues a M20 code after sending M30, see [here](https://github.com/OctoPrint/OctoPrint/blob/d79a0d20f3f1c7f2edb56dedda3b70267a937e65/src/octoprint/util/comm.py#L2041), so I think we should be fine.

src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
```diff
@@ -119,7 +121,12 @@ void GcodeSuite::M29() {
 
 // M30 - Delete a file on the SD card
 void GcodeSuite::M30() {
-    // TODO
+    ArrayStringBuilder<FILE_PATH_MAX_LEN> filepath;
```
CZDanol, line 124: FILE_PATH_MAX_LEN really only considers SFN paths, so it will quite possibly not be able to fit LFNs.

src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
```diff
@@ -119,7 +121,12 @@ void GcodeSuite::M29() {
...
+    filepath.append_printf("/usb/%s", parser.string_arg);
+    DeleteResult result = remove_file(filepath.str());
```
CZDanol, line 126: We should check `builder.is_ok()` and prevent deletion attempt if the string was cropped

src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
bkerler: ok, using FF_MAX_LFN instead.

src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
bkerler: ok, checking via is_ok as well

CZDanol (approved):

CZDanol:
Internal ticket to process this: BFW-6693

PR #4451, opened 2025-02-02
Fix handling of .tar.xzg in bootstrap

rcloran:
The macOS build includes a `.tar.xzg` file, which 1bfe80a67 does not properly handle. This simply adds this extension to the list which should be extracted with tar instead of zip.

### DIFF
--- a/utils/bootstrap.py
+++ b/utils/bootstrap.py
@@ -173,7 +173,9 @@ def download_and_unzip(url: str, directory: Path):
     print('Extracting ' + file, end=" ")
 
     # Check if tar or zip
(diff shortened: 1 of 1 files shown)


rcloran:
I'm unable to view the CI failure, so I have no way of checking whether it was caused by my change or not. 🤷‍♂️

Tomcus:
Hi thanks for the contribution. Sorry for the error, but none of us use Macs, so this slipped our cracks. The CI fails with formatting check, you probably didn't install `pre-commit` into your local repository.

Tomcus (approved):
Thanks for your contribution. I will now merge your changes.

PR #4471, opened 2025-02-10
beautify: get_dwarf_nr for Tool pickup, cleanup code

bkerler:
Using `get_dwarf_nr() - 1` for internal/logging use vs `get_dwarf_nr()` for Marlin/Buddy usage might cause issues, so cleaning it up might help


lib/Marlin/Marlin/src/module/prusa/toolchanger_utils.cpp
```diff
@@ -418,9 +418,9 @@ bool PrusaToolChangerUtils::load_tool_offsets_from_usb() {
 }
 
 const PrusaToolInfo &PrusaToolChangerUtils::get_tool_info(const Dwarf &dwarf, bool check_calibrated) const {
-    assert(dwarf.get_dwarf_nr() <= tool_info.size());
-    assert(dwarf.get_dwarf_nr() > 0);
-    const PrusaToolInfo &info = tool_info[dwarf.get_dwarf_nr() - 1];
+    assert(dwarf.get_dwarf_nr() < tool_info.size());
+    assert(dwarf.get_dwarf_nr() >= 0);
```
CZDanol, line 422: This will be always true, dwarf_nr is uint8_t

lib/Marlin/Marlin/src/module/prusa/toolchanger_utils.cpp
```diff
@@ -441,10 +441,10 @@ bool PrusaToolChangerUtils::is_tool_info_valid(const Dwarf &dwarf, const PrusaTo
 }
 
 void PrusaToolChangerUtils::set_tool_info(const buddy::puppies::Dwarf &dwarf, const PrusaToolInfo &info) {
-    assert(dwarf.get_dwarf_nr() <= tool_info.size());
-    assert(dwarf.get_dwarf_nr() > 0);
+    assert(dwarf.get_dwarf_nr() < tool_info.size());
+    assert(dwarf.get_dwarf_nr() >= 0);
```
CZDanol, line 445: Ditto, always true

bkerler: Agree.

lib/Marlin/Marlin/src/module/prusa/toolchanger_utils.cpp
bkerler: Agree.

include/puppies/Dwarf.hpp
```diff
@@ -295,7 +295,7 @@ void ToolsMappingBody::windowEvent([[maybe_unused]] window_t *sender, GUI_event_
     uint16_t get_heatbreak_fan_pwr();
 
     inline uint8_t get_dwarf_nr() const {
-        return dwarf_nr;
+        return dwarf_nr - 1;
```
CZDanol, line 298: Disrepance between the public getter and private member is even worse than the current state. If we wanna do this, we will need to go further

CZDanol: Alternatively, we could rename `get_dwarf_nr` to `dwarf_index`

bkerler: Done.

CZDanol:
Internal ticket: BFW-6775

bkerler:
was merged in commit 71839391567990a64c91872881005acee81e2fdf

PR #4477, opened 2025-02-14
[BFW-6776] Fix FAT issues, port to r0.15

bkerler:
This PR partly ports to [STM FatFS r0.15](https://github.com/STMicroelectronics/stm32-mw-fatfs/releases/tag/r0.15_stm32cube_20230818), solving issues with inproper FAT init, GPT Partition table parsing, missing filename escape sequences and file lock count. 

**This is NO full port to STM FatFS r0.15, as there are Prusa changes in the FatFS that shouldn't be patched (such as unicode parsing bypass) and thus were left as they are.**


CZDanol:
Internal ticket: BFW-6776

Issue #4478, opened 2025-02-14
[BFW-6789] [ENHANCEMENT] Minor: Log closing of log file to log file just as it does when opening the file.


ulab:
### Printer model

All

### Describe the enhancement

If `file_log_enable()` writes `Started logging to file: %s` to the log file, `file_log_disable()` should write `Stopped logging to file: %s` too. :-)

danopernis:
Yeah, I agree it is inconsistent. May I suggest, even better would be to stop logging the start of logging. Each string/code we add to firmware costs precious FLASH, especially on MINI, so I would like to keep it minimal if I can. Patches welcome 🙂

ulab:
I'm fine with removing it, it doesn't really help a lot anyway?

That said, my coding knowledge is not good enough to not break things - especially since I can't just test it easily.

ulab:
That said, it looks like the string [kind of exists already](https://github.com/prusa3d/Prusa-Firmware-Buddy/blob/310b7f890f09c1318e264c46aed80e7ecf94dd93/src/logging/log_dest_file.cpp#L136). It just doesn't log it to the file?

danopernis:
Yeah no worries, not everybody has to be a coder. Thanks for reporting anyway, I left the "good first issue" label so perhaps somebody will grab it. If nobody does, we might implement it, but I like to leave tasks such as this to encourage community  contributions.

As for logging the disabling of the log, good catch. Either that's a race condition and log gets disabled before actually logging stuff, or it is deliberate mechanism to not log to the same log that is being disabled. I can't remember right now.

CZDanol:
I would argue that it is consistent and makes sense.

1. The log "Started logging" is emitted after we start logging. It makes sense that this gets captured in the file.
2. The log "Stopped logging" is emitted when we end logging. It makes sense that this does not get captured.

ulab:
While I agree that it is minor, I'd like to make the case that having a proper "end" marking in the log can be important to decide whether the printer crashed or the logging was ended properly.

It can also help a support member to see if he received a partial file.

CZDanol:
Ok, those are very good points

CZDanol:
Internal ticket: BFW-6789

bkerler:
I'd say as this was already merged, the issue can be closed ? :)

ulab:
If it's fixed, I agree. 

[...]

PR #4491, opened 2025-02-19
logging: Ensure file logging end message is written

bkerler:
This implements two fixes according to issue #4478:

1. Make sure that file logging is stopped if already opened on enabling file log.
2. Make sure that the message "Stopped logging to file" is written to the log as well so that a crash can be ruled out.

### DIFF
--- a/src/logging/log_dest_file.cpp
+++ b/src/logging/log_dest_file.cpp
@@ -28,6 +28,19 @@ namespace file {
         unique_file_ptr file;
 
         BufferChunk wip_chunk;
+
+        void write_buffer() {
+            while (!buffer.isEmpty()) {
+                BufferChunk chunk = buffer.dequeue();
+                fwrite(chunk.data.data(), 1, chunk.size, file.get());
+            }
+        }
+
+        ~Data() {
+            if (file) {
+                write_buffer();
+            }
+        }
     };
 
     std::atomic_bool is_enabled = false;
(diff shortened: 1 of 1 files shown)


src/logging/log_dest_file.cpp
```diff
@@ -136,6 +139,10 @@ void file_log_disable() {
     log_info(FileSystem, "Stopped logging to file");
 
     std::scoped_lock guard(mutex);
+    while (!data->buffer.isEmpty()) {
```
CZDanol, line 142: Code duplication with `file_log_write`. Please unify this in a dedicated `file_log_write_buffer`. The overflow checking should be also included (but not the ferror)

CZDanol, line 142: It is also necessary to call `flush_chunk` before

CZDanol, line 142: A better place for this cleanup logic would be in `~Data()`. Don't forget to check for the file not being null

src/logging/log_dest_file.cpp
```diff
@@ -115,6 +115,9 @@ bool file_log_enable(const char *filepath) {
     if (!d) {
         return false;
     }
+    if (is_enabled) {
```
CZDanol, line 118: This if will not be necessary once you move the flush logic into `~Data()`

src/logging/log_dest_file.cpp
CZDanol: I was wrong, disregard this

CZDanol: You forgot to check the file for being null in the destructor

CZDanol (approved):
Good job 👍

Issue #4553, opened 2025-03-23
[ENHANCEMENT] Up to 1.8mm nozzles selectable in firmware


FyrbyAdditive:
### Printer model

XL

### Describe the enhancement

Currently it is possible to select up to 1.2mm nozzles on the printer. It is possible to use larger nozzles by overriding this in the slicer, and then going to the printer and manually interacting with it to accept the settings and then confirm.

1.4 and 1.8mm nozzles for the V6 adaptor are available and usable with a little effort (as pictured) - a great enhancement would be to allow up to 1.8mm nozzles in firmware to alleviate us of the need to go and manually poke the printer when we would like to use them.

Hopefully this is "low hanging fruit" and just editing a variable somewhere!

### Expected functionality

![Image](https://github.com/user-attachments/assets/0fd395f4-7782-4a29-ac33-76eead5b7cc4)
![Image](https://github.com/user-attachments/assets/9a4ef420-cfe9-436f-9415-6ec384f62789)

FyrbyAdditive:
Thank you!

FyrbyAdditive:
A workaround for anyone else reading this that wants to see this reflected on the printer seems to be to set it using Prusa Connect, and the XL seems to display the nozzle size correctly then.

CZDanol:
Internal ticket: BFW-7007

Prusa-Support:
This feature has been implemented in FW 6.3.3 for Prusa CO and should be ported to Prusa XL with the future universal firmware upgrade FW 6.4.

> Nozzle size can now be set up to 1.8 mm. Thanks to @bkerler!

https://github.com/prusa3d/Prusa-Firmware-Buddy/releases/tag/v6.3.3

Michele Moramarco
Prusa Research

schrodlm:
Closing this issue, thanks again @bkerler!

PR #4565, opened 2025-03-28
[BFW-7007] nozzle_diameter: Change max nozzle diameter to 1.8mm

bkerler:
This solves issue #4553, and changes the max nozzle diameter from 1.2mm to 1.8mm.

### DIFF
--- a/include/common/nozzle_diameter.hpp
+++ b/include/common/nozzle_diameter.hpp
@@ -4,7 +4,7 @@
 
 static constexpr NumericInputConfig nozzle_diameter_spin_config {
     .min_value = 0.2,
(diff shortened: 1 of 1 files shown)


CZDanol (approved):

CZDanol:
Internal ticket: BFW-7007

bkerler:
> Do we want to only change it to 1.8? What if someone else comes a week later, asking for 1.9?
I don't think 1.9 does make any sense with 1.75mm filament ? :D 
1.8 clearly is the max.

CZDanol:
Lol, yep, makes sense

PR #4595, opened 2025-04-12
[BFW-7072] gcode: Fix issues on gcode parsing with inline comments

bkerler:
Using this gcode:
`M862.3 P "MK4S" ; Printer model check`
instead of
`M862.3 P "MK4S" ; printer model check`
will lead to an error due to incorrect printer type.

The reason is that the "GCodeInfo::parse_gcode" function only skips comments at the beginning, but not inline comments,
thus leading in an incorrect parsing of comments as parameters, which in this case results that the printer model is parsed incorrectly.

### DIFF
--- a/src/common/gcode/gcode_info.cpp
+++ b/src/common/gcode/gcode_info.cpp
@@ -420,6 +420,13 @@ void GCodeInfo::parse_gcode(GcodeBuffer::String cmd, uint32_t &gcode_counter) {
     if (cmd.is_empty() || cmd.front() == ';') {
         return;
     }
+
+    // Make sure we don't parse inline comments
+    const auto comment = std::find(cmd.begin, cmd.end, ';');
+    if (comment != cmd.end) { // only if comments are found
+        cmd = GcodeBuffer::String(cmd.begin, comment);
+    }
+
     gcode_counter++;
 
     // skip line number if present


bwees:
Small clarification, Orca does not "use" capitalized gcode comments in the sense that is the default for every profile. I am developing CORE One profiles for OrcaSlicer and stumbled across this bug when I accidentally capitalized the comments at some point. 

Doesn't change the fact that its a bug and should be fixed though! Thanks @bkerler for helping debug and putting this together. :)

CZDanol:
Internal ticket: BFW-7072

src/common/gcode/gcode_info.cpp
```diff
@@ -497,6 +497,13 @@ void GCodeInfo::parse_gcode(GcodeBuffer::String cmd, uint32_t &gcode_counter) {
     if (cmd.front() == ';' || cmd.is_empty()) {
         return;
     }
+
+    // Make sure we don't parse inline comments
+    auto equal = std::find(cmd.begin, cmd.end, ';');
```
CZDanol: nitpick: const

CZDanol: Why `equal`? What does it say to the reader?

CZDanol:
Ah, this is in GCodeInfo. GCode parsing in GCodeInfo is long overdue for a rewrite - I will try proposing that and see if we get space for it. Based on that, we would go through with this PR review.

src/common/gcode/gcode_info.cpp
bkerler: Done :)

bkerler: changed it to comment_line

CZDanol: Why line?

bkerler: It's now only "comment" :)

Issue #4542, opened 2025-03-17
[BFW-7076] [FEATURE REQUEST] G-code to get loaded filament and nozzle details by serial


L10nhunter:
### Printer model

Prusa MK4(S)

### Describe the feature

I'm writing my own version of prusalink as a fun little side project, and i realized that I don't have any way to check if the filament in the g-code is the same as the filament in the printer. then i realized there was also no way to check ensure nozzle size or filament diameter either. the thing is, the printer already has the functionality inside itself, because if i try to run a print that has a mismatched filament material, diameter, or nozzle size, the printer screen will warn me about it. why can't there also be a way to check those over serial?

### Expected functionality

if an MK4(S) printer is attached to a computer by an open serial port, the computer should be able to write to the serial port with a g or m code, and the printer would reply over the same serial port with that information. optional improvements would be to add arguments to only return one object in the list, instead of all of them.

bkerler:
For nozzle diameter, there is the gcode "M862.1 Q".

bkerler:
For printing the filament type, parameter H has been added in PR #4600

CZDanol:
Internal ticket: BFW-7076

PR #4600, opened 2025-04-13
[BFW-7076] gcode: rework M865

bkerler:
This solves issue #4542 by adding an option to print the filament type for a specific tool slot:

`M865 I0` will print for tool slot 0 (Extruder 1):

image

Furthermore:
- Rename F<type> to S<type> due to conflict with F (Required filtration), update documentation
- Fix max value of Extruder slot for L<ix>, update documentation
- Change behavior of I<ix> for changing/getting the current filament, update documentation

### DIFF
--- a/src/marlin_stubs/M865.cpp
+++ b/src/marlin_stubs/M865.cpp
@@ -13,8 +13,7 @@
  * - `I<ix>` - Configure parameters of a filament currently loaded to the specified tool (indexed from 0)
  * - `U<ix>` - Configure parameters of a User filament (indexed from 0)
  * - `X` - Configure parameters of a Custom filament type that will be loaded using `M600 F"##"` (or similar filament change gcode)
- * - `F"<preset>"` - Configure parameters of User filament with this name  (or select Preset filament for `L`)
- *
+ * - `S"<preset>"` - Configure parameters of filament with this name  (or select Preset filament for `L`)
  * - `L<ix>` - Set currently loaded filament for the given tool to the selected filament
  *
  * - `R` - Reset parameters not specified in this gcode to defaults
@@ -23,12 +22,13 @@
  * - `P` - Nozzle preheat temperature
  * - `B` - Bed temperature
  * - `A` - Is abrasive
- * - `F` - Requries filtration
+ * - `F` - Requires filtration
  * - `N"<string>"` - New filament name
  *
  * Ad-hoc/custom filaments can the be referenced in other gcodes using adhoc_filament_gcode_prefix.
  * For example `M600 S"#0"` will load ad-hoc filament previously set with `M865 I0`.
- * The current filament settings of the specified tool (indexed from 0) are printed to the serial console
+ * The filament_type settings are printed to the serial console. For option I, the current filament_type
+ * setting of the specified tool (indexed from 0) is displayed instead.
  */
 void PrusaGcodeSuite::M865() {
     GCodeParser2 p;
@@ -47,7 +47,7 @@ void PrusaGcodeSuite::M865() {
     } else if (const auto slot = p.option<uint8_t>('U', static_cast<uint8_t>(0), static_cast<uint8_t>(user_filament_type_count - 1))) {
         filament_type = UserFilamentType { .index = *slot };
 
-    } else if (const auto ft = p.option<FilamentType>('F')) {
+    } else if (const auto ft = p.option<FilamentType>('S')) {
         filament_type = *ft;
 
     } else {
--- a/src/marlin_stubs/M865.cpp
+++ b/src/marlin_stubs/M865.cpp
@@ -28,6 +28,7 @@
  *
  * Ad-hoc/custom filaments can the be referenced in other gcodes using adhoc_filament_gcode_prefix.
  * For example `M600 S"#0"` will load ad-hoc filament previously set with `M865 I0`.
+ * The current filament settings of the specified tool (indexed from 0) are printed to the serial console
  */
 void PrusaGcodeSuite::M865() {
     GCodeParser2 p;
@@ -96,6 +97,18 @@ void PrusaGcodeSuite::M865() {
 
     if (auto load = p.option<uint8_t>('L', static_cast<uint8_t>(0), static_cast<uint8_t>(EXTRUDERS))) {
         config_store().set_filament_type(*load, filament_type);
+    }
+
+    if (filament_type != FilamentType::none) {
+        ArrayStringBuilder<filament_name_buffer_size + 82> filament_info;
+        filament_info.append_printf("name:%s\n", params.name);
+        filament_info.append_printf("nozzle_temperature:%d\n", params.nozzle_temperature);
+        filament_info.append_printf("heatbed_temperature:%d\n", params.heatbed_temperature);
+        filament_info.append_printf("is_abrasive:%d\n", params.is_abrasive);
+#if HAS_CHAMBER_API()
+        filament_info.append_printf("requires_filtration:%d\n", params.requires_filtration);
+#endif
+        SERIAL_ECHOLN(filament_info.str());
     }
 }
 
--- a/src/marlin_stubs/M865.cpp
+++ b/src/marlin_stubs/M865.cpp
@@ -10,7 +10,7 @@
  *### M865: Configure filament parameters
  *
  *#### Parameters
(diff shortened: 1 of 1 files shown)


src/marlin_stubs/M865.cpp
```diff
@@ -36,6 +37,10 @@ void PrusaGcodeSuite::M865() {
     }
 
     FilamentType filament_type;
+    if (const auto slot = p.option<uint8_t>('H', static_cast<uint8_t>(0), static_cast<uint8_t>(adhoc_filament_type_count - 1))) {
```
CZDanol: `adhoc_filament_type_count` is not the right variable to check against here. You wanna do `EXTRUDERS`

src/marlin_stubs/M865.cpp
```diff
@@ -25,6 +25,7 @@
  * - `A` - Is abrasive
  * - `F` - Requries filtration
  * - `N"<string>"` - New filament name
+ * - `H`<i>` - Print current filament type loaded to the specified tool (indexed from 0)
```
CZDanol: ˙H˙ is already taken in our trunk code unfortunately

CZDanol (changes requested):
While the idea is noble, the implementation is rather unconceptual. Let's try to redesign this feature in the spirit of this gcode:

1. Add a "selector" for a filament loaded in a specific tool. There is already the `I<ix>` for the "Custom" filament type in a specified tool - we could simply change that to the current filament loaded to the specified tool, only adding functionality
2. Do we need a print option? Why not print things about the filaments always, in a parseable manner? And why limit ourselves only to the filament name?

src/marlin_stubs/M865.cpp
```diff
@@ -51,7 +48,6 @@ void PrusaGcodeSuite::M865() {
 
     } else if (const auto ft = p.option<FilamentType>('F')) {
         filament_type = *ft;
-
```
CZDanol, line 54: Please keep this line

src/marlin_stubs/M865.cpp
```diff
@@ -37,12 +38,8 @@ void PrusaGcodeSuite::M865() {
 
     FilamentType filament_type;
 
-    if (const auto slot = p.option<uint8_t>('I', static_cast<uint8_t>(0), static_cast<uint8_t>(adhoc_filament_type_count - 1))) {
-        filament_type = AdHocFilamentType { .tool = *slot };
-        if (config_store().get_filament_type(*slot) != filament_type) {
-            SERIAL_ERROR_MSG("The selected tool does not have the ad-hoc filament loaded. Changes will have no effect.");
-        }
-
+    if (const auto slot = p.option<uint8_t>('I', static_cast<uint8_t>(0), static_cast<uint8_t>(EXTRUDERS))) {
```
CZDanol: We also need to change the documentation

src/marlin_stubs/M865.cpp
```diff
@@ -37,12 +38,8 @@ void PrusaGcodeSuite::M865() {
 
     FilamentType filament_type;
 
-    if (const auto slot = p.option<uint8_t>('I', static_cast<uint8_t>(0), static_cast<uint8_t>(adhoc_filament_type_count - 1))) {
-        filament_type = AdHocFilamentType { .tool = *slot };
-        if (config_store().get_filament_type(*slot) != filament_type) {
-            SERIAL_ERROR_MSG("The selected tool does not have the ad-hoc filament loaded. Changes will have no effect.");
-        }
-
+    if (const auto slot = p.option<uint8_t>('I', static_cast<uint8_t>(0), static_cast<uint8_t>(EXTRUDERS - 1))) {
```
CZDanol, line 41: The documentation for this parameter should also be updated

src/marlin_stubs/M865.cpp
```diff
@@ -1,4 +1,5 @@
 #include "PrusaGcodeSuite.hpp"
+#include "marlin_vars.hpp"
```
CZDanol: We don't need this include

src/marlin_stubs/M865.cpp
bkerler: Commit format tool doesn't like this line :)

src/marlin_stubs/M865.cpp
CZDanol, line 40: Changing behavior (and documentation) of the I parameter should be a separate commit

src/marlin_stubs/M865.cpp
```diff
@@ -11,6 +11,7 @@
  *
  *#### Parameters
  * - `I<ix>` - Configure parameters of a Custom filament currently loaded to the specified tool (indexed from 0)
+ * - `I<i>`  - Show the parameters of the filament currently loaded to the specified tool (indexed from 0)
```
CZDanol: - The rest is using the "Configure parameters" text. I'd suggest either updating all texts, or keeping it unified.
- The rest is using "ix", not "i"
- Also, you've duplicated the documentation for I.

src/marlin_stubs/M865.cpp
```diff
@@ -28,6 +28,7 @@
  *
  * Ad-hoc/custom filaments can the be referenced in other gcodes using adhoc_filament_gcode_prefix.
  * For example `M600 S"#0"` will load ad-hoc filament previously set with `M865 I0`.
+ * The current filament settings of the specified tool (indexed from 0) are printed to the serial console
```
CZDanol: >  The current filament settings of the specified tool
Only if you use the `I` selector. Is not true if you use any other selector.

src/marlin_stubs/M865.cpp
```diff
@@ -13,7 +13,7 @@
  * - `I<ix>` - Configure parameters of a filament currently loaded to the specified tool (indexed from 0)
  * - `U<ix>` - Configure parameters of a User filament (indexed from 0)
  * - `X` - Configure parameters of a Custom filament type that will be loaded using `M600 F"##"` (or similar filament change gcode)
- * - `F"<preset>"` - Configure parameters of User filament with this name  (or select Preset filament for `L`)
```
CZDanol, line 16: This commit needs a description explaining why this change was needed.

src/marlin_stubs/M865.cpp
```diff
@@ -13,8 +13,7 @@
...
- *
+ * - `S"<preset>"` - Configure parameters of filament with this name  (or select Preset filament for `L`)
```
CZDanol: Nitpick: You left `<preset>` here - should be `<name>` now

src/marlin_stubs/M865.cpp
```diff
@@ -23,12 +22,13 @@
...
  * - `B` - Bed temperature
  * - `A` - Is abrasive
- * - `F` - Requries filtration
+ * - `F` - Requires filtration
  * - `N"<string>"` - New filament name
  *
  * Ad-hoc/custom filaments can the be referenced in other gcodes using adhoc_filament_gcode_prefix.
  * For example `M600 S"#0"` will load ad-hoc filament previously set with `M865 I0`.
- * The current filament settings of the specified tool (indexed from 0) are printed to the serial console
+ * The filament_type settings are printed to the serial console. For option I, the current filament_type
```
CZDanol: Nitpick: Why do you keep referring to the `I` parameter here, like it's something special?

CZDanol (approved):

CZDanol:
I have now checked and the buffer size you selected in `ArrayStringBuilder<filament_name_buffer_size + 82> filament_info;` doesn't even cover the key names length. Rather than going another review round, I will fix this privately and close this PR.

CZDanol (changes requested):
I have now checked and the buffer size you selected in ArrayStringBuilder<filament_name_buffer_size + 82> filament_info; doesn't even cover the key names length. Rather than going another review round, I will fix this privately and close this PR.

src/marlin_stubs/M865.cpp
bkerler: Done

src/marlin_stubs/M865.cpp
bkerler: Correct, removed that.

bkerler:
Meanwhile I've updated the PR following your comments and tested it as well :)

CZDanol:
Well, I have fixed it and merged in our codebase a month ago, so that was pointless. I just forgot to close the PR here - so doing it now. Sorry.

PR #4604, opened 2025-04-14
[BFW-7076] gcode: M865 fix EXTRUDER param max size for L option

bkerler:
This corrects an issue in gcode M865, correcting the wrong max size for the EXTRUDER specified parameter in the 'L' option.


CZDanol (approved):

CZDanol:
Internal ticket: BFW-7076

PR #4647, opened 2025-05-11
Restore spreadCycle in G28

espr14:
G28 can switch to stealthChop but did not revert the changes.


bkerler:
Hi ! Can you please explain why this is needed ?

espr14:
1. Either Original Prusa printers use stealthChop all the time (XY) so there's no reason to add extra code that sets stealthChop again. M569 should be disabled as well.
2. Or user can switch to spreadCycle but G28 switches to stealthChop and does not return to spreadCycle. So user has to add extra G code after each G28. That's not nice.

This change adds few bytes to the FW size and does not alter behavior or performance of original printers which uses original profiles.

bkerler:
Thanks !

bkerler:
@espr14 Please update to latest master. The current master seems to have incorporated your fixed at first sight, so if your PR is no longer needed, please close it. Thanks !

espr14:
Master is 4 months old, release 2 months. And there's no change. The rest is private. So I'll wait for the next release.

bkerler:
You need to check the tags. Current firmware is https://github.com/prusa3d/Prusa-Firmware-Buddy/tree/v6.5.3. Not sure how soon master will be updated. Porting to current master is required for reviewing the PRs, but in the end it's your choice if you want to do that of course :)

espr14:
Rebased to current master.

Issue #4479, opened 2025-02-14
[BFW-6788] [BUG] Core One has no CHAMBER_TEMPERATURE capability and does not report C temperature


ulab:
### Printer model

Core One

### Firmware version

6.3.0+10073

### Upgrades and modifications

None

### Printing from...

does not apply

### Describe the bug

The Core One reports it has no CHAMBER_TEMPERATURE capability on `M115`:

```
Send: M115
FIRMWARE_NAME:Prusa-Firmware-Buddy 6.3.0+10073 (Github) SOURCE_CODE_URL:https://github.com/prusa3d/Prusa-Firmware-Buddy PROTOCOL_VERSION:1.0 MACHINE_TYPE:Prusa-COREONE EXTRUDER_COUNT:1 UUID:cede2a2f-41a2-4748-9b12-c55c62f367ff
                                                                                                                                                                                                                                  Cap:SERIAL_XON_XOFF:0
[...]
Cap:CHAMBER_TEMPERATURE:0
ok
```

It also does not report the chamber temperature (`C`) on `M105` or `M155`?
```
T:26.50/0.00 B:59.83/90.00 X:26.98/36.00 A:40.91/0.00 @:0 B@:255 HBR@:0
```



### How to reproduce

Send the above GCodes `M115`, `M105` and `M155 ...` over the serial interface.

### Expected behavior

`CHAMBER_TEMPERATURE` should show `1` and the temperature reports should have a `C` with the chamber temperatures.


### Files

_No response_

ulab:
I am guessing the cause for this is that Marlin [expects active heating for the chamber](https://github.com/prusa3d/Prusa-Firmware-Buddy/blob/310b7f890f09c1318e264c46aed80e7ecf94dd93/lib/Marlin/Marlin/src/gcode/host/M115.cpp#L169)?

`HAS_TEMP_CHAMBER` instead of `HAS_HEATED_CHAMBER` might be enough, but I am not sure if changing the capabilities this way will mess with something else.


But then I am not sure why it doesn't print `C` on `M105` or `M155`. The code [looks like it should](https://github.com/prusa3d/Prusa-Firmware-Buddy/blob/310b7f890f09c1318e264c46aed80e7ecf94dd93/lib/Marlin/Marlin/src/module/temperature.cpp#L3726-L3728).

CZDanol:
Oh I was wondering when someone would come up with this :)

[...]

PR #4671, opened 2025-05-28
octoprint: Fix serial output from chamber power to expected chamber temp format

bkerler:
This changes the C@: output which shall be used for chamber power only to C:x.x/m.m where x is current chamber temp and m is target chamber temp in order to match the expected behaviour, see [reference](https://reprap.org/wiki/G-code#M105:_Get_Extruder_Temperature). Also separated from chamber power define to allow chamber temp being used if chamber api and chamber temp is supported, even if chamber power isn't defined.

As HAS_HEATER_CHAMBER and HAS_TEMP_CHAMBER are old Merlin defines and HAS_CHAMBER_API() is used in favor, we prioritize HAS_CHAMBER_API for now until the old Merlin code has been removed.
Solves issues #4800 and #4479

### DIFF
--- a/lib/Marlin/Marlin/src/module/temperature.cpp
+++ b/lib/Marlin/Marlin/src/module/temperature.cpp
@@ -1014,6 +1014,29 @@ void Temperature::isr() {
       print_heater_state(degBed(), degTargetBed(), H_BED);
     #endif
 
+    #if HAS_CHAMBER_API()
+      {
+        const auto chamber_capabilities = buddy::chamber().capabilities();
+        if (chamber_capabilities.temperature_reporting) {
+          const auto current_chamber_temperature = buddy::chamber().current_temperature();
+          const auto target_chamber_temperature = buddy::chamber().target_temperature();
+          SERIAL_ECHOPAIR(" C:", current_chamber_temperature.value_or(0.0f));
+          SERIAL_ECHOPAIR("/", target_chamber_temperature.value_or(0.0f));
+        }
+      }
+    #else
+      #if HAS_TEMP_CHAMBER
+        print_heater_state(degChamber()
+        #if HAS_HEATED_CHAMBER
+          , degTargetChamber()
+        #else
+          , 0
+        #endif
+        , H_CHAMBER
+      );
+      #endif
+    #endif // HAS_CHAMBER_API()
+
     #if HAS_TEMP_HEATBREAK
       print_heater_state(degHeatbreak(target_extruder)
           , degTargetHeatbreak(target_extruder)
@@ -1037,9 +1060,8 @@ void Temperature::isr() {
     #if HAS_HEATED_BED
       SERIAL_ECHOPAIR(" B@:", getHeaterPower(H_BED));
     #endif
-    #if HAS_CHAMBER_API()
-      auto current_chamber_temperature = buddy::chamber().current_temperature();
-      if (current_chamber_temperature.has_value()) SERIAL_ECHOPAIR(" C@:", current_chamber_temperature.value());
+    #if HAS_HEATED_CHAMBER
+      SERIAL_ECHOPAIR(" C@:", getHeaterPower(H_CHAMBER));
     #endif
 
     #if HAS_TEMP_HEATBREAK
--- a/src/marlin_stubs/feature/chamber/M141_M191.cpp
+++ b/src/marlin_stubs/feature/chamber/M141_M191.cpp
@@ -80,6 +80,7 @@ void PrusaGcodeSuite::M141_no_parser(const M141Args &args) {
 
     if (!chamber().capabilities().temperature_control()) {
         SERIAL_ERROR_MSG("Chamber does not allow temperature control");
+        return;
     }
 
     auto target = args.target_temp;


ulab:
I am not an expert on these values, but does software expect it to be in the form `C:1.0/0.0` where `1.0` is the current chamber temp and `0.0` the target temp?

All other temperatures seem to be in that format.
```
Recv: T:230.00/230.00 B:60.02/60.00 X:37.45/36.00 A:48.65/0.00 @:137 B@:73 C@:31.61 HBR@:255
```
https://reprap.org/wiki/G-code#M105:_Get_Extruder_Temperature seems to be a little vague on this:
> A temperature report will usually include actual and target temperature for all available heaters, with the format being "actual/target" or - for some firmware variants - "actual /target". During a blocking heatup some firmware variants only report the temperature tuple for the heater that is currently in blocking heatup state. 

The ambient temperature is being reported as `A:48.65/0.00` too and it hopefully doesn't have a target temp set by the printer ;-).

bkerler:
@ulab Good point, thanks ! I've changed the PR to reflect the expected format.

bkerler:
The format seems to accept both C:49.3 and C:49.3/39.3 .... but as we already have a function to get the target chamber temperature, it does sound right to also add it to be parsed :)

lib/Marlin/Marlin/src/module/temperature.cpp
```diff
@@ -3846,9 +3846,11 @@ void Temperature::isr() {
     #endif
     #if HAS_HEATED_CHAMBER
       SERIAL_ECHOPAIR(" C@:", getHeaterPower(H_CHAMBER));
-    #elif HAS_CHAMBER_API()
+    #endif
+    #if HAS_CHAMBER_API()
       auto current_chamber_temperature = buddy::chamber().current_temperature();
-      if (current_chamber_temperature.has_value()) SERIAL_ECHOPAIR(" C@:", current_chamber_temperature.value());
+      auto target_chamber_temperature = buddy::chamber().target_temperature();
+      if (current_chamber_temperature.has_value()) SERIAL_ECHOPAIR(" C:", current_chamber_temperature.value(), "/", target_chamber_temperature.value());
```
CZDanol: Do you think ECHOPAIR will work with four parameters?

CZDanol: Not checking whether target_chamber_temperature has value

lib/Marlin/Marlin/src/module/temperature.cpp
```diff
@@ -3846,9 +3846,11 @@ void Temperature::isr() {
     #endif
     #if HAS_HEATED_CHAMBER
       SERIAL_ECHOPAIR(" C@:", getHeaterPower(H_CHAMBER));
-    #elif HAS_CHAMBER_API()
+    #endif
```
CZDanol, line 3849: Why this change?

lib/Marlin/Marlin/src/module/temperature.cpp
bkerler: Yes, at least there are several codelines using multiple arguments (see binary_protocol.h, line 170).
But for readability, I've now split it up to two SERIAL_ECHOPAIR :)

bkerler: Absolutely right. Fixed that.

lib/Marlin/Marlin/src/module/temperature.cpp
```diff
@@ -3846,9 +3846,14 @@ void Temperature::isr() {
...
+    #if HAS_CHAMBER_API()
       auto current_chamber_temperature = buddy::chamber().current_temperature();
-      if (current_chamber_temperature.has_value()) SERIAL_ECHOPAIR(" C@:", current_chamber_temperature.value());
+      auto target_chamber_temperature = buddy::chamber().target_temperature();
+      if (current_chamber_temperature.has_value()){
+        SERIAL_ECHOPAIR(" C:", current_chamber_temperature.value());
+        SERIAL_ECHOPAIR("/", target_chamber_temperature.has_value() ? target_chamber_temperature.value() : 0.0f);
```
CZDanol: There is `std::optional::value_or`. In that case, you don't even need to store target_chamber_temperature into a separate variable

lib/Marlin/Marlin/src/module/temperature.cpp
```diff
@@ -3846,9 +3846,14 @@ void Temperature::isr() {
     #endif
     #if HAS_HEATED_CHAMBER
       SERIAL_ECHOPAIR(" C@:", getHeaterPower(H_CHAMBER));
-    #elif HAS_CHAMBER_API()
+    #endif
+    #if HAS_CHAMBER_API()
       auto current_chamber_temperature = buddy::chamber().current_temperature();
-      if (current_chamber_temperature.has_value()) SERIAL_ECHOPAIR(" C@:", current_chamber_temperature.value());
+      auto target_chamber_temperature = buddy::chamber().target_temperature();
+      if (current_chamber_temperature.has_value()){
```
CZDanol: Nitpick/tip: You can do `if(auto temp = buddy::chamber().current_temperature())`

lib/Marlin/Marlin/src/module/temperature.cpp
bkerler: Cool :) Thanks for the tip !

lib/Marlin/Marlin/src/module/temperature.cpp
bkerler: will do :)

bkerler: In this case, we only want to print the chamber temperature if it has a value. The if on std::optional doesn't seem to check if it does contain a value, thus this isn't applicable. Otherwise this would have been great :)

TLMcNulty:
Would love to see this merged and released! In discussion with some octo contributors it looks good.

graelo:
Hi, thanks for this work!
I'm curious about why this PR isn't making progress anymore, can you share? Oh, by the way, unfortunately we currently can't access the Jenkins logs.

graelo:
Thank you @bkerler ❤️

3d-gussner:
>Oh, by the way, unfortunately we currently can't access the Jenkins logs.

FYI: Jenkins has been disabled on public PRs.

3d-gussner:
@bkerler Can you please check if that hasn't been implemented in 6.4.0-alpha or 6.4.0-rc?

bkerler:
Sure ! Will have a look

BlueFyre:
The RC seems to still have the old one: https://github.com/prusa3d/Prusa-Firmware-Buddy/blob/9b4e96e276304894175ab9e8030bf213df8546b7/lib/Marlin/Marlin/src/module/temperature.cpp#L3012

bkerler:
@3d-gussner as @BlueFyre correctly already stated, the RC still does have the old behaviour and thus isn't fixed.

3d-gussner:
@bkerler @BlueFyre Thanks for reviewing, forwarded the information to the developers.

ruebarb:
Same issue for C1L

bkerler:
Reworked the PR to be much cleaner and with octoprint in mind

tzahari:
Can someone merge this please!?

o-vesely:
Hey hey, we finaly found time to implement this, thanks a bunch. Will be in 6.11 onwards.

PR #4674, opened 2025-05-29
XL: Disable temp sensor chamber as it doesn't exist, causing wrong chamber temp report

bkerler:
This solves issue #3848.

Issue: If the device doesn't have a heated chamber, then chamber temps shouldn't be reported via gcode M155.

Solution: on XL, TEMP_SENSOR_CHAMBER was defined as 2000, which was invalid. Setting the value to 0 disables the TEMP_SENSOR_CHAMBER, thus M155 no longer will report a chamber temperature that doesn't exist.

### DIFF
--- a/include/marlin/Configuration_XL.h
+++ b/include/marlin/Configuration_XL.h
@@ -154,6 +154,10 @@
 //===========================================================================
 
 #define TEMP_SENSOR_BED 5
+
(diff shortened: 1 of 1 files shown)


Ro3Deee:
prusa XL has temp chamber senzor:
image
You can put a heater thermistor as I did:
image

bkerler:
@Ro3Deee Thanks for pointing out, but the chamber temp sensor used in Marlin isn't used in Prusa XBuddy Firmware, as Prusa uses its own Chamber API for that. Also you are referring to the ambient temp sensor.

Issue #4246, opened 2024-10-10
[BUG] MK4S / 6.1.3 Wrong count on display during bed probe (cosmetical issue)


Mike-FUT:
### Printer model

MK4S

### Firmware version

6.1.3

### Upgrades and modifications

None

### Printing from...

Prusa Connect

### Describe the bug

When starting a print before the actual extrusion start teh printer runs bed levelling.
During levelling it probes different points, for example 1/20, 2/20, 3/20, etc. until it's finished and it ends with 20/20.

After that it probes the front area of the bed (nozzle clean/purge area) with very few points (depending on MMU3 usage).
When doing so it let't say it probes two points it will still use the old "20" from earlier. So for the two points it will start with probing 1/20, 2/20. Then it's done.
20 is obviously wrong in this case and I guess the firmware is accessing a variable from earlier.

This is plain cosmetical. It causes zero issues. This is just if you don't want wrong information on the screen to be displayed then you might be looking into it.

### How to reproduce

Start any print and look at the screen during bed levelling.

### Expected behavior

After main bed levelling, when probing the front part (nozzle purge/clean area) either display the correct amount of points to be probed or just ditch the amount of points completely at that stage, so for the two points in the example:

Probing point 1
Probing point 2

or

Probing point 1/2
Probing point 2/2

(instead of the wrong 1/20, 2/20 from the main bed probing points count)

cdbfoster:
Bugged me too!  Felt simple enough that I dug around for it this morning.

PR #4733, opened 2025-07-27
[BFW-7536] Report the correct number of probe points during bed leveling within a specified area.

cdbfoster:
### What
Calculate the point total in the `Probing Point <current>/<total>` UI message from the passed-in area instead of the current print footprint.

### Why
The current UI behavior works fine for probing the print footprint itself, but when the slicer default gcode then goes on to probe the area in the corner for its purge line, it's still reporting a total calculated from the print footprint even though it only needs to probe 3 points (in the CORE One case).

### How
Use the specified area in the point calculation instead of always using `print_area`.  Still default to `print_area` when no area is specified.  I haven't tested this change, since I haven't made the physical modifications necessary to flash custom firmware, but the change is simple enough that maybe someone else can verify that it works.

Addresses https://github.com/prusa3d/Prusa-Firmware-Buddy/issues/4246

### DIFF
--- a/lib/Marlin/Marlin/src/feature/bedlevel/ubl/ubl_G29.cpp
+++ b/lib/Marlin/Marlin/src/feature/bedlevel/ubl/ubl_G29.cpp
@@ -314,7 +314,7 @@
   #if PRINTER_IS_PRUSA_MK3_5() || PRINTER_IS_PRUSA_MINI()
 
   // Apply weighted correction on each point based on it's location.
-  // The whole correction is then conversed from µm to mm.
+  // The whole correction is then converted from µm to mm.
 
   float x_axis_correction(int x, int y) {
     int32_t left_correction_um{config_store().left_bed_correction.get()};
--- a/lib/Marlin/Marlin/src/feature/bedlevel/ubl/ubl_G29.cpp
+++ b/lib/Marlin/Marlin/src/feature/bedlevel/ubl/ubl_G29.cpp
@@ -314,14 +314,14 @@
   #if PRINTER_IS_PRUSA_MK3_5() || PRINTER_IS_PRUSA_MINI()
 
   // Apply weighted correction on each point based on it's location.
-  // The whole correction is then conversed from µm to mm. 
+  // The whole correction is then conversed from µm to mm.
 
   float x_axis_correction(int x, int y) {
     int32_t left_correction_um{config_store().left_bed_correction.get()};
     int32_t right_correction_um{config_store().right_bed_correction.get()};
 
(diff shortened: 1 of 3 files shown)


3d-gussner:
@cdbfoster Thanks for the PR, we will look into that. @CZDanol What do you think?

lib/Marlin/Marlin/src/feature/bedlevel/ubl/ubl_G29.cpp
```diff
@@ -314,14 +314,14 @@
   #if PRINTER_IS_PRUSA_MK3_5() || PRINTER_IS_PRUSA_MINI()
 
   // Apply weighted correction on each point based on it's location.
-  // The whole correction is then conversed from µm to mm. 
+  // The whole correction is then conversed from µm to mm.
```
CZDanol: Nitpick: typo conversed/converted

CZDanol (commented):
Hey, looking at the code, it is a nice work :) We have however already made a refactor of the probing point reporting for our next major release where we've fixed the count issue (and got rid of the count_points_to_probe function altogether) - so this PR is not fully mergeable into our codebase.

The refactorings are nice though and it would be a pity not to sue them, so I have merged most of this PR under the BFW-7536 ticket.

I would need to take more attention into the "fixing" part of the code though and since it is already rewritten, I don't want to spend time reviewing it.

But you should see your commits in the future versions of the firmware :) (not the next major release though, that one is already unbranched)

lib/Marlin/Marlin/src/feature/bedlevel/ubl/ubl_G29.cpp
cdbfoster: Ah, yeah, my IDE just stripped the trailing whitespace.  Looks like that's been there for more than a year.  Fixed it in any case.

cdbfoster:
@CZDanol Thanks for reviewing.  I guess I'm not sure how this works -- is this repo a public-facing mirror of a private internal repo or something?  Related, if the useful bits have been taken somewhere internal, should I close this PR?

Ro3Deee:
>  I guess I'm not sure how this works -- is this repo a public-facing mirror of a private internal repo or something? 

Prusa has a private repo of buddy fw, so we the people (🎯 other printer manufacturers) don't see how it is developed... I understand that some future features may be kept private until released, but for most of issues it not ok... I really enjoyed following prusa mk3 and mmu fw daily commits (the optimizations and fight for every byte are crazy in that fw)

3d-gussner:
@cdbfoster Please close the PR for the moment. We can reopen it when the changes in the next firmware release still have issues.

PR #4752, opened 2025-08-11
Fix German i18n for Z-Offset question

MartinEmrich:
The original question translated back to "Did you already run a Z-offset calibration". When truthfully answering "no", it will continue printing (on the slanted/badly aligned bed).

The English original text was more like "Do you want to run the Z-Offset now?", where the correct answer would be reversed.
The new German translation now asks the same question.


3d-gussner (dismissed):
Translation is correct in :de:

3d-gussner:
@MartinEmrich Danke für den PR. Thanks for the PR.

smoe (approved):

smoe:
Commit ff5b614af80e8d3af5a4e1bd5cea71cbec7f5564 (the translation) is fine and important. I am just slightly surprised about using the .po files for the translation directly as websites like weblate.org  are doing a fine job to orchestrate (and automate) translations.

MartinEmrich:
Did not notice this was still not merged. In the meantime my branch became non-rebaseable due to a force-push on master; and the commit even is not cherry-pickable: Apparently someone fixed the mistake 4 months ago with a slightly different, but also now correct translation. Closing this now.

@smoe I was not aware the translations are managed with a 3rd-party cloud service.

Issue #4754, opened 2025-08-11
[BFW-7600][BUG] Absorbing heat % wrong value


dmognaschi:
### Printer model

Core ONE

### Firmware version

6.3.4

### Upgrades and modifications

None

### Printing from...

Prusa Connect

### Describe the bug

Printing on flex material.
Ambient temperature: 34° C (summer).
Starting print, the 'absorbing heat' phase never ends and the percentage is huge (maybe a wrong cast of a negative number). See picture.
![Image](https://github.com/user-attachments/assets/851b347a-a416-49a7-8b2f-2a04d2c34d09)

f4ni:
Hi, can I work on this?

3d-gussner:
> Hi, can I work on this?

What do you mean? If you want to prepare a PullRequest it is always appreciated.

paynterf:
Any progress on this?  Had the same thing happen to me printing PP with heat bed set to zero.  'Absorbing Heat' shows the same (impossibly large) percentage as in OP's photo, and never ends.  Have to select 'Skip' to proceed

3d-gussner:
@paynterf FYI: The PR https://github.com/prusa3d/Prusa-Firmware-Buddy/pull/4765 has been reviewed yesterday and need some changes.

bkerler:
@paynterf can you please state exactly what you did and when this happens? I cannot reproduce it on my setup and in order to do the PR properly, it would be really helpful.

paynterf:
I just used the Prusa default settings for PolyPropylene, which has the bed temp set at zero.  When I attempt to print this, I got the 'large' heat soak percentage.  I'm in the middle of a long print right now, but I'll try to duplicate the problem in a way that you can reproduce

3d-gussner:
FYI: This issue should be solved in the next final release.

3d-gussner:
Please try firmware [6.4.0-RC2](https://github.com/prusa3d/Prusa-Firmware-Buddy/releases/tag/v6.4.0-RC2) and report back.

PR #4765, opened 2025-08-19
[BFW-7600]gui: Remaining preheat time - Fix int32 vs uint32 type confusion

bkerler:
This fixes issue #4754. 
The function required_preheat_time may return a negative int32_t value due to a multiplication of int16_t values and compiler optimizations happening in std:max ignoring a possible overflow . Returning the negative int32_t value as uint32_t does lead to an overflow of -1 minutes (uint32_t) as shown in issue #4754 as 4294967295%. Also fixes the printf format from being signed %i to unsigned %u to match the parameters.


lib/Marlin/Marlin/src/feature/bed_preheat.cpp
```diff
@@ -38,11 +38,12 @@ void BedPreheat::update() {
...
 
 uint32_t BedPreheat::required_preheat_time() {
-    if (thermalManager.degTargetBed() < minimal_preheat_temp) {
+    const uint32_t target_temp = uint32_t(thermalManager.degTargetBed());
+    if (target_temp < uint32_t(minimal_preheat_temp)) {
         return 0;
     }
 
-    return std::max((180 + (thermalManager.degTargetBed() - 60) * (12 * 60 / 50)) * 1000, 0);
+    return std::max<uint32_t>((180 + (target_temp - 60) * (12 * 60 / 50)) * 1000, int32_t(0));
```
CZDanol: This will not work properly, `std::max<uint32_t>` will cast the values to uint32_t before maxing, so the left side would always be >= 0.

bkerler: Fixed that issue.

bkerler:
Closed as a patch has been added to 6.4.0 RC

Issue #4338, opened 2024-12-09
[FEATURE REQUEST] stealth modes status via gcode


mogorman:
### Printer model

prusa xl

### Describe the feature

I would like to be able to query the current status of the stealthmode of the printer. like the other prusa firmware.
like described here https://github.com/prusa3d/Prusa-Firmware/pull/4388

### Expected functionality

`M915
M915 Q
ok
Mode: Silent
ok

M914
M914 Q
ok
Mode: Normal
ok

<Normal on display>
M914 Q
Mode: Normal
ok

<Change to Stealth on display>
M914 Q
Mode: Silent
ok`

CZDanol:
Okay, so apparently there is already M1940 for disabling the stealth mode and M9150 for enabling it :)

CZDanol:
Oh, we're missing a functionality to query it - I guess we will add it to one of these.

bkerler:
Sorry, took some time, but finally, here is the PR :)

3d-gussner:
Please verify if [Firmware 6.4.0-alpha](https://github.com/prusa3d/Prusa-Firmware-Buddy/releases/tag/v6.4.0-alpha) solves the issue.

Please use `M9141` to check it see https://github.com/prusa3d/Prusa-Firmware-Buddy/blob/v6.4.0-alpha/src/marlin_stubs/M9140_M9150.cpp#L29-L47

PR #4766, opened 2025-08-19
[BFW-6445]gcode: Add query of silent status using Q param

bkerler:
This solves issue #4338 by adding a "Q" parameter for M9140 and M9150 which does display if the silent mode is enabled or disabled.


3d-gussner:
@bkerler Thanks for the PR. Internally some work has been done as well. Added your PR to internal ticket.

bkerler:
Closed in favor of M9141

Issue #1485, opened 2021-06-24
[ENHANCEMENT] LCD Info diplayed - ADD Layer Number printing out of the total number of layers


Martenz:
**Printer type** - [MINI]
**Optional upgrades** - [LCD info displayed]

Would it be possible to add some more info in the LCD of the print status?

I really like to use Octoprint but I have to say that I now prefer to use only Prusa LCD to tune a bit velocity temp etc while printing from there.. What I still miss is just where the print is, to compare for instance with the PrusaSlicer sliced model.. the percentage is fine but would also be nice to see the layer number that is printing out of the total number of layers so if something is weird I can go to the layer and see if it is a problem of settings or of the 3D model I can improve for better prints.

Is this a feature I can ask here?

Does it make sense also for someone else?

Thanks a lot,

Martino.

bdwheele:
That would be great to have -- often I'm printing something and there's a "gotcha" point in the print so I'll track where I am in PrusaSlicer and comparing it to what's coming out of the printer.

Without layer information I'm subtracting the z-axis height and doing a best guess.  This feature would take away the math, so I'm all for it.

Martenz:
Exactly I would like to have ot for the same reason!
Finger crossed.

jh-vo:
Yes, this is really needed function! Add it, please!

x10102:
Yes please, this would be so much better than estimating the layer from the Z-height.

yetisnack1:
+1 for this feature.  Besides being beneficial for troubleshooting, it's useful for manual color swaps, adding magnets, providing temporary supports (ie. sticking your finger under a long overhang ;) .  I find the mini's zlayer height useless due to it constantly changing and would much prefer to have the actual layer number displayed.

Prusa-Support:
Hello! Thank you for the request and for the comments, sorry for the late response from my side. Request noted and brought to our devs attention, in case of more info from our side we'll update in this thread. 

Alessandro Pantaleo
Prusa Research

CZDanol:
Hey, this is being considered as a part of internal ticket BFW-5403. Not promising it will pass through though. Also, not sure if it will work with Octoprint.

jim-roberts:
There has not been update on this open issue for over year. What is the status?

bkerler:
[...]

PR #4775, opened 2025-08-25
gui: Add footer info for current z layer

bkerler:
This adds a footer that displays the current layer number, layer height is calculated using gcode metadata (includes layer_height and first_layer_height variables). Refers to issue #1485. 

Example:
image


CZDanol (changes requested):
I'm sorry, but this is imo not admissible. The algorithm used for calculating the layer height is not reliable. It report wrong numbers if variable layer length is used, and also during toolchanges, zhops, probings and so on.

The only proper solution that comes to my mind would require asking slicer to start emitting gcodes that would report the layer number.

bkerler:
@CZDanol yeah, understood. Moving this request to draft for now. Will keep you updated.

Issue #4157, opened 2024-08-27
[BUG] M862 command not functioning over serial.


Java-SE-Binary:
### Printer model

MK4s

### Firmware version

6.1.2

### Upgrades and modifications

MMU3 Installed

### Printing from...

Octoprint

### Describe the bug

When sending the M862 command to the printer, specifically with the parameter 'P', the printer does not actually perform the appropriate check for the feature that the command passed. For example, sending 'M862.6 P "MMU5" does not trigger a check on the printer for a supposed MMU5 to exist, and trigger the appropriate warning prompt.

'M862.X P' is not documented to be disabled when printing over a serial connection, and I feel it would be beneficial to remain enabled during serial printing, similar to how the MK3 printers behaved.

M862.X Q seems to work as intended, with the exception of M862.6 Q. When passing M862.6 Q to my MK4s with Input Shaper enabled and a MMU3 unit installed, the printer reports "Input shaper" as the only installed feature, despite MMU3 being present.

### How to reproduce

Simply sending any variations of M862 with the 'P' parameter will show this behavior.
M862.1 P<nozzle_diameter>
M862.2 P<model_code>
M862.3 P"<model_name>"
M862.4 P<fw_version>
M862.5 P<gcode_version>
M862.6 P"<installed_features>"

### Files

[Cube_0.4n_0.15mm_PLA_26.4228_1h49m.zip](https://github.com/user-attachments/files/16755081/Cube_0.4n_0.15mm_PLA_26.4228_1h49m.zip)
This G-Code file has manually edited M862 checks that will cause an incompatibility prompt when printed directly on the printer. When printing through Octoprint, this prompt _does not_ appear.

Java-SE-Binary:
This is still an issue with firmware 6.1.3. Calling the M862.X P command from OctoPrint still causes no response, and M862.6 Q still replies with incomplete information.

danopernis:
Assigning myself to prevent stalebot from closing. This is not a priority, but I feel like it should be fixed at some point.

DerMouse:
Hello,
Any progress on this annoying point ? Just sent a 0.4-sliced-gcode file to a 0.25-equipped MK4 (via OctoPrint), and the printer happilly started printing without any warning.
Firmware is 6.2.4.

bkerler:
[...]

PR #4777, opened 2025-08-25
gcode: Add serial information for M862.x for P and Q parameters

bkerler:
This partly solves #4157.
1. When being run remotely, the M862.x P commands should return a response whether it is being compatible or not.
2. `M862.3 P "MK3S"` and `M862.3 P"MK3S"` both should work, but didn't. This behaviour was fixed as well.


src/marlin_stubs/M862_2_3.cpp
```diff
@@ -28,6 +28,24 @@ static void setup_gcode_compatibility(const PrinterModelInfo *gcode_printer) {
...
+static void print_compatibility(PrinterGCodeCompatibilityReport compatibility) {
+    SERIAL_ECHO("Printer is ");
+    if (!compatibility.is_compatible) {
+        SERIAL_ECHO("in");
+    }
+    SERIAL_ECHOLN("compatible.");
+    if (compatibility.is_compatible) {
+        SERIAL_ECHO("Compatibility mode: ");
+        if (compatibility.mk3_compatibility_mode) {
+            SERIAL_ECHOLN("MK3");
```
CZDanol: I feel like by putting this on separate lines, these will be rather hard to parse by a machine

src/marlin_stubs/M862_2_3.cpp
```diff
@@ -61,19 +79,24 @@ static void setup_gcode_compatibility(const PrinterModelInfo *gcode_printer) {
...
 void PrusaGcodeSuite::M862_2() {
-    // Handle only Q
     // P is ignored when printing (it is handled before printing by GCodeInfo.*)
-    if (parser.boolval('Q')) {
+    GCodeParser2 p;
+    if (!p.parse_marlin_command()) {
+        return;
+    }
+
+    if (p.option<bool>('Q').value_or(false)) {
```
CZDanol: Nitpick/tip: `== true` would work as well (but `value_or` is also fine)

src/marlin_stubs/M862_2_3.cpp
```diff
@@ -61,19 +79,24 @@ static void setup_gcode_compatibility(const PrinterModelInfo *gcode_printer) {
...
         SERIAL_ECHO_START();
         char temp_buf[sizeof("  M862.2 P0123456789")];
         snprintf(temp_buf, sizeof(temp_buf), PSTR("  M862.2 P%lu"), (unsigned long)PrinterModelInfo::current().gcode_check_code);
         SERIAL_ECHO(temp_buf);
         SERIAL_EOL();
     }
 
-    if (parser.boolval('P')) {
-        setup_gcode_compatibility(PrinterModelInfo::from_gcode_check_code(parser.value_int()));
+    if (const auto gcode = p.option<int>('P').value_or(0)) {
```
CZDanol: Here, value_or is not fine. Technically speaking, `0` could also be a valid printer model. The point of the parser returning `std::optional` is to avoid these double meanings.

```
if (const auto gcode = p.option<int>('P')) {
print_compatibility(setup_gcode_compatibility(PrinterModelInfo::from_gcode_check_code(*gcode)));
}
```

src/marlin_stubs/M862_2_3.cpp
```diff
@@ -105,28 +128,23 @@ void PrusaGcodeSuite::M862_2() {
...
-        }
-
-        const char *arg_end = arg;
-        while (isalnum(*arg_end) || *arg_end == '.') {
-            arg_end++;
-        }
-
-        setup_gcode_compatibility(PrinterModelInfo::from_id_str(std::string_view(arg, arg_end)));
+    std::array<char, 64> id_str;
+    if (const auto opt = p.option<std::string_view>('P', id_str)) {
```
CZDanol: Good usage here 👍

src/marlin_stubs/M862_2_3.cpp
```diff
@@ -105,28 +128,23 @@ void PrusaGcodeSuite::M862_2() {
...
+        print_compatibility(setup_gcode_compatibility(PrinterModelInfo::from_id_str(opt->data())));
```
CZDanol: `std::string_view` does not guarantee null termination, you cannot just pass the data from it.

from_id_str already takes std::string_view, so there is no need to take the const char* pointer from the view and cast it back to std::string_view. You can just do `print_compatibility(setup_gcode_compatibility(PrinterModelInfo::from_id_str(*opt))));`

src/common/gcode/gcode_info.hpp
```diff
@@ -49,7 +49,7 @@ class GCodeInfo {
 public:
     static constexpr uint32_t gcode_level = GCODE_LEVEL;
 
-    static constexpr auto supported_features = std::to_array({ "Input shaper" });
+    static constexpr auto supported_features = std::to_array({ "Input shaper", "Phase stepping" });
```
CZDanol, line 53: Not all our printers support phase stepping.

src/marlin_stubs/M862_6.cpp
```diff
@@ -24,17 +24,53 @@
...
+    #if HAS_PHASE_STEPPING()
+    SERIAL_ECHOLN(" M862 P\"Phase stepping\"");
+    #endif
+    if (p.option<bool>('Q').value_or(false)) {
         for (auto &feature : GCodeInfo::supported_features) {
-            SERIAL_ECHO_START();
-            snprintf(temp_buf, sizeof(temp_buf), PSTR("  M862.6 P\"%s\""), feature);
-            SERIAL_ECHO(temp_buf);
-            SERIAL_EOL();
+            SERIAL_ECHOLNPAIR(" M862.6 P\"", feature, "\"");
```
CZDanol: This would print out Phase stepping twice.

src/marlin_stubs/M862_6.cpp
```diff
@@ -24,17 +24,53 @@
...
-    // P is ignored when printing (it is handled before printing by GCodeInfo.*)
-    if (parser.boolval('Q')) {
-        char temp_buf[sizeof("  M862.6 P\"01234567890123456789\"")];
+    GCodeParser2 p;
+    if (!p.parse_marlin_command()) {
+        return;
+    }
+
+    #if ENABLED(PRUSA_MMU2)
+    if (MMU2::mmu2.Enabled()) {
```
CZDanol, line 38: Let's not add more hacks to showing & checking for the MMU support. If we wanna do this, the `GCodeInfo::supported_features` should be turned into a function that would dynamically generate an array of supported features based on the current printer state. (for example using stdext::inplace_array<const char*>)

src/marlin_stubs/M862_6.cpp
```diff
@@ -24,17 +24,53 @@
...
+    std::array<char, 64> req_feature_buffer = { 0 };
+    if (const auto opt = p.option<std::string_view>('P', req_feature_buffer)) {
+        bool found = false;
+        SERIAL_ECHOPAIR("\"", opt->data(), "\"");
+        for (auto &feature : GCodeInfo::supported_features) {
+            if (opt->find(feature) != std::string_view::npos) {
+                found = true;
+                break;
+            }
+    #if ENABLED(PRUSA_MMU2)
```
CZDanol: Yep, sorry, this is way too much code duplication. This MMU3/Phase stepping conditioning should all be implemented on a single place.

src/marlin_stubs/M862_6.cpp
```diff
@@ -24,17 +24,53 @@
...
+            SERIAL_ECHOLNPAIR(" M862.6 P\"", feature, "\"");
+        }
+    }
+
+    std::array<char, 64> req_feature_buffer = { 0 };
+    if (const auto opt = p.option<std::string_view>('P', req_feature_buffer)) {
+        bool found = false;
+        SERIAL_ECHOPAIR("\"", opt->data(), "\"");
+        for (auto &feature : GCodeInfo::supported_features) {
+            if (opt->find(feature) != std::string_view::npos) {
```
CZDanol: Find? Find is finding a substring. The match should either be 1:1 or there should be some clearly documented separators.

src/marlin_stubs/M862_1.cpp
```diff
@@ -30,28 +30,46 @@
...
+        // Fetch diameter from EEPROM
         ArrayStringBuilder<64> sb;
-        sb.append_printf("  M862.1 T%u P%.2f A%i F%i", tool, static_cast<double>(config_store().get_nozzle_diameter(tool)), config_store().nozzle_is_hardened.get().test(tool), config_store().nozzle_is_high_flow.get().test(tool));
+        sb.append_printf("  M862.1 T%u P%.2f A%i F%i", tool.value(), nozzle_diameter, config_store().nozzle_is_hardened.get().test(tool.value()), config_store().nozzle_is_high_flow.get().test(tool.value()));
         SERIAL_ECHO(sb.str());
-        SERIAL_EOL();
     }
+
+    if (const std::optional<double> requested_nozzle_diameter = p.option<float>('P')) {
+        if (requested_nozzle_diameter.value() < 0) {
```
CZDanol: This can be auto-checked by passing a min value to p.option. Also, what is the point of even checking this? If the user enters a negative value, he would get a mismatch anyway.

src/marlin_stubs/M862_5.cpp
```diff
@@ -28,15 +28,29 @@
...
+    if (p.option<bool>('Q').value_or(false)) {
         SERIAL_ECHO_START();
         char temp_buf[sizeof("  M862.5 P0123456789")];
-        snprintf(temp_buf, sizeof(temp_buf), PSTR("  M862.5 P%lu"), GCodeInfo::gcode_level);
+        snprintf(temp_buf, sizeof(temp_buf), PSTR("  M862.5 P%lu"), (unsigned long)GCodeInfo::gcode_level);
         SERIAL_ECHO(temp_buf);
         SERIAL_EOL();
     }
+
+    if (const std::optional<int> req_gcode_level = p.option<int>('P')) {
```
CZDanol: nitpick: auto would be nicer

src/marlin_stubs/M862_5.cpp
```diff
@@ -28,15 +28,29 @@
...
+        SERIAL_ECHO("Gcode level ");
+        if (req_gcode_level == (int)GCodeInfo::gcode_level) {
```
CZDanol: Nitpick: Why cast to int here? Would it be better to take the P option as the same type as gcode_level?

CZDanol (commented):
Still not a full review, sorry, not much time and this PR is on the bigger side.

Besides from nitpick/comments in the review, there is also a higher-level problem that we're duplicating here what GCodeInfo does. And GCodeInfo does it in a completely different way, so we could end up with a situation where the gcode outputs are telling us one thing and the GCodeInfo is reporting something else.

The way things are implemented in GCodeInfo currently is terrible, because it basically parses the gcodes "by hand". But to allow this extension of the gcodes, we might first need to clean up the GCodeInfo to make things consistent (which would be a big task).

I am not sure.

src/marlin_stubs/M862_2_3.cpp
bkerler: Agree, changed that.

src/marlin_stubs/M862_2_3.cpp
bkerler: Done :)

src/marlin_stubs/M862_2_3.cpp
bkerler: Changed that as well, good point.

src/marlin_stubs/M862_2_3.cpp
bkerler: Done.

src/marlin_stubs/M862_6.cpp
bkerler: Removed for now.

src/marlin_stubs/M862_6.cpp
bkerler: Removed for now.

src/marlin_stubs/M862_1.cpp
bkerler: Done, you're right, negative testing doesn't make sense as we force it to be bigger than 0.1

src/marlin_stubs/M862_5.cpp
bkerler: Agree

src/marlin_stubs/M862_5.cpp
bkerler: Totally right. Changed that.

bkerler:
I think you are right regarding the redesign and also not happy about the way the mmu3 and feature output/check is happening in M862_6. In order to allow the other fixes to happen, I decided that M862_6 isn't put in this PR, but instead in a separate PR (either with proper changes or after full gcodeinfo patches happened).

PR #4818, opened 2025-09-14
gcode: Improve M106 (uint16_t vs. float)

bkerler:
According to the reprap gcode and the prusa firmware, the M106 S Parameter is parsed as a uint16_t value between 0 and 255.
However, the gcodes that PrusaSlicer is sending are reprap firmware compatible and thus are represented as a real value.

Example, expected is:
```M106 S107```
PrusaSlicer sends:
```M106 S107.1```

The current firmware does parse this as an unsigned short value, which using inline function is parsed with strtol but instead it should be parsed as float value (strdold or strtod) and then used as a uint16_t value with range between 0 and 255.

Also, in order to improve speed, the S parameter is fixed to only be parsed once and not twice and instead of the NOMORE define we use std::clamp instead. The description of A was also missing and has been added according to what the parameter is used for.


o-vesely:
Closing this, thanks, we've implemented this and merged, will be in 6.11 onwards

PR #4869, opened 2025-09-30
Use wait alert sound for filament stuck

jcs:
My XL was getting repeated "filament stuck" errors but despite the printer's sound setting being on "assist", it did not beep repeatedly as it does when the filament runs out.

I think this change is how this should be fixed but I haven't tested it.


jcs:
I have no idea how all of those commits got added to this...

PR #4895, opened 2025-10-11
Increase axis precision to 3 decimals in status API

JeroMul:
Axis values have a 3-digits precision in prusa-slicer and in gcode files.
PrusaLink displays Z-Height uses this API, and should be able to display non truncated height values


Ro3Deee:
> should be able to display non truncated height values

why?

PR #4927, opened 2025-11-06
gui: Add option to hide file extensions

### LINKED ISSUE #4426: [ENHANCEMENT] Add option to hide file extensions
### Printer model

All

### Describe the enhancement

In some cases, (I'm struggling to think of them) it is useful to view the file extension on a print file either in the `Print` menu, or during a print. However, most of the time it doesn't matter, and the file extension only takes up unnecessary screen space. Only two types of files (`.gcode` and `.bgcode`) are going to be shown anyway, so most of the time it is not necessary to know the exact filetype, as long as it is a print file.

### Expected functionality

There would be an option in the `User interface` menu to hide file extensions. If enabled, the extension would be hid from both the `Print` menu, and the printing screen.

me25542:
Resolves #4426.

Adds "Show File Extensions" toggle to user interface menu, directly below "Sort Files".
- If this toggle is enabled, behavior is unchanged.
- If this toggle is disabled, file extensions are hid from the file browser, print preview, and printing screens.

Increases RAM usage slightly; at least ~104 extra bytes could be used at any given time.

Tested and seems to work on a MK4S.

BFW-6704


3d-gussner:
@me25542 Thanks for the PR. Added a comment in the internal ticket.
As it uses quite a bit of extra resources it may become an issue on MINI.

me25542:
Thanks.
You're right that this might not work on the MINI; resource usage is increased somewhat. However, it *does* now seem to build OK for the MINI, after some optimizations (it didn't build for the MINI when this PR was first opened), so I don't know for sure.

me25542:
Extended testing revealed that optimizations made to enable this feature on the MINI broke things, so I have switched to working code, and disabled this feature on the MINI.

me25542:
I accidentally closed this somehow when updating my branch.

bkerler:
@me25542 Does this still break things on MINI ?

me25542:
This still doesn't work on MINI; It does compile for everything else and seems to work on my MK4S.
I added guards so building the MINI firmware still works, but doesn't have my changes.

me25542:
After looking into it more, it seems like I won't be able to get this working on the MINI, and other printers are stuck with increased memory usage (up to around 728 extra bytes).

Issue #4946, opened 2025-11-18
[BUG] M92 configuration is not saved to EEPROM and reset to defaults on G28


imax9000:
### Printer model

CORE one

### Firmware version

6.4.0-RC2

### Printing from...

serial console

### Describe the bug

I'm trying to switch to 1.5GT belts, for which a slight adjustment to [M92](https://help.prusa3d.com/article/prusa-firmware-specific-g-code-commands_112173) is needed (21T at 1.5mm is a hair shorter than 16T at 2mm).

However, I found that changes to M92 are getting reset to defaults by almost anything, even just looking at a printer wrong:

* Homing (both from LCD menu and G28)
* Attempting X or Y axis calibration
* Rebooting after M500
* M501 after M500

### How to reproduce

```
M92 X98.4375 Y98.4375
M92                   
M500                         
M92                          
M501  
M92
```

The result:

```
ok
echo: M92 X98.44 Y98.44 Z400.00 E380.00
ok
ok
echo: M92 X98.44 Y98.44 Z400.00 E380.00
ok
Resetting stepper drivers
echo:  G21    ; Units in mm (mm)
[...]
ok
echo: M92 X100.00 Y100.00 Z400.00 E380.00
ok
```

### Expected behavior

```
echo:M92 X98.44 Y98.44 Z400.00 E380.00
```

### Files

https://github.com/prusa3d/Prusa-Firmware-Buddy/issues/4675 might be related, but I haven't even got to checking if the printhead is moving correctly.

Runklestiltskin:
I have the same issue on 6.4.0+11974 on my 5T XL, I print in shiny black PETG and the VFAs are killing me, so I switched to 1.5GT which are known to reduce the issue dramatically. If I ask my XL to run a GCODE just containing: M92 X101.59 Y101.59, it tells me something about  the Gcode being incompatible. Please advice.

fosterish:
@imax9000 Your EEPROM-related issues are working-as-intended.  Saving to EEPROM [is disabled](https://github.com/prusa3d/Prusa-Firmware-Buddy/blob/master/lib/Marlin/Marlin/src/module/configuration_store.cpp#L200-L203).

Issue #4675, opened 2025-05-31
 [BFW-7094] [BUG] M92 command not work in XL


michcio56:
### Printer model

Prusa XL 

### Firmware version

6.2.4.

### Printing from...

pronterface

### Describe the bug

Any ideas how to make M92 work as in MK4S?

I wanted to test 16t stepper pulleys instead of 20t (VFA on low speeds on 20t).

I have my own working procedure how to use other pulley size in MK4, so I used it in XL:

1 step - via pronterface:
-Reset printer
-G28 home tool
-M92 X100 Y100 - from reprap calc
-Not changing M350 (1/16 is hardcoded with interpolation to fake 256 steps)
-Testing G0 X360 Y360
-Toolhead moves to around 3/4 of the value
-position check via M114 - X360 Y360
-Turning off the soft stop - M211 S0
-Moving little steps to the end of X Y axes 
-M114 again - with values around 440

So the printer doesn’t use new value of M92, which is strange because it works in MK4S

Any ideas?

Runklestiltskin:
Can we prioritize fixing this?

Runklestiltskin:
This is from may and it is still not fixed end of november and the following firmware revision was already released, you can't imagine my disappointment.

Issue #4619, opened 2025-04-22
[BFW-7094] [BUG] M92 being ignored.


kenour:
### Printer model

MK3.5S

### Firmware version

6.2.4+8909

### Printing from...

Prusa Connect

### Describe the bug

I use M92 to calibrate my machine for dimensionally accurate prints. It has been working flawlessly for years.
I've been having general issues since 6.2.3 I think it was, and I just checked the size of my prints, and it's undersized.
I then ran my calibration again, and even with the below line in my 'Start G-code' section;
M92 X100.29 Y100.53 Z401.06 ; calibration
My calibration shape printed as if it was ignored. I then considered it may have drifted over the years, so changed it to;
M92 X100.64 Y101.05 Z402.08 ; calibration
But it printed exactly the same size, which leads me to think the M92 command it being ignored. I confirmed it was in the produced gcode by inspecting it with the gcode viewer, and it was present as expected.

All my models are now printing undersized since M92 has started to be ignored.

### How to reproduce

Print a test shape, use M92 to calibrate it's size, print it again to confirm no change to size.

CZDanol:
Internal ticket: BFW-7094

kenour:
> Thank you for your contribution to our project. This issue has not received any updates for 60 days and may be considered "stale." If this issue is still important to you, please add an update within the next 7 days to keep it open. Administrators can manually reopen the issue if necessary.

Hello Mr Bot. Yes it's still important, and from what I understand it's now been moved to an internal ticket. Is there anyway to follow this internal ticket to see the outcome/updates?

kenour:
Have requested #4697 to have this calibration added to the advanced menu after the issue has been fixed.

kenour:
> Internal ticket: BFW-7094

Any chance of finding out what stage the fix is up to?

bkerler:
I just checked and I don't seem to see anything off in the code, except that the settings aren't persistent ... as you wrote M92 was used as start gcode, the written values should be used in the planner (variable planner.settings.axis_setps_per_mm per axis). Did your gcode maybe overwrite the values ?
Will try to emulate this behaviour to see if anything interfers with the planner ...

kenour:
[...]

PR #4953, opened 2025-11-20
fix: eliminate one case of possible mismatch between axis_steps_per_mm and mm_per_step

imax9000:

PR #5062, opened 2025-12-31
Fix incorrect XY scaling when `axis_steps_per_mm` is not 100.

fosterish:
# Problem
Issuing `M92` (for instance, to compensate for 1.5mm pitch belt pulleys on the CORE One) produces incorrect results.  The nature of the error depends on where in the gcode the command is issued in relation to a `G28` command.

## Inverted scaling
If `M92` is issued **before** `G28`, the resulting scale factor will be inverted (values lower than 100 produce a _larger_ print, and values higher than 100 produce a _smaller_ print).
### Cause
This happens because `G28` calls `MarlinSettings::reset_motion`, which resets the planner's `axis_steps_per_mm` (and not `mm_per_step` et al.), which never gets restored during the call to `Motion_Parameters::load` at the end of the command.  So then we're using the default 100 steps per mm to generate step counts in `Planner::buffer_segment`, but then consume them with a scaled `mm_per_mstep` in `Planner::_populate_block`.
### Solution
I've unified `motion_parameters_t` and `planner_settings_t`.  Maybe this causes issues elsewhere, but the two were so similar I didn't understand why they were separate.  At the very least, this causes `axis_steps_per_mm` to get saved and loaded with the rest of the motion parameters.  I also call `Planner::refresh_positioning` anytime we change that value.

## Ignored scaling
If `M92` is issued **after** `G28`, there is no effect on the scale of the print.
### Cause
A print move gets converted into motor movements through a convoluted chain of conversions: mm -> ministeps -> mm -> duration@speed -> rotation speed.  The final step of this chain relies on `AxisMotorParams::mm_to_rev`, which gets calculated only once, at startup, and not anytime `axis_steps_per_mm` changes.
### Solution
I've created `phase_stepping::update_axis_motor_params`, which I call from inside `Planner::refresh_positioning`.  It does the same thing as the initialization, but pulls its values from the planner and not the defaults.

**Note**: I would never claim to be familiar with concurrent programming in C++, so someone who actually knows what they're doing should look at this.  I noticed that `axis_motor_params` is getting read thousands of times per second by the step ISR, so I figured we'd be dabbling in some light UB if we didn't at least disable interrupts before modifying it.

# Background
I have a Prusa CORE One, and have recently swapped the stock belts and pulleys for 1.5mm pitch belts and pulleys, as many people are experimenting with, to combat VFAs.  Because the pitch diameter of the new pulleys is slightly smaller than that of the stock pulleys, one needs to configure the machine's steps per mm to compensate.

The common solution found in the [forums](https://forum.prusa3d.com/forum/postid/756028/) and in [videos](https://www.youtube.com/watch?v=9doyX_2YBkg) is to issue `M92 X98.44 Y98.44` at the start of your gcode, which does indeed have an enlarging effect on the print.  However, this solution has problems:
1. It makes no sense.  The new pulleys are slightly smaller than the stock, so they need to rotate _more_ to cover the same distance.  This command does the opposite by telling the machine to take _fewer_ steps per mm.  This is clearly just a bug.
2. It leaves the printer in a strange state where calibration of the X or Y axes, or home, fails.

With this PR, I hope to get the ball rolling on a fix.  I've tested it on my own CORE One, and it works very well.  With this fix, one can correctly issue the appropriate `M92` command, and get properly sized prints using 1.5mm pitch belts and pulleys without leaving your printer in a weird state:
```
scale = old pitch diameter / new pitch diameter
scale = (16 * 2 / pi) / (21 * 1.5 / pi)
scale = 32 / 31.5
scale ~= 1.015873
```
```M92 X101.5873 Y101.5873```

I believe this solves https://github.com/prusa3d/Prusa-Firmware-Buddy/issues/4619, https://github.com/prusa3d/Prusa-Firmware-Buddy/issues/4675, and the first part of https://github.com/prusa3d/Prusa-Firmware-Buddy/issues/4946, though I haven't tested it on any printer other than the CORE One.


bkerler:
No duplicate of #4953, but linked.

fosterish:
@bkerler The PR you mention addresses only part of the first issue mentioned in this PR ("Inverted Scaling" above).  Specifically, this PR addresses the issue that `G28` will reset `axis_steps_per_mm` and then never restore it.  Both PRs ensure that `mm_per_step` is in sync with `axis_steps_per_mm`, but the mentioned PR does not restore `axis_steps_per_mm` after what should be a temporary reset.

This PR also addresses another issue with the scale system: "Ignored Scaling" above.

bkerler:
Thanks, this helps a lot in understanding and distinguishing the PRs.

imax9000:
> @bkerler The PR you mention addresses only part of the first issue mentioned in this PR ("Inverted Scaling" above). Specifically, this PR addresses the issue that `G28` will reset `axis_steps_per_mm` and then never restore it. Both PRs ensure that `mm_per_step` is in sync with `axis_steps_per_mm`, but the mentioned PR does not restore `axis_steps_per_mm` after what should be a temporary reset.

Yeah, I've poked through the code while trying to switch my Core One to 1.5GT belts. I never got around to put some work into fixing the code to make it work properly, but #4953 seemed self-contained and simple enough to send a PR right away.

PR #5080, opened 2026-01-11
Spelling consistency: Gcode/G-Code/GCode -> G-code

smoe:
It is a tough call, but whatever spelling of "G-code" is used - it should be used consistently. Wikipedia, RepRap and LinuxCNC on the hyphen, a capital "G" and the lowercase "c". And also in header lines, I would treat "G-codes" as one term and not capitalize the "c", especially so when other words in that header are not capitalized..

This PR got triggered while exploring the menus of the printer, so it is the user visible changes that I initially wanted to be more close to perfect. While at it, the patch now also addresses comment lines. In retrospect, those invisible changes may even be more important for a grepping in the source code because of the often missing hyphen.

I have not changed the Doxygen group name "G-Codes", anticipating to disturb some downstream processing. Direct me if you prefer those adapted, too.


bkerler:
@smoe Can you please rebase your PR to the latest master ? Thanks !

smoe:
@bkerler Maybe there is some other master somewhere, or I am a bit clumsy, but this contributed a bit over two months ago and https://github.com/prusa3d/Prusa-Firmware-Buddy is older. So, those changes should be directly applicable.

bkerler:
Ah my fault, seems to be fine.

PR #5082, opened 2026-01-12
pause: Fix infinite loop in filament stuck logic with MMU

### LINKED ISSUE #4791: [BUG] MMU3 filament change halting a print
### Printer model

Core One

### Firmware version

6.3.5-alpha+10514

MMU3 Enclosed kit

### Printing from...

Prusa connect (usb inserted) 

### Describe the bug

The printer stucks in the loading filament screen (like when you load filaments on mmu) and doesn't do anything else, failing the current print. 

### How to reproduce

- Run a print
- Make the mmu3 finish the filament mid print
- Make it so that the mmu cannot unload the filament properly, causing an error on the screen (i think it tries 3 times before showing)
- Now "fix" the issue and let filament unload
- Load new filament
- The loading filament screen doesn't go away. 

### Expected behavior

Filament loads and proceeds to resume printing 

### Files

N/A

m42e:
When the filament is stuck on a printer with MMU (e.g. MK4S), the firmware would enter an infinite loop in the pause state machine because 'LoadType::filament_stuck' was not handled in 'mmu_unload_start_process'.

This change adds the missing case to allow the MMU to unload and eject the stuck filament.

Fixes #4791 

IMHO: I am not sure if you would want to handle this error case differently and maybe try to reload the filament a few times more to automatically resolve the filament stuck error in the case it happens mid-print. This will now lead to a pause but the user will be able to continue after removing and reinserting the filament into the MMU.


m42e:
~Ok, resuming does not work correctly, I will have another look.~
resuming works.

thomasloven:
I can confirm that this works as intended on MK3.9S with MMU3.
I recommend it is merged as soon as possible seeing as this is an entirely breaking bug which severely impacts the trademark reliability of the MMU3.


<details>
<summary>Off topic</summary>

Thank you thank you thank you thank you thank you!!
My MMU3 is at all useable again!

</details>

JGeiger2024:
Does anyone know when this type of fix is going to be provided to the public by Prusa?

Prusa-Support:
Thanks for your contribution. The Prusa official fix should be included in [FW 6.5.3](https://github.com/prusa3d/Prusa-Firmware-Buddy/releases/tag/v6.5.3).

Michele Moramarco
Prusa Research

PR #5129, opened 2026-02-12
isr: Qualify shared state and snapshot ISR-visible handles

MaxGit001:
## Summary
- qualify ISR-shared state as volatile where asynchronous updates occur
- snapshot volatile callback/task handles into locals before use
- keep atomic synchronization for BufferedSerial pendingWrite

## Why this matters
These variables are accessed from both ISR and non-ISR contexts. Without
volatile/atomic semantics, compiler optimizations may cache stale values and
miss asynchronous ISR updates. In embedded firmware this can surface as
nondeterministic failures: invalid state transitions, wrong behavior after ISR
return, and symptoms that look like stack or return-path corruption.

## Scope of changes
- WDT callback pointer now treated as ISR-shared and read via local snapshot
- USB MSC worker task handle treated as ISR-shared and snapped before
  vTaskNotifyGiveFromISR
- RS485 receive callback pointer treated as ISR-shared and snapped in IRQ path
- Modularbed timebase/PWM ISR-shared counters and pointer qualified
- BufferedSerial idle marker qualified as ISR-shared (pendingWrite remains
  atomic)
- MMU early overcurrent trigger timestamp qualified for ISR/task sharing

## Validation
- Targeted rebuilds of all modified objects completed successfully:
  - xl_release_boot main firmware targets
  - xl_release_boot/modularbed-build targets
  - xl_release_boot/dwarf-build targets


danopernis:
This looks mighty AI generated and isn't motivated by any real issue.

MaxGit001:
Still some crashes of Heads are the issues.
Variables which are used (written) in interrupt routines should be defined as Volatile.
Otherwise the compiler can optimise and dont see that this variables can get changed by the interrupts. This leads to spurious crashes because the variables maybe got optimised and then written over wrang registers.
I did find volatile mistakes in in other codes.

It did solve spurious crashes.
Better 1 varialbe is unnecessary defined as volatile, then 1 should be volatile and is not.

danopernis:
Please provide reproducer, minimal patch that demonstrably fixes the problem, and cleanup the commit message. And also stop using AI and stop wasting other people's time. Why do you expect human to read your code if you didn't care enough to write it?

PR #5139, opened 2026-02-21
gui: add "Home Screen Set Ready Mode" option

DomiiBunn:
Adds a new toggle in Settings > User Interface that replaces the Print button on the home screen with a "Set Ready" action. When enabled:

- The home screen Print button calls set_printer_ready(true) instead of starting a local print, and its label changes to "Set Ready".
- wui_start_print() returns early without showing a print dialog.

Gated behind BUDDY_ENABLE_CONNECT so it only appears on Connect-capable builds.


DomiiBunn:
Some claude was used but mainly to tidy it up and make the simulator work while debugging

PR #5178, opened 2026-03-19
Fix: chamber filtration fan does not resume after re-enabling backend

### LINKED ISSUE #5173: [BUG] Toggling advanced filtration during a print will not re-enable filtration
### Printer model

Core One+

### Firmware version

4.6.0, 6.5.3-RC

Advanced Filtration

### Printing from...

USB Drive, PrusaLink

### Describe the bug

Disabling chamber filtration (setting to none), and then re-enabling DIY or advanced filtration will not reactivate the filtration fan and requires manually setting the filtration fan

### How to reproduce

Disable Advanced Filtration or DIY filtration during a print.  Filtration fan turns off as expected
Re-enable DIY or Advanced Filtration during print and the filtration fan will not go back on

### Expected behavior

Re-enabling advanced filtration or DIY should turn back on the filtration fan

### Files

_No response_

holdenkilbride:
# Problem                                                                                                                                                                
When the chamber filtration backend is changed to None during an active print and then restored to Advanced Filtration or DIY, the filtration fan never comes back on for the remainder of that print.                                                                                                                        
                                                                                                                                                                  
  In ChamberFiltration::step(), the disabled path (!is_enabled()) was setting needs_filtration_ = false. When the backend is re-enabled, update_needs_filtration() is called but immediately returns early due to the has_value() guard (intended to preserve M147/M148 gcode overrides). Since needs_filtration_ holds false, the fan stays off.                                                                                                                                                 
                  
 # Fix                                                                                                                                                             
Set needs_filtration_ = std::nullopt instead of false in the disabled path. nullopt means "unknown — re-evaluate on next print cycle", which bypasses the M147/M148 guard and allows update_needs_filtration() to re-evaluate filament types and restore filtration correctly.
                                                                                                                                                                  
# Why 
False is a valid evaluated state meaning "this print does not require filtration." Using it here conflated two distinct states: "evaluated and not needed" vs "not yet evaluated." nullopt is the correct sentinel for the latter.
                                                                                                                                                                  
  Behavior unchanged for:                                                                                                                                         
  - Normal print start/end
  - M147/M148 overrides (these set needs_filtration_ via set_needs_filtration(), not through the disabled path)                                                   
  - Post-print filtration
Fixes #5173


holdenkilbride:
Closing — this is already resolved on master. The chamber_filtration: Tie to nozzle temperature refactor (BFW-7026, shipped in v6.6.0) removed the cached needs_filtration_/is_printing_prev_ state that caused the fan to not resume after re-enabling mid-print; needs_filtration() is now recomputed each step(), so the stale-state path this PR patched no longer exists. My branch was based on 6.5.3, which predates that refactor.

PR #5180, opened 2026-03-19
Improve Chamber Lighting Functionality

### LINKED ISSUE #5179: [ENHANCEMENT] Enhance options for chamber lighting during interaction, printing, and idle
### Printer model

Core One+

### Describe the enhancement

Previously, chamber lighting had a single "dimmed" brightness level controlled by a three-way toggle (Never / Always / On Idle). This made it impossible to independently configure how bright the lights should be while printing versus when the printer is idle. Users printing in a dark room want the chamber well-lit during a print, but dim when idle — and a different brightness again when actively interacting with the printer (opening the door, turning the knob). The old toggle didn't allow this nuance.

A specific use case: you want full brightness when you open the door to check on a print, a comfortable medium brightness while the print runs unattended, and near-off when the printer is idle overnight.

### Expected functionality

 The lighting settings menu gains three independent brightness sliders, replacing the old dimming toggle:

  1. Chamber Brightness — brightness when you interact with the printer (door open, knob turn, touch). Acts as the ceiling; 0–100%.
  2. Print Brightness — brightness during an active print when you're not interacting. 0–100%.
  3. Idle Brightness — brightness when the printer is not printing and idle. Automatically clamped to not exceed Chamber Brightness. 0–100%.

  Step-by-step behavior:
  - User interacts (door, knob, touch) → lights jump to Chamber Brightness, stay there for the activity timeout
  - Activity timeout expires during a print → lights settle to Print Brightness
  - Activity timeout expires while idle → lights settle to Idle Brightness
  - Print ends → lights transition from Print Brightness to Idle Brightness on next activity timeout

holdenkilbride:
## Summary
Replaces the binary `DimmingEnabled` setting (Never / Always / On Idle) with three independent brightness levels for chamber lighting: interaction brightness, print brightness, and idle brightness. This gives users granular control over lighting behavior across all printer states without sacrificing simplicity.

  ## Problem

 The previous implementation had two configurable values — a maximum brightness and a dimmed brightness — gated by a three-way enum toggle. This created several UX problems:

  - No way to set a distinct brightness for active printing vs. true idle; "On Idle" dimming applied the same dimmed level to both
  - "Never" and "Always" modes were confusing — "Never" disabled dimming entirely (always full brightness), "Always" dimmed regardless of print state
  - The dimmed brightness slider was hidden/disabled when dimming was set to "Never", leading to settings that silently did nothing
  - Users wanting comfortable printing illumination without blinding idle brightness had no clean way to configure it

  ## Changes

  **`side_strip_handler`**
  - Adds a new `printing` state to `SideStripState` alongside the existing `dimmed` and `active` states
  - Replaces `DimmingEnabled dimming_enabled` member with `uint8_t print_brightness`
  - Updates `update()` logic: when activity timeout expires, transitions to `printing` state if a print is active, otherwise `dimmed`; the `active` state is only
  entered on user interaction if `max_brightness > baseline` (prevents a pointless flash to active when interaction brightness is already at or below current
  brightness)
  - `get_color_for_state()` handles the new `printing` state, clamping to `print_brightness`

  **Config store**
  - Removes `side_leds_dimming_enabled` (migrated to `DeprecatedStore`)
  - Adds `side_leds_print_brightness` (default 255 — full brightness during prints, preserving existing behavior for users upgrading)
  - `side_leds_dimmed_brightness` and `side_leds_max_brightness` retain their existing hash keys for backward compatibility

  **Menu items**
  - Removes `MI_SIDE_LEDS_DIMMING_ENABLE` (the three-way toggle)
  - Adds `MI_SIDE_LEDS_PRINT_BRIGTHNESS` spinner (0–100%)
  - Renames labels for clarity: "Chamber Lights" → "Chamber Brightness", "Chamber Lights Dimmed" → "Idle Brightness", new item "Print Brightness"
  - `MI_SIDE_LEDS_DIMMED_BRIGTHNESS::Loop()` now enforces that idle brightness cannot exceed chamber brightness, auto-clamping and saving if needed

  ## Behavior after this change

  | Printer state | Brightness used |
  |---|---|
  | User interacting (door, knob, touch) | Chamber Brightness |
  | Printing, activity timeout elapsed | Print Brightness |
  | Idle, activity timeout elapsed | Idle Brightness (≤ Chamber Brightness) |

  ## Upgrade path

  Existing users upgrading from firmware with `DimmingEnabled`:
  - `side_leds_dimming_enabled` is deprecated and migrated away; existing max/dimmed brightness values are preserved
  - Print brightness defaults to 255 (full), matching the previous behavior of "On Idle" mode where printing was never dimmed

  ## Testing

  - Verify all three brightness levels apply correctly by transitioning between idle, printing, and interaction states
  - Verify idle brightness is auto-clamped when set above chamber brightness via the menu
  - Verify settings persist across power cycles
  - Verify upgrade from previous firmware does not reset brightness preferences

Fixes #5179 
IMG_4317


bkerler:
Some things I noticed:
1. The upgrade path does not actually preserve prior “Always” dimming behavior. side_leds_dimming_enabled is only deprecated, not migrated into the new side_leds_print_brightness, and the new print-brightness item defaults to 255. Users who previously had DimmingEnabled::always will therefore be changed from “dim during printing” to “full brightness during printing” after upgrade.

2. The new “idle brightness must not exceed chamber brightness” rule is enforced only in the menu widget loop. load_config(), set_max_brightness(), and set_dimmed_brightness() never clamp persisted/runtime values, so upgraded configs or connect-driven brightness changes can still leave idle brightness above chamber brightness until someone opens that menu screen.

holdenkilbride:
@bkerler Would you happen to know when the master branch gets updated with the release code?

bkerler:
The latest code is always released in the tags, but community development mainly happens in master. Master is normally updated after major releases, I'd expect the next master to be 6.6.0.

holdenkilbride:
@bkerler I updated the code per your notes/observations.  I tested on my Core One and verified the setting upgrade paths worked correctly.  Thank you for your review!

Issue #4974, opened 2025-11-24
[BFW-8122] [BUG] CoreOne L -  FW 6.5.1 ; Command M115 indicates that there is no thermal protection.


MenschM:
### Printer model

CoreOne L

### Firmware version

6.5.1+12574 (Github)

### Upgrades and modifications

None

### Printing from...

Octoprint

### Describe the bug

When I connect to the printer with Octoprint, Octoprint shows the printer as dangerous.  
This page [(https://community.octoprint.org/t/warning-message-pops-up-for-mandatory-safety/51347](url) mentioned to check the output of the M115 Command. And yes I've got this line from that command:
Recv: Cap:THERMAL_PROTECTION:0 
That's why Octoprint classify this printer as dangerous - missing basic security feature.

This article also suggests that the printer has thermal runaway protection: [https://help.prusa3d.com/article/hotend-thermal-runaway-31204-core-one-35204-core-one-l-26204-mk4s-13204-mk4-27204-mk3-9s-21204-mk3-9_401076](url)

### How to reproduce

Open Octoprint and connect to the printer. 
Next open the Terminal-Tab and send the M115 command and look at the results you've got from the printer. 

### Expected behavior

I expect this printer to be classified as safe, just like its predecessors.

3d-gussner:
@MenschM Thanks for reporting. I will create an internal ticket.

bkerler:
This should be solved in v6.5.3-RC, can you please verify @MenschM ? Update: Identified the root cause and provided a fix.

bkerler:
I just checked and something is wrong :

<img width="1493" height="780" alt="Image" src="https://github.com/user-attachments/assets/8d011db4-8cbe-4e78-ad5d-d5869525b232" />

bkerler:
Let me check the issue ...

bkerler:
The root cause is due to switching from past Marlin GPIO chamber management to Prusas new Chamber API. As the previous defines aren't managed anymore, the thermal protection for the chamber was reported incorrectly. I wrote a PR which fixes this by checking if the Chamber API is in use.

jsgf:
I'm seeing this with Octoprint and mk4s with release 6.5.3 firmware.

```
[...]

PR #5186, opened 2026-03-22
GCODE: Fix incorrect report of no thermal protection on Core One / Core One L

bkerler:
This solves issue #4974.
Command M115 does report THERMAL_PROTECTION:0 due to improper checking the Chamber as no longer Marlin's implementation is used for Chamber protection but the chamber api instead, leading to an incorrect report of chamber protection.


bkerler:
Closed as already in master

Issue #4779, opened 2025-08-26
Filtration fan in DIY mode does not turn on after print end even though set to do so.


CpuNut:
### Printer model

Core One

### Firmware version

6.3.5-alpha

### Upgrades and modifications

None

### Printing from...

USB drive

### Describe the bug

I added a DIY filtration fan.  If I set to a specific speed to 1% to 100% the fan works fine.  In the Settings, Chamber Filtration, menu I set to DIY mode, 5% minimum, Post-Print Filtration to on, Post-print Duration to 10 Minutes, Post-print power to 40%.  In the Control, Temperature, menu I have Filtration Fans set to Auto.

At the end of a print the chamber fan stops, but the filtration does not start as asked for with the above settings.

This is most of the reason I added a filtration fan so to keep extracting fumes after a print ends for so many minutes.  This should be a one or two line fix and should be in the next production code.

If it's not going to in the next production release, maybe someone can point me to the file name and lines of code to add or change so I can at least make this fix myself in the meantime time.

### How to reproduce

Set Filtration fan to Auto under the Control, Temperature menu, and as above in the Settings, Temperature menu.  At the end of the print the filtration fan does not turn on.  Very simple to reproduce!

### Expected behavior

When the filtration fan is set to Auto in the Control menu and DIY as above in the Settings menu, the filtration fan should turn on after print end for the duration selected.

3d-gussner:
@CpuNut Can you please explain bit more about your DIY filtration fan?
Which fan? Where is it connected to? ...

CpuNut:
It's a 24V four wire 800 Ma fan connected to “Fan3”.  The Filtration Fan speed control works.  In Adv Filtration mode the fan runs fine, in DIY it seemed not to come on so I run it manually in this case.
I did find in DIY or Adv Filtration modes with the Filtration Fan set to Auto, the Fan3 didn't run after the print was finished as I selected in the sub menu, probably a simple firmware fix?
The other thing I found is with the Filtration Fan set to anything but None, the Fan Test doesn't pass, it never starts the Chamber Fan so it fails.
I welcome other questions :-)
Craig
    On Wednesday, August 27, 2025 at 05:22:11 AM EDT, 3d-gussner ***@***.***> wrote:  
 
 3d-gussner left a comment (prusa3d/Prusa-Firmware-Buddy#4779)
@CpuNut Can you please explain bit more about your DIY filtration fan?
Which fan? Where is it connected to? ...

[...]

PR #5187, opened 2026-03-22
chamber_filtration: Fix post-print filtration skipped for non-volatile filament

bkerler:
This fixes issue #4779.
"needs_filtration_" is set at print start based on filament type and is false for non-volatile filaments (PLA, PETG, etc.).

This caused two problems:
- "last_print_s_" was only updated inside the gated branch, so it was never set when printing with non-volatile filament.
- The post-print condition was unreachable because execution always fell into the first (!needs_filtration_) branch and set "output_pwm_" to 0.

As a result, post-print filtration never activated for non-volatile filaments even when explicitly configured by the user, regardless of the backend (DIY or Advanced/official filter).


bkerler:
is fixed on current master, thus closed

Issue #5152, opened 2026-03-06
[BUG] MK4S WiFi disconnects every ~75 seconds due to aggressive 5-second inactivity timeout in ESP32-NIC-FW


bharrison6:
### Printer model

MK4S

### Firmware version

6.4.0+11974

### Upgrades and modifications

None

### Printing from...

USB, Prusa Connect

### Describe the bug

  Printer: MK4S (3 units in fleet)
  Firmware: PrusaLink server 2.1.2
  WiFi module: ESP32-WROOM running Prusa-ESP32-Nic-FW

  Problem

  MK4S printers experience periodic WiFi disconnects approximately every 75 seconds. In a fleet of 10 Prusa printers (4
  Minis, 3 MK4S, 2 Core Ones, 1 XL), only the MK4S units exhibit this behavior. The Minis, Core Ones, and XL are stable.

  Over a 4-day observation window (March 2-5, 2026), the MK4S printers accounted for ~83% of all WiFi disconnects:

  ┌──────────────────────────────┬────────────────────────────────────────┐
  │           Printer            │         4-day disconnect total         │
  ├──────────────────────────────┼────────────────────────────────────────┤
  │ MK4 #4                       │ 272                                    │
  ├──────────────────────────────┼────────────────────────────────────────┤
  │ MK4 #3                       │ 100                                    │
  ├──────────────────────────────┼────────────────────────────────────────┤
  │ MK4 #1                       │ 65                                     │
  ├──────────────────────────────┼────────────────────────────────────────┤
  │ All other printers (7 units) │ ~91 (almost entirely from AP restarts) │
  └──────────────────────────────┴────────────────────────────────────────┘

  The worst unit (MK4 #4) would sometimes exhaust its reconnection retries and go completely offline for hours,
  requiring a power cycle to recover.

  Root cause

  I traced this to Prusa-ESP32-Nic-FW (main/uart_nic.c, branch feature/esp32).

  Line 69:
  static const uint32_t INACTIVE_PACKET_SECONDS = 5;

  If the ESP32 receives no inbound WiFi packet for just 5 seconds, check_online_status() (line 620) triggers a probe
  scan. For an idle printer not actively printing or being polled, 5-second gaps in inbound traffic are normal and
  expected.

  The probe scan (SCAN_TYPE_PROBE) takes the radio off-channel for 120-300ms per channel across ~13 channels. If the
[...]

Issue #5090, opened 2026-01-18
[ENHANCMENT] Core One+ connects to weak wifi signal AP instead of strong one


mavu:
### Printer model

Core One+

### Firmware version

6.4.0+11974

### Upgrades and modifications

none

### Printing from...

Prusa Connect

### Describe the bug

when switching on, the printer connects to the configured WIFI, but it chooses a random AP in our WIFI mesh, instead of the strong one just meters away.

This leads to problems connecting/reaching the printer over Prusa connect or directly, until it completely loses connection to WIFI and reconnects (randomly) to the correct/strong AP.

This is very annoying.

### How to reproduce

Switch on printer.
Observer it randomly connecting to weak AP even though 100% signal strength AP is mere meters away. 
Wait until it loses connection and re-connects with a 50/50 chance to the strong (closeby) wifi AP.

### Expected behavior

It should notice that there are more than one Access Point with the same (configured) SSID and it should chose to connect to the one with the best wifi signal strength of those with the same SSID which is configured.

3d-gussner:
@mavu I don't think this is a bug more an enhancement as it connects to the WiFi, frankly doesn't choose the most power full close by.
Which AP do you use, which bandwidth?

Easiest workaround would be probably if you setup an additional dedicated SSID for the printer on the closest AP only. I assume that printer doesn't change the location that often.

The implementation would need to search for some time which APs are answering with which strength and that may take some additional time. Users may then complain that the WiFi connections takes so long.

mman:
I know absolutely nothing about actual Prusa implementation, but if it is true that the WiFi modules is based on ESP8266, and if it true that it is used mostly as is, then it may possibly look like this:

1. ESP8266 scans using  `WIFI_FAST_SCAN` (https://docs.espressif.com/projects/esp8266-rtos-sdk/en/latest/api-reference/wifi/esp_wifi.html#_CPPv418wifi_scan_method_t) and will connect to the first AP found. That may in your case be the wrong AP because it sends beacons more frequently.
2. ESP8266 scans the channels from lower to higher, so workaround may be to put stronger signal AP on a lower channel (if your AP allows that)
3. And the proper solution should be to use `WIFI_ALL_CHANNEL_SCAN` and `WIFI_CONNECT_AP_BY_SIGNAL`. Which as @3d-gussner may take a longer time to find all APs and sort them before trying to connect to the best one.

[...]

Issue #3864, opened 2024-03-29
[BUG] 6.0.0-RC1 - Wifi doesn't work


Gyrxiur:
Please, before you create a new bug report, please make sure you searched in open and closed issues and couldn't find anything that matches.

**Printer type** - [MK4]

**Printer firmware version** - [6.0.0-RC1]

**Original or Custom firmware** - [Original]

**Optional upgrades** - [none]

**Describe the bug**
  After updating to 6.0.0-RC1 Wifi have worked, but only till restart, then it says No AP. Only way to make it work is factory reset, but it again works only till restart. Also, in czech language, you have typo, instead of Pokus o připojení you have Pokus o piřpojení ;c)

**How to reproduce**
  Update firmware to 6.0.0-RC1 and restart.

**Expected behavior**
  Working wifi every time, as it was before updating.

**G-code**
  N/A

**Crash dump file**
  N/A

  Do not share the file publicly, as the crash dump contains a raw snapshot of the printer's memory and may include unencrypted sensitive information.

**Video**
  N/A

Gyrxiur:
So I've reverted to old firmware and it works, so it is bug in new firmware for sure.

bkerler:
Do you have a prusa-printer-settings.ini file on your stick ? Sounds like a ini parsing issue maybe, trying to sort that out.

Gyrxiur:
> Do you have a [prusa-printer-settings.ini](http://prusa-printer-settings.ini) file on your stick ? Sounds like a ini parsing issue maybe, trying to sort that out.

Yes, and when I load settings, it asks to delete files, so I confirm it. And even when I try to make new file and load it again, it won't connect. Older firmware connects ok with same file, so no mistakes in that file for sure.

H0Z0:
Interesting. I tried to reproduce the error, but after setting up the network, my MK4 connected to the Wi-Fi network every time. It also works after switching off or disconnecting the power cable or pressing the reset button. I also deleted the .ini file from the usb drive, so it can't even read from there. Both PrusaLink and Prusa Connect can connect to the printer.

salmander:
I confirm that I also experienced this. However, I have flashed the RC1 firmware and restarted the printer a couple of times and now it seems to be working. Another issue I experienced is when I am using binary code (bgcode), it can't seem to download the file from Prusa Connect. I tried this a few times. However, changing the legacy gcode in PrusaSlicer worked the first time.

[...]

PR #5188, opened 2026-03-23
WIFI: Esp32/Esp8266 - Priorize AP by signal strength, fix WPA2/WPA3 transitioning issues, prevent reconnect if associated.

bkerler:
This PR solves issue #5090 by enforcing to select AP by signal, thus preventing to choose AP with same SSID but weaker signal (which is the case for repeated signals).

This PR also solves https://github.com/prusa3d/Prusa-Firmware-Buddy/issues/3864 :

ESP32: sae_pwe_h2e defaults to WPA3_SAE_PWE_UNSPECIFIED (0), which fails with routers that require the H2E (Hash-to-Element) PWE method especially for Wifi 6 Routers. IDF v5.0 introduced WPA3_SAE_PWE_BOTH = 3 to support both methods. Without it, SAE authentication fails in transition mode. Additionally, CONFIG_ESP32_WIFI_NVS_ENABLED=1 causes WPA3 PMKSA entries to persist in NVS across reboots, and esp_wifi_restore() called at startup may not clear these PMKSA entries, causing stale-PMK authentication failures on subsequent boots.

ESP8266: The ESP8266 RTOS SDK's WPA3 SAE implementation doesn't reliably handle WPA3/WPA2 transition mode (the SDK is based on IDF v3.4 with backported WPA3 patches, without H2E support). The fix is to disable WPA3 SAE for ESP8266 entirely (CONFIG_ESP8266_WIFI_ENABLE_WPA3_SAE=n), which forces WPA2 PSK connections. WPA3/WPA2 transition mode routers must accept WPA2 PSK by definition.

This PR solves #5193 and #5152 by preventing reconnection if a link is associated but idle (on isolated local wifi) due to no inbound traffic.

Tested and confirmed on mikrotik hAP ax S V7.22.1 and ubiquity AP7.


lib/esp32-nic/main/uart_nic.c
```diff
@@ -48,7 +48,7 @@
 // Externals with no header
 esp_err_t mac_init(void);
 
-#define FW_VERSION 13
+#define FW_VERSION 14
```
Tomcus, line 51: You need to bump the FW version also on printers side IIRC. Otherwise the printer will fail the communication.

Tomcus (changes requested):
Also you need to build the esp fw manually and move to resources. So the code is actually used. And or you can try to integrate it into our build system so it is done automatically.

lib/esp32-nic/main/uart_nic.c
bkerler: Done.

Tomcus (commented):
LGTM. I will merge it as soon as I have a little time.

bkerler:
I've merged two other PRs into this one and tested them with two different network setups (one user also confirmed it to finally work on his mikrotik router).

joachimklug:
Thanks @bkerler - Would be really nice to get this in.

PR #5189, opened 2026-03-23
Build-System: Add ESP32/ESP8266 build integration into Cmake

bkerler:
This integrates building the esp32/esp8266 wifi firmware into the existing cmake ecosystem, but using Docker for compiling the wifi firmware. For that, ESP_FW_BINARY_DIR was introduced. By default is uses the already precompiled wifi binary paths and thus the docker build system won't be run. In order to run Docker and force building the wifi firmware, it is thus required to use ESP_FW_BINARY_DIR=""

Example:
```shell
cmake --preset coreone_release_boot -B build -DESP_FW_BINARY_DIR=""
cmake --build build
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -390,6 +390,8 @@ if(BOARD IN_LIST BUDDY_BOARDS AND HAS_PUPPIES_BOOTLOADER)
   endif()
 endif()
 
+include(lib/AddEspNicFw.cmake)
+
 # Unittests
 if(UNITTESTS_ENABLE)
   enable_testing()
--- a/lib/esp32-nic/Dockerfile
+++ b/lib/esp32-nic/Dockerfile
@@ -1,6 +1,7 @@
 FROM ubuntu:22.04
 
-RUN apt-get update && apt-get -y install \
+RUN apt-get update \
+    && apt-get -y install \
         bash \
         git \
         cmake \
@@ -22,5 +23,13 @@ RUN apt-get update && apt-get -y install \
     && rm -rf /var/lib/apt/lists/*
 
 RUN git clone -b v5.0.3 --recursive https://github.com/espressif/esp-idf.git
-RUN cd /esp-idf && ./install.sh
-RUN echo "source /esp-idf/export.sh" >> ~/.bashrc
+
+# Install tools into a world-accessible path so the container can be run as
+# an arbitrary non-root user (--user flag) without losing access to the tools.
+# Pin setuptools<82: pkg_resources was removed in setuptools v82, ESP-IDF v5.0 still needs it.
+# Set git safe.directory so export.sh works for non-root users (git >= 2.35.2).
+ENV IDF_TOOLS_PATH=/opt/espressif
+RUN cd /esp-idf && ./install.sh \
+    && /opt/espressif/python_env/idf5.0_py3.10_env/bin/pip install "setuptools<82" \
+    && chmod -R a+rX /opt/espressif \
+    && git config --system --add safe.directory /esp-idf
--- a/lib/AddEspNicFw.cmake
+++ b/lib/AddEspNicFw.cmake
@@ -1,0 +1,136 @@
+# AddEspNicFw.cmake
+#
+# Handles ESP NIC firmware for printers with HAS_ESP_FLASH_TASK.
+#
+# By default the pre-built binaries checked into src/resources/<chip> are used. Set
+# ESP_FW_BINARY_DIR to empty to build the firmware from source via Docker.
+#
+# The Docker image is rebuilt only when the Dockerfile changes; the firmware is rebuilt only when
+# the ESP NIC source files change.
+
+if(NOT HAS_ESP_FLASH_TASK)
+  return()
+endif()
+
+# Derive chip and in-container SDK path from the hardware feature flags.
+if(HAS_EMBEDDED_ESP32)
+  set(ESP_CHIP "esp32")
+  set(ESP_SDK_EXPORT_SH "/esp-idf/export.sh")
+else()
+  set(ESP_CHIP "esp8266")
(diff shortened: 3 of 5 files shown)


lib/esp32-nic/Dockerfile
```diff
@@ -22,5 +23,15 @@ RUN apt-get update && apt-get -y install \
...
 
 RUN git clone -b v5.0.3 --recursive https://github.com/espressif/esp-idf.git
+
+# Install tools into a world-accessible path so the container can be run as
+# an arbitrary non-root user (--user flag) without losing access to the tools.
+ENV IDF_TOOLS_PATH=/opt/espressif
 RUN cd /esp-idf && ./install.sh
+# Pin setuptools to <82 to avoid pkg_resources removal (breaking change in setuptools v82)
+# ESP-IDF v5.0 still relies on pkg_resources which was removed in setuptools v82.0.0
+RUN /opt/espressif/python_env/idf5.0_py3.10_env/bin/pip install "setuptools<82"
```
Tomcus: IIRC, Consecutive `RUN`s increase the image size significantelly. It is prefered to do the `&& \\` trick

lib/esp32-nic/Dockerfile
```diff
@@ -1,6 +1,7 @@
 FROM ubuntu:22.04
 
-RUN apt-get update && apt-get -y install \
+RUN apt-get -o Acquire::AllowInsecureRepositories=true update \
+    && apt-get -y --allow-unauthenticated install \
```
Tomcus: Why all this unauthicated stuff?

utils/build_esp_fw.sh
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env bash
```
Tomcus: We are still support multiplatform project, even if I don't agree, so I don't think shell script is much portable for Windows.

Tomcus: IIRC `add_custom_command` generates a "shell" script, so you might be able to do that in a "portable" way. Also you might be able to detect docker etc.

CMakeLists.txt
```diff
@@ -414,6 +414,107 @@ if(BOARD IN_LIST BUDDY_BOARDS AND HAS_PUPPIES_BOOTLOADER)
...
+      "${_esp_output_dir}/partition-table.bin" ${_esp_copy_stubs}
+    DEPENDS ${_esp_sources} "${_esp_docker_stamp}"
+    COMMENT "Building ${ESP_CHIP} NIC firmware"
+    USES_TERMINAL VERBATIM
+    )
+
+  add_custom_target(
+    esp-nic-fw ALL DEPENDS "${_esp_output_dir}/uart_wifi.bin" "${_esp_output_dir}/bootloader.bin"
+                           "${_esp_output_dir}/partition-table.bin" ${_esp_stub_outputs}
+    )
```
Tomcus: 1) Does this needs to be this complicated?
2) Can we move to separate file?
3) Can we move it from root `CMakeLists.txt`?

ProjectOptions.cmake
```diff
@@ -342,6 +342,33 @@ set_feature_for_printers(HAS_FILAMENT_SENSORS_MENU "XL" "COREONE")
...
+    set(ESP_FW_SOURCE_DIR
+        "${CMAKE_SOURCE_DIR}/lib/${ESP_CHIP}-nic"
+        CACHE PATH "Source directory for the ESP NIC firmware (lib/esp32-nic or lib/esp8266-nic)."
+        )
+    set(ESP_FW_BUILD_DIR
+        "${CMAKE_BINARY_DIR}/esp-fw-build"
+        CACHE PATH "Build directory for ESP NIC firmware built from source."
+        )
+  endif()
+endif()
```
Tomcus: These don't define any Project Options. I would personally put them in the `lib` subfolder. Maybe I would add `AddEspNicFw.cmake` with most of the changes and ifs here.

CMakeLists.txt
```diff
@@ -414,6 +414,107 @@ if(BOARD IN_LIST BUDDY_BOARDS AND HAS_PUPPIES_BOOTLOADER)
   endif()
 endif()
 
+#
+# Build ESP NIC Firmware
+#
+
+if(BOARD IN_LIST BUDDY_BOARDS
+   AND HAS_ESP_FLASH_TASK
+   AND NOT ESP_FW_BINARY_DIR
```
Tomcus: Maybe we can simply it to the point `AND NOT DEFINED ESP_FW_BINARY_DIR`?

Tomcus (changes requested):
This needs a lot of work.

Sidenote: I hate our cmake formatter.

lib/esp32-nic/Dockerfile
bkerler: Yep, good point.

lib/esp32-nic/Dockerfile
bkerler: This was a left over due to clock drift issue and gpg failing in Docker. Will remove it (unneeded)

ProjectOptions.cmake
bkerler: Agreed :)

utils/build_esp_fw.sh
bkerler: using add_custom_command now, should be more portable this way

CMakeLists.txt
bkerler: Will move it to lib/AddEspNicFw.cmake

CMakeLists.txt
bkerler: Can't use DEFINED — it's a CACHE PATH with a default so it's always defined. NOT ESP_FW_BINARY_DIR correctly detects the user explicitly setting it empty to opt into the source build. Except you have any other ideas around it

Issue #5005, opened 2025-12-03
[BUG] M30 command does not delete a file on the USB drive


timur-tabi:
### Printer model

Core One

### Firmware version

6.4

### Upgrades and modifications

None

### Printing from...

USB Drive

### Describe the bug

<img width="1542" height="1375" alt="Image" src="https://github.com/user-attachments/assets/96e8738a-8607-410b-ad25-a1c2e319ee26" />

You can see here that I'm trying to delete a file with M30 command, but it always returns a failure message.  No matter how I format the M30 command, it just responds with an error.

ttabi:
It's still a real bug.

bkerler:
And for real bugs, we provide real fixes :)

ttabi:
Not yet merged.

PR #5190, opened 2026-03-23
gcode: M30 - Properly handle absolute paths and relative paths

bkerler:
This fixes issue #5005.

The M30 command (delete file on SD/USB) was failing because it didn't properly handle:

- Leading whitespace in the filename argument
- Simplify3D format (file size appended after space)
- Absolute paths that already include /usb/ or /sd/ prefix

### DIFF
--- a/src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
+++ b/src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
@@ -1,4 +1,5 @@
 #include <dirent.h>
+#include <string.h>
 
 #include "../../lib/Marlin/Marlin/src/gcode/gcode.h"
 #include "../src/common/print_utils.hpp"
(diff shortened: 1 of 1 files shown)


src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
```diff
@@ -199,14 +200,41 @@ void GcodeSuite::M29() {
  *    M30 [filename]
  */
 void GcodeSuite::M30() {
+    // Skip leading whitespace
+    const char *filename = parser.string_arg;
```
ttabi, line 203: Do you need to test for NULL first?

src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
```diff
@@ -199,14 +200,41 @@ void GcodeSuite::M29() {
...
+    while (*filename == ' ') {
+        filename++;
+    }
+
+    // Handle Simplify3D format which includes file size after a space (same as M23)
+    for (char *fn = const_cast<char *>(filename); *fn; ++fn) {
+        if (*fn == ' ') {
```
ttabi, line 219: Does the firmware handle escape characters?  That is, if I type in hello\ world.gcode, will it recognize it the entire filename?  Does the firmware allow the user to upload files with spaces in their names?

src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
```diff
@@ -199,14 +200,41 @@ void GcodeSuite::M29() {
...
+            *fn = '\0';
+            break;
+        }
+    }
+
+    // Skip empty filename
+    if (!filename || !*filename) {
```
ttabi: The !filename test is too late here.

src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
```diff
@@ -199,14 +200,41 @@ void GcodeSuite::M29() {
...
+        SERIAL_ECHOLNPGM("Deletion failed: No filename provided");
+        return;
+    }
+
     ArrayStringBuilder<FF_MAX_LFN> filepath;
-    filepath.append_printf("/usb/%s", parser.string_arg);
+
+    // Handle paths that may already have /usb/ or /sd/ prefix
+    if (strncmp(filename, "/usb/", 5) == 0 || strncmp(filename, "/sd/", 4) == 0) {
+        filepath.append_string(filename);
```
ttabi, line 235: You need to test whether the filename is too long to fit into filepath.

bkerler: ArrayStringBuilder already tracks overflows, but yeah, will check the is_ok function.

src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
bkerler: Right.

src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
bkerler: if we need to check for nullptr, then yes

src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
bkerler: The gcode parser has no escape character mechanism. The truncation-at-space is intentional — it mirrors what M23 does for Simplify3D compatibility (where the host appends a file size, e.g. M30 filename.gcode 12345).

Issue #5004, opened 2025-12-03
[BUG] M24 command does not start print


timur-tabi:
### Printer model

Core One

### Firmware version

6.4

### Upgrades and modifications

None

### Printing from...

USB Drive

### Describe the bug

<img width="1542" height="1375" alt="Image" src="https://github.com/user-attachments/assets/c56134a5-34c9-434e-b88e-58e05b6a5448" />

You can see here that I'm trying to start a print with the M24 command (M30 also does not work), and it's just ignored.  No matter how I format the M24 command, it just responds with Ok but does not actually start any print.

ttabi:
It's still a real bug.

bkerler:
And here's another fix for another real bug.

bkerler:
PR will be merged in upcoming FW.

ttabi:
Not yet merged.

PR #5191, opened 2026-03-23
[BFW-8560] gcode: M24 provide proper file path for print_start

bkerler:
This fixes issue #5004.

The bug is in `src/common/marlin_server.cpp` in the print_resume() function (line 1766):
When M24 is called and the printer is not in a paused/resuming state (i.e., starting a new print), the code falls through to the else branch which calls:
`print_start(nullptr, GCodeReaderPosition(), marlin_server::PreviewSkipIfAble::all);`
However, print_start() has an early return at line 1243-1245:

```
if (filename == nullptr) {
    return;
}
```

This causes the print to silently fail because:

M23 <filename> correctly stores the selected file path in "marlin_vars().media_SFN_path"
M24 then calls print_resume() which calls print_start(nullptr, ...)
print_start() immediately returns due to the nullptr check and thus the print never actually starts

### DIFF
--- a/src/common/marlin_server.cpp
+++ b/src/common/marlin_server.cpp
@@ -1763,7 +1763,7 @@ void print_resume(void) {
         server.print_state = State::PowerPanic_Resume;
 #endif
     } else {
-        print_start(nullptr, GCodeReaderPosition(), marlin_server::PreviewSkipIfAble::all);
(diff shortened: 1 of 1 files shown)


bkerler:
Closing as it is being merged.

PR #5195, opened 2026-03-23
WIFI: Fix connectivity issues on Wifi 6 Routers and WPA2/WPA3 transitioning mode

bkerler:
This issue solves #3864.

Explanation

**ESP32**: sae_pwe_h2e defaults to WPA3_SAE_PWE_UNSPECIFIED (0), which fails with routers that require the H2E (Hash-to-Element) PWE method especially for Wifi 6 Routers. IDF v5.0 introduced WPA3_SAE_PWE_BOTH = 3 to support both methods. Without it, SAE authentication fails in transition mode. Additionally, CONFIG_ESP32_WIFI_NVS_ENABLED=1 causes WPA3 PMKSA entries to persist in NVS across reboots, and esp_wifi_restore() called at startup may not clear these PMKSA entries, causing stale-PMK authentication failures on subsequent boots.

**ESP8266**: The ESP8266 RTOS SDK's WPA3 SAE implementation doesn't reliably handle WPA3/WPA2 transition mode (the SDK is based on IDF v3.4 with backported WPA3 patches, without H2E support). The fix is to disable WPA3 SAE for ESP8266 entirely (CONFIG_ESP8266_WIFI_ENABLE_WPA3_SAE=n), which forces WPA2 PSK connections. WPA3/WPA2 transition mode routers must accept WPA2 PSK by definition.

**toDo before merging:**
1. Wait for adoption of PR #5189 (for wifi build system)
2. Test PR against several WIFI setups (mikrotik ? wifi 6, 6e and 7 hotspots ?)
2. Update FW version in both ESP32/ESP8266 and printer fw
3. Build fw blob


bkerler:
This PR is being merged into #5188

Issue #5174, opened 2026-03-19
[BUG]Error while printing with Prusa Connect, "G-code transfer running too slow", print stops/fails 6.5.3-RC


safetychief8:
### Printer model

Core One+

### Firmware version

6.5.3-RC

### Upgrades and modifications

None

### Printing from...

Prusa Connect

### Describe the bug

Error while printing with Prusa Connect, "G-code transfer running too slow", print stops/fails. It's happened twice with this firmware. 

### How to reproduce

Print via Connect. I'll close connect and print directly from USB.

safetychief8:
The Prusa connect transfer gets to 87-88% and stops with this error message.

safetychief8:
I flashed back to the 6.5.3-Beta firmware otherwise I won't be able to print anything.

Kleins4T3D:
This exact issue occurred on my MK4S and MMU3 after updating both to new firmware. Has never occurred in the past. Internet speed not an issue. Was 1/3 of the way into print. Seemed upload to printer just froze. Tried forcing a power failure, print restarted, nozzle Reisen’s come to temp, but printer moves occurred in X,Y, & Z.

Alienbliep:
Look while printing if your printer is offline in connect/app.
Or on the printer screen in settings-> network and then connect.
You will propably see an net fail error.
It is the connector dropping out when an print is started and thus cannot complete the transfer.

I have the same issue.

Alienbliep:
![Image](https://github.com/user-attachments/assets/7d6b87b5-0d5e-4e54-81ca-6c70e48d8086)

Alienbliep:
![Image](https://github.com/user-attachments/assets/c4b87d1e-0801-4ee8-acf7-b863c8f3b9f6)

![Image](https://github.com/user-attachments/assets/f6424603-6e26-4d3b-ba23-ee0aacacdb8f)

Just happened again

Alienbliep:
Reverting back firmware does not help. So at this moment it is waiting to use my 5 days old printer.

Only manual transfer to usb will work or click only trasnfer to prjnter and wait for it to 100% complete transfer before starting the print

safetychief8:
Alienbliep, the other option was just to disable Connect and print from USB, but the 6.5.3 Beta build had no issues so far, so I reverted to that.

Alienbliep:
Ok maybe you and i have different problems.

safetychief8:
[...]

Issue #5167, opened 2026-03-15
[BUG] CORE One L 6.5.3. beta connect "net fail" drop out


Alienbliep:
### Printer model

CORE One L

### Firmware version

6.5.3 beta

### Upgrades and modifications

MMU3

### Printing from...

USB

### Describe the bug

When the print starts "Connect" drops out(net fail in connect network menu). In the app and on connect the printer is offline.
Local ip / prusalink works.

When i stop the print the connection cames back online again.

This can be repeated over and over

![Image](https://github.com/user-attachments/assets/4daee5df-1fdd-40a7-b741-eb1d18285344)

### How to reproduce

Start an print and see in menu during print if the (network-> connect) if "net fail"appears.
Stop the print to see it coming back online.

### Expected behavior

To just work.

Alienbliep:
I downgraded to the stable version 6.5.2. and it did not help. The bootloader also does nto go back to 2.5.0

Alienbliep:
6.5.3. RC also does not solve the problem for me.

Alienbliep:
Downgraded to 6.5.1. does not help either.
Is this going to be fixed or can i ask for an new xbuddy board?
I need to remotely control the printer

bkerler:
I suspect the following:

The connect task (osPriorityBelowNormal) is starved of CPU during active printing on CORE One due to:

Media prefetch temporarily elevating to osPriorityNormal
Phase stepping and motion ISRs consuming significant CPU
TLS handshake timeouts when Connect can't get enough cycles

https://github.com/prusa3d/Prusa-Firmware-Buddy/blob/master/include/buddy/priorities_config.h#L64

```
// Currently :
#define TASK_PRIORITY_CONNECT                osPriorityBelowNormal
#define TASK_PRIORITY_TCPIP_THREAD       osPriorityBelowNormal

// After :
#define TASK_PRIORITY_CONNECT                osPriorityNormal
#define TASK_PRIORITY_TCPIP_THREAD       osPriorityNormal
```

bkerler:
Just trying 6.5.3 and I cannot reproduce this issue on Core One L, but I haven't got MMU connected. It would be helpful if you could sniff your local network (seems like you are connected via ethernet ?) to rule out possible network issues. Maybe related to issue #5115.

Alienbliep:
Sorry. I just updated to the 6.5.3. stable and 3.0.4. and it is still not working.

[...]

PR #5205, opened 2026-03-25
task priorities: raise CONNECT and TCPIP_THREAD to osPriorityNormal

bkerler:
During active printing, media_prefetch temporarily raises the async executor thread to TASK_PRIORITY_MEDIA_PREFETCH (osPriorityNormal) for the duration of each fetch_routine call.  With TASK_PRIORITY_CONNECT and TASK_PRIORITY_TCPIP_THREAD both at osPriorityBelowNormal, this caused two distinct failures:

1. "net fail" / Connect drops offline while printing (issue #5167). The TCPIP thread could not process incoming TCP segments or send ACKs while prefetch was running.  The Connect task was also starved, missing the 4-second WebSocket telemetry window and eventually the 15-second ping-inactivity threshold, causing the server to mark the printer offline.

2. "G-Code transfer running too slow" when printing from Connect (issue #5174).  When printing via inline WebSocket streaming, the Connect task receives chunks from the server and writes them to the partial download file.  Media prefetch reads that same file.  With prefetch elevated above Connect, Connect could not write new chunks while prefetch was reading — a priority inversion where prefetch hit RESULT_OUT_OF_RANGE on data that the starved Connect task had not yet delivered.

Fix: raise both tasks to osPriorityNormal so they are peers of the elevated prefetch thread rather than subordinates to it.  They now share CPU fairly with media prefetch during I/O, which is the same relationship the old dual-priority scheme (TASK_PRIORITY_MEDIA_PREFETCH _WHILE_FREAD) provided before BFW-5653 removed it.

TASK_PRIORITY_WUI is left at osPriorityBelowNormal (local PrusaLink remains functional and is latency-tolerant).  TASK_PRIORITY_ASYNC_JOB_ EXECUTOR remains at osPriorityBelowNormal; the executor only needs elevation during active fetch work, which fetch_routine already handles via osThreadSetPriority.


bkerler:
Moved to draft as Prusa devs said that raising priorities (race print vs. tcpip) is not an option

Issue #5115, opened 2026-02-03
[BUG] prusalink network stack problems with big UDP-broadcasts -> 408 ERROR


boecko:
**EDIT**: See post https://github.com/prusa3d/Prusa-Firmware-Buddy/issues/5115#issuecomment-3977273507 for the problem description.

### Printer model

MK4S (meanwhile COREOne)

### Firmware version

6.4.0+11974

### Upgrades and modifications

None

### Printing from...

PrusaLink, PrusaConnect

### Describe the bug

Hi,

i'm experiencing constant 408-errors if I try to upload a file via PrusaLink locally. Sometimes it is pure luck, that the upload finishes. 
Currently i'm trying to upload a 11MB print  via connect. It lasts *about an hour* to do that, which is very frustrating.
It seems to be worse, when a print job is running atm.
<img width="882" height="203" alt="Image" src="https://github.com/user-attachments/assets/7721d8d5-69a4-4165-9dc8-ccadeec3a454" />

The printer is connected via ethernet. So it can't be any wifi issue. 
```
$ ping prusa-mk4s
PING prusa-mk4s.empire.glx (192.168.66.71) 56(84) bytes of data.
64 bytes from prusa-mk4s.empire.glx (192.168.66.71): icmp_seq=1 ttl=255 time=3.72 ms
64 bytes from prusa-mk4s.empire.glx (192.168.66.71): icmp_seq=2 ttl=255 time=1.80 ms
64 bytes from prusa-mk4s.empire.glx (192.168.66.71): icmp_seq=3 ttl=255 time=3.58 ms
64 bytes from prusa-mk4s.empire.glx (192.168.66.71): icmp_seq=4 ttl=255 time=1.09 ms
64 bytes from prusa-mk4s.empire.glx (192.168.66.71): icmp_seq=5 ttl=255 time=3.22 ms
64 bytes from prusa-mk4s.empire.glx (192.168.66.71): icmp_seq=6 ttl=255 time=1.78 ms
64 bytes from prusa-mk4s.empire.glx (192.168.66.71): icmp_seq=7 ttl=255 time=1.53 ms
64 bytes from prusa-mk4s.empire.glx (192.168.66.71): icmp_seq=8 ttl=255 time=2.33 ms
64 bytes from prusa-mk4s.empire.glx (192.168.66.71): icmp_seq=9 ttl=255 time=2.50 ms
[...]

PR #5208, opened 2026-03-28
network: Drop non unicast UDP packets

bkerler:
This PR solves issue #5115
This change addresses upload stalls and 408 Request Timeout failures caused by large fragmented UDP broadcast/multicast traffic overwhelming the networking stack.

The issue report showed that the printer was receiving repeated large OpenWrt usteer UDP broadcasts, attempting IPv4 fragment reassembly for them, and eventually emitting ICMP "fragment reassembly time exceeded" replies. While that reassembly state was active, PrusaLink / PrusaConnect traffic could stall badly enough to break uploads.

Instead of increasing the RX pool size globally as suggested in PR #5185, this PR drops fragmented non-unicast IPv4 UDP traffic before it reaches lwIP reassembly:

Ethernet path drops it before pbuf allocation
ESP/Wi-Fi path drops it before handing the packet to lwIP
This keeps normal unicast traffic unchanged and avoids a large always-on RAM increase.

### DIFF
--- a/lib/WUI/espif.cpp
+++ b/lib/WUI/espif.cpp
@@ -38,6 +38,9 @@
 #include <lwip/def.h>
 #include <lwip/ethip6.h>
 #include <lwip/etharp.h>
+#include <lwip/prot/ethernet.h>
+#include <lwip/prot/ip.h>
+#include <lwip/prot/ip4.h>
 #include <lwip/sys.h>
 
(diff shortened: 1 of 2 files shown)


jabdoa2:
I definitely have this issue. Prusa send me a new Wifi module (which obviously did not fix the issue)

PR #5194, opened 2026-03-23
Prusalink: Add chamber temps to status

bkerler:
This addresses issue #5148.

Added a chamber top-level object to the /api/v1/status JSON response, conditionally compiled per printer model:

CoreONE (HAS_XBUDDY_EXTENSION()):
"chamber": {
    "temp": 35.5,
    "target_temp": 35,
    "fan_1_rpm": 1234,
    "fan_2_rpm": 1234
}
XL with enclosure (XL_ENCLOSURE_SUPPORT(), when enclosure is enabled):

"chamber": {
    "temp": 35.5,
    "fan_rpm": 1234
}
Other printers: no chamber object (compile-time excluded).

PR #4780, opened 2025-08-26
prusalink: Add mcu/board/sandwich/splitter temperature values to PrusaLink

bkerler:
Refers to issue #4078.
This adds mcu and board temp to "printer" api request of PrusaLink V1. For XL sandwich and splitter temp is added. If Chamber Temp exists, this is being returned as well.

Example json output (on MK4 simulator):
`{"telemetry": {"temp-bed": 18.6, "temp-nozzle": 22.0, "print-speed": 100, "z-height": 0.0, "material": "---"}, "temperature": {"tool0": {"actual": 22.0, "target": 0.0, "display": 0.0, "offset": 0}, "bed": {"actual": 18.6, "target": 0.0, "offset": 0}, "board": {"actual": 24.5}, "mcu": {"actual": 0}}, "state": {"text": "Operational", "flags": {"operational": true, "paused": false, "printing": false, "cancelling": false, "pausing": false, "error": false, "sdReady": false, "closedOnError": false, "ready": true, "busy": false}}}
`

PR #5229, opened 2026-04-06
feat(wui): Add extended fields to /api/v1/status

packerlschupfer:
## Summary

Add commonly needed telemetry fields to the `/api/v1/status` printer object that are already available in `marlin_vars` but missing from the local PrusaLink API.

## New fields

| Field | Type | Description |
|-------|------|-------------|
| `temp_heatbreak` | float | Heatbreak temperature [°C] |
| `target_heatbreak` | float | Heatbreak target temperature [°C] |
| `print_fan_speed` | int | Print fan PWM value (0-255) |
| `job_id` | int | Current print job ID |
| `print_duration` | int | Print time [ms] |
| `time_to_end` | int | Estimated remaining time [s] (only during print) |
| `progress` | int | Print completion (0-100%) |

## Motivation

These fields are already exposed through Prusa Connect telemetry but missing from the local API, forcing local integrations (Home Assistant, custom dashboards) to either use the cloud API or poll multiple endpoints. This is a simple addition with no breaking changes.

## Test plan

- [ ] `GET /api/v1/status` returns new fields while idle
- [ ] `time_to_end` appears only during active print
- [ ] `progress` updates during print
- [ ] No regressions on existing fields

### DIFF
--- a/lib/WUI/nhttp/status_renderer.cpp
+++ b/lib/WUI/nhttp/status_renderer.cpp
@@ -74,7 +74,16 @@ json::JsonResult StatusRenderer::renderState(size_t resume_point, json::JsonOutp
             JSON_FIELD_INT("flow", VirtualToolIndex::currently_selected_opt().transform([] (auto tool) { return marlin_vars().virtual_tools[tool].flow_factor.get(); }).value_or(0)) JSON_COMMA;
             JSON_FIELD_INT("speed", marlin_vars().print_speed) JSON_COMMA;
             JSON_FIELD_INT("fan_hotend", marlin_vars().active_hotend().heatbreak_fan_rpm) JSON_COMMA;
-            JSON_FIELD_INT("fan_print", marlin_vars().active_hotend().print_fan_rpm);
+            JSON_FIELD_INT("fan_print", marlin_vars().active_hotend().print_fan_rpm) JSON_COMMA;
+            JSON_FIELD_FFIXED("temp_heatbreak", marlin_vars().active_hotend().temp_heatbreak, 1) JSON_COMMA;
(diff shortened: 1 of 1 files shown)


packerlschupfer:
For visibility — this PR touches `status_renderer.cpp` and so does **#5194** (chamber object — `temp` / `target_temp` / fan RPMs to `/api/v1/status`). The two add different fields so there's no semantic overlap; just a likely merge conflict in the same file.

Happy to rebase #5229 onto whichever lands first. No gating in either direction.

(#4780 is unrelated to this PR — it modifies `basic_gets.cpp` for the legacy `/api/v1/printer` endpoint, not the V1 status shape.)

jabdoa2:
What do you think about also exposing the "gcode" variable? You can also get this var via metrics but it will never be repeated. So if you the printer is waiting on a dialog you will never get it (unless you already listened before the dialog).

packerlschupfer:
Great point about `gcode`, @jabdoa2 — the metrics-only-emit-once behavior is a real gap for exactly the dialog-pause case you flagged. When the printer stops on an M0 or similar, the last-seen gcode fires down the metrics stream once and then goes silent forever, so any client that connected after the stop has no way to recover it.

Adding it to `/api/v1/status` would fix that (status is polled, so freshness is free). It would fit either as `marlin_vars().gcode` alongside the other extended fields in this PR, or as a follow-up. I'd lean toward putting it in a follow-up so this PR stays a focused scope — happy to open one right after this lands (or before if you'd rather chase it in parallel).

Also just cleaned up the PR branch here — it had accumulated a lot of unrelated history. Now it's a single commit touching one file (`status_renderer.cpp`), should be trivial to review.

packerlschupfer:

On the `gcode` variable follow-up I offered: I won't be picking that up. If someone else wants to extend this PR (or open a follow-up) once it lands, please do.

Thanks for the reviews!

PR #5228, opened 2026-04-06
feat(wui): Add unified POST /api/v1/control endpoint

packerlschupfer:
## Summary

Add a single `POST /api/v1/control` endpoint for controlling the printer via the local PrusaLink HTTP API. This enables local network automation and Home Assistant integration without depending on Prusa Connect.

**Motivation:** The local PrusaLink API currently provides read-only status and job control, but no way to set temperatures, speed, fan settings, or trigger filament operations. Users wanting local automation (e.g. Home Assistant, custom scripts) must use the Prusa Connect cloud API. This PR adds a local-first alternative.

## API

`POST /api/v1/control` with a JSON body. All fields optional — omitted fields leave the state unchanged.

```json
{
  "nozzle": 215,          "bed": 60,              "chamber": 40,
  "speed": 100,           "flow": 100,
  "fan_print": 255,       "chamber_fan": -1,      "led": 100,
  "home": true,           "move_z": 5.0,          "feedrate": 1000,
  "load_filament": true,  "unload_filament": true,
  "cooldown": true,       "motors_off": true,     "reboot": true,
  "gcode": "M220 S100"
}
```

## Implementation

- New `ControlCommand` handler (same pattern as `JobCommand`/`FileCommand`)
- JSON parsing via existing `json_parser.h` infrastructure
- Marlin interaction via `marlin_client::gcode()` for standard commands
- Chamber/fan/LED use native APIs (`buddy::chamber()`, `buddy::xbuddy_extension()`, `leds::SideStripHandler`) gated behind `HAS_XBUDDY_EXTENSION()` and `HAS_SIDE_LEDS()`
- Route registered in `prusa_link_api_v1.cpp`

## Files changed

| File | Change |
|------|--------|
| `control_command.h` | Handler class declaration |
| `control_command.cpp` | JSON parsing + dispatch |
| `control_command_marlin.cpp` | Marlin/hardware interaction (separated for test isolation) |
| `prusa_link_api_v1.cpp` | Route registration (+12 lines) |
| `handler.h` | Variant type registration (+2 lines) |
| `CMakeLists.txt` | Build files (+2 lines) |

## Test plan

- [ ] `curl -X POST ... -d '{"speed": 150}'` sets speed to 150%
- [ ] `curl -X POST ... -d '{"nozzle": 215, "bed": 60}'` sets targets
- [ ] `curl -X POST ... -d '{"cooldown": true}'` turns off all heaters
- [ ] `curl -X POST ... -d '{"home": true}'` homes all axes
[...]

PR #5230, opened 2026-04-06
feat(wui): Add dialog state to /api/v1/status + normal load/unload dialogs

packerlschupfer:
## Summary

Expose active printer dialogs in the `/api/v1/status` response and implement the upstream TODO for normal filament load/unload dialog reporting.

## Changes

1. **Dialog in status API**: When the printer shows a dialog (filament prompt, crash recovery, warning), the status response includes a `dialog` object with the dialog ID, error code, and available button names.

2. **Normal load/unload dialogs** (`printer_state.cpp`): Implements the TODO at line 393 — dialogs for non-printing load/unload operations are now exposed with their available response buttons, not just while-printing ones.

## Response format

```json
{
  "printer": { "state": "BUSY", ... },
  "dialog": {
    "id": 12345,
    "code": 31829,
    "button0": "Stop",
    "button1": "",
    "button2": "",
    "button3": ""
  }
}
```

The `dialog` object is only present when a dialog is active. Button fields are empty strings when not available.

## Motivation

This enables remote printer management tools (Home Assistant, custom dashboards) to detect and display printer dialogs. Combined with a control endpoint (separate PR #5228), this allows fully remote dialog response.

## Test plan

- [ ] Trigger filament unload → dialog appears in status with "Stop" button
- [ ] No dialog object when printer is idle
- [ ] Crash recovery dialog shows correct buttons
- [ ] Warning dialogs reported correctly

### DIFF
--- a/lib/WUI/nhttp/status_renderer.cpp
+++ b/lib/WUI/nhttp/status_renderer.cpp
@@ -6,6 +6,8 @@
 #include <transfers/monitor.hpp>
 #include <segmented_json_macros.h>
 
+#include <marlin_server_types/general_response.hpp>
+
 #include <option/buddy_enable_connect.h>
 #if BUDDY_ENABLE_CONNECT()
     #include <connect/connect.hpp>
@@ -27,7 +29,12 @@ json::JsonResult StatusRenderer::renderState(size_t resume_point, json::JsonOutp
 
     uint32_t time_to_end = marlin_vars().time_to_end;
     uint32_t time_to_pause = marlin_vars().time_to_pause;
-    auto link_state = printer_state::get_state(false);
+    auto state_with_dialog = printer_state::get_state_with_dialog(false);
+    auto link_state = state_with_dialog.device_state;
+    bool has_dialog = state_with_dialog.dialog.has_value();
+    uint32_t dialog_id = has_dialog ? state_with_dialog.dialog->dialog_id.to_uint32_t() : 0;
+    uint32_t dialog_code = state_with_dialog.code_num();
+    const Response *dialog_buttons = state_with_dialog.buttons();
 
     // Keep the indentation of the JSON in here!
     // clang-format off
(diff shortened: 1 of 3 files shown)


jabdoa2:
We need this! Would it be possible to also show the actual message from `M0` G-Codes? Currently, this is also not available in PrusaConnect. I guess the exact message (i.e. "Insert magnets") would also be helpful in Home Assistant. They certainly would be helpful in my scripts.

packerlschupfer:
Good catch, @jabdoa2 — you're right that the current PR doesn't include the message body. It exposes:

```json
{
  "dialog": {
    "id": 12345,
    "code": 31829,      // numeric error code, not human text
    "button0": "Stop",
    "button1": "",
    "button2": "",
    "button3": ""
  }
}
```

The "Insert magnets" text lives in the fsm state's message field (populated from M0's payload or from the error registry lookup). It's not currently piped through to the JSON. Adding it would be a clean extension — I'd want to bound the JSON field (probably 128 or 256 chars max) to avoid stack blowups on long custom messages.

Happy to add it — either extend this PR to include `"message": "Insert magnets"` in the `dialog` object, or land this first and follow up with a message-text PR. Let me know your preference.

Also just cleaned up this PR branch — it had picked up a lot of unrelated history from a stale fork base. Now it's 3 files, 29 lines added, all directly related to the dialog reporting logic.

packerlschupfer:
Filed the follow-up as **#5400**: [feat(wui): expose M0/QuickPause dialog message in /api/v1/status](https://github.com/prusa3d/Prusa-Firmware-Buddy/pull/5400). It adds a `"message": "Insert magnets"` field to the dialog object here (small change since the M0 handler was already stashing the message pointer in the FSM data — just needed to unpack and emit).

Base is `master`, so its diff currently includes this PR's changes plus the new field. When this one lands, #5400's diff cleanly reduces to just the two-line message plumbing.

Verified builds. Bench test on my Core One+ once this PR (#5230) is merged.

packerlschupfer:

On the M0 message text follow-up: I did file that as **#5400** already (which cherry-picked cleanly onto the rebuilt dev-allpatches). It's still a fine PR to review, just won't get further iteration from me.

Thanks for the reviews!

Issue #5249, opened 2026-04-15
[BUG] sdpos value does not match file offset for (at least) G0 and G1 commands with BGCODE files


jabdoa2:
### Printer model

Original Core One+ (likely affects all Buddy-based models)

### Firmware version

6.5.3+12780

### Upgrades and modifications

None

### Printing from...

PrusaConnect/USB

### Describe the bug

The `sdpos` value as reported by the firmware or via the network does not always correctly represent the file offset in the G-code file being executed when printing BGCODE files. For most G-code commands, sdpos matches the actual byte position in the file, but for G0 and G1 commands there is an offset of 3 bytes for every such command (in my tests). This is not caused by comments or any dynamic source content; it happens even on a minimal G-code file specifically crafted to have no comments and no extra whitespace. The offset does NOT happen for M73, M201, M203, M204, M205, M486, M17, M862, M115, M555, M83, M84 (and possibly others). The mismatch defeats the purpose of the sdpos metric, as it does not allow external tools to seek to the right position for timelapses or job preview.

Root cause appears to be the `compact_gcode()` function in the `media_prefetch` system, which removes whitespace exclusively for Gx commands (G0, G1, etc.) before sdpos is finalized, causing the offset. This behavior is unique to Gx commands because the code strips out spaces between parameters (e.g. `G1 X10 Y20` becomes `G1X10Y20`), making the file positions no longer line up.

### How to reproduce

1. Upload a BGCODE file to the printer (Core One or similar)
2. Use a G-code file with only G0/G1 moves and no comments or extra whitespace, e.g.:
   G1 X10 Y10 Z1
   G1 X20 Y20 Z1
   G1 X30 Y30 Z1
3. Observe reported sdpos as each command is executed (via network/serial metric or logs)
4. Compare to expected file offset for each command start in the original file
5. The reported sdpos will be 3 bytes higher than the expected value for each G0/G1 command.
6. Try with other commands (e.g. M73, M201) and observe those do not have the offset.

### Expected behavior

The sdpos/position metrics reported by the firmware should always exactly match the byte offset for the current command in the G-code/BGCODE file being executed, with no extra offset for any instruction type. This is crucial for tools which wish to observe the current position of the print (i.e. for triggering timelapses at the right time).

### Files

Example minimal G-code file to reproduce (BGCODE encoded):
G1 X10 Y10 Z1
G1 X20 Y20 Z1
G1 X30 Y30 Z1

Actual offset difference:
[...]

PR #5290, opened 2026-05-23
link_content: Add byte_position to get_job_v1 response

ky095n:
## Summary

Adds two new fields to `GET /api/v1/job`:

- `byte_position` — current position in the loaded media file
- `byte_size_estimate` — estimated total size of the media file

Both come from existing lock-free atomics maintained by the media
subsystem (`marlin_vars().media_position` /
`.media_size_estimate`). No new state is introduced; this commit
only exposes data that already exists.

## Motivation

Closes the long-standing request in
[prusa3d/Prusa-Link-Web#444](https://github.com/prusa3d/Prusa-Link-Web/issues/444)
(open since 2023).

`GET /api/v1/job` currently exposes only `progress` (1 % granularity
→ ~2 layers of uncertainty on a 200-layer print). Fine for a
progress bar, insufficient for tools that need to align external
actions with print position (timelapse triggers, multi-device
coordination, recovery without re-sync from scratch).

The underlying values are already maintained, lock-free, in
`marlin_vars`. This PR is a two-line additive exposure of them.

## Accuracy

- **Plain-text `.gcode`:** position is byte-exact.
- **Binary `.bgcode`:** `media_position` can drift a few bytes per
  G0/G1 due to whitespace compaction in `compact_gcode()` — see
  [#5249](https://github.com/prusa3d/Prusa-Firmware-Buddy/issues/5249).
  The field is still useful as a progress indicator on BGCODE and
  exact on text gcode. A consumer needing byte-exact on BGCODE would
  need #5249 fixed first; this PR doesn't make that situation worse.

A code comment on the change explains both points so future readers
don't have to chase it.

## Test plan

- Built locally for the `coreone` preset (`python utils/build.py
  --preset coreone`) — clean build, both `coreone_release_boot` and
  `coreone_release_noboot` succeed.
- Verified the new field names landed in the build product:
  ```
  $ strings build/coreone_release_noboot/firmware \
      | grep -E 'byte_position|byte_size_estimate|time_printing'
  byte_position
  byte_size_estimate
  time_printing
  ```
  Both new strings sit alongside the existing `time_printing` field
  (same `JSON_FIELD_INT` macro), confirming the emission code path
  expands as expected and isn't dead-code-eliminated.
- **No new host unit test added.**
  `lib/WUI/link_content/basic_gets.cpp` is not currently covered by
  `tests/unit/`, and the file's transitive dependencies on
  `marlin_client.hpp` (FSM types, selftest, Marlin core, …) make
  standalone testing nontrivial. Happy to add test coverage in a
  follow-up PR if welcome.
- Did not flash on hardware.

## Consumer impact

Purely additive. Existing clients see the same fields they always
did, plus two new ones.


jabdoa2:
>     * **Binary `.bgcode`:** `media_position` can drift a few bytes per
>       G0/G1 due to whitespace compaction in `compact_gcode()` — see
>       [#5249](https://github.com/prusa3d/Prusa-Firmware-Buddy/issues/5249).
>       The field is still useful as a progress indicator on BGCODE and
>       exact on text gcode. A consumer needing byte-exact on BGCODE would
>       need [[BUG] sdpos value does not match file offset for (at least) G0 and G1 commands with BGCODE files #5249](https://github.com/prusa3d/Prusa-Firmware-Buddy/issues/5249) fixed first; this PR doesn't make that situation worse.

Unfortunately, a few bytes is quite an understatement. This can be easily 30% off after a long print. I tried to fix this in firmware and there are definitely more places where this goes wrong. I tried to get this accurate but after fixing multiple places it was still off and I gave up eventually. Tracking BGCode is simply not possible with the current firmware. I am currently trying to hack some kind of callbacks from the firmware to external services (via MQTT) to get some kind of reliable behaviour.

PR #5306, opened 2026-05-31
fix(M303): reject C < 3 instead of silently emitting zero Kp/Ki/Kd

packerlschupfer:
## Summary

`M303 C2` (or any C ≤ 2) runs the autotune to completion and prints `#define DEFAULT_Kp 0.00` / `Ki 0.00` / `Kd 0.00` with no indication that anything went wrong. The M303 docstring already says \"C — Cycles minimum 3\" but the parser does not enforce it. This PR adds the missing range check so the failure mode is loud instead of silent.

## Repro (before the fix)

```bash
$ curl -X POST -d '{\"gcode\":\"M303 E0 S225 C2 U1\"}' \\
       http://printer/api/v1/control
```

```
PID Autotune start
 T:30.72/0.00 ... heating ...
 bias: 245 d: 245 min: 223.11 max: 230.08 ...
 bias: 235 d: 235 min: 223.50 max: 230.93PID Autotune finished! Put the last Kp, Ki and Kd constants from below into Configuration.h
#define DEFAULT_Kp 0.00
#define DEFAULT_Ki 0.00
#define DEFAULT_Kd 0.00
```

The cycles ran, the bias / d / min / max were measured correctly, and yet the output Kp/Ki/Kd are all zero.

## Why it happens

`Temperature::PID_autotune()` has two relevant `cycles > 2` guards:

```cpp
if (!heating && current_temp < target) {
  if (ELAPSED(...)) {
    heating = true;
    ...
    if (cycles > 0) {
      // measure bias/d/min/max  (printed in our log above)
      if (cycles > 2) {
        // compute Ku, Tu, tune_pid.Kp/Ki/Kd  ←  never reached for C ≤ 2
      }
    }
    cycles++;
  }
}
...
if (cycles > ncycles && cycles > 2) {
  // emit \"PID Autotune finished!\" and #define DEFAULT_K{p,i,d}
}
```

The compute branch is checked **before** `cycles++`, so the highest value `cycles` ever has at that check is `ncycles`. For `C=2` that's 2, fails `cycles > 2`, and `tune_pid` stays at its initial `{0, 0, 0}`. The termination check on the next valley *also* checks `cycles > 2`, but by then `cycles == 3` and we exit — the inner compute simply never ran.

Rewriting the algorithm to make `C=2` work is risky (changes Ziegler–Nichols semantics) and not really meaningful: with two oscillation periods you don't have enough data to converge anyway. The honest fix is to enforce what the docstring already claims.

## The fix

```cpp
const int c = parser.intval('C', 5);
if (c < 3) {
  SERIAL_ECHOLNPGM(MSG_PID_AUTOTUNE_FAILED \" C must be >= 3\");
  return;
}
```

Default is unchanged at 5, so this only affects users who explicitly pass a low `C`. The new failure message uses the existing `MSG_PID_AUTOTUNE_FAILED` prefix so error parsing on the host side (PrusaSlicer, OctoPrint, etc.) sees a familiar shape.

## Test plan

- [x] `M303 E0 S225 C2 U1` now responds `echo:PID Autotune failed! C must be >= 3` and returns immediately, without heating.
- [x] `M303 E0 S225 C5 U1` (default) runs the full tune unchanged.
- [ ] CI smoke on all printer models — no platform-specific code paths touched.

## Notes

- Discovered while debugging a separate issue over the WUI gcode passthrough. Running `M303 ... C2` to get a quick smoke test gave back the zero-valued output above; tracing into `PID_autotune()` showed the gating bug.
- Same code is in mainline Marlin and presumably has the same behavior. Worth carrying upstream as well.

### DIFF
--- a/src/feature/nozzle_cleaner/nozzle_cleaner.cpp
+++ b/src/feature/nozzle_cleaner/nozzle_cleaner.cpp
@@ -69,11 +69,11 @@ static constexpr EnumArray<Sequence, GCodeFile, static_cast<int>(Sequence::_cnt)
         { Sequence::purge_clean, {
                                      .filename = "purge_clean",
                                      .directory = directory,
-                                     .default_gcode = "G750 Y85 F21000 A\n" // Eject poop and move back to purge position
+                                     .default_gcode = "G750 Y86 F21000 A\n" // Eject poop and move back to purge position
                                                       "G750 Y93 F21000 A\n"
-                                                      "G750 Y82 F21000 A\n"
-                                                      "G750 Y98.5 F21000 A\n"
-                                                      "G750 Y75 F21000 A\n"
+                                                      "G750 Y83 F21000 A\n"
+                                                      "G750 Y93 F21000 A\n"
+                                                      "G750 Y76 F21000 A\n"
                                                       "G750 Y93 F21000 A\n"
                                                       "G750 Y86.5 F21000\n"
                                                       "M906 P1\n" // Increase E current for purge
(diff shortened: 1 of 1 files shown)


packerlschupfer:
Closing this as no longer applicable — upstream removed the entire PID_AUTOTUNE feature in commit [`37f1f8aa0`](https://github.com/prusa3d/Prusa-Firmware-Buddy/commit/37f1f8aa0) (ticket BFW-8271, 2026-01-29) with the candid rationale:

> *The code was an unmanageable mess, the PID EEPROM was also not working.*
>
> *Don't have the budget to implement it properly now, so it's better to just remove it to spare us the burden of having to walk around the code.*

That commit stripped the entire `PID_autotune()` function body (~310 lines from `temperature.cpp`) and undefined `PID_AUTOTUNE` in every `Configuration_*_adv.h`. `M303.cpp` still exists but is gated by `#if HAS_PID_HEATING && ENABLED(PID_AUTOTUNE)` — with PID_AUTOTUNE undefined everywhere it's dead code, and M303 becomes an unknown command.

The bug this PR fixed (Ziegler-Nichols leaving `tune_pid` zeroed when `cycles < 3`) has no code left to apply to. Follow-up commits under the same ticket (`9bd932e64`, `487a48a50`, `da8ee5fe5`, `fc4c783da`, `9b7efe983`, `582f6e0e8`) refactored the internal PID stack to compile-time-baked `HotendRegulator`/`HeatbreakRegulator` classes with no user-tuning surface — a deliberate design direction, not something to replace.

Thanks for looking!

PR #5336, opened 2026-06-16
gui: Draw partial character in front of rolling text

espr14:


espr14:
Tested on MINI. `store_char_part_in_buffer` could be merged with `store_char_in_buffer`. Less code but a little slower. Draws the first character. The last character must be done differently (will be different PR).

PR #5357, opened 2026-06-29
Fix unsupported feature text termination

Old-Ding:
﻿## Summary

Fixes the unsupported feature text buffer so it is always NUL-terminated when the first unsupported feature fills the fixed-size display buffer.

`ValidPrinterSettings::add_unsupported_feature()` receives a feature name and explicit length parsed from G-code comments. The previous first-write path copied up to `sizeof(unsupported_features_text)` bytes with `strncpy()`, which can leave the buffer without a trailing NUL when the feature name is at least as long as the buffer. The GUI later passes this field as a C string.

## Changes

- Reserve one byte in `unsupported_features_text` for the trailing NUL.
- Explicitly terminate the buffer after the bounded first copy.
- Keep the existing append/truncation behavior otherwise unchanged.

## Validation

- `git diff --check origin/master..HEAD`
- `git diff --stat origin/master..HEAD`

Not run locally: unit tests/build. This Windows environment does not have a C/C++ compiler available (`gcc`, `g++`, `clang`, `cl`, and `arm-none-eabi-gcc` are not in PATH); `cmake` and `ninja` are also not installed system-wide.


Old-Ding:
I refreshed this against the current master and found that the original target code has been removed by the recent upstream history rewrite/refactor: `unsupported_features_text` and `ValidPrinterSettings::add_unsupported_feature()` no longer exist on master.

Since the unterminated copy path this PR fixed is gone, rebasing this PR would reintroduce old code instead of applying a useful fix. Closing it as obsolete.

PR #5369, opened 2026-07-06
INDX: Fix reset polarity on older xBuddy revisions

ScottKolo:
**Note to the team:** Feel free to use any part of this or none of it at all. It's just my attempt at trying to fix the problem. I have no way to safely test this or verify its correctness. It compiles, and it seems to fix a logic error - but beyond that, I have to leave it to you.

**Problem:** On xBuddy boards with BOM ID < 37 (rev 0.2.7, found in MK4 -> Core One upgrades), the INDX head fails at boot with a `Puppy INDX_HEAD not responding` redscreen.

The head reset is driven from PE9 through the former FAN1 output stage, whose polarity differs by board revision (the same quirk `has_inverted_fans()` already compensates for). But `indx_head_reset` uses the new-sytle polarity only:

https://github.com/prusa3d/Prusa-Firmware-Buddy/blob/e96ce2b92adcdafae19c132e76a81397b717e057/src/common/hwio_pindef.h#L370-L372

**Fix:** Use new helper functions `Configuration::activate/deactivate_indx_head_reset()` to write out to PE9 that adjust high/low levels depending on board revision.

Basically the same approach as `has_inverted_mmu_reset()` and `activate/deactivate_ext_reset()`


o-vesely:
Hey, good catch, we will fix, the fix will be very similar to this. Thanks a bunch

o-vesely:
Will be in the next indx fw release 🥇

PR #3976, opened 2024-05-03
gcode: Add Marlin / Octoprint SD card compatible behavior

thess:
This PR adds enough SD card gcode compatibility to list USB contents, including sub-directories, and start/stop/pause/resume printing from Octoprint or any other gcode serial interface application. 

- Implement M20 compatible with Marlin (long-file name, size & timestamp)
- Align M23 & M24 compatibility with Marlin
- Add gcode M27 auto report from #3739
- Updates M2x commands for use with **EXTENDED_M20** support
- Hides **/usb** root path

Addresses issue #3054
Alternative solution to #3765 (includes sub-directories and M23/M24 actually work)

Note: Installing the `BGCode` Octoprint plugin makes `.bgcode` files visible in the file listings. The entries can then be selected and printing started, etc. thus increasing the usability of Octoprint with the Buddy.

PR #5374, opened 2026-07-11
gcode: Add Marlin / Octoprint SD card compatible behavior

thess:
This PR adds enough SD card gcode compatibility to list USB contents, including sub-directories, and start/stop/pause/resume printing from Octoprint or any other gcode serial interface application. 

- Implement M20 compatible with Marlin (long-file name, size & timestamp)
- Align M23 & M24 compatibility with Marlin
- Add gcode M27 auto report from #3739
- Updates M2x commands for use with **EXTENDED_M20** support
- Hides **/usb** root path

Addresses issue #3054
Alternative solution to #3765 (includes sub-directories and M23/M24 actually work)
Supersedes #3976

Note: Installing the `BGCode` Octoprint plugin makes `.bgcode` files visible in the file listings. The entries can then be selected and printing started, etc. thus increasing the usability of Octoprint with the Buddy.


src/marlin_stubs/sdcard/M20-M30_M32-M34.cpp
```diff
@@ -1,12 +1,100 @@
...
+#define MAX_RECURSION_DEPTH 4
+// Device root name for FatFS
+#define ROOT_PREFIX "/usb"
+
+// Routine to output single dirent info in Marlin M20 format
+// Note: Use of 'alloca' prohibits inline
+static void __attribute__((noinline)) list_single_entry(const char *dir_path, struct dirent *entry, struct ListControl *lc) {
+    // Construct path to sub-dir.
+    int len = strlen(dir_path) + strlen(entry->d_name) + 2;
+    char *path = reinterpret_cast<char *>(alloca(len));
```
ttabi, line 33: Shouldn't you test for NULL here?

thess: Short answer - not necessary. `alloca` (a built-in compiler intrinsic) always returns a stack-pointer. Subsequent use of the storage may result in a stack overflow, but that is entirely a separate issue.

Issue #5386, opened 2026-07-20
[BUG] selftest_COREONE.cpp  `.heat_max_temp` too low or `.heat_time_ms` too long. Selftest Fail -> Printer not Usable


jsf-h:
### Printer model

Core One INDX

### Firmware version

6.6.2+15578

### Upgrades and modifications

None

### Printing from...

Not Relevant - Calibration ISSUE

### Describe the bug

C1 Indx Fails Nozzle Heating Calibration as it is Heating too Fast - within the 5700ms `.heat_time_ms` it exceeds the `.heat_max_temp` of 255 and fails the Issue. My INDX Head reaches 269 within the 5700ms and thus trips the Upper limit

See Excerpt of LOG of an affected Printer.

```
60.430s [INFO  - Selftest:10] Bed Started
360.430s [INFO  - Selftest:10] Bed target: 110 current: 56.745193
360.430s [INFO  - Selftest:10] Bed heater PID regulator changed to P regulator
360.530s [INFO  - Selftest:10] Bed cooling down, target: 110 current: 56.758411
360.580s [INFO  - Selftest:10] Nozzle Started
360.580s [INFO  - Selftest:10] Nozzle target: 290 current: 46.700001
360.580s [INFO  - Selftest:10] Nozzle heater PID regulator changed to P regulator
360.630s [INFO  - Selftest:10] Nozzle took control of fans
360.780s [INFO  - Selftest:10] Nozzle cooldown not needed, target: 290 current: 46.580002
360.830s [INFO  - Selftest:10] Nozzle returned control of fans
360.880s [INFO  - Selftest:10] Nozzle set target, target: 290 current: 46.580002
360.930s [INFO  - Selftest:10] Nozzle wait, run: target: 290 current: 46.580002
361.930s [INFO  - Selftest:10] Nozzle wait start temp reached: target: 290 current: 85.739998
362.530s [INFO  - Selftest:10] Bed cooling down, target: 110 current: 56.679089
364.530s [INFO  - Selftest:10] Bed cooling down, target: 110 current: 56.586536
366.530s [INFO  - Selftest:10] Bed cooling down, target: 110 current: 56.493992
367.630s [ERROR - Selftest:10] Nozzle 269 out of range (185 - 255)

368.530s [INFO  - Selftest:10] Bed cooling down, target: 110 current: 56.361778
369.680s [INFO  - Selftest:10] Nozzle finish, target: 290 current: 292.529999
[...]

PR #5387, opened 2026-07-20
Increase .heat_max_temp for Core One INDX Selftest

jsf-h:
Suggested Solution for #5386 

Increase .heat_max_temp to prevent failing Heating Selftest.


jsf-h:
Fixed with 6.6.3

